PR #74, opened 2025-10-24
fix: add schema validation to CI pipeline

bokelley:
## Problem

PR #72 was merged without regenerated TypeScript types, causing the release workflow to fail when trying to create the Release PR. The workflow's pre-push hook detected that generated files were out of sync with the latest AdCP schemas.

## Root Cause

The CI workflow only validated schema sync when schema-related files were modified. Regular library code changes (like ) didn't trigger schema validation, allowing out-of-date generated files to slip through.

## Solution

Added a schema validation step to the main CI pipeline that runs on **every PR**:

```yaml
- name: Validate generated files are in sync
  run: |
    npm run generate-types
    
    if \! git diff --exit-code src/lib/types/*.generated.ts; then
      echo "❌ Generated TypeScript files are out of sync\!"
      echo "🔧 Run: npm run sync-schemas && npm run generate-types"
      exit 1
    fi
```

This runs after syncing schemas and checks if regenerating types would produce any changes. If so, the PR fails with clear instructions.

## Benefits

✅ **Prevents release failures**: Schema mismatches caught before merge
✅ **Clear error messages**: Developers know exactly how to fix it
✅ **Runs on all PRs**: Not limited to schema file changes
✅ **Fast feedback**: Fails early in CI before other checks

## Testing

This will be tested when the PR CI runs - the validation should pass since generated files are currently in sync on main.

## Related

- Fixes the issue that caused release workflow failure after PR #72
- Complements existing schema-sync.yml validation
- Works alongside pre-push hooks for local development

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,29 @@ jobs:
     - name: Sync schemas from AdCP
       run: npm run sync-schemas
 
+    - name: Validate generated files are in sync
+      run: |
+        # Regenerate types from schemas
+        npm run generate-types
+
+        # Check if any generated files changed
+        if ! git diff --exit-code src/lib/types/*.generated.ts src/lib/agents/index.generated.ts 2>/dev/null; then
+          echo ""
+          echo "❌ Generated TypeScript files are out of sync with AdCP schemas!"
+          echo ""
+          echo "📋 Changes detected in:"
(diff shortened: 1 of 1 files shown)


bokelley:
Closing in favor of PR #75 which has a cleaner approach:
- Doesn't include the Changesets release commit
- Just has schema regeneration + CI validation
- Lets Changesets naturally create a new Release PR after merge

PR #91, opened 2025-10-27
fix: correct MCP Accept header handling for Headers objects

EmmaLouise2018:
## Problem
The MCP client in ADCP v2.5.1+ was not sending the required `Accept` header to MCP servers. The MCP protocol requires:
```
Accept: application/json, text/event-stream
```

This caused MCP servers to reject requests.

## Root Cause
Bug in `customFetch` function in `src/lib/protocols/mcp.ts`:

```typescript
} else {
  existingHeaders = { ...init.headers };  // BUG: Returns {} for Headers objects!
}
```

When the MCP SDK passes headers as a `Headers` object, JavaScript's object spread operator `{...headers}` returns an empty object `{}`, losing all headers including the critical `Accept` header.

## Solution
Changed header extraction logic to properly iterate through all header types:
- **Headers objects**: Use `forEach()` to extract
- **Array**: Iterate through tuples
- **Plain object**: Use `for...in` loop with `hasOwnProperty` check

This ensures the SDK's `Accept` header is preserved and sent to the server.

## Timeline
- Bug introduced: Oct 20 (commit 086be48) in v2.3.2
- Bug exposed: Between v2.5.0 and v2.5.1 (likely due to SDK behavior change)
- Fix: Now (this PR)

## Testing
✅ Build passes
✅ Lint passes

### DIFF
--- a/.changeset/fix-mcp-headers-object.md
+++ b/.changeset/fix-mcp-headers-object.md
@@ -1,0 +1,21 @@
+---
+"@adcp/client": patch
+---
+
+Fixed MCP Accept header handling for Headers objects
+
+The customFetch function in mcp.ts was incorrectly handling Headers objects by using object spread syntax (`{...init.headers}`), which returns an empty object for Headers instances. This caused the MCP SDK's required `Accept: application/json, text/event-stream` header to be lost.
+
+**Changes:**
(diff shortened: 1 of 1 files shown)


bokelley:
I have fixed this so many times. fucking Claude just reverts it and then changes the tests so it passes. can you swear in GitHub?

PR #101, opened 2025-10-28
fix(validation): handle framework response wrappers in ADCP schema validation

prateakb:
Updated response validation to check extracted data instead of protocol wrapper.

fix: handle framework response wrappers in ADCP schema validation

  **Problem:**
  Schema validation was failing with "formats: Required" errors when calling
  ADK-based agents, even though the agents were returning valid ADCP responses.
  The validator was checking the wrong object layer.

  **Root Cause:**
  When ADK tools have a ToolContext parameter, ADK wraps the response in the
  A2A FunctionResponse format: { id, name, response: {...} }. This is the
  correct behavior per the A2A protocol specification - FunctionResponse is
  the standard wrapper for structured tool outputs.

  The @adcp/client was validating the wrapper object instead of extracting
  the actual ADCP data from the `response` field.

  **Solution:**
  Updated extractResponseData() to detect framework wrappers by checking for
  the { id, name, response } pattern and extracting the nested data before
  validation. Updated validateResponseSchema() calls to validate the extracted
  ADCP data instead of the protocol wrapper.

  **Why This Is a Good Solution:**
  1. Protocol-Agnostic: Works with any framework that follows the wrapper pattern (A2A, MCP, or custom frameworks), not hardcoded to A2A specifics
  2. Standards-Compliant: ADK correctly implements A2A FunctionResponse, so the fix should be in the client, not the agent tools
  3. Zero Breaking Changes: Existing unwrapped responses still work; the fix only activates when a wrapper is detected
  4. Benefits All Users: Any A2A-compliant agent using structured responses will now work correctly with @adcp/client

  **A2A Protocol Context:**
  The A2A (Agent-to-Agent) protocol defines FunctionResponse as the standard
  format for tool outputs. When frameworks like ADK serialize tool responses,
  they wrap the actual data in this envelope:
```
    {
      "id": "tool_call_id",
      "name": "tool_name",
      "response": { /* actual ADCP data here */ }
    }
```
  This wrapper provides metadata for the agent framework while preserving
  the tool's structured output. The @adcp/client must unwrap this envelope
  before validating against ADCP schemas.

  Testing:
  - All 10 ADCP tools pass schema validation
  - list_creative_formats: ✅ Returns format objects
  - get_media_buy_delivery: ✅ Returns delivery data
  - get_products: ✅ Returns product data
  - Backward compatible with direct (unwrapped) responses

  Files Changed:
  - src/lib/core/TaskExecutor.ts:
    - extractResponseData() - Added wrapper detection and extraction
    - validateResponseSchema() calls - Now validate extracted data

  Fixes: Schema validation failures with ADK-based A2A agents
  Related: ADK ToolContext parameter, A2A FunctionResponse format

### DIFF
--- a/src/lib/core/TaskExecutor.ts
+++ b/src/lib/core/TaskExecutor.ts
@@ -245,8 +245,8 @@ export class TaskExecutor {
         // Some agents return { error: "..." } without success field
         const operationSuccess = completedData?.success !== false && !completedData?.error;
 
-        // Validate response against AdCP schema
-        const validationResult = this.validateResponseSchema(response, taskName, debugLogs);
+        // Validate response against AdCP schema - validate extracted data, not protocol wrapper
+        const validationResult = this.validateResponseSchema(completedData, taskName, debugLogs);
 
         // In strict mode, schema validation failures cause task to fail
         const finalSuccess = operationSuccess && validationResult.valid;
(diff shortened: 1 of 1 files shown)


prateakb:
Thank you for taking that over. Cheers

Pratik

PR #147, opened 2025-11-21
fix: A2A async task handling and webhook configuration

EmmaLouise2018:


src/lib/core/ResponseValidator.ts
bokelley: isn't this redundant with the response unwrapper code I added? we should combine - can be a different PR

PR #274, opened 2026-02-26
fix: v2/v3 backwards compat for create_media_buy, update_media_buy, get_products

EmmaLouise2018:
## Summary

- Strip legacy `brand_manifest` before strict Zod validation fires for `create_media_buy` and `update_media_buy` — v2.5 callers no longer hit \"Request validation failed: Unrecognized key: brand_manifest\"
- Convert `brand` (BrandReference) → `brand_manifest` bare domain URL when adapting requests for v2 servers — previously `brand` passed through and v2 servers rejected it
- Throw a clear error when `proposal_id` is sent to a v2 server (proposal mode is v3-only; v2 requires explicit `packages`)
- Strip `catalog` from package items for v2 servers (`optimization_goals` was already stripped; `catalog` was missed)
- Fix dead-code delete: `account_id` → `account` in `adaptGetProductsRequestForV2`
- Align brand URL format: `get_products` and `create_media_buy` both now use `https://<domain>` (bare domain) consistently

## Test plan

- [ ] New tests in `request-validation.test.js`: `brand_manifest` stripped pre-validation for `create_media_buy`, `brand` wins when both supplied, mock-based test verifying conversion at the transport layer
- [ ] Updated + new tests in `v3-compatibility.test.js`: brand → brand_manifest conversion, brand-with-no-domain preserved, proposal_id throws, catalog stripped from packages, `adaptUpdateMediaBuyRequestForV2` brand passthrough documented
- [ ] All existing tests pass
- [ ] Build and typecheck clean

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5787,16 +5787,16 @@
       }
     },
     "node_modules/minimatch": {
-      "version": "10.2.1",
-      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-10.2.1.tgz",
-      "integrity": "sha512-MClCe8IL5nRRmawL6ib/eT4oLyeKMGCghibcDWK+J0hh0Q8kqSdia6BvbRMVk6mPa6WqUa5uR2oxt6C5jd533A==",
+      "version": "10.2.4",
+      "resolved": "https://registry.npmjs.org/minimatch/-/minimatch-10.2.4.tgz",
+      "integrity": "sha512-oRjTw/97aTBN0RHbYCdtF1MQfvusSIBQM0IZEgzl6426+8jSC0nF1a/GmnVLpfB9yyr6g6FTqWqiZVbxrtaCIg==",
       "dev": true,
       "license": "BlueOak-1.0.0",
       "dependencies": {
         "brace-expansion": "^5.0.2"
       },
       "engines": {
-        "node": "20 || >=22"
+        "node": "18 || 20 || >=22"
       },
       "funding": {
         "url": "https://github.com/sponsors/isaacs"


src/lib/core/SingleAgentClient.ts
```diff
@@ -2014,12 +2014,30 @@ export class SingleAgentClient {
...
+    // Strip brand_manifest → brand for any task that uses strict validation with a brand field.
+    // brand takes precedence if both are supplied.
+    if (taskType === 'get_products' || taskType === 'create_media_buy' || taskType === 'update_media_buy') {
+      if (normalized.brand_manifest && !normalized.brand) {
+        const brand = brandManifestToBrandReference(normalized.brand_manifest);
+        if (brand) {
+          normalized.brand = brand;
+        }
+      }
+      delete normalized.brand_manifest;
```
nastassiafulconis: this now normalizes brand_manifest -> brand for update_media_buy, but adaptUpdateMediaBuyRequestForV2 (creative-adapter.ts:92-98) doesn't convert brand back to brand_manifest for v2 servers. so if someone passes brand_manifest through update_media_buy, normalization promotes it to brand, then the v2 adapter passes brand straight through to the v2 server which won't know what to do with it.

either add the brand -> brand_manifest conversion to adaptUpdateMediaBuyRequestForV2 (consistent with create), or drop update_media_buy from this block if the schema never has brand.

test/lib/request-validation.test.js
```diff
@@ -103,6 +103,52 @@ describe('SingleAgentClient Request Validation', () => {
...
+          });
+        } catch (err) {
+          if (err.message.includes('Request validation failed')) {
+            throw err;
+          }
+        }
+      }, 'brand_manifest should be stripped before strict validation, not cause a validation error');
+    });
+
+    test('should prefer explicit brand over brand_manifest when both are supplied', async () => {
```
nastassiafulconis: these tests prove that validation doesn't reject the request, but they don't actually verify the conversion is correct — like that brand ends up with the right value or that brand_manifest was actually removed from the normalized params. the doesNotReject + swallowing non-validation errors pattern means a lot of things could go wrong silently and these would still pass.

would be nice to mock the network layer or spy on what params actually get sent, to assert the conversion result too.

test/lib/v3-compatibility.test.js
```diff
@@ -335,11 +349,59 @@ describe('Creative Assignment Adapter', () => {
       const result = adaptCreateMediaBuyRequestForV2(v3Request);
```
nastassiafulconis: no tests for update_media_buy with brand_manifest anywhere in the PR — the normalization block was updated to include it but there's nothing verifying that path works end to end. worth adding at least a basic test for adaptUpdateMediaBuyRequestForV2 with the new brand semantics.

EmmaLouise2018:
Addressed all 5 review comments — pushed the fixes:

**#1 (`creative-adapter.ts:97`) — `brand?.domain` guard**
Changed `brand ? \`https://\${brand.domain}\`` → `brand?.domain ? \`https://\${brand.domain}\`` so a brand object without a `domain` field doesn't produce `"https://undefined"`.

**#2 (`creative-adapter.ts:98`) — preserve brand when no domain**
When `brand` is present but has no `domain` (so can't be converted to a `brand_manifest` URL), brand is now re-added to the return value — consistent with `adaptGetProductsRequestForV2` which also leaves `brand` on the object when it can't be converted. Added a test for this case.

**#3 (`SingleAgentClient.ts:2032`) — remove `update_media_buy` from normalization block**
Dropped `update_media_buy` from the `brand_manifest → brand` normalization condition. Neither the v2 nor v3 `update_media_buy` schemas have a `brand` field, so the normalization was wrong to include it. Added a comment to the condition explaining the deliberate exclusion.

**#4 (`request-validation.test.js:130`) — verify actual conversion, not just no-throw**
Added a new test that mocks `A2AClient.fromCardUrl` to capture the parameters actually sent to the server. It asserts that `brand_manifest` is not forwarded as an object (i.e., normalization happened). The existing no-throw tests are kept for the validation-layer coverage.

**#5 (`v3-compatibility.test.js:349`) — `adaptUpdateMediaBuyRequestForV2` brand tests**
Added a test that verifies the adapter passes `brand` through unchanged and doesn't produce a `brand_manifest` field — documenting the "no conversion" contract since neither schema has that field.

src/lib/utils/creative-adapter.ts
```diff
@@ -74,13 +74,29 @@ export function adaptPackageRequestForV2(pkg: PackageRequestV3): PackageRequestV
...
+        'The connected server only supports AdCP v2. Provide an explicit packages array instead.'
+    );
+  }
+
+  // Convert v3 BrandReference → v2 brand_manifest URL (bare domain, consistent with get_products).
+  // normalizeRequestParams has already stripped any incoming brand_manifest and
+  // promoted it to brand, so brand is the canonical source here.
+  const brand_manifest = brand ? `https://${brand.domain}` : undefined;
 
   return {
```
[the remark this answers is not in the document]
nastassiafulconis: Fixed in ca7fc3b5 — looks good.

src/lib/utils/creative-adapter.ts
```diff
@@ -74,13 +74,29 @@ export function adaptPackageRequestForV2(pkg: PackageRequestV3): PackageRequestV
...
     ...rest,
```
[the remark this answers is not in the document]
nastassiafulconis: Fixed in ca7fc3b5 — now preserves brand when no domain, consistent with get_products adapter.

src/lib/core/SingleAgentClient.ts
nastassiafulconis: Fixed in ca7fc3b5 — update_media_buy removed from the normalization block since neither schema has brand. Clean solution.

test/lib/v3-compatibility.test.js
nastassiafulconis: Fixed in ca7fc3b5 — added tests for both brand-with-no-domain and adaptUpdateMediaBuyRequestForV2 brand semantics.

nastassiafulconis (approved):
All 5 comments addressed in ca7fc3b5. Looks good to merge.

PR #314, opened 2026-03-11
fix: use raw body bytes for webhook HMAC verification

### LINKED ISSUE #313: Bug: verifyWebhookSignature re-serializes payload instead of using raw body bytes
## Summary

`SingleAgentClient.verifyWebhookSignature()` computes the expected HMAC over `JSON.stringify(payload)` (a re-serialized parsed object) instead of the raw HTTP body bytes. This is the same class of bug as https://github.com/adcontextprotocol/adcp-client-python/issues/151 on the Python signing side.

## The Problem

```javascript
// SingleAgentClient.js line 653
const message = `${ts}.${JSON.stringify(payload)}`;
```

`payload` here is the **parsed** JavaScript object from `req.body`. `JSON.stringify` re-serializes it, which may not produce the same bytes that were on the wire. For example:

- Sender (Python) sends: `{"key": "value",  "num": 1.0}` 
- `JSON.stringify(JSON.parse(...))` produces: `{"key":"value","num":1}`
- Key ordering, spacing, number formatting, and unicode escaping all vary by language

The HMAC will only match if `JSON.stringify` happens to produce byte-identical output to the sender's serializer. This is fragile and breaks cross-language interop.

## The Correct Approach

HMAC verification must always use the raw HTTP body bytes — the exact bytes received on the wire, before JSON parsing. This is how GitHub, Stripe, and Slack all implement webhook HMAC verification.

The AdCP spec itself says: `HMAC-SHA256(secret, rawBody)` — not `HMAC-SHA256(secret, JSON.stringify(parsed))`.

## Current Workaround

We bypass this method entirely by passing `enforceWebhookSignature=false` and implementing our own raw-body verification:

```typescript
// Our workaround in adcp-webhook.ts
const message = `${ts}.${rawBody}`;  // rawBody = actual wire bytes
const expected = `sha256=${createHmac('sha256', secret).update(message).digest('hex')}`;
```

## Suggested Fix

Change `verifyWebhookSignature` to accept raw body bytes instead of a parsed object:

```typescript
// BEFORE (broken):
verifyWebhookSignature(payload: object, signature: string, timestamp: string | number): boolean {
    const message = `${ts}.${JSON.stringify(payload)}`;
    // ...
}

// AFTER (correct):
verifyWebhookSignature(rawBody: string, signature: string, timestamp: string | number): boolean {
    const message = `${ts}.${rawBody}`;
    // ...
}
```

This is a breaking API change, so it could be offered alongside the old method during a deprecation period:

```typescript
verifyWebhookSignatureRaw(rawBody: string, signature: string, timestamp: string | number): boolean
```

## Related

- Python client has the same issue on the **signing** side: https://github.com/adcontextprotocol/adcp-client-python/issues/151
- Spec PR to clarify raw body requirement: https://github.com/adcontextprotocol/adcp/pull/1382

EmmaLouise2018:
## Summary
- `verifyWebhookSignature()` was computing HMAC over `JSON.stringify(payload)` (re-serialized parsed JSON) instead of the raw HTTP body bytes
- This breaks cross-language interop because different JSON serializers produce different byte representations (spacing, key ordering, number formatting)
- Now accepts raw body string as first argument for correct verification; parsed objects still work for backward compat

## Changes
- **`SingleAgentClient.ts`**: `verifyWebhookSignature` accepts `string | any` — uses raw string directly, falls back to `JSON.stringify` for objects. `handleWebhook` passes raw body through. `createWebhookHandler` captures raw body before parsing.
- **`AgentClient.ts`** / **`ADCPMultiAgentClient.ts`**: Updated wrappers to pass `rawBody` through the call chain.
- **`server.ts`**: Added `preParsing` hook on `/webhook/` routes to capture raw body bytes before Fastify parses them.
- **Tests**: Validated against all 10 official AdCP spec test vectors from [`webhook-hmac-sha256.json`](https://github.com/adcontextprotocol/adcp/blob/main/static/test-vectors/webhook-hmac-sha256.json) (adcontextprotocol/adcp#1383). Covers compact JSON, Python `json.dumps()` spacing, unicode, pretty-printed, empty body, and timestamp edge cases.

Closes #313

## Test plan
- [x] All 10 official spec HMAC test vectors pass (raw computation)
- [x] All 10 spec vectors pass through `verifyWebhookSignature` end-to-end
- [x] Python `json.dumps()` cross-language interop test confirms raw body succeeds and re-serialized object fails
- [x] Backward compat test confirms parsed objects still work when sender uses same serializer
- [x] All existing tests pass (29 total, 0 failures)
- [x] Build and typecheck pass

### DIFF
--- a/test/webhook-signature-verification.test.js
+++ b/test/webhook-signature-verification.test.js
@@ -201,3 +201,120 @@ describe('Webhook Signature Verification (PR #86 Spec)', () => {
     assert.strictEqual(isValid, false);
   });
 });
+
+/**
+ * Official AdCP spec test vectors from:
+ * https://github.com/adcontextprotocol/adcp/blob/main/static/test-vectors/webhook-hmac-sha256.json
+ *
+ * These are static, known-good HMAC signatures that all AdCP implementations
+ * must produce identically. Validates cross-language interop.
+ */
+describe('AdCP Spec Test Vectors (webhook-hmac-sha256.json)', () => {
+  const secret = 'test-secret-key-minimum-32-characters-long';
+
+  const vectors = [
+    {
+      description: 'compact JSON (JS-style JSON.stringify)',
+      timestamp: 1700000000,
+      raw_body: '{"event":"creative.status_changed","creative_id":"creative_123","status":"approved"}',
+      expected_signature: 'sha256=c4faf82609efe07621706df0d28c801de2b5145f427e129f243a3839df891a4e',
+    },
+    {
+      description: 'spaced JSON (Python-style json.dumps with default separators)',
+      timestamp: 1700000000,
+      raw_body: '{"event": "creative.status_changed", "creative_id": "creative_123", "status": "approved"}',
+      expected_signature: 'sha256=4acce503547a93922a2b41c32f5f0e646b71a36572fd1536d3d7fcd88a4e5c5f',
+    },
+    {
+      description: 'empty object',
+      timestamp: 1700000000,
+      raw_body: '{}',
+      expected_signature: 'sha256=fc66235ab6cf0a5927d76d88194036fa99c7e08c75d55c9de5008288d448f1a0',
+    },
+    {
+      description: 'nested objects and arrays',
+      timestamp: 1700000000,
+      raw_body:
+        '{"task_id":"task_456","operation_id":"op_789","result":{"media_buy_id":"mb_001","packages":[{"package_id":"pkg_1"},{"package_id":"pkg_2"}]}}',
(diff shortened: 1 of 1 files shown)


EmmaLouise2018:
Thanks for the review! Added the cross-language interop test in e7daa7c — it signs with Python `json.dumps()` output (spaces after `:` and `,`, `1500.0` float) and verifies both that raw body verification passes and that re-serialized parsed object verification fails (proving the bug). Agree on the Python client needing the same treatment on their verification side (#152).

PR #344, opened 2026-03-18
fix: send brand_manifest as object for v2 servers

nastassiafulconis:
## Summary

Fixed validation errors when sending `get_products` and `create_media_buy` requests to v2 servers. The v2 server expects `brand_manifest` as an object with `name` and `url` fields, not a bare URL string.

## Changes

- Updated `adaptGetProductsRequestForV2()` in `pricing-adapter.ts` to produce `{ name, url }` object
- Updated `adaptCreateMediaBuyRequestForV2()` in `creative-adapter.ts` with same fix
- Updated corresponding tests to assert on object structure instead of string

## Test Results

All 135 tests in v3-compatibility suite pass.

### DIFF
--- a/src/lib/utils/pricing-adapter.ts
+++ b/src/lib/utils/pricing-adapter.ts
@@ -241,7 +241,7 @@ export function normalizeProductPricing(product: any): any {
  * Adapt a get_products request for a v2 server.
  *
  * Converts v3 fields to their v2 equivalents:
- * - brand (BrandReference) → brand_manifest (base domain URL string)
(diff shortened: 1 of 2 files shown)


bokelley:
This PR has an unresolved comment from @EmmaLouise2018 about whether the v2 spec actually accepts both string and object forms for `brand_manifest`. If it does, this is working around a non-compliant server rather than fixing the server.

Also agree with the suggestion to use imported AdCP types rather than inline `{ name: string; url: string }` annotations. Please resolve the review thread before this can merge.

PR #500, opened 2026-04-13
feat: compliance status APIs for buyer integration

### LINKED ISSUE #493: feat: compliance status APIs for buyer integration
## Use case

A buyer platform (e.g., Scope3) registers agents they work with on the AgenticAdvertising.org registry. They need to:

1. Query compliance status for specific agents (single + bulk)
2. See per-storyboard pass/fail detail (members-only)
3. Subscribe to compliance changes via the change feed
4. Tie compliance changes back to operators

The server-side infrastructure for this is being built in adcontextprotocol/adcp#2078+. This issue tracks the SDK surface needed to make it easy for platforms to consume.

## New server endpoints

These are being added to the registry API:

- `GET /registry/agents/:url/compliance` — now includes `storyboards_passing` and `storyboards_total` counts (public)
- `GET /registry/agents/:url/storyboard-status` — per-storyboard detail (members-only)
- `POST /registry/agents/storyboard-status` — batch query, accepts `{ agent_urls: string[] }` (members-only, max 100)
- Change feed event: `agent.compliance_changed` — emitted on status transitions, includes operator_domain, tracks, storyboard counts

### RegistryClient

- [ ] `getAgentCompliance(agentUrl)` — dedicated compliance status query (wraps existing endpoint with typed response)
- [ ] `getAgentStoryboardStatus(agentUrl)` — members-only storyboard detail
- [ ] `getAgentStoryboardStatusBulk(agentUrls[])` — batch query (max 100)
- [ ] `lookupOperator(domain)` — needs TypeScript types exported
- [ ] `lookupPublisher(domain)` — needs TypeScript types exported

### RegistrySync

- [ ] Include `compliance_summary` in the agent index (`AgentSearchResult` currently omits compliance data)
- [ ] Emit typed `compliance_changed` events when processing `agent.compliance_changed` feed events
- [ ] `findAgents()` should accept `compliance_status` filter parameter

### Feed integration

- [ ] Add typed event interfaces for `agent.compliance_changed` payload
- [ ] `getFeed({ types: ['agent.compliance_changed'] })` already works with glob filtering — just needs types

## Target DX

From the Scope3 perspective:

```typescript
const sync = new RegistrySync({ client, apiKey: 'sk_...' });
await sync.start();

// Find all passing seller agents
const sellers = sync.findAgents({ 
  type: 'sales', 
  compliance_status: 'passing' 
});

// React to compliance changes
sync.on('agent.compliance_changed', (event) => {
  const { agent_url, operator_domain, current_status, tracks } = event.payload;
  // Update internal routing decisions
});

// One-off detailed lookup (members-only)
const detail = await client.getAgentStoryboardStatus(agentUrl);
// → [{ id: 'media_buy_seller', status: 'passing', last_passed_at: '...' }, ...]
```

## Context

This enables a key ecosystem workflow: when an agent registers on the AAO registry, buyer platforms can automatically consume its compliance status without duplicating testing. The change feed means platforms get push-style notifications (via polling or future webhooks) rather than having to re-test every agent themselves.

bokelley:
## Summary

- Added `getAgentCompliance()`, `getAgentStoryboardStatus()`, `getAgentStoryboardStatusBulk()` to RegistryClient for querying agent compliance and per-storyboard pass/fail detail
- Added `lookupOperator()` and `lookupPublisher()` to RegistryClient with typed responses
- RegistrySync now processes `agent.compliance_changed` feed events, emits typed `compliance_changed` events, and supports `compliance_status` filter in `findAgents()`
- Synced `registry.yaml` from upstream after adcontextprotocol/adcp#2119 added OpenAPI spec coverage
- Exported types: `AgentCompliance`, `AgentComplianceDetail`, `StoryboardStatus`, `OperatorLookupResult`, `PublisherLookupResult`, `ComplianceChangedPayload`

Closes #493

## Test plan

- [x] 26 new tests covering all new methods, auth gating, input validation, URL encoding, deduplication, event handling, and compliance filtering
- [x] All 2642 tests pass (0 failures)
- [x] TypeScript compiles clean
- [x] Smoke tested against prod registry — all 5 endpoints return expected responses
- [x] Code review and security review addressed (runtime payload validation, schema-derived types, bulk dedup)
- [ ] CI passes


bokelley (commented):
## Code Review

Clean PR -- the compliance APIs are well-structured and the RegistrySync integration is solid. A few items:

### API Design (looks good)

The three-tier access model is right:
- `getAgentCompliance()` — public, summary-level (status + storyboard counts)
- `getAgentStoryboardStatus()` — members-only, per-storyboard detail
- `getAgentStoryboardStatusBulk()` — members-only, batch for dashboard views

The `compliance_summary` on `AgentSearchResult` is exactly what we need for storefront onboarding — we can show compliance status inline in the agent picker without extra API calls.

### Agent Type Enum Change

The enum went from `sales/creative/signals/governance/si/buyer/unknown` to `brand/rights/measurement/governance/creative/buying/signals/unknown`. This is a breaking change for consumers using the old enum values. The changeset says `minor` but this should be `major` (or at minimum flagged clearly in the changelog).

Specifically, `sales` → `buying` will break any code filtering by `type === 'sales'`. We have this in our storefront UI (`typeMap` in AddSourceDialog). Not blocking, but worth documenting.

### RegistrySync compliance_changed Event

The event shape is good:
```typescript
compliance_changed: [{ agentUrl: string; previousStatus: AgentCompliance['status']; currentStatus: AgentCompliance['status'] }]
```

For our use case (storefront dashboard showing agent health), we'll subscribe to this to update the Operations bucket in real-time. The event doesn't include the storyboard detail — that's correct, we'd fetch it on-demand when the user drills in.

### lookupOperator / lookupPublisher

These are the endpoints from adcontextprotocol/adcp#2076. Good to see them typed in the client. One note: `lookupOperator` returns `null` on 404 via `nullOn404: true`, which means "domain not found" and "domain found with no agents" are distinguishable (the latter returns `{ domain, member: null, agents: [] }`). That's the right behavior.

### Questions for Integration

1. Does `AgentSearchResult.compliance_summary` get populated automatically when calling `findAgents()`, or does it require a separate enrichment step?

2. For the `compliance_status` filter on `findAgents()` — does `unknown` mean "never tested" or "opted out"? For our storefront agent picker, we want to show untested agents (they're new) but flag opted-out agents differently.

3. The `StoryboardStatus.category` field — does this map to the storyboard YAML `category` field (e.g., `media_buy_seller`, `creative_template`)? We use these categories to determine archetype compatibility (a buying agent needs media_buy storyboards to pass).

### Summary

Ship it. The typed APIs are exactly what we need for the storefront compliance integration. The agent type enum change is the only thing to flag for downstream consumers.

bokelley (commented):
Thanks for the review. Addressing the items:

### Agent Type Enum Change

Good catch. Updated the changeset to `major` and documented the breaking change: `sales` → `buying`, `si` → removed, `buyer` → removed, new types: `brand`, `rights`, `measurement`. This came from the upstream registry spec sync (adcontextprotocol/adcp#2119 added 20+ endpoints and updated the enum to match the current server).

### Answers to Integration Questions

1. **`compliance_summary` on `AgentSearchResult`** — It gets populated two ways: (a) if the server includes it in the `searchAgents` response during bootstrap, it's there immediately; (b) as `agent.compliance_changed` feed events arrive during polling, the sync updates it in-place. No separate enrichment step needed. For agents that haven't been tested yet, the field will be `undefined` (treated as `'unknown'` by the `compliance_status` filter).

2. **`unknown` vs opted-out** — Currently `unknown` means "no compliance data available" (never tested or no data from the server). The upstream spec added `opted_out` as a status on `AgentComplianceDetail` (the endpoint returns `status: 'opted_out'` for agents that opted out). The `AgentCompliance` type on `AgentSearchResult` has the same enum, so once the server populates it, you'll be able to distinguish: `unknown` = never tested, `opted_out` = explicitly opted out. The `compliance_status` filter would work with both values.

3. **`StoryboardStatus.category`** — Yes, it maps to the storyboard YAML `category` field. The upstream schema defines it as a nullable string. You can use it to filter storyboards by archetype (e.g., check that all `media_buy_*` storyboards pass for a buying agent).

PR #502, opened 2026-04-13
docs: add Implementation Traps section to build-seller-agent skill

tescoboy:
## Summary

Adds a new **Implementation Traps** section to `skills/build-seller-agent/SKILL.md` documenting nine production gotchas discovered during a real seller agent implementation. These are issues that the storyboards will catch but the existing docs don't warn you about.

## Traps Documented

1. **Terminal status responses must omit `structuredContent`** — `mediaBuyResponse()` on a `canceled` status causes MCP protocol errors; use plain text-only JSON instead
2. **IO-signing setup URL belongs in `account.setup`** — not at the top level of the media buy response
3. **`valid_actions` required in three places** — `create_media_buy`, `update_media_buy`, AND `get_media_buys`; the storyboard checks all three
4. **`canceled_at`/`canceled_by` must be persisted** — the storyboard polls `get_media_buys` after cancellation and checks these fields
5. **`revision` field required in `create_media_buy`** — storyboard does a `field_present` check
6. **`media_buy.reporting` is a nested capability** — not a `supported_protocols` enum value
7. **Accept header interception for evaluation harnesses** — Addie and other harnesses send `Accept: */*`; must patch `rawHeaders` before the framework reads them
8. **`z.record(z.unknown())` crashes Zod v4 JSON schema** — use `z.object({}).passthrough()` instead
9. **Pin `@adcp/client` version** — storyboard YAML changes between patch releases; local vs harness results may diverge

## Context

These were found while building a production TypeScript MCP seller agent and iterating through full Addie evaluation + all 29 bundled storyboards. Each trap caused a storyboard failure that wasn't obvious from the existing docs.

**Author contact:** harvinsgupta@gmail.com


bokelley (dismissed):
## Review: Implementation Traps Section

Thanks for documenting real-world pain points — this kind of field feedback is valuable. However, several of these traps have accuracy issues that would mislead agent builders if merged as-is.

### Issues that need resolution before merging

1. **Trap #1 (structuredContent)** — The claim that terminal states must omit `structuredContent` isn't supported by the MCP spec. The example also uses the wrong response builder (`mediaBuyResponse` instead of `updateMediaBuyResponse`) and references `textOnlyTaskResponse()`, which doesn't exist in this SDK.

2. **Trap #6 (reporting not in supported_protocols)** — Directly contradicts `src/lib/testing/scenarios/media-buy.ts:489-490`, which checks `supported_protocols?.includes('reporting')`. Following this advice would trigger warnings from our own comply engine.

3. **Traps #4 and #5 (canceled_at, revision)** — These fields don't appear in any bundled storyboard YAML. If they're Addie-specific checks, that distinction needs to be explicit.

### Items that don't belong in protocol documentation

4. **Trap #7 (Accept header patching)** — This is specific to one HTTP server configuration. Most frameworks handle `Accept: */*` without intervention.

5. **Trap #8 (z.record Zod bug)** — A transient library compatibility issue, not an AdCP trap. Better suited to a troubleshooting FAQ or SDK issue.

6. **Trap #9 (pin version)** — Advising version pinning works against semver and means users miss fixes. The real solution is lockfile consistency.

### Items that look accurate

- **Trap #2 (account.setup nesting)** — Looks correct, though would benefit from a schema reference.
- **Trap #3 (valid_actions in 3 places)** — Storyboards do mention this in expected output, though calling it a `field_present` failure may overstate the enforcement.

### Suggestion

Consider distilling the verified items (traps #2, #3) into the existing "Common Mistakes" table at the bottom of the SKILL.md rather than a standalone section. That keeps the guide focused and avoids the risk of documenting one implementation's quirks as protocol truth.

bokelley (commented):
One more thought on placement: rather than a standalone "Implementation Traps" section, consider:

- **Traps #2 and #3** (account.setup nesting, valid_actions in 3 places) → fold into the existing **Common Mistakes** table at the bottom of SKILL.md. They're the same kind of content and that's where agent builders already look.
- **Traps #7 and #8** (Accept header, z.record crash) → if these belong anywhere, a cross-cutting `docs/troubleshooting.md` or a section in `BUILD-AN-AGENT.md` would be better since they affect all agent types, not just sellers.
- **Traps #4, #5, #9** (canceled_at, revision, version pinning) → file as GitHub Issues. The first two may be Addie-specific checks not reflected in bundled storyboards; version drift is a release process concern. Issues get tracked and resolved rather than permanently documented as workarounds.
- **Trap #1 and #6** → these need accuracy fixes before they land anywhere (see inline comments).

The SKILL.md should teach the happy path + common mistakes. Environment-specific debugging belongs elsewhere.

tescoboy:
Updated per review feedback:

Removed the standalone "Implementation Traps" section
Folded the 6 accepted traps into the existing Common Mistakes table as new rows
Dropped traps #7 (Accept header patching), #8 (Zod z.record bug), and #9 (pin version) — agreed these don't belong in the skill file

Ready for another look.

tescoboy:
Apologies for the previous push — it incorrectly included traps #1, #4, #5, #6. Fixed now. PR contains only the two reviewer-approved items:

#2 (account.setup nesting)
#3 (valid_actions in 3 places)

Will hold off from filing #4 and #5 as separate GitHub issues unless you suggest otherwise.

bokelley:
PR #526 adds response builders that structurally eliminate several of these traps:

- **`valid_actions` trap** → `mediaBuyResponse()` and `updateMediaBuyResponse()` now auto-populate `valid_actions` from status via `validActionsForStatus()`. Builders can't forget it.
- **`revision` trap** → `mediaBuyResponse()` defaults `revision: 1` when not provided.
- **`canceled_at`/`canceled_by` trap** → `cancelMediaBuyResponse()` requires `canceled_by` and auto-sets `canceled_at`.

This PR's documentation additions are still valuable — especially the setup URL nesting trap and the general "things storyboards catch" framing. The two PRs are complementary.

bokelley:
Quick update on where this PR lands given recent changes:

**Row 1 (IO-signing setup URL nesting) — keep.** Worth landing. PR #556 just added a runtime guard in `mediaBuyResponse` / `updateMediaBuyResponse` / `getMediaBuysResponse` that throws a descriptive error when `setup` is placed at the top level, but docs + runtime error is two layers of defense. Docs catch it at design-time; the runtime error catches devs who skip the docs. Optionally tighten the fix wording to reinforce: "Builders reject this at runtime — nest it in `account.setup`."

**Row 2 (`valid_actions` in all three responses) — mostly redundant, please drop or rephrase.** PR #526 already makes `createAdcpServer` builders auto-populate `valid_actions` from `status`, and `SKILL.md:317` documents this. A "must appear in all three" mistake row contradicts that message. If you want to keep it, rephrase to cover only the bypass case, e.g.: "Bypassing response builders and forgetting `valid_actions` on update/get responses. Use the builders — they set it automatically from status."

**Also needs rebase** — branch is conflicting against main.

Separately, the storyboard narrative wording that caused the trap in the first place is tracked at adcontextprotocol/adcp#2270 (storyboards are moving to that repo).

bokelley:
Pushed a rebased + tightened version directly to your branch (force-push — your local will need a reset/pull).

**Changes vs your last push:**
- Rebased onto current `main` (resolves the merge conflict).
- Squashed to a single commit.
- **Row 1 (setup nesting)** — tightened to cite the runtime guard: *"Response builders reject a top-level `setup` at runtime."* (See `src/lib/server/responses.ts:76` `assertNoTopLevelSetup`.)
- **Row 2 (`valid_actions`)** — rephrased to target the builder-bypass case, since `createAdcpServer` builders now auto-populate `valid_actions` from `status` (PR #526). New wording: `mediaBuyResponse` and `updateMediaBuyResponse` auto-populate it; `get_media_buys` callers should use `validActionsForStatus()` per buy.
- Added a patch changeset (`skills/**/*` is published, so SKILL.md edits need one per `CLAUDE.md`).

**Still blocked on:**
- CLA Assistant — the original commits were authored as `Your Name <your.email@example.com>`, which the CLA bot can't resolve to a GitHub user. My squash commit is attributed to me with you as `Co-Authored-By`, which should unblock it; if the bot still flags it, adding your GitHub-verified email to your local git config and rebasing would fix it cleanly.

Thanks for the iteration on this — the two surviving rows are genuinely useful guidance.

bokelley (approved):
Rebased onto current main (post-#624 regen), my changes-requested review dismissed, diff is now the two surviving rows in the Common Mistakes table — accurate, useful, minimal. LGTM.

PR #522, opened 2026-04-14
fix: reconcile v2 brand adapter with schema filter for misdetected v3 servers

EmmaLouise2018:
## Summary

- When a v3 server (e.g. Content Ignite) is misdetected as v2 because its `get_adcp_capabilities` call fails, the v2 adapter renames `brand` → `brand_manifest`. The schema filter then strips `brand_manifest` because the v3 tool schema declares `brand` — resulting in empty args and a `VALIDATION_ERROR`.
- Added adapter alias reconciliation in the schema filter: if `brand_manifest` is present but undeclared, and `brand` IS declared, maps the value back to `brand`.
- Improved logging when `get_adcp_capabilities` fails to surface why a server falls back to v2 detection, including agent ID, error details, and tool list.

## Test plan

- [ ] Verify existing tests pass (181/181 passing)
- [ ] Test against Content Ignite to confirm `brand` is no longer stripped
- [ ] Check warn logs to identify root cause of `get_adcp_capabilities` failure for Content Ignite

### DIFF
--- a/.changeset/fix-v2-brand-adapter.md
+++ b/.changeset/fix-v2-brand-adapter.md
@@ -1,0 +1,5 @@
+---
+"@adcp/client": patch
+---
+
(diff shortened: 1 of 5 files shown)


nastassiafulconis (approved):
thanks!

PR #523, opened 2026-04-14
feat: add filterInvalidArrayItems validation option

EmmaLouise2018:
## Summary
- Adds `validation.filterInvalidProducts` client config option (default: `false`)
- When enabled, invalid products in `get_products` responses are filtered out individually instead of rejecting the entire response
- Only applies to `get_products` — all other tool responses use standard validation
- Includes 3 new tests covering filtering, empty result, and passthrough scenarios

## How it works
When `get_products` whole-response Zod validation fails and this option is `true`:
1. Validates each product individually against `ProductSchema`
2. Keeps valid products, drops invalid ones
3. Re-validates the filtered response against the full `GetProductsResponseSchema`
4. Returns it if valid — otherwise falls through to the normal error path

## How this differs from `strictSchemaValidation: false`

| | `strictSchemaValidation: false` | `filterInvalidProducts: true` |
|---|---|---|
| Invalid product in array | Returns the **entire raw response** unmodified (all 10 products, including the 9 bad ones) | Returns only the **1 valid product**, drops the 9 bad ones |
| Missing required top-level field | Logs warning, returns response anyway | Still throws — only filters products |
| Type guarantee | **None** — you get unvalidated data | **Validated** — returned data passes the full Zod schema |
| Scope | All tool responses | `get_products` only |

`strictSchemaValidation: false` already lets data through when some products are bad, but passes everything unvalidated. `filterInvalidProducts` gives you Zod-validated output with only the bad products removed — the response you get back is guaranteed to match the schema.

## Usage
```ts
const client = new AdCPClient(agents, {
  validation: {
    filterInvalidProducts: true,
  },
});
```

## Test plan
- [x] Existing 44 response-unwrapper tests pass
- [x] New test: filters invalid products, returns valid ones
- [x] New test: returns empty array when all products invalid
- [x] New test: passes through fully valid responses unchanged
- [x] TypeScript compiles cleanly
- [x] Lint passes (0 errors)

### DIFF
--- a/src/lib/utils/response-unwrapper.ts
+++ b/src/lib/utils/response-unwrapper.ts
@@ -66,7 +66,7 @@ export function unwrapProtocolResponse(
   protocolResponse: any,
   toolName?: string,
   protocol?: 'mcp' | 'a2a',
-  options?: { filterInvalidArrayItems?: boolean }
+  options?: { filterInvalidProducts?: boolean }
 ): AdCPResponse & { _message?: string } {
   if (!protocolResponse) {
     throw new Error('Protocol response is null or undefined');
@@ -105,10 +105,10 @@ export function unwrapProtocolResponse(
       const { _message: _msg, ...dataToValidate } = unwrapped as Record<string, unknown>;
       const result = schema.safeParse(dataToValidate);
       if (!result.success) {
-        // When filterInvalidArrayItems is enabled, try filtering invalid items from
-        // array fields individually rather than rejecting the entire response.
-        if (options?.filterInvalidArrayItems) {
-          const filtered = filterInvalidItems(schema, dataToValidate);
+        // When filterInvalidArrayItems is enabled and this is a get_products response,
+        // try filtering invalid products individually rather than rejecting the entire response.
+        if (options?.filterInvalidProducts && toolName === 'get_products') {
+          const filtered = filterInvalidProducts(schema, dataToValidate);
           if (filtered) {
             const validated = filtered as unknown as AdCPResponse & { _message?: string };
             if (_msg) validated._message = _msg as string;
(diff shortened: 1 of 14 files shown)


nastassiafulconis (commented):
## Code Review: feat: add filterInvalidArrayItems validation option

Nice feature -- the ability to gracefully degrade on partially-invalid `get_products` responses instead of rejecting everything is a practical need. The implementation approach is sound: validate each product individually, keep the good ones, re-validate the whole response. A few issues to address before merging.

---

### Critical

**1. CI is failing -- generated files are out of sync**

The "Validate generated files are in sync" step and the "Validate Schema Synchronization" check are both failing. The generated files in this PR appear to have been generated against a slightly stale schema snapshot. You'll need to run `npm run sync-schemas && npm run generate-types` and recommit the generated files.

**2. `reporting_capabilities` becoming required in ProductSchema is a breaking change**

The generated schema changes `reporting_capabilities` from `.nullish()` to required on `ProductSchema`. This means any seller that does not return `reporting_capabilities` in their products will now cause validation failures. This is a significant behavioral change bundled into what's labeled as a `minor` changeset. Two sub-issues:

- The `createTestProduct()` test fixture does not include `reporting_capabilities`. This means products from `createTestProduct()` will now fail validation against the updated ProductSchema. Ironically, the "should pass through fully valid responses unchanged" test is probably passing *because* the products are actually *invalid* under the new schema and filtering kicks in (or it may throw entirely). The fixture needs to be updated.
- This should be called out in the PR description and changeset -- consumers need to know that `reporting_capabilities` is now a required field.

**3. Naming inconsistency: `filterInvalidProducts` vs `filterInvalidArrayItems`**

The public config uses `filterInvalidProducts` (`SingleAgentClient.ts`), but the internal option parameter is `filterInvalidArrayItems` (`response-unwrapper.ts`). The PR title also references `filterInvalidArrayItems`. Since the feature is scoped exclusively to `get_products` and hardcoded to the `products` array, the generic name `filterInvalidArrayItems` is misleading -- it suggests it would work on any array in any tool response, which it does not. Pick one name and use it consistently. I'd suggest `filterInvalidProducts` everywhere since that accurately describes the scope.

---

### Important

**4. No logging/observability when products are silently filtered**

When invalid products are dropped, there is no log output. In production, silently swallowing data will make debugging difficult. At minimum, log a warning with the count of filtered products (e.g., "Filtered 3 of 10 invalid products from get_products response"). The client already has `logSchemaViolations` infrastructure -- this should use it.

**5. Test coverage gaps**

The three new tests cover the happy paths well, but a few edge cases are missing:

- **Non-`get_products` tool with filtering enabled**: Verify that filtering does NOT apply to other tools (e.g., `get_formats` with an invalid format in the array should still throw).
- **A2A protocol path**: All three tests use MCP format. At least one should verify the filtering works through the A2A unwrapping path.
- **`products` field is not an array**: What happens if a malformed response has `products: "not-an-array"`? The code handles this (returns null, falls through to error), but a test would document the behavior.
- **Mixed with `strictSchemaValidation: false`**: The PR description explains how these two options interact, but there is no test for the combination.

**6. The `validProducts.length === products.length` early return may mask non-product errors**

When all products are individually valid but the *response-level* validation still fails (e.g., a required top-level field is missing), `filterInvalidProducts` returns `null` because nothing was filtered. This is correct -- the error gets re-thrown via the normal path. But it is worth adding a brief inline comment explaining this design choice, since the "Nothing was filtered -- the error is elsewhere" comment could be read as "all products valid = response must be valid", which is not the case.

---

### Suggestions

**7. Consider returning the validation errors for filtered products**

Rather than just a count, it would be useful for debugging if the function could surface *which* products were filtered and *why*. This could be a debug-level log with the product_id and first Zod error for each filtered product. Not blocking, but would significantly improve production debuggability.

**8. The `as z.ZodObject<any>` and `as z.ZodArray<any>` casts are fragile**

The `filterInvalidProducts` function casts the schema to `z.ZodObject` and then accesses `.shape.products`. If the schema is wrapped (e.g., in a `z.ZodEffects` from `.transform()` or `.refine()`), this will silently fail (return null). This is acceptable today since the generated schemas don't use transforms, but worth noting as a maintenance concern. A comment like "// Assumes unwrapped ZodObject -- will return null (fallback to normal error) if schema is wrapped" would help future readers.

---

### What was done well

- The re-validation of the full response after filtering is the right call -- it catches issues beyond just individual products.
- The fallback behavior (return null -> fall through to normal error path) is a safe design.
- The PR description and comparison table with `strictSchemaValidation: false` is excellent.
- The changeset is correctly scoped as `minor`.

nastassiafulconis (commented):
Inline comments on specific code locations (see above review for full context).

src/lib/utils/response-unwrapper.ts
```diff
@@ -58,13 +58,15 @@ export type AdCPResponse =
...
  * @param protocol - Protocol type ('mcp' or 'a2a'), if known. If not provided, will auto-detect.
+ * @param options - Optional validation behavior overrides
  * @returns Raw AdCP response data matching schema exactly
  * @throws {Error} If response doesn't match expected schema for the tool
  */
 export function unwrapProtocolResponse(
   protocolResponse: any,
   toolName?: string,
-  protocol?: 'mcp' | 'a2a'
+  protocol?: 'mcp' | 'a2a',
```
nastassiafulconis, line 68: Naming inconsistency: this parameter is called `filterInvalidArrayItems` but the public config in `SingleAgentClient.ts` uses `filterInvalidProducts`. Since the feature only applies to `get_products` and is hardcoded to the `products` array, the generic name is misleading. Suggest renaming to `filterInvalidProducts` here as well for consistency.

src/lib/utils/response-unwrapper.ts
```diff
@@ -130,6 +143,36 @@ export function unwrapProtocolResponse(
...
+  if (!(ProductSchema instanceof z.ZodArray)) return null;
+
+  const elementSchema = (ProductSchema as z.ZodArray<any>).element;
+  const validProducts: unknown[] = [];
+  for (const product of products) {
+    if (elementSchema.safeParse(product).success) {
+      validProducts.push(product);
+    }
+  }
+
```
nastassiafulconis, line 167: Nit: This comment could be misread as "all products valid means response is valid", which is not the case -- a top-level required field could still be missing. Consider: "Nothing was filtered -- all products are individually valid, so the validation error is at the response level (not caused by invalid products). Fall through to the normal error path."

src/lib/utils/response-unwrapper.ts
```diff
@@ -130,6 +143,36 @@ export function unwrapProtocolResponse(
...
+  // Nothing was filtered — the error is elsewhere
+  if (validProducts.length === products.length) return null;
+
+  const filtered = { ...data, products: validProducts };
```
nastassiafulconis, line 173: No observability when products are silently dropped. Consider logging a warning here with the count of filtered products. The client already has `logSchemaViolations` infrastructure that could be leveraged.

test/lib/response-unwrapper.test.js
```diff
@@ -633,6 +633,59 @@ describe('Response Unwrapper', () => {
...
+      };
+
+      const result = unwrapProtocolResponse(mcpResponse, 'get_products', 'mcp', {
+        filterInvalidArrayItems: true,
+      });
+      assert.strictEqual(result.products.length, 2);
+    });
   });
 
   describe('Protocol Auto-Detection Edge Cases', () => {
```
nastassiafulconis, line 741: Missing test cases to consider:
- Non-`get_products` tool with filtering enabled (verify it does NOT filter other tools)
- A2A protocol path (all 3 tests use MCP only)
- `products` field is not an array (malformed response)
- Interaction with `strictSchemaValidation: false`

nastassiafulconis (approved):
lgtm! one small suggestion inline

src/lib/utils/response-unwrapper.ts
```diff
@@ -130,6 +143,36 @@ export function unwrapProtocolResponse(
...
+  const revalidated = schema.safeParse(filtered);
```
nastassiafulconis, line 174: nit: might be nice to log something when products get dropped here — right now it's totally silent which could make debugging tricky. even just a debug log with how many were filtered out would help

PR #550, opened 2026-04-16
fix: protocol-aware connection cleanup for storyboard A2A testing

### LINKED ISSUE #549: feat: storyboard A2A testing parity
## Problem

Storyboards (`npx @adcp/client storyboard run URL`) only test agents via MCP. There is no `--protocol a2a` flag. A2A agents cannot be validated with storyboards today.

This means:
- Agents serving A2A only cannot run compliance tests
- Dual-protocol agents (MCP + A2A) only get their MCP path tested
- Protocol-specific bugs in A2A serialization/routing go undetected

## Proposed solution

Add A2A protocol support to the storyboard runner:

```bash
npx @adcp/client storyboard run https://agent.example.com media_buy_seller --protocol a2a
```

The storyboard scenarios are protocol-agnostic (they test tool semantics, not transport). The runner just needs to use A2A task submission instead of MCP tool calls.

## Context

- prebid/salesagent serves both MCP and A2A and cannot validate their A2A path
- The Python SDK is adding A2A server support (adcontextprotocol/adcp-client-python#175)
- The JS client already has `callA2ATool` and `createA2AClient` utilities

Ref: prebid/salesagent#1183

bokelley:
## Summary

- Adds `closeConnections(protocol)` to `src/lib/protocols/index.ts` that routes to `closeMCPConnections()` for MCP and is a no-op for A2A (ephemeral clients)
- Replaces unconditional `closeMCPConnections()` calls in storyboard runner and compliance engine with protocol-aware `closeConnections(options.protocol)`
- Fixes MCP-specific comments to be protocol-neutral

## Context

The `--protocol a2a` flag already flows through CLI -> TestOptions -> StoryboardRunOptions -> client creation. The storyboard scenarios are fully protocol-agnostic. The only gap was connection cleanup unconditionally calling MCP-specific teardown.

Closes #549

## Test plan

- [x] TypeScript compiles cleanly
- [x] 1291 storyboard tests pass
- [x] Compliance abort tests pass
- [x] Governance stub tests pass


EmmaLouise2018 (approved):

bokelley:
Tested the storyboard runner against our Python A2A server (`adcp-client-python` on `bokelley/issue-175`). Overall results are positive — two minor issues found.

**Results**: capability_discovery passed 2/2, media_buy_seller passed 7/9 (failures were server-side handler issues, not runner bugs). Protocol detection, skill dispatch via DataPart, response extraction from artifacts, `_message` stripping from TextParts, and context validation all worked correctly.

**Issue 1: Agent card fetched per request**

Each `callA2ATool` invocation creates an ephemeral `A2AClient.fromCardUrl()`, which means a `GET /.well-known/agent.json` is issued for every tool call. For a 9-step storyboard this produced ~18 agent card fetches (one for discovery, one for the call, per step). The card doesn't change between calls.

Relevant code: `src/lib/protocols/a2a.ts` — `callA2ATool` constructs a new client each time rather than reusing one.

**Issue 2: `[AdCP]` log line on stdout breaks `--json` mode**

When an agent has `comply_test_controller` but doesn't declare `compliance_testing` in capabilities, the runner prints:

```
[AdCP] Agent has comply_test_controller but does not declare compliance_testing. Adding automatically.
```

This goes to stdout, which means `--json` output is no longer valid JSON (the log line is interleaved with the JSON output). This should either go to stderr or be suppressed when `--json` is active.

bokelley:
Both issues fixed in 9d89a7c6.

**Issue 1**: A2AClient is now cached by (agentUrl, authToken hash) — agent card fetched once per (url, token), not per call. Per-call state (debugLogs, customHeaders, got401) flows through `AsyncLocalStorage` so concurrent callers sharing a cached client are isolated. Cache entry is evicted on 401. `closeConnections('a2a')` clears the whole cache.

**Issue 2**: `[AdCP]` auto-augment log moved from `console.debug` (stdout) to `console.error` (stderr). `--json` output is no longer corrupted.

PR #558, opened 2026-04-17
feat(server): optimistic concurrency — putIfMatch, getWithVersion, patchWithRetry

bokelley:
## Summary

Ships the second upstream RFC from the seller batch: an atomic compare-and-swap primitive on `AdcpStateStore` so read-modify-write loops (counters, accumulators, guarded state transitions) don't silently lose writes.

### New surface

- **`patchWithRetry(store, collection, id, updateFn, options?)`** — the primitive most callers actually want. Runs get → compute → `putIfMatch` → retry on conflict. Throws `PatchConflictError` after `maxAttempts` (default 5).
- **`putIfMatch(collection, id, data, expectedVersion)`** — raw CAS. Returns `{ok: true, version}` on success or `{ok: false, currentVersion}` on conflict. `expectedVersion: null` is insert-only.
- **`getWithVersion(collection, id)`** — read a document with its row version.
- **`PatchConflictError`** — thrown when `patchWithRetry` exhausts retries.
- **`VersionedDocument<T>`** and **`PutIfMatchResult`** types.

Both built-in stores (`InMemoryStateStore`, `PostgresStateStore`) track a monotonically increasing `version` per row. Every `put` / `patch` / `putIfMatch` bumps it. Sessioned stores (`createSessionedStore` / `store.scoped(key)`) proxy the new methods automatically, with method references captured via `.bind` at wrap time.

### Postgres migration

Additive and idempotent:

```sql
ALTER TABLE adcp_state ADD COLUMN IF NOT EXISTS version INTEGER NOT NULL DEFAULT 1;
```

Existing rows start at version 1. No data rewrite. `getAdcpStateMigration()` emits both the `CREATE TABLE` and the `ALTER TABLE` so new and upgrading deployments are covered by a single call.

### Compatibility

No breaking changes. Both `getWithVersion` and `putIfMatch` are optional on the `AdcpStateStore` interface; custom store implementations that don't implement them keep working (with `patchWithRetry` emitting a `BACKEND_ERROR` StateError that names the missing methods).

### Reviewed

Code-reviewer pass surfaced three issues, all addressed:
- **Conflict-report race**: documented that `currentVersion` is a best-effort post-CAS snapshot. Fine for `patchWithRetry`; direct callers using it for anything beyond "retry if it changed" should account for the race.
- **`patchWithRetry` mutation + throw semantics**: documented. Closure exceptions propagate immediately without retry; nested in-place mutation is a caller bug.
- **Sessioned-store proxy `this`-binding**: hardened with `.bind(inner)` at wrap time so `PostgresStateStore.getWithVersion` (which closes over `this.db`) works correctly through the scope wrapper.

## Test plan

- [x] `test/server-put-if-match.test.js` — 23 tests covering version tracking, CAS success/conflict paths, `patchWithRetry` (create, update, null-abort, retry on intervening writes, exhaustion, propagating exceptions, missing-method error), sessioned-store proxying, and `_session_key` rejection on CAS payload.
- [x] Concurrent-insert test: two `Promise.all` inserters with `expectedVersion: null` — exactly one wins; loser sees `currentVersion: 1`.
- [x] `npm test` — 3181 / 3183 (2 intentional skips, 0 failures).
- [x] `npx tsc --noEmit` — clean.
- [x] `npx prettier --check` — clean.

### DIFF
--- a/.changeset/put-if-match.md
+++ b/.changeset/put-if-match.md
@@ -1,0 +1,23 @@
+---
+'@adcp/client': minor
+---
+
+Optimistic concurrency primitives on `AdcpStateStore`.
+
+**New**
+
+- `putIfMatch(collection, id, data, expectedVersion)` — atomic compare-and-swap. Returns `{ok: true, version}` on success, `{ok: false, currentVersion}` on conflict. `expectedVersion: null` means insert-only.
+- `getWithVersion(collection, id)` — read a document with its row version.
+- `patchWithRetry(store, collection, id, updateFn, options?)` — get → compute → putIfMatch → retry loop for read-modify-write updates. Throws `PatchConflictError` after `maxAttempts` (default 5).
+- Both built-in stores (`InMemoryStateStore`, `PostgresStateStore`) track a monotonically increasing `version` per row. Every `put`/`patch`/`putIfMatch` bumps it.
+- Sessioned stores (`createSessionedStore` / `store.scoped(key)`) proxy the new methods through so scoped views get CAS for free.
+
+**Postgres migration**
+
+- `getAdcpStateMigration()` adds `version INTEGER NOT NULL DEFAULT 1` via `ADD COLUMN IF NOT EXISTS`. Existing rows start at version 1. No data rewrite.
+
+**Docs**
+
+- `docs/guides/CONCURRENCY.md` gains a section covering `patchWithRetry`, `putIfMatch`, and when to reach for each.
+
+No breaking changes. Both new methods are optional on the `AdcpStateStore` interface; custom stores that don't implement them keep working.
--- a/docs/guides/CONCURRENCY.md
+++ b/docs/guides/CONCURRENCY.md
@@ -169,9 +169,67 @@ const server = createAdcpServer({
 - `list()` cursors are valid **only within the session** that produced them.
   Don't pass a cursor from session A to session B.
 
-## Coming soon: optimistic concurrency
+## Optimistic concurrency with `putIfMatch`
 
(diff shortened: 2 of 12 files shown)


bokelley:
Review follow-ups pushed as `f892cb64`:

1. **Split schema regen** → PR #559 (mechanical regen, no handwritten code). Merge that first; I'll rebase this PR on top and the schema-drift diff will disappear naturally.
2. **Postgres integration test** → `test/lib/postgres-state-store.test.js`, gated on `DATABASE_URL`. Covers real concurrent-client CAS (two separate pool clients racing both null-expected and version-matched `putIfMatch`), 50 concurrent `patchWithRetry` increments landing exactly once each, and the `ADD COLUMN IF NOT EXISTS` upgrade path from a pre-version table.
3. **Cosmetic concurrency test dropped** — renamed and clearly commented that IMS is deterministic by microtask order; real concurrency lives in the Postgres test.
4. **Backoff clamp** — `Math.pow(2, Math.min(attempt, 20))` so `maxAttempts >= 31` doesn't silently lose backoff to signed-int overflow.
5. **Shallow-copy doc** — dropped the IMS-vs-Postgres divergence and replaced with "always build a fresh object in `update`."

Also: re-exported `patchWithRetry` / `PatchConflictError` / `isPutIfMatchConflict` / `scopedStore` / `createSessionedStore` / `requireSessionKey` / `structuredSerialize` / `structuredDeserialize` / `StateError` / `SESSION_KEY_FIELD` from the top-level `@adcp/client` package (they were only on `@adcp/client/server` before).

Issue #565, opened 2026-04-18
Implement security_baseline storyboard runner + middleware helpers


bokelley:
## Background

adcp#2298 adds `universal/security.yaml` — a new conformance storyboard that every AdCP agent must pass regardless of specialism. It verifies:

1. Unauthenticated requests are rejected with 401/403 (and `WWW-Authenticate` on 401 per RFC 6750 §3).
2. At least one of API key or OAuth is implemented.
3. For OAuth, the `resource` advertised at `/.well-known/oauth-protected-resource/<path>` equals the agent URL under test (RFC 9728). This catches the real-world audience-mismatch bug from #563.

The storyboard YAML uses two additive directives that the runner in this repo needs to implement. This issue tracks that runner work plus the SDK-side helpers sellers need to pass the test.

### 1. `auth: none` step directive

Force the runner to strip transport credentials for a single step even when the client has a valid token configured, so the unauth-rejection probe actually exercises the unauthenticated path.

```yaml
steps:
  - id: probe_unauth
    task: get_adcp_capabilities
    auth: none
    expect_error: true
    validations:
      - check: http_status_in
        allowed_values: [401, 403]
      - check: on_401_require_header
        value: \"www-authenticate\"
```

### 2. `contributes_to` + `any_of` accumulator

Lightweight flag accumulator so a downstream phase can require that at least one of several optional phases marked a contribution. Used to encode \"API key OR OAuth must be verified\" without runtime branching.

```yaml
# An earlier optional step marks a flag on success:
- id: probe_api_key
  contributes_to: auth_mechanism_verified

# A later assertion step requires any flag was set:
- id: assert_mechanism
  task: assert_contribution
  validations:
    - check: any_of
      allowed_values: [auth_mechanism_verified]
```

### 3. New step tasks (raw HTTP probes, not MCP tools)

- \`protected_resource_metadata\` — GET \`/.well-known/oauth-protected-resource/<path>\` against the agent URL, parse JSON, apply validations.
- \`oauth_auth_server_metadata\` — GET \`/.well-known/oauth-authorization-server\` for the first issuer listed in the previous step's response.
- \`assert_contribution\` — synthetic step that evaluates accumulated flags, no network call.

### 4. New validation checks

- \`http_status\` — exact status code match.
- \`http_status_in\` — status code in allowed_values (already used in the storyboard).
- \`on_401_require_header\` — if the response was 401, the named header must be present (case-insensitive). Silent pass on non-401 responses.
[...]

Issue #563, opened 2026-04-18
OAuth DX: make authenticated agent access trivial


bokelley:
## Background

Debugging a recent OAuth cross-wiring between `test-agent.adcontextprotocol.org` and the AAO authorization server took hours of manual wire-level probing, despite PR #562 having correctly added OAuth plumbing. The fix was in a server config (adcp-1#2297). The debug time was spent because the SDK gives consumers no way to see what's actually happening on the wire when OAuth misbehaves.

We shipped the correct client-side fix. We should also close the DX gap so the next incident takes seconds instead of hours — and, where possible, make OAuth work without the consumer having to understand resource indicators, JWKS, PKCE, refresh semantics, or `WWW-Authenticate` parsing at all.

### 1. `adcp diagnose-auth <alias>` — new CLI command

End-to-end diagnostic that traces the full OAuth handshake and produces structured, ranked hypotheses about what's failing. This is the single highest-leverage addition.

What it should do:
- Wrap fetch and capture wire-level HTTP for the entire round-trip
- Probe `.well-known/oauth-protected-resource/<path>` and `.well-known/oauth-authorization-server`
- Compare the advertised resource URL to the agent's actual host
- Attempt a token refresh if saved tokens are expired
- Decode access_token JWT claims pre/post refresh and diff them
- Attempt a `listTools()` call and capture the response headers + body
- Output ranked hypotheses:
  - H1: Resource URL mismatch between well-known and agent host
  - H2: Refresh grant missing `resource` parameter (RFC 8707)
  - H3: `mcp-session-id` not preserved across retry
  - H4: Agent endpoint returns 401 with no `WWW-Authenticate` (RFC 6750 violation)
  - H5: Token `aud` claim missing or doesn't match agent URL
  - H6: Agent endpoint doesn't implement OAuth validation at all (accepts token but ignores it)

Reference implementation already exists — the one-off probe we wrote to diagnose the recent bug is ~200 lines and directly maps to these hypotheses. Polishing it into `adcp diagnose-auth` is maybe a day.

### 2. `parseWWWAuthenticate(header)` — exported utility

The SDK has no parser for `WWW-Authenticate` challenges. Every consumer that wants to read `resource_metadata=` (the point of RFC 9728 §5.3 discovery!) has to roll their own regex. Add `parseWWWAuthenticate(header): { scheme, realm?, error?, error_description?, scope?, resource_metadata? }`. ~20 lines.

[...]

PR #567, opened 2026-04-18
feat: security_baseline storyboard runner — directives, HTTP probes, SSRF guardrails

bokelley:
## Summary

Implements the runner-side work tracked in #565 so the upstream `universal/security.yaml` storyboard from [adcontextprotocol/adcp#2298](https://github.com/adcontextprotocol/adcp/issues/2298) can actually execute. No storyboard YAML ships here — the real one arrives via the tarball sync when the upstream PR lands.

Stacked conceptually on #564 (server-side middleware + skills): the middleware helps sellers *pass* the baseline; this PR teaches the runner how to *test* the baseline.

### Step / phase directives

- `auth: 'none'` — force a single step to hit the agent with no credentials (unauth probe).
- `auth: { type: 'api_key', value | from_test_kit: true }` — per-step Bearer override for the invalid-key and valid-key probes.
- `contributes_to: '<flag>'` — mark a step as contributing a flag on success.
- `contributes_if: 'prior_step.<id>.passed'` — conditional contribution (so `auth_mechanism_verified` only lights up when *both* the valid-key and invalid-key steps pass).
- Phase-level `skip_if: '!test_kit.auth.api_key'` — skip optional phases based on test-kit data.

### New raw-HTTP task handlers

- `protected_resource_metadata` — GET `/.well-known/oauth-protected-resource<path>` (RFC 9728).
- `oauth_auth_server_metadata` — GET `<issuer>/.well-known/oauth-authorization-server`. **SSRF-hot path** — the URL comes from agent-supplied metadata.
- `assert_contribution` — synthetic accumulator check, no network.

### New validation checks

- `http_status` / `http_status_in` — status code match.
- `on_401_require_header` — silent pass on non-401; require named header on 401 (RFC 6750 §3).
- `resource_equals_agent_url` — normalized comparison of RFC 9728 `resource` against the URL under test. Catches the audience-mismatch bug from #563. **Error message does not echo the advertised value** — compliance reports are shareable and detailed diffs help attackers probe victim agents.
- `any_of` — at least one listed flag must be in the accumulator.

### Safety

- **HTTPS enforcement**: `comply()` refuses `http://` agent URLs by default. Pass `{ allow_http: true }` (or `--allow-http` on the CLI) for local dev. The CLI prints a non-publishable banner when the flag is used.
- **SSRF guardrails** on auth-server discovery fetches:
  - HTTPS only.
  - DNS-resolved address check: blocks loopback, RFC 1918, link-local (169.254.x, fe80:), IPv6 ULA (fc/fd), and the `0.0.0.0` / `::` sentinels.
  - 10 s total timeout, 64 KiB body cap, `redirect: 'manual'` (no cross-host redirect following).

### Test-kit schema

`TestOptions.test_kit.auth.api_key` and `auth.probe_task` (default `list_creatives`) are now recognized. Storyboard phases read these via `skip_if` and the `from_test_kit: true` auth directive.

## Breaking (internal)

`runValidations(validations, taskName, taskResult)` became `runValidations(validations, validationContext)` to carry the HTTP probe result, agent URL, and contribution accumulator alongside the MCP task result. Existing internal callers updated. External callers who import `runValidations` directly must pass a `ValidationContext` object.

## Deferred to v1.1 (noted on #565)

- `adcp diagnose-auth` CLI for local debugging (#563).
- Authenticated-call audience check (one live call after OAuth discovery to verify the validator works, not just the metadata).
- Test-kit YAML migration (`auth.api_key` / `auth.probe_task` values for existing kits — upstream adcp repo).

## Test plan

- [x] `test/lib/storyboard-security.test.js` — 22 tests covering `isPrivateIp` (IPv4/IPv6), `fetchProbe` SSRF refusals + happy path via throwaway HTTP server, `PROBE_TASKS` registry, each new validation check with positive and negative cases, validation-context discrimination (HTTP-only vs MCP-only check mismatch surfaces cleanly), and `comply()` HTTPS enforcement.
- [x] `test/lib/storyboard-validations.test.js` — updated for the new `runValidations(validations, ctx)` signature.
- [x] `npm test` — 3328 / 3330 (2 intentional skips, 0 failures).
- [x] `npx tsc --noEmit` — clean.
- [x] `npx prettier --check` — clean.


bokelley:
Pushed runner support for the post-2304 scope. Now implements everything the shipped storyboard needs:

**Task resolution**
- `task: "$test_kit.<path>"` + `task_default` — probe tasks parameterize across agent specialisms.

**Auth value strategies**
- `auth: { type: api_key, value_strategy: 'random_invalid' }` → `invalid-<32 hex bytes>`, fresh per run.
- `auth: { type: oauth_bearer, value_strategy: 'random_invalid_jwt' }` → three base64url segments with valid JSON header/payload and random signature. Well-implemented validators fail at signature verification (→401); strict parse-time validators 400. Both conformant per RFC 6750 §3.1.
- `auth: { type: oauth_bearer, value: '<literal>' }` also works for explicit test cases.

**Phase `optional: true`**
- Failing steps in optional phases are recorded in the step-level report but do not fail overall. The storyboard's `assert_contribution` is the gate.

**Auth-override dispatch**
- When `step.auth` is set, the runner bypasses the MCP SDK and does a raw JSON-RPC POST via `rawMcpProbe`. Required because the SDK can't strip credentials or send arbitrary Bearer values, and validations need the raw HTTP status + `WWW-Authenticate` header. A synthetic `TaskResult` is built from the JSON-RPC body for `field_present` / `field_value` compatibility. HTTP-level 4xx/5xx pins `taskResult.success=false` regardless of envelope shape.

**`--allow-http` threads into probes**
- Previously `--allow-http` only relaxed the pre-flight URL check in `comply()`. Now also relaxes the SSRF guard on `protected_resource_metadata` and `oauth_auth_server_metadata` probes so localhost dev loops work end to end.

Tests: 31 in `test/lib/storyboard-security.test.js` (was 22), including real-HTTP coverage of $test_kit.* substitution, value_strategy injection, auth: none, and phase `optional: true` pass/fail semantics. Full suite 3336/3338 (2 intentional skips, 0 failures).

bokelley:
**End-to-end demo complete.** Ran the security_baseline storyboard (fetched directly from adcontextprotocol/adcp@main) through three local demo agents to validate the full loop.

| Case | OAuth metadata | API key path | \`resource_equals_agent_url\` | Overall |
|---|---|---|---|---|
| A. correct + working key | ✓ | ✓ | ✓ | **PASS** |
| B. broken + working key (#563 repro) | ✗ (caught) | ✓ | ✗ | **PASS** (API key is sufficient) |
| C. broken + no test-kit key | ✗ (caught) | skipped | ✗ | **FAIL** (no mechanism verified) |

Case B is the money shot: the runner catches the RFC 9728 audience-mismatch bug (the same one that bit test-agent.adcontextprotocol.org → #563) but still reports overall pass because the agent has a working API key path. Case C shows that an OAuth-only agent with the same bug correctly fails overall.

**Fix from the demo**: \`field_present\` / \`field_value\` now resolve against either the MCP taskResult data OR the HTTP probe body. Without this, the storyboard's \`probe_protected_resource\` step (which uses \`field_present: path: resource\`) was failing with "requires MCP task result" — not what the upstream storyboard intended. Also exempted the synthetic tasks (\`protected_resource_metadata\`, \`oauth_auth_server_metadata\`, \`assert_contribution\`) and \`$test_kit.*\` task references from the \`storyboard-completeness\` and \`storyboard-drift\` test suites so new storyboards using them don't trip structural checks.

Public \`test-agent.adcontextprotocol.org\` still can't run the storyboard because \`comply()\` short-circuits with \`auth_required\` when capability discovery 401s — the agent wrongly requires auth on \`get_adcp_capabilities\` (public per spec). Probably worth a separate issue: comply() should proceed into universal/security.yaml when capability discovery fails with auth, since security.yaml exists precisely to diagnose this class of agent.

Demo runner lives at \`.context/demo-run.mjs\` (gitignored). \`node .context/demo-run.mjs\` reproduces the three-case table above.

Full suite now 3277/3279 (2 intentional skips, 0 failures).

bokelley:
Expert review follow-ups pushed. Three parallel reviews (code / DX / security) flagged real merge-blockers; all addressed.

**Security hardening (code reviewer #2, security M1/M2/M3)**
- `fetchProbe` resolves every A/AAAA record, rejects if any is private, then **pins the connection to the validated IP** via an undici `Agent` with a custom `connect.lookup` hook. Defeats DNS rebinding (short-TTL authoritative server returning different records to guard vs connect) and multi-record attacks.
- Scheme gate: explicit http/https whitelist; `file:`, `data:`, `ftp:` rejected even under `--allow-http`.
- `isAlwaysBlocked()` keeps IMDS (169.254/16, IPv6 fe80:*) blocked **even under `--allow-http`** — no legitimate dev use for probing cloud-metadata endpoints.
- `isPrivateIp()` expanded: CGNAT (RFC 6598 100.64/10), broadcast, multicast (v4 + v6), and unwraps IPv4-mapped IPv6 / NAT64 / 6to4 before classifying.

**Correctness (code reviewer #1, #5)**
- `rawMcpProbe` no longer advertises `text/event-stream` — SSE-framed responses were slipping past `JSON.parse` and falling through as silent passes on `expect_error` steps. JSON-only Accept forces the server's hand.
- MCP session-not-initialized (-32002) errors surfaced distinctly so strict-server rejects-before-auth aren't mistaken for valid auth probes.
- Non-JSON response bodies now produce a failing taskResult with a clear error instead of silent passthrough.

**Runner invariants (code reviewer #4)**
- `overall_passed` requires at least one non-optional phase with executed steps to have passed. The fragile math (`failedCount===0 && passedCount>0`) was technically correct for the current storyboard's structure but would pass vacuously for storyboards that restructure their gate phase. New invariant is explicit about the gate.

**DX (review #2)**
- `resource_equals_agent_url` failure message now names the agent's own URL + explains the likely fix. Still redacts the advertised value (public-reports hygiene).
- `rawMcpProbe` JSDoc spells out: args MUST NOT contain secrets (server response flows to report) + session-init limitation + expected conformant behavior.

**Tests**: +10 new SSRF/classification tests. Full suite 3328/3330 (2 intentional skips, 0 failures). End-to-end demo still produces the three expected outcomes.

**Deferred to follow-up PRs** (out of scope for this PR, will land separately):
- #570 — comply() should execute universal/security.yaml against agents that 401 on capability discovery. The DX reviewer was right that this is the most impactful gap, but it's a clean separate change.
- Options leak (`_client`/`_profile` on TestOptions) — pre-existing, not introduced here.
- Redact Authorization header from `httpResult.headers` before report serialization — defense in depth, filing as its own issue.

I think this is ship-ready. Happy to run another round if you want a second look.

bokelley:
## Bundled #570 (fix comply() short-circuit on 401 discovery)

Per DX review, splitting the comply() fix from the storyboard runner ships a feature that works in CI but silently no-ops against the exact agents security.yaml is designed to diagnose. Bundled here instead.

**What changed in this commit** (`9afda39`):

- Extracted `detectAuthRejection` from `buildUnreachableResult` so both the "unreachable fallback" and the new "degraded profile" paths share the same truth.
- Case-insensitive keyword match — caught a real failure: the MCP client wraps 401s as `"Authentication required ..."` which the previous lowercase-only heuristic missed.
- Fallback `fetch(agentUrl)` probe now always runs when keyword matching fails, not only when the error contains `"Failed to discover"`.
- New `runWithDegradedProfile` runs storyboards with `track: 'security'` or `required_tools: []` against a 401-on-discovery agent. Other storyboards stay gated (empty `agentTools` would make them meaningless) and `buildUnreachableResult` is the fallback for network errors + no-security-storyboards cases.
- Added 2 tests to `test/lib/storyboard-security.test.js`:
  - 401-on-discovery + `storyboards: ['security_baseline']` → `security_baseline` is in `storyboards_executed`, auth observation present, `overall_status` is not `'auth_required'`.
  - 401-on-discovery + non-security tool-dependent storyboard → `overall_status === 'auth_required'` (fallback invariant).

**Test plan (local)**:
- [x] `test/lib/storyboard-security.test.js` — 42 / 42 pass.
- [x] `test/lib/{compliance,comply-abort,storyboard-*}.test.js` — 1188 / 1188 pass.
- [x] `npx tsc --noEmit` — clean.

bokelley:
## Expert review round (code-reviewer + security-reviewer)

### Security review: no L1 blockers

Architecturally sound — SSRF guardrails, credential containment in serialized reports, token-collision math, redaction of `resource_equals_agent_url` error, and `--allow-http` flag scope all check out.

### Fixed in `c06dc08`

**sec L2-2 — fallback POST in `detectAuthRejection` bypassed SSRF guardrails.** My new auth-detection fallback probe was plain `fetch(agentUrl, { method: 'POST' })` — no `redirect: 'manual'`, no timeout. A hostile agent that fails discovery could respond 302 → `http://169.254.169.254/...` and the fallback would follow it. Now pinned to `redirect: 'manual'` + 5s `AbortSignal.timeout` combined with the outer `signal` via `AbortSignal.any`.

**sec L2-1 — `test_kit.auth.api_key` CRLF/control-char leak.** `authHeadersForStep` passed the raw value into `\`Bearer ${value}\``. If a key contained `\r\n`, undici's header validator would throw and the exception text (including the secret) would land in the serialized compliance report. Added an explicit guard that throws a non-echoing error before the header is built.

**code L1 #2 (latent) — vacuous pass when every required step skips.** `requiredPhasesPassed` checked `p.passed && p.steps.length > 0` — but `p.passed` is vacuously true when all steps skipped. Tightened to require at least one non-skipped passing step per required phase. Not triggered by any currently-selected storyboard, but defensive against future storyboards whose `required_tools` filter strips everything.

### Deferred to follow-ups

- **sec L2-3**: `rawMcpProbe` uses plain `fetch` (no DNS pinning) for auth-override dispatch. Operator URL is trusted so this is lower-risk than the external `fetchProbe`, but routing through the pinned undici `Agent` closes a TOCTOU/rebind gap. File a follow-up.
- **sec L2-4**: `capabilities_probe_error` field contains agent-controlled text that flows into `AdvisoryObservation.message`. Downstream LLM summarizers need to know this is untrusted. File a follow-up to fence / label.
- **code-review L1 #1 was a false positive** — the reviewer read my filter as unparenthesized. The actual code is `(sb.required_tools?.length ?? 0) === 0 || sb.track === 'security'` (line 728), which evaluates correctly. Tests confirm.

### CI

Pre-existing failures on earlier commits were:
- Prettier drift → fixed in `ebcbb4a2`.
- Schema sync racing with upstream `latest.tgz` changes → committed regen in `ebcbb4a2` + `654fec49`.
- Agent docs out of sync → `b181e0ce`.
- `webidl.util.markAsUncloneable` on Node 20.x → downgraded `undici` to `^6` in `3b2aee3d` (v6 supports Node 18.17+, v8 needs 20.18.1+).

Running now on `c06dc08`.

Issue #599, opened 2026-04-19
Conform runner output to adcp runner-output-contract (failure detail, skip reasons, extraction path)


bokelley:
## Summary

Make the storyboard runner and `comply()` output conform to the new **runner-output contract** in the spec: [`static/compliance/source/universal/runner-output-contract.yaml`](https://github.com/adcontextprotocol/adcp/blob/main/static/compliance/source/universal/runner-output-contract.yaml) (introduced in adcontextprotocol/adcp#2364 / adcontextprotocol/adcp#2352).

The contract defines the minimum failure-result shape every AdCP compliance runner MUST emit. Today the runner returns generic messages like `"Check agent capabilities"` with no actionable detail, which is the root complaint in adcontextprotocol/adcp#2352.

## What the contract requires

Per the spec YAML, each **failed** `validation_result` MUST carry:

| Field | Required on failure | Notes |
|---|---|---|
| `check` | always | `response_schema` / `field_present` / `field_value` / etc. |
| `passed` | always | boolean |
| `description` | always | copied from the storyboard validation entry |
| `request` | on failure | exact payload the runner sent |
| `response` | on failure | exact payload the agent returned |
| `json_pointer` | on failure | RFC 6901 pointer to the failing field |
| `expected` | on failure | machine-readable expected value / schema `$id` |
| `actual` | on failure | machine-readable observed value / AJV error array |
| `schema_id` | on `response_schema` | the `$id` actually applied |
| `schema_url` | on `response_schema` | resolvable URL for local re-validation |

Skip results MUST distinguish:
- `not_applicable` — agent didn't claim the protocol.
- `no_phases` — storyboard is a placeholder.
- `prerequisite_failed` — cite the prior step id.
- `missing_tool` — cite the tool name from `required_tools`.
- `missing_test_controller` — for deterministic-testing phases.
- `unsatisfied_contract` — for harness contracts that permit SKIP.

Every step result MUST record the MCP extraction `path` (`structured_content` / `text_fallback` / `error` / `none`) so runner extraction bugs are separable from agent bugs. See [`docs/building/implementation/mcp-response-extraction.mdx`](https://adcontextprotocol.org/docs/building/implementation/mcp-response-extraction) for the normative extraction algorithm.

## Today's gap

From the reporter in adcontextprotocol/adcp#2352:

> ```
> FAILED: schema_validation/capability_discovery
>   - Check agent capabilities
> ```
>
[...]

PR #602, opened 2026-04-19
fix: replayed:false on fresh execs + skill idempotency gaps (#569)

### LINKED ISSUE #569: Server-side idempotency middleware with pluggable storage
Tracking issue for the **server-side** implementation helper for [AdCP #2308](https://github.com/adcontextprotocol/adcp/issues/2308) — idempotency_key is now required on every mutating request. The spec defines the contract; this issue is about giving sellers a drop-in middleware so they don't have to build the storage/atomicity layer themselves.

Client ergonomics tracked separately in #568.

## Why a middleware

Sellers have to implement four things correctly, and the atomicity bit is what trips people up:

1. Extract `idempotency_key` from the request
2. Look up `(principal_id, idempotency_key)` in a store with TTL
3. If hit: compare canonical payload hash → return cached response, OR raise `IDEMPOTENCY_CONFLICT`
4. If miss: execute the handler, then **atomically** commit side effects + cache entry. A crash between \"fire webhook\" and \"cache response\" means the retry fires the webhook again.

A middleware that gets this right once saves every implementer from getting it wrong.

## Proposed API

```ts
import { createIdempotencyStore, canonicalJsonSha256 } from '@adcp/client/server';

const idempotency = createIdempotencyStore({
  backend: redisBackend(redisClient),   // or pgBackend, or memoryBackend for tests
  ttl: 86400,                            // 24h
  hash: canonicalJsonSha256,             // pluggable payload fingerprint
});

// Wrap any mutating handler
export const createMediaBuy = idempotency.wrap(async (req, ctx) => {
  // Your business logic — same transaction as the cache write
  const mediaBuy = await db.insertMediaBuy(req, ctx.tx);
  await fireWebhook(mediaBuy);
  return mediaBuy;
});
```

The wrapper:
- Extracts `idempotency_key` from `req`
- Scopes lookups by `ctx.principal.id` (per-principal namespace — security requirement)
- On hit with matching payload hash: returns cached response, skips handler
- On hit with different hash: throws `IdempotencyConflictError` → mapped to `IDEMPOTENCY_CONFLICT`
- On miss: runs handler inside a transaction, writes `{key, hash, response}` alongside business writes, commits together

## Backends

Ship three implementations, MVP just needs the first two:

- `memoryBackend()` — in-process Map, useful for tests and the reference training agent
- `pgBackend(pool, { schema?, table? })` — Postgres with a dedup table indexed on `(principal_id, key)` with `expires_at`. Matches the transaction guarantees sellers need if their business data is already in Postgres.
- `redisBackend(client)` — Redis with TTL-keyed hashes. Faster, but harder to coordinate transactions across Redis and the business DB. Document the caveat.

## Capabilities integration

One-line opt-in to declare the replay window:

```ts
const idempotency = createIdempotencyStore({ backend, ttl: 86400 });

capabilities.adcp.idempotency = idempotency.capability();
// → { replay_ttl_seconds: 86400 }
```

## Docs

- \"Why idempotency_key and what it costs\" narrative (pointing at [security.mdx](https://github.com/adcontextprotocol/adcp/blob/main/docs/building/implementation/security.mdx#idempotency))
- Snippet for each backend
- Gotchas: the atomicity guarantee across backend choice, per-principal scoping rationale, TTL implications

## Acceptance

- [ ] `createIdempotencyStore` with `memoryBackend` + `pgBackend`
- [ ] `canonicalJsonSha256` hash helper (stable key order, whitespace-normalized)
- [ ] Typed `IdempotencyConflictError` mapped to `IDEMPOTENCY_CONFLICT`
- [ ] `.wrap()` method composes with existing MCP/A2A handler wrapping
- [ ] `.capability()` method returns the capabilities fragment
- [ ] Reference training agent (in this repo or adcp) adopts the middleware as proof of correctness
- [ ] Server-side integration passes the [idempotency compliance storyboard](https://github.com/adcontextprotocol/adcp/blob/main/static/compliance/source/universal/idempotency.yaml) end-to-end

Related: adcontextprotocol/adcp#2308, adcontextprotocol/adcp#2315, #568 (client side)

bokelley:
Closes #569.

### Framework: `replayed: false` on fresh executions

The server-side idempotency middleware was stamping `replayed: true` on replays but omitting the field on fresh executions. The spec (and the universal idempotency compliance storyboard) expects `replayed: false` on every fresh mutating response — a buyer that checks `metadata.replayed` to decide whether side effects already fired needs the field present either way.

Caught by running `storyboard run ... idempotency` against a fresh retail-media build: `replay_same_payload/create_media_buy_initial` failed because `replayed` was absent. After the fix it passes cleanly.

The cache still stores the envelope WITHOUT the `replayed` field; the middleware stamps it after the save so a subsequent replay cleanly overwrites with `true`.

Existing test tightened from `assert.notEqual(result.replayed, true)` to `assert.equal(result.replayed, false)` so future regressions fail loud.

### Skills: 5 SKILL.md files had the wiring gap

Each skill documented idempotency in a dedicated section but the main copy-paste-complete Implementation code block didn't wire `createIdempotencyStore` into `createAdcpServer`. Fresh agents built from those skills logged the v3 non-compliance error at startup.

Fixed in: `build-creative-agent`, `build-signals-agent`, `build-brand-rights-agent`, `build-retail-media-agent`, `build-generative-seller-agent`.

Seller/SI/governance skills were already complete — no changes.

### Test harness: full coverage + `--allow-http`

`test-agents/test-agent-build.sh`:
- Extended from 4 agent types (seller, signals, si, governance) to all 8 (+ creative, brand-rights, retail-media, generative-seller)
- Added the universal `idempotency` storyboard as a second check on every run — catches regressions in any domain
- Passes `--allow-http` so local storyboard runs actually execute (the CLI refuses non-HTTPS URLs by default)

## Validation

Ran `test-agent-build.sh claude <type>` in parallel for all 5 updated skills. All 5 produce idempotency-compliant agents on first try. Build times: 29s–229s.

Built-agent storyboards (with the fix):
- signals × signal_marketplace: ✅ passing
- signals × idempotency: ✅ passing (create_media_buy not implemented, tests skipped)
- creative × idempotency: ✅ passing
- brand-rights × idempotency: ✅ passing
- retail-media × idempotency: 1 pass / 1 partial. `create_media_buy_initial` now ✅; `create_media_buy_replay` still fails because the CLIENT (from PR #590) auto-injects a fresh idempotency_key per request, defeating the replay test. Harness-level issue.
- generative-seller × idempotency: same partial as retail-media (same root cause)

`node --test test/server-idempotency.test.js test/lib/idempotency*.test.js test/lib/jcs.test.js test/skill-examples-compile.test.js` — 89/89 pass.

## Follow-up issues worth filing

1. **Client auto-inject defeats storyboard missing_key test.** The client's TaskExecutor auto-injects idempotency_key on mutating requests. The storyboard's `missing_key` test can't actually send a missing key through the client. Either the harness needs an opt-out or the test needs a flag.
2. **Alias placeholder resolution inconsistent across scenarios.** The storyboard's `$generate:uuid_v4#replay_key` placeholder sometimes resolves to a fresh UUID per step (defeating replay) and sometimes passes through as a literal `$generate:...` string (tripping server pattern validation).

Neither is in scope for #569; both are compliance-harness/client issues filed separately once I've repro'd them in isolation.

## Test plan

- [x] `npm run build:lib` — clean compile
- [x] `node --test test/server-idempotency.test.js` — 19/19 pass
- [x] `node --test test/lib/idempotency*.test.js test/lib/jcs.test.js` — all pass
- [x] `node --test test/skill-examples-compile.test.js` — 19/19 skill TS blocks compile
- [x] `test-agent-build.sh claude <type>` for all 5 updated skills — compile clean, wire idempotency correctly
- [x] Retail-media `create_media_buy_initial` passes idempotency storyboard (was failing)

### DIFF
--- a/.changeset/skills-idempotency-coverage.md
+++ b/.changeset/skills-idempotency-coverage.md
@@ -1,0 +1,24 @@
+---
+'@adcp/client': patch
+---
+
+Idempotency middleware now stamps `replayed: false` on fresh executions (not
+just `replayed: true` on replays). Caught by running the universal
+`idempotency` compliance storyboard against a retail-media agent built from
+the updated skill — the `replay_same_payload/create_media_buy_initial` step
+expects `replayed: false` on the envelope, and the previous implementation
+omitted the field entirely. The cache still stores the envelope WITHOUT the
+`replayed` field; it's stamped after the save so a subsequent replay cleanly
+overwrites it with `true`. Existing test asserting `replayed !== true` on
+fresh executions tightened to `replayed === false`.
+
+Skills: wire `createIdempotencyStore` into the main Implementation code block
+for `build-creative-agent`, `build-signals-agent`, `build-brand-rights-agent`,
+`build-retail-media-agent`, and `build-generative-seller-agent`. Each skill
+already documented idempotency but the copy-paste-complete Implementation
+example was missing the wiring, so a fresh agent built from the skill would
+log the "v3 non-compliance" error at startup. Seller, SI, and governance
+skills were already complete. Also extends `test-agents/test-agent-build.sh`
+to cover all 8 agent types, adds the universal `idempotency` storyboard as
+a second check on every run, and passes `--allow-http` so local storyboard
+runs actually execute.
--- a/.changeset/skills-idempotency-coverage.md
+++ b/.changeset/skills-idempotency-coverage.md
@@ -2,23 +2,50 @@
 '@adcp/client': patch
 ---
 
-Idempotency middleware now stamps `replayed: false` on fresh executions (not
-just `replayed: true` on replays). Caught by running the universal
-`idempotency` compliance storyboard against a retail-media agent built from
-the updated skill — the `replay_same_payload/create_media_buy_initial` step
-expects `replayed: false` on the envelope, and the previous implementation
-omitted the field entirely. The cache still stores the envelope WITHOUT the
-`replayed` field; it's stamped after the save so a subsequent replay cleanly
-overwrites it with `true`. Existing test asserting `replayed !== true` on
-fresh executions tightened to `replayed === false`.
+**Idempotency storyboard end-to-end compliance** — the universal
+`idempotency` compliance storyboard now passes 1/0/0/0 against agents
+built from all 8 skills. Three framework fixes were required to get there:
 
-Skills: wire `createIdempotencyStore` into the main Implementation code block
-for `build-creative-agent`, `build-signals-agent`, `build-brand-rights-agent`,
-`build-retail-media-agent`, and `build-generative-seller-agent`. Each skill
-already documented idempotency but the copy-paste-complete Implementation
-example was missing the wiring, so a fresh agent built from the skill would
-log the "v3 non-compliance" error at startup. Seller, SI, and governance
-skills were already complete. Also extends `test-agents/test-agent-build.sh`
-to cover all 8 agent types, adds the universal `idempotency` storyboard as
-a second check on every run, and passes `--allow-http` so local storyboard
-runs actually execute.
+1. **`replayed: false` on fresh executions.** Middleware now stamps the
+   field on every mutating response, not just replays. Buyers that branch
+   on `metadata.replayed` to decide whether side effects already fired
+   need the field present either way. Cached envelopes are stamped after
+   the save so replays cleanly overwrite with `true`.
+
+2. **Replay echoes the current retry's `context`, not the first caller's.**
+   Previously, cached response envelopes baked in the first caller's
+   `correlation_id` and replays returned that to every subsequent retry,
+   breaking end-to-end tracing. The middleware now strips `context` from
+   the formatted response before caching; on replay, `finalize()`
+   re-injects the current request's context.
+
+3. **MCP-level `idempotency_key` relaxed to optional when the framework
+   has an idempotency store wired.** The middleware is authoritative for
+   this field and returns a structured `adcp_error` with `code` + `field`.
+   If the MCP SDK's schema validator rejected first, buyers got a text-
+   only `-32602` error that failed the storyboard's `error_code` check.
+
+**Storyboard harness fixes** so the runner actually exercises replay semantics:
+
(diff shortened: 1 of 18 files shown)


bokelley:
Dropping observations from the runner-output review thread on adcontextprotocol/adcp#2352 so the context isn't lost.

**This PR resolves both of its stated follow-up issues internally** — worth updating the "Follow-up issues worth filing" section to say so:

1. *"Client auto-inject defeats storyboard missing_key test"* → solved by the new `skipIdempotencyAutoInject` flag with the narrow trigger `expect_error === true && isMutatingTask && no idempotency_key`. That's cleaner than a global opt-out; the missing-key storyboard step is now the only thing that suppresses auto-inject.
2. *"Alias placeholder resolution inconsistent"* → solved by `forwardAliasCache()` in `context.ts`. `$generate:uuid_v4#<alias>` now resolves consistently across steps in the same run.

**Relationship to adcontextprotocol/adcp-client#599 (runner-output conformance).** That issue is *not* resolved by this PR — different layer. #599 is about the *failure result shape* the runner emits (exact request/response, JSON Pointer to the failing field, structured expected/actual, schema `\$id`). But the request-builder changes here are a prerequisite: now that `idempotency_key` is forwarded through from `sample_request` instead of being silently replaced by the client's auto-inject, capturing "the exact request the runner sent" in a failure result will actually reflect what went on the wire. Without this PR, a failure report claiming `idempotency_key: "<uuid>"` could be misleading. Worth noting in the #599 conversation so whoever picks it up layers on top of this.

**One upstream spec gap this PR surfaces.** Storyboard `sample_request` blocks in the spec repo that *should* declare `idempotency_key` currently don't — e.g. `specialisms/signal-owned/index.yaml` activate steps, and the signals protocol baseline PR I have open at adcontextprotocol/adcp#2365. Given this PR's idempotency_key forward-through now makes storyboard-declared keys authoritative, I'll update #2365 to include `idempotency_key: "\$generate:uuid_v4#activate_key"` and file a separate issue for the signal-owned omission so they can be fixed together.

LGTM on the substance — nothing to change here. Just linking the threads.

PR #629, opened 2026-04-19
feat(webhooks): receiver-side idempotency_key dedup on AsyncHandler

bokelley:
## Summary

Closes the receiver-side half of AdCP [#2417](https://github.com/adcontextprotocol/adcp/pull/2417) in the client library. Every MCP webhook payload now carries a required `idempotency_key`; this PR plumbs it through the pipeline and adds a drop-in dedup layer so at-least-once retries no longer trigger duplicate side effects in handler code.

## What's new

- **`AsyncHandlerConfig.webhookDedup?: { backend, ttlSeconds? }`** — reuses `IdempotencyBackend` from `@adcp/client/server`, so the same `memoryBackend()` / `pgBackend(...)` used for request-side idempotency can back webhook dedup. Defaults to 24h.
- **`WebhookMetadata.idempotency_key?: string`** — extracted from the MCP envelope and passed to every `onXxxStatusChange` handler so application code can log, trace, or layer its own dedup.
- **`Activity` union gains `'webhook_duplicate'`** — emitted via `onActivity` when a repeat key is dropped. Typed handlers are NOT called for duplicates (no side effects).

## Behavior

- Scope is **per-agent** (`webhook\u001f{agent_id}\u001f{idempotency_key}`) — matches the spec's "scoped to authenticated sender identity" rule, keys from different senders never collide.
- `putIfAbsent` closes the concurrent-retry race: when two retries race on the same fresh key, exactly one wins the claim and dispatches; the rest surface as `webhook_duplicate`.
- Payloads without `idempotency_key` (non-conforming senders, or A2A transport which doesn't carry the field) dispatch without dedup and log a `console.warn`. No silent drops, no tuple fallback.

## Schema sync

Pulled from AdCP `latest` — `idempotency_key` now required on:
- `MCPWebhookPayload`
- `CollectionListChangedWebhook`
- `PropertyListChangedWebhook`
- `ArtifactWebhookPayload`
- `RevocationNotification` (brand rights — not previously flagged in #2417)

## Usage

```typescript
import { AdCPClient } from '@adcp/client';
import { memoryBackend } from '@adcp/client/server';

const client = new AdCPClient(agents, {
  webhookUrlTemplate: 'https://your-app.com/adcp/webhook/{task_type}/{agent_id}/{operation_id}',
  webhookSecret: process.env.WEBHOOK_SECRET,
  handlers: {
    webhookDedup: { backend: memoryBackend() },
    onCreateMediaBuyStatusChange: async (result, metadata) => {
      // First delivery runs here; publisher retries are dropped.
    },
  },
});
```

## Out of scope (follow-ups)

- **Governance list-change / artifact / revocation webhooks** — not currently routed through `AsyncHandler`; dedup for those payload types is a follow-up.
- **Sender-side webhook emission helper** — `@adcp/client` doesn't currently emit webhooks for sellers. When added, it will need to generate one `idempotency_key` per event and reuse across retries.
- **RFC 9421 webhook signing** — AdCP [#2423](https://github.com/adcontextprotocol/adcp/pull/2423) landed upstream; client implementation will follow in a separate PR.

## Changeset

`minor` — new optional config + new metadata field + `Activity` union gains a new string literal. See `.changeset/webhook-receiver-dedup.md`.

## Test plan

- [x] `npm test` — 4067 pass / 0 fail
- [x] 7 new tests in `test/lib/async-handler-webhook-dedup.test.js`:
  - duplicate delivery dropped
  - distinct keys dispatched independently
  - per-agent scoping prevents cross-sender collisions
  - missing key warns + dispatches
  - back-compat when `webhookDedup` unset
  - TTL expiry re-dispatches
  - end-to-end: `idempotency_key` propagates from MCP envelope → `WebhookMetadata` → handler

### DIFF
--- a/.changeset/webhook-receiver-dedup.md
+++ b/.changeset/webhook-receiver-dedup.md
@@ -1,0 +1,50 @@
+---
+'@adcp/client': minor
+---
+
+Webhook receiver-side deduplication via `AsyncHandlerConfig.webhookDedup`.
+
+AdCP webhooks use at-least-once delivery — publishers retry until they see a 2xx, so the same event can arrive more than once. The spec now requires an `idempotency_key` on every MCP, governance, artifact, and revocation webhook payload so receivers have a canonical dedup field. This release plumbs that key through the client pipeline and ships a drop-in dedup layer for the MCP envelope path.
+
+**New**
+
+- `AsyncHandlerConfig.webhookDedup?: { backend: IdempotencyBackend; ttlSeconds?: number }` — drop duplicate deliveries with a single config. Reuses `IdempotencyBackend` from `@adcp/client/server`, so the same `memoryBackend()` or `pgBackend(...)` used for request-side idempotency can back webhook dedup. Defaults to 24h retention.
+- `WebhookMetadata.idempotency_key?: string` — extracted from the MCP envelope and passed to every `onXxxStatusChange` handler so application code can log, trace, or build its own dedup on top.
+- `WebhookMetadata.protocol?: 'mcp' | 'a2a'` — transport that delivered the webhook; useful for handler code that branches on protocol (A2A lacks `idempotency_key`).
+- `Activity` union gains `'webhook_duplicate'` — surfaced via `onActivity` when a repeat key is dropped. The typed handler is NOT called for duplicates.
+- `Activity.idempotency_key?: string` — surfaced on both `webhook_received` and `webhook_duplicate` for correlation.
+
+**Type changes (strict-TS callers may need to update)**
+
+- The `Activity.type` union gains `'webhook_duplicate'`. TypeScript users doing exhaustive `switch (activity.type)` with a `never`-check will see a new missing-case error. Treat `webhook_duplicate` the same as `webhook_received` in `onActivity` logging, or branch on `activity.type` to suppress side effects for duplicates.
+
+**Behavior**
+
+- Scope is per-agent under a reserved prefix (`adcp\u001fwebhook\u001fv1\u001f{agent_id}\u001f{idempotency_key}`) — keys from different senders are independent, and the prefix guarantees no collision with request-side idempotency entries when sharing a backend.
+- `putIfAbsent` closes the concurrent-retry race: when two retries race on the same fresh key, exactly one wins the claim and dispatches; the rest surface as `webhook_duplicate`.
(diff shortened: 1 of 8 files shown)


bokelley:
Addressed review feedback in 3913076f. Summary:

**Accepted**
- **sec T7** — `webhook_duplicate` activity now omits `payload`; both events carry `idempotency_key` for correlation
- **sec T2+T6** — receiver validates the spec regex `^[A-Za-z0-9_.:-]{16,255}$`; invalid/empty/too-short keys treated as missing (no scoped-key construction from arbitrary sender bytes)
- **code #1** — reserved scoped-key prefix `adcp\u001fwebhook\u001fv1\u001f...` so backend sharing with request-side idempotency is collision-proof
- **code #4** — `WebhookMetadata.protocol` plumbed through; missing-key warn only fires for MCP (A2A lacks the field by protocol design)
- **dx #3+#4** — warn message now names the spec regex, describes fallback behavior, links to docs; JSDoc on `handlers` field points at dedup section
- **dx #5+#7** — docs cover activity double-fire, migration-from-ad-hoc dedup, A2A/invalid-key behavior
- **code #6** — changeset has explicit "Type changes" subsection for strict-TS exhaustive `Activity.type` switches
- **code #8a+c** — concurrent-race test (5 parallel retries → 1 dispatch) and handler-exception test (claim sticks)
- **code #7** — grepped; no downstream references to `RevocationNotification.notification_id`

**Rejected with rationale**
- **code #2 (release claim on handler error)** — incoherent with current behavior. `handleCompletion` catches handler exceptions and returns normally, so `handleWebhook` already 200s the publisher on handler failure. Releasing the claim would trigger infinite retry storms against deterministic handler bugs. Added a test documenting this at-most-once contract rather than changing it.
- **dx #1 (`webhookDedup: true` shortcut)** — would force `core/` to runtime-import `server/`, inverting module layering. Explicit `{ backend: memoryBackend() }` is 2 lines and a clear mental model.
- **dx #2 (re-export `memoryBackend` from root)** — fragments module layout.
- **sec T3 (memoryBackend LRU cap)** — pre-existing backend behavior, not introduced by this PR. Filing separately.
- **dx #5 (`onMissingKey` throw option)** — YAGNI; no user has asked for strict mode.

Test suite: 4084 pass / 0 fail. New tests bring dedup coverage to 12 cases.

PR #642, opened 2026-04-20
feat(cli): --brand / --brand-manifest for storyboard run

### LINKED ISSUE #639: storyboard CLI: expose --brand so applyBrandInvariant fires from `adcp storyboard run`
## Summary

`applyBrandInvariant` in `src/lib/testing/storyboard/runner.ts` (added in #586) is a correctness safety net: when `options.brand` is set, every outgoing step's request gets `brand` overwritten to a single, storyboard-run-scoped value so a buyer can't accidentally land in two sessions on a brand-scoped seller. It guards:

```ts
if (!options.brand && !options.brand_manifest) return request;
```

The CLI (`bin/adcp.js` → `storyboard run`) never sets `options.brand`, so the invariant is a no-op when invoked from the command line. All authoring mistakes like [adcp#2526](https://github.com/adcontextprotocol/adcp/pull/2526) — storyboard step omits brand, so it lands in `open:default` instead of the tenant it should — slip through any CLI-driven run.

## Proposed fix

1. Accept `--brand <domain>` (or `--brand '{"domain":"...","agent_url":"..."}'` for full `BrandReference` JSON) on `adcp storyboard run`.
2. Accept `--brand-manifest <path|url>` for the manifest variant (parity with the runner field).
3. Thread the resolved brand into `TestOptions.brand` on the run-configuration path so `applyBrandInvariant` activates.
4. Document in `storyboard run --help`:

```
  --brand DOMAIN|JSON      Force every step onto this brand for per-tenant session isolation
                           (matches sellers that scope state by brand.domain; same behavior as
                           setting options.brand programmatically).
  --brand-manifest PATH    Brand manifest URL / local file; mutually exclusive with --brand.
```

## Tangentially related

While looking at this, also noticed the `--file <path>` CLI parsing bug tracked in #637 — should be fixed in the same pass since the file-path arg needs to be registered in `flagValues` the same way `--brand` will.

## Why now

Without this, any future storyboard author who forgets to include `brand` on a sub-step will re-introduce the cross-session bug that #2526 just fixed — and the runner's safety net won't catch it from the CLI path. Belt-and-suspenders: storyboard lint ([adcp#<this-followup-issue>](https://github.com/adcontextprotocol/adcp/issues)) at author time, `--brand` passthrough at run time.

## Acceptance

- [ ] `adcp storyboard run test-mcp --brand acmeoutdoor.example media_buy_seller/inventory_list_targeting` passes against the test agent even if a step's `sample_request.brand` is missing.
- [ ] `--brand` and `--brand-manifest` are documented in `storyboard run --help`.
- [ ] Mutually exclusive enforcement between the two flags.
- [ ] Covered by a unit test in `test/lib/storyboard-brand-invariant.test.js` (follows PR #586's pattern) asserting `options.brand` is populated from the CLI flag.

Filed from adcontextprotocol/adcp#2526 as non-blocking follow-up.

bokelley:
## Summary
- Fixes #639 — `applyBrandInvariant` (PR #586) was a no-op for CLI-driven storyboard runs because `options.brand` / `options.brand_manifest` was never threaded. Any step that omitted `brand` on its `sample_request` would fall into the seller's `open:default` session instead of the tenant under test — exactly the cross-session bug adcontextprotocol/adcp#2526 just fixed.
- Accepts `--brand DOMAIN|JSON` (bare domain, inline `BrandReference` JSON, or `@file.json`) and `--brand-manifest JSON|@file.json`. Mutually exclusive.
- Threaded into `TestOptions.brand` / `TestOptions.brand_manifest` on all three paths: single-instance `storyboard run`, multi-instance `storyboard run`, and capability-driven full assessment.
- Also threads `--allow-http` in single-instance storyboard run for parity with the other two paths (closes a gap I noticed while writing the test).

## Implementation notes
- `bin/adcp.js` now exports `parseAgentOptions` / `parseBrandFlag` when `require()`d (guarded by `require.main === module`) so unit tests exercise the parser directly instead of spawning the CLI against a fake MCP server. No new files — same binary.
- Stacked on #640 (`bokelley/issue-637`) because the issue explicitly noted the `--file` parsing bug needs to land alongside the `--brand` work. PR is based on that branch; will retarget to `main` once #640 merges.

## Test plan
- [x] `test/lib/cli-storyboard-brand-flag.test.js` — 11 tests covering `parseBrandFlag` forms, `parseAgentOptions` threading, positional-arg non-collision, mutual-exclusion error path, invalid-JSON rejection.
- [x] `npm test` — full suite passes (4275 pass, 12 skipped, 0 fail).
- [x] Manual: `adcp storyboard run test-mcp --file scenario.yaml --brand acme.example --dry-run` loads YAML and runs; `--brand + --brand-manifest` emits the mutual-exclusion error; `--brand '{"brand_id":"x"}'` rejects JSON missing `domain`.

## Acceptance criteria (from the issue)
- [x] `adcp storyboard run <agent> --brand acmeoutdoor.example <id>` passes the brand through even when sub-step `sample_request.brand` is missing.
- [x] `--brand` / `--brand-manifest` documented in `storyboard run --help`.
- [x] Mutual-exclusion enforcement between the two flags.
- [x] Unit test asserting `options.brand` is populated from the CLI flag.

### DIFF
--- a/.changeset/storyboard-run-brand-flag.md
+++ b/.changeset/storyboard-run-brand-flag.md
@@ -1,0 +1,5 @@
+---
+'@adcp/client': minor
+---
+
+Expose `--brand DOMAIN|JSON` and `--brand-manifest JSON|@file.json` on `adcp storyboard run` (single-instance, multi-instance, and full capability-driven assessment). The runner's `applyBrandInvariant` was previously a no-op for CLI-driven runs because `options.brand` was never threaded — any storyboard step that omitted `brand` on its sample_request could slip into the seller's `open:default` session instead of the tenant under test. `--brand` and `--brand-manifest` are mutually exclusive. Also threads `--allow-http` in single-instance storyboard run for parity with multi-instance and full assessment. (adcp-client#639)
--- a/bin/adcp.js
+++ b/bin/adcp.js
@@ -609,6 +609,34 @@ function parseAgentOptions(args) {
     if (eqArg) file = eqArg.slice('--file='.length);
   }
 
+  // --brand DOMAIN|JSON and --brand-manifest JSON|@file.json — threaded into
+  // TestOptions.brand / .brand_manifest so the runner's applyBrandInvariant
+  // fires for CLI-driven runs (otherwise storyboards that omit brand on a
+  // sub-step land in the seller's `open:default` session instead of the
+  // tenant under test). Mutually exclusive.
+  const brandIndex = args.indexOf('--brand');
+  let brandRawValue = null;
+  if (brandIndex !== -1 && brandIndex + 1 < args.length && !args[brandIndex + 1].startsWith('--')) {
+    brandRawValue = args[brandIndex + 1];
+  }
+  const brand = brandRawValue != null ? parseBrandFlag(brandRawValue) : null;
+
+  const brandManifestIndex = args.indexOf('--brand-manifest');
+  let brandManifestRaw = null;
+  if (
+    brandManifestIndex !== -1 &&
+    brandManifestIndex + 1 < args.length &&
+    !args[brandManifestIndex + 1].startsWith('--')
+  ) {
+    brandManifestRaw = args[brandManifestIndex + 1];
+  }
+  const brandManifest = brandManifestRaw != null ? parseJsonFlag('--brand-manifest', brandManifestRaw) : null;
+
+  if (brand && brandManifest) {
+    console.error('ERROR: --brand and --brand-manifest are mutually exclusive. Use one or the other.');
+    process.exit(2);
+  }
+
   const jsonOutput = args.includes('--json');
   const debug = args.includes('--debug') || process.env.ADCP_DEBUG === 'true';
   const dryRun = args.includes('--dry-run');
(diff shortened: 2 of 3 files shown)


bokelley:
Closing in favor of deciding the upstream primitive first.

Stepping back, the ergonomic win this PR advertises ("safety net for authoring mistakes") is backwards: if a storyboard step omits `brand` and the runner silently heals it, the authoring bug is hidden — exactly the class the storyboard lint (adcontextprotocol/adcp#2527) was filed to prevent. Shipping `--brand` as safety actively competes with the lint.

The only legitimate use case for a CLI `--brand` is tenant-routing — running one ad-hoc `--file foo.yaml` against several tenants without forking the YAML. Narrow, probably not what most CLI users hit day-to-day. And as long as `applyBrandInvariant` silently rewrites, threading `--brand` into `options.brand` inherits that behavior — the flag can't cleanly be "tenant routing only" while the primitive underneath is a silent rewriter.

Right order: settle the primitive first (warn vs overwrite vs remove). If `applyBrandInvariant` goes warn-mode, `--brand` becomes safe to reintroduce as pure tenant routing. If it gets removed, `--brand` was never needed.

Follow-up issue filed for the `applyBrandInvariant` semantics question. The `--file` parsing fix is unrelated and stands on its own in #640.

PR #647, opened 2026-04-20
fix(parser): disambiguate MCP task status from domain status in getStatus (#646)

### LINKED ISSUE #646: ProtocolResponseParser.getStatus() collides MCP task status with domain status enums (MediaBuyStatus)
## Summary

`ProtocolResponseParser.getStatus()` classifies an MCP response as a **task-status** response whenever `structuredContent.status` matches any value in the `ADCP_STATUS` enum. Several `ADCP_STATUS` literals (`completed`, `canceled`, `failed`, `rejected`) collide with **domain** status literals defined by AdCP v3 schemas (e.g. `MediaBuyStatus`). When a server returns a perfectly spec-compliant domain envelope whose `status` happens to be `"canceled"` or `"completed"`, the client misinterprets it as an MCP task state and routes it through `TaskExecutor`'s terminal-failure / terminal-success branches — bypassing the Zod validators and often returning `data: undefined` with `success: false`.

## Affected code

- `src/lib/core/ProtocolResponseParser.ts` — `getStatus()`, the `structuredContent.status` branch (currently around line 91, compiled `dist/.../ProtocolResponseParser.js:79`).
- `src/lib/core/TaskExecutor.ts` — the `case ADCP_STATUS.FAILED / REJECTED / CANCELED` short-circuit (compiled `dist/.../TaskExecutor.js:319`). When `hasStructuredError` is false (which is the case for spec-compliant non-error domain envelopes), it returns:
  ```
  { success: false, data: undefined, error: 'Task canceled', ... }
  ```

## Repro

Against any AdCP v3 seller that implements `cancel_media_buy` per spec:

```js
import { ADCPClient } from '@adcp/client';
const client = new ADCPClient({ agent: seller });
const res = await client.callTool('cancel_media_buy', {
  media_buy_id: 'mb_abc',
  reason: 'buyer_cancel',
});
// Server returns a correct AdCP v3 envelope:
//   structuredContent: {
//     status: 'canceled',              // MediaBuyStatus per AdCP v3 schema
//     media_buy: { media_buy_id: 'mb_abc', status: 'canceled', ... },
//     adcp_version: '3.0.0'
//   }
// Expected: res.success === true, res.data.media_buy populated
// Actual:   res.success === false, res.data === undefined, res.error === 'Task canceled'
```

Triggered in practice by the official `media_buy_state_machine` storyboard (`cancel_buy` step, and chain-effect `resume_canceled_buy`).

## Root cause

The enum literals `completed` / `canceled` / `failed` / `rejected` are overloaded: they mean one thing when they appear as the MCP task status for an async MCP task envelope, and a different thing when they appear as the `status` field of a returned domain object (e.g. `MediaBuy`, `Deal`, etc.). `getStatus()` cannot distinguish the two from the string alone.

## Proposed fix

In the `structuredContent.status` branch of `getStatus()`, only treat the status as an ADCP task status when the shape of the envelope looks like an MCP **task** wrapper — i.e. when no additional domain payload is present alongside `status`.

Concretely: if `structuredContent` contains any domain-payload keys beyond the task-envelope allowlist (`status`, `message`, `messages`, `errors`, `warnings`, `adcp_version`, `context_id`, `task_id`, `task_status`, `replayed`), fall through so the response is classified as `COMPLETED` and Zod validators run on the domain payload. Shared-literal values (`completed`/`canceled`/`failed`/`rejected`) should never be treated as an ADCP task state via `structuredContent.status` alone, because they are ambiguous with domain enums.

This preserves the existing fast-paths for `submitted` / `working` / `input-required` / `auth-required` (which are exclusive to the task lifecycle and never appear in domain enums) and does not change behavior for non-MCP transports (A2A wrapped and top-level `status` remain in an MCP task context by construction).

## Impact

- Any AdCP v3 seller whose domain resources have a `status` field sharing a literal with `ADCP_STATUS`. In the current schema that's `MediaBuy` (status literals include `completed`, `cancelled`/`canceled` depending on revision).
- Blocks the `media_buy_state_machine` storyboard on `cancel_buy` and `resume_canceled_buy`, regardless of server correctness.

Happy to send a PR.

fgranata:
## Summary

Fixes #646.

`ProtocolResponseParser.getStatus()` classified any `structuredContent.status` matching an `ADCP_STATUS` literal as an MCP task state. Several of those literals (`completed` / `canceled` / `failed` / `rejected`) also appear as values in AdCP v3 domain status enums (e.g. `MediaBuyStatus`). A spec-compliant domain envelope like `{ status: 'canceled', media_buy: {...} }` was misrouted through `TaskExecutor`'s terminal-state branches, which return `data: undefined` and skip Zod validation.

## Fix

In the `structuredContent.status` branch of `getStatus()`:

- Only treat the value as an ADCP task status when either
  - (a) the envelope contains no domain-payload keys (i.e. only task-envelope keys: `status`, `message`, `messages`, `errors`, `warnings`, `adcp_version`, `context_id`, `task_id`, `task_status`, `replayed`), **or**
  - (b) the value is an unambiguous task-only lifecycle state (`submitted`, `working`, `input-required`, `auth-required`).
- Shared literals (`completed`/`canceled`/`failed`/`rejected`) alongside a domain payload fall through to the existing "has `structuredContent` → `COMPLETED`" branch so Zod validators run on the payload.

This preserves behaviour for bare task envelopes, A2A-wrapped responses, top-level `status`, `isError` errors, and all task-only lifecycle states.

## Test plan

- [x] New test file `test/lib/protocol-response-parser-status.test.js` — 13 cases covering the collision (canceled/completed/rejected with a domain payload → COMPLETED), bare task envelopes (→ CANCELED/COMPLETED), task-only states with a domain payload (working/submitted/input-required preserved), and the A2A-wrapped / top-level / isError / bare-content / null paths.
- [x] Full suite (`CI=1 npm test`): 2251 pass / 27 fail / 19 cancelled before and after the change — identical counts, and the 27 failures are pre-existing environmental issues (missing `compliance/cache/latest/test-vectors/request-signing/keys.json` fixtures not present in the repo). No regressions introduced by this change.

## Impact

Unblocks the `media_buy_state_machine` storyboard (`cancel_buy` and the chain-effect `resume_canceled_buy`) against any AdCP v3 seller that correctly returns `MediaBuyStatus` values in the envelope `status` field.

### DIFF
--- a/src/lib/core/ProtocolResponseParser.ts
+++ b/src/lib/core/ProtocolResponseParser.ts
@@ -87,9 +87,53 @@ export class ProtocolResponseParser {
       return response.status as ADCPStatus;
     }
 
-    // Check MCP structuredContent.status
-    if (response?.structuredContent?.status && Object.values(ADCP_STATUS).includes(response.structuredContent.status)) {
-      return response.structuredContent.status as ADCPStatus;
+    // Check MCP structuredContent.status.
+    //
+    // Several ADCP_STATUS literals (completed, canceled, failed, rejected) also
+    // appear as values in domain status enums (e.g. MediaBuyStatus). When a
+    // seller returns a spec-compliant domain envelope like
+    //   { status: "canceled", media_buy: {...}, adcp_version: "3.0.0" }
+    // we must NOT classify that as an MCP task status, or TaskExecutor's
+    // terminal-state branches short-circuit the response with data:undefined
+    // and skip Zod validation. See issue #646.
+    //
+    // Heuristic: only treat structuredContent.status as an ADCP task status
+    // when the envelope looks like a plain task wrapper — i.e. contains only
+    // task-envelope keys. Any domain payload key present alongside `status`
+    // means this is a domain response; fall through so it's classified as
+    // COMPLETED and validators run on the payload.
+    const sc = response?.structuredContent;
+    if (sc?.status && Object.values(ADCP_STATUS).includes(sc.status)) {
+      const TASK_ENVELOPE_KEYS = new Set([
+        'status',
+        'message',
+        'messages',
+        'errors',
+        'warnings',
+        'adcp_version',
+        'context_id',
+        'task_id',
+        'task_status',
+        'replayed',
+      ]);
+      const hasDomainPayload = Object.keys(sc).some((k) => !TASK_ENVELOPE_KEYS.has(k));
+      if (!hasDomainPayload) {
+        return sc.status as ADCPStatus;
+      }
+      // Exception: preserve the task-lifecycle states that are never domain
+      // enum values, even when a domain payload is attached (e.g. a server
+      // returning partial data while still working).
+      const TASK_ONLY_STATES: string[] = [
+        ADCP_STATUS.SUBMITTED,
+        ADCP_STATUS.WORKING,
+        ADCP_STATUS.INPUT_REQUIRED,
+        ADCP_STATUS.AUTH_REQUIRED,
+      ];
+      if (TASK_ONLY_STATES.includes(sc.status)) {
+        return sc.status as ADCPStatus;
+      }
+      // Shared literal (completed/canceled/failed/rejected) + domain payload
+      // → fall through to the "has structuredContent → COMPLETED" branch.
     }
 
     // Check for MCP error responses
--- a/test/lib/protocol-response-parser-status.test.js
+++ b/test/lib/protocol-response-parser-status.test.js
@@ -1,0 +1,133 @@
+// Regression tests for ProtocolResponseParser.getStatus() — see issue #646.
+//
+// ADCP_STATUS (MCP task lifecycle) shares literals with AdCP v3 domain status
+// enums (e.g. MediaBuyStatus: "completed", "canceled"). The parser must NOT
+// misclassify a domain response whose `status` happens to be a shared literal
+// as an MCP task state, because TaskExecutor's terminal-state branches skip
+// Zod validation and return data:undefined.
+
+const { test, describe } = require('node:test');
+const assert = require('node:assert');
+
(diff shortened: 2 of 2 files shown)


fgranata:
I have read the IPR Policy

bokelley:
oops I saw the issue and fixed it -  too fast! will let you do the next one

bokelley:
Thanks for the clean root-cause and the fix here, @fgranata — this is exactly right, and the test cases you wrote drove the final design.

Closing in favor of #649, which landed as cca6c57a. It builds directly on your approach with a few small tweaks from protocol review against `ProtocolEnvelope` in `src/lib/types/core.generated.ts`:

- **Allowlist derived from `ProtocolEnvelope` + async envelope types** rather than hand-curated. Dropped `messages`, `task_status`, `warnings` (not spec envelope fields at the `structuredContent` root); added `timestamp`, `push_notification_config`, `governance_context`, `context`, `ext` (present on every task envelope type — `CreateMediaBuySubmitted`, `CreateMediaBuyError`, etc.).
- **Named the two constants** (`EXCLUSIVE_TASK_STATUSES`, `TASK_ENVELOPE_FIELDS`) as module-level so the allowlist is a single source of truth instead of inlined in the branch.
- Kept your `submitted`/`working`/`input-required`/`auth-required` fast-path exactly as you had it.
- Tests expanded to 26 cases covering the allowlist keys, `rejected` bare-envelope path, and an end-to-end `TaskExecutor` integration test proving `cancel_media_buy` now returns `{ success: true, data: { media_buy: {...} } }` instead of the old terminal-failure shape.

Credit for report and root-cause goes to you in the release notes. Unblocks `media_buy_state_machine` `cancel_buy` against compliant v3 sellers — thank you.

bokelley:
Superseded by #649 (merged as cca6c57a). See above.

fgranata:
Thanks @bokelley!

Issue #627, opened 2026-04-19
Regen for AdCP 3.0 GA: new `custom` pricing variant + `experimental_features[]` + `per_unit` catchup


bokelley:
AdCP 3.0 GA ships three schema changes affecting generated types. All additive to the wire format, but consumer code patterns may need updates.

Upstream PR: adcontextprotocol/adcp#2422

## Changes to regenerate

1. **New `custom` variant on vendor pricing `oneOf`.**
   ```ts
   {
     model: 'custom';
     description: string;      // required, human-readable
     metadata: object;         // required, minProperties: 1, vendor-defined
     currency?: string;        // ISO 4217 when monetary
   }
   ```
   Consumers doing exhaustive `switch (p.model) { case 'cpm': ... }` need a `case 'custom':` handler. Recommended default: route to operator review rather than auto-select.

2. **New optional `experimental_features[]` on `get_adcp_capabilities` response.**
   Dot-namespaced feature ids sellers declare when they implement experimental surfaces (e.g. `brand.rights_lifecycle`, `governance.campaign`, `trusted_match.core`). Consumers: `capabilities.experimental_features?.includes(id)` before relying on experimental surfaces. Full contract: https://adcontextprotocol.org/docs/reference/experimental-status

3. **`x-status: experimental` annotations on Brand Rights, Campaign Governance, and TMP schemas.**
   Metadata only — `json-schema-to-typescript` ignores `x-*` extensions by default; type output unchanged. Optionally propagate to JSDoc if you want consumers to see the experimental flag inline.

## Pre-existing backlog this regen also resolves

- `per_unit` pricing variant (added in an earlier 3.0 rc) is currently missing from `@adcp/client` v4.7.1 types. This regen picks it up alongside `custom`.

## Suggested release

- Minor version bump (4.7.1 → 4.8.0)
- Changelog: "Handle `model: 'custom'` in pricing consumers. New `experimental_features[]` helper on capabilities. Adds previously-missing `per_unit` pricing variant."
- No wire-format breaks — no migration guide required

bokelley:
Closing — the three schema changes this issue tracked are all resolved:

1. **`custom` pricing variant** — already in generated types at `src/lib/types/core.generated.ts:7832` (VendorPricing union) and `:8133` (CustomPricing interface). Picked up in #645 (governance-aware-seller regen).
[...]

PR #650, opened 2026-04-20
feat: idempotency auto-inject (#625) + experimental_features helper (#627)

### LINKED ISSUE #625: Storyboard runner should auto-inject idempotency_key for mutating sample_requests (or yamls need one)
## Problem

AdCP v3 requires `idempotency_key` on every mutating request, and the SDK rejects requests that lack it with `INVALID_REQUEST` **before the handler runs**. Storyboard `sample_request` blocks don't include `idempotency_key` for mutating steps. This blocks legitimate handler logic from being exercised and causes false failures that look like handler bugs.

## Concrete case from brand-rights E2E test

`brand_rights/rights_enforcement/acquire_expired_rights` — runner sends `acquire_rights` with `campaign.end_date` in the past. The agent's `acquireRights` handler is written to detect the expired campaign and return `rejected`. But the SDK rejects the request at the idempotency layer first (no key present), so the agent returns a generic `INVALID_REQUEST` and the step "passes" only because the scenario treats any 4xx as success. The handler's expired-campaign logic never runs.

`brand_rights/governance_denied/rights_denied/acquire_rights_denied` — same pattern. Request lacks `idempotency_key` AND lacks `revocation_webhook`; agent returns whatever comes first, which isn't the `GOVERNANCE_DENIED` the storyboard expected.

## Request

Pick one:

1. **Runner auto-injects** `idempotency_key` on every mutating operation when the `sample_request` doesn't carry one. Matches how real buyers would operate; storyboard authors don't need to remember to include the key on every mutating step. My preference — the runner already knows which tools are mutating (`MUTATING_TASKS` from `@adcp/client/server`).
2. **Add `idempotency_key` to every mutating `sample_request`** in the upstream spec storyboards. Brittle — easy to forget when adding new scenarios — but keeps the runner simple.

Option 1 recommended. Either way, the fix is probably 20 lines in the runner vs hundreds of edits across storyboard yamls in `adcontextprotocol/adcp`.

## Discovered via

End-to-end test `@adcp/client` 5.1.0 + `skills/build-brand-rights-agent/SKILL.md`, real `adcp storyboard run`. The false passes hid actual handler behavior and one false failure (wrong expected error code).

bokelley:
## Summary

Resolves #625 and #627 — AdCP 3.0 GA alignment in one PR.

### #625 — idempotency_key auto-inject on mutating requests

- **Storyboard runner**: new `applyIdempotencyInvariant` helper mints a UUID v4 on every mutating step whose `sample_request` omits one, so compliance runs exercise handler logic instead of short-circuiting on the server's required-field check. Skipped when `step.expect_error === true` so missing-key rejection scenarios still flow through unchanged.
- **Underlying SDK fix**: `normalizeRequestParams` now derives its mutating-task set from the Zod request schemas (`MUTATING_TASKS` in `utils/idempotency`) rather than a hand-maintained list that was missing `acquire_rights`, `update_media_buy`, `si_initiate_session`, `si_send_message`, and the property / collection / content-standards writes. Untyped `client.executeTask(...)` now auto-injects the same key that typed methods already minted via `executeAndHandle`.

### #627 — experimental_features helper + capabilities ergonomics

- `AdcpCapabilities` carries an `experimentalFeatures?: string[]` field populated from the AdCP 3.0 GA `experimental_features` envelope.
- New `supportsExperimentalFeature(caps, id)` helper. `resolveFeature` handles the `experimental:<id>` namespace so `require()` / `supports()` flows behave the same way as `ext:<name>`.
- The `custom` vendor-pricing variant and `per_unit` catchup were picked up in a prior regen (#645) — no type-surface changes ship here.

## Test plan

- [x] `npm test` — 4328 pass, 12 skipped, 0 fail
- [x] New unit tests for `applyIdempotencyInvariant` (6 cases: mutating task, read-only, BYOK key, expect_error skip, untyped-task, no-mutation)
- [x] New integration test: wire-level assertion that mutating steps carry `idempotency_key` and `get_products` does not, plus a `expect_error` regression
- [x] New unit tests for `experimentalFeatures` parsing + `supportsExperimentalFeature` helper + `resolveFeature('experimental:<id>')` + `listDeclaredFeatures` surfacing
- [x] `npm run lint` clean
- [x] `npx tsc -p tsconfig.lib.json --noEmit` clean


bokelley:
## Expert review pass (code-reviewer + ad-tech-protocol-expert + nodejs-testing-expert)

Addressed in c1889314:

**Blocker (ad-tech-expert + tracked in #625):** auto-skip on all `expect_error` mutating steps was too broad. Compliance storyboards use `expect_error` for GOVERNANCE_DENIED, UNAUTHORIZED, brand_mismatch, etc. — leaving the key absent meant the server short-circuits on INVALID_REQUEST before reaching the actual error path. Issue #625 cites this exact case (`acquire_rights_denied`). Fixed by inverting the default:
- `applyIdempotencyInvariant` now injects on `expect_error` mutating steps
- New `step.omit_idempotency_key: true` flag opts out when a storyboard explicitly tests missing-key rejection
- `testsMissingIdempotencyKey` now keys off `step.omit_idempotency_key` so runner and SDK layers agree

**Coverage (code-reviewer):** added a table-driven test that every `MUTATING_TASKS` member gets auto-injected by `normalizeRequestParams`. Locks the Zod-introspection behavior against schema drift. Plus negative coverage for read-only tasks.

**build_creative (code-reviewer):** confirmed it belongs in `MUTATING_TASKS` — `schemas/cache/latest/media-buy/build-creative-request.json` has `idempotency_key` in its `required` array. Old hardcoded list was wrong. Changeset now calls out the additions explicitly.

**Edge cases (nodejs-testing-expert):** added unit tests for BYOK keys, `$generate:uuid_v4#alias` pre-resolved UUIDs passing through, and empty-string `idempotency_key` being treated as unset.

**Not addressed (deferred as out-of-scope polish):**
- Experimental feature-id format validation (warn on non-`^[a-z][a-z0-9_]*(\.[a-z][a-z0-9_]*)+$` — spec doesn't mandate enforcement client-side; can add if seller drift shows up)
- `server.close(r)` awaiting in `finally` — existing pattern from `storyboard-brand-invariant.test.js`, no observed flake

Tests: 4337 pass, 12 skipped, 0 fail.

PR #651, opened 2026-04-20
feat(client): 5.4 downstream ergonomics — AdcpServer opaque handle + dispatchTestRequest

bokelley:
## Summary

Bundles five fixes surfaced while porting a training agent onto 5.3. One public-type change (breaking for pre-release consumers only — the type never reached a stable release), four additive.

- **BREAKING (pre-release only)** — `createAdcpServer()` returns the opaque `AdcpServer` type instead of SDK `McpServer`. Eliminates the ESM/CJS dual-package hazard where downstream consumers couldn't assign the return value to the SDK type due to the private `_serverInfo` field.
- **`AdcpServer.dispatchTestRequest({ method, params })`** — one encapsulated cast replaces downstream `(server as any)._requestHandlers.get('tools/call')` harness hacks. `tools/call` overload is typed `Promise<CallToolResult>`; generic fallback is `Promise<unknown>`.
- **`McpToolResponse.structuredContent` is optional** — error responses no longer need to fabricate empty objects to satisfy the type.
- **`SingleAgentClient.validateRequest` drops `.strict()`** — lets the runner's `brand`/`account` injection pass client validation on tools whose schema declares neither, so `adaptRequestForServerVersion` can strip by schema downstream.
- **Storyboard runner `request_signing.transport: 'raw' | 'mcp'`** — plumbs the existing grader option through the runner so MCP-only agents can grade `signed-requests` vectors via JSON-RPC `tools/call` envelopes.

Migration guide: [`docs/migration-5.3-to-5.4.md`](./docs/migration-5.3-to-5.4.md)

Review rounds (code-reviewer + dx-expert) addressed: unused import, `assert.doesNotReject` regression-guarded with explicit dispatch verification, `stripInternal` enabled in lib tsconfig, `dispatchTestRequest` typed overload for `tools/call`, double-wrap guard in `wrapMcpServer`, migration doc authored.

## Test plan

- [x] `npm test` — 4374 pass, 0 fail, 12 skipped
- [x] `npm run lint` — pre-existing warnings only; no new errors
- [x] New tests pin dispatch through `AdcpServer.dispatchTestRequest`, symbol wiring through `getSdkServer()`, and brand/account injection surviving client validation (the real scenario the strict() drop enables)


bokelley:
Landed ⇒ 5 upstream asks closed. Huge unblock for our training-agent migration. One follow-up request worth raising before we adopt the `AdcpServer` opacity downstream.

## Use case the migration guide doesn't cover

We have a server that registers both spec tools (through `createAdcpServer`'s domain groups) AND custom tools outside `AdcpToolMap` — specifically nine of them on the training agent: `creative_approval`, `update_rights`, `comply_test_controller`, `validate_property_delivery`, and five `*_collection_list` endpoints. Some are extensions the framework's AdcpToolMap doesn't model yet (collection_lists are a governance specialism); `comply_test_controller` is a test-harness tool the storyboard runner calls; `creative_approval` and `update_rights` were explicitly carved out of the framework per a comment in 5.3 pending JSON schemas.

Today our scaffold calls `server.registerTool(...)` nine times after `createAdcpServer()` returns. That's about to break against the opaque `AdcpServer` type.

The migration guide's guidance — *"For non-AdCP tools, keep using `createTaskCapableServer()` directly; `serve()` accepts either"* — implies one server per process, not mixed. But a production seller with spec tools + a handful of domain-specific extensions is a common shape; splitting into two transports at different mount paths isn't equivalent behavior.

The escape hatch exists: `getSdkServer(server)!.registerTool(...)`. But `getSdkServer` is exported at runtime and stripped from the `.d.ts` by `stripInternal`. Downstream consumers calling it are reaching through a deliberate opacity boundary, which defeats the whole point of the `AdcpServer` wrapper.

## Proposed addition

Add a `customTools` field on `AdcpServerConfig`:

```ts
interface AdcpServerConfig<TAccount = unknown> {
  // ... existing fields ...

  /**
   * Register tools outside `AdcpToolMap`. These bypass the framework's
   * idempotency/governance/response-wrap pipeline — the handler receives
   * the validated args and returns a `CallToolResult` directly. Useful
   * for seller-specific extensions (e.g., collection lists), test
   * harnesses (e.g., a compliance controller), or AdCP surfaces whose
   * JSON schemas haven't landed yet (`creative_approval`,
   * `update_rights`).
   *
   * Keys are the public tool names. Descriptions, Zod input schemas,
   * and callbacks match what `McpServer.registerTool` accepts.
   */
  customTools?: Record<string, {
    description?: string;
    inputSchema?: ZodRawShapeCompat;
    outputSchema?: ZodRawShapeCompat;
    annotations?: ToolAnnotations;
    handler: ToolCallback<ZodRawShapeCompat | undefined>;
  }>;
}
```

Internally `createAdcpServer()` calls the same `server.registerTool(...)` it already uses for spec tools — just from a single declarative config instead of a post-construction API. Keeps the `AdcpServer` wrapper opaque (no `.tool()` on the public interface) while giving consumers a supported path for extensions.

## Why this feels worth the 30 lines

Right now every downstream agent that has both spec + custom tools will end up calling `getSdkServer()`. The whole point of the opaque wrapper is to prevent that. Adding `customTools` closes the escape hatch on the public-API side so the wrapper's guarantee holds.

Shape mirrors the existing pattern: `taskStore`, `taskMessageQueue`, `idempotency`, `webhooks`, `signedRequests` are all pluggable bits passed through config. `customTools` fits the same mold.

Happy to send a follow-up PR against 5.4 if the API shape above looks right to you. Want to get your take before we write it — worth noting any concerns about custom tools bypassing the framework's conventions (validation, response wrap, idempotency) before the API is locked.

bokelley:
Landed `customTools` on `AdcpServerConfig` in d88108d2. Shape matches your proposal:

```ts
createAdcpServer({
  name: 'Publisher',
  version: '1.0.0',
  mediaBuy: { getProducts: async () => ({ products }) },
  customTools: {
    creative_approval: {
      description: '…',
      inputSchema: { creative_id: z.string(), approved: z.boolean() },
      handler: async ({ creative_id, approved }) => ({
        content: [{ type: 'text', text: `…` }],
        structuredContent: { creative_id, approved },
      }),
    },
    // ... eight more
  },
});
```

Internally calls `server.registerTool(...)` — same path as the spec tools — so no `getSdkServer()` reach-through required downstream.

**Guardrails.** Name collisions with `AdcpToolMap` or `get_adcp_capabilities` throw at construction time. Custom tools **bypass the framework pipeline** (no idempotency middleware, no governance pre-check, no account resolution, no response wrap) — handler documentation calls this out explicitly, and framework helpers (`checkGovernance`, `adcpError`, `capabilitiesResponse`) are importable from inside the handler if you want any of those behaviors.

Migration doc updated with the pattern at `docs/migration-5.3-to-5.4.md`. Tests cover happy-path dispatch + both collision throws (66 pass, 0 fail in `test/server-create-adcp-server.test.js`; full suite 4352 pass, 0 fail).

Ready for your adoption pass — flag anything the handler signature doesn't give you.

Issue #665, opened 2026-04-20
Proposal: requireAuthenticatedOrSigned — presence-gated signing + required_for pre-check with fallback bypass


bokelley:
## Problem

`requireSignatureWhenPresent` (shipped in 5.6, closes [#659](https://github.com/adcontextprotocol/adcp-client/issues/659)) solves the "invalid signature should never fall through to bearer" case. That's half the composition story.

The other half — **`required_for` enforcement for the no-signature path** — is left to the caller.

In our training agent we had to hand-roll:

```ts
const fallback: Authenticator = async (req) => {
  if (bearerChain) {
    const result = await bearerChain(req);
    if (result) return result;
  }
  const op = resolveOperationFromRequest(req);
  if (op && REQUIRED_FOR_OPERATIONS.includes(op)) {
    throw new AuthError(`Signature required for ${op}.`, {
      cause: new RequestSignatureError('request_signature_required', 0, ...),
    });
  }
  return null;
};

return requireSignatureWhenPresent(lazySigningAuthenticator(), fallback);
```

Two things the SDK should own:

### 1. `required_for` semantics with fallback credentials

The spec question — what happens when op is in `required_for`, no signature is present, but bearer is valid? — is filed upstream at [adcp#2586](https://github.com/adcontextprotocol/adcp/issues/2586). Consensus should be "fallback bypass is valid" (bearer-authed callers pass; unauthenticated callers on a required_for op get `request_signature_required`).

Encoding that by hand is a security footgun. Every seller who declares `required_for` and composes with bearer will reinvent this, and some will get it wrong.

### 2. `RequestSignatureError` → `WWW-Authenticate: Signature error="..."` mapping

Conformance graders probe the 401 response and read `WWW-Authenticate` for the spec error code. `respondUnauthorized` hardcodes `Bearer` as the challenge scheme; we had to emit the Signature challenge manually:

```ts
const sigCause = err instanceof AuthError && err.cause instanceof RequestSignatureError ? err.cause : null;
if (sigCause) {
  res.set('WWW-Authenticate', `Signature realm="mcp", error="${sigCause.code}"`);
  res.status(401).json({ error: sigCause.code, error_description: sigCause.message });
  return;
}
```

Every seller rediscovers this too.

### A. Extend `requireSignatureWhenPresent` with `requiredFor`

```ts
requireSignatureWhenPresent(
  signatureAuth,
  fallbackAuth,
  {
[...]

Issue #668, opened 2026-04-20
Proposal: capability-driven vector skip in request-signing grader


bokelley:
## Problem

The request-signing conformance grader runs all 40 negative vectors regardless of the agent's declared capability. Vectors that assert behaviors the agent didn't advertise (`covers_content_digest: 'required'` vs agent-declared `'either'`) fail without meaningful signal.

Current workaround: agents must know which vectors to exclude and pass `skipVectors` manually:

```ts
request_signing: {
  transport: 'mcp',
  skipVectors: [
    '007-missing-content-digest',       // vector asserts covers_content_digest=required
    '018-digest-covered-when-forbidden', // vector asserts covers_content_digest=forbidden
    '025-jwk-alg-crv-mismatch',         // exercises SDK library, not the agent
  ],
  skipRateAbuse: true, // vector 020
}
```

`skipVectors` is documented as "operator_skip" in the grader. The SDK docs even say 007/018 are the canonical cap-mismatch cases. But the agent operator still has to know to skip them.

## Proposal

Pre-flight the agent's capability before running vectors:

1. Grader fetches `get_adcp_capabilities` from the agent (or reads it from an optional `agentCapability:` option).
2. For each vector, compare `vector.verifier_capability` to the agent's capability:
   - `covers_content_digest: 'required'` on vector but `'either'` or `'forbidden'` on agent → auto-skip with `skip_reason: 'capability_profile_mismatch'`.
   - `required_for: [op]` on vector but op not in agent's `required_for` → auto-skip.
3. `skipVectors` remains available for operator-driven skips (e.g., 025's SDK-internal case).

```ts
// After this change:
request_signing: { transport: 'mcp' }
// and we're done.
```

## Why this is a grader-side problem

Every signing-implementing seller advertises a specific capability profile. The cross-product of cap profiles (3 digest modes × `required_for` as set-valued) is large; forcing operators to maintain a skip list per profile is fragile. The grader already reads `vector.verifier_capability` — matching that against the agent's declared cap is a small diff away.

## Backward compatibility

- Behavior change is additive: vectors that previously failed due to cap mismatch now `skipped: true, skip_reason: 'capability_profile_mismatch'`. Downstream clients already handle `skipped` (storyboard runner distinguishes skipped from failed).
- `skipVectors` stays available for operator-driven skips.
- An opt-out (`strictMode: true`) can preserve today's run-everything behavior for sellers that want to see all failures explicitly.

## Related

[...]

Issue #667, opened 2026-04-20
Proposal: generate zod schemas for non-AdcpToolMap tools (check_governance, acquire_rights, creative_approval, comply_test_controller, ...)


bokelley:
## Problem

`@adcp/client` generates zod schemas for `AdcpToolMap` tools (the 23 core media-buy/signals/creative tasks). Every other AdCP tool — governance, rights, brand identity, compliance testing, and all the custom surfaces — doesn't get generated zod.

In our training agent, `customTools` ships 9 tools outside `AdcpToolMap` that MUST publish an `inputSchema` via `tools/list` per MCP spec. Without SDK-generated zod, we had to hand-write 9 schemas (~120 lines):

```ts
const CREATIVE_APPROVAL_SCHEMA = {
  rights_id: z.string().optional(),
  creative_url: z.string().optional(),
  creative: z.object({ ... }).passthrough().optional(),
  account: ACCOUNT_REF,
  brand: BRAND_REF,
  context: CONTEXT_REF,
};
// ... and 8 more like this
```

Every MCP-based seller that registers these tools reinvents the same schemas. They will drift from the spec. `tools/list` contracts will disagree across implementations.

## Proposal

Extend the codegen pipeline (`scripts/generate-types.ts`) to emit zod raw shapes for every tool whose request schema lives in `static/schemas/source/**/*-request.json`, not just the AdcpToolMap subset.

```ts
import {
  CheckGovernanceInputShape,
  AcquireRightsInputShape,
  CreativeApprovalInputShape,
  ComplyTestControllerInputShape,
  // ... every tool with a request schema
} from '@adcp/client/schemas/shapes';
```

Then `customTools:` registration collapses to:

```ts
customTools: {
  creative_approval: customToolFor('creative_approval', 'Submit creative...', CreativeApprovalInputShape, handler),
  // ... 8 more, same pattern
}
```

No hand-authored schemas. Drift from spec impossible by construction.

## Why this unlocks framework parity

The 5.4 framework path refused to become default in adcontextprotocol/adcp#2561 because 9 custom tools hand-wrote zod schemas that didn't quite match the generated ones for AdcpToolMap — tools/list contracts diverged across dialects. Framework-generated shapes would close that last gap and let us flip the default.

Currently (adcontextprotocol/adcp#2582): legacy path 37/55 clean, framework path 23/55 clean. The ~14 storyboard delta includes several zod-shape drifts on non-AdcpToolMap tools.

## Scope

[...]

Issue #666, opened 2026-04-20
Proposal: AdcpServer.compliance.reset() — test-harness hook for session + idempotency cache flush


bokelley:
## Problem

Conformance storyboards share brand domains across test kits. A governance plan seeded by `media_buy_seller/governance_denied` (brand `acmeoutdoor.example`, budget $10K) silently intercepts a `$50K` buy in `sales_guaranteed` (same brand) when run in sequence, because the session store is keyed by brand domain.

Symptoms: random GOVERNANCE_DENIED failures, random idempotency cache hits on "different" storyboards sharing an idempotency key prefix.

## Current workaround

In our training-agent runner:

```ts
const { clearSessions } = await import('../../src/training-agent/state.js');
// ...
for (const storyboard of all) {
  await clearSessions();
  await runStoryboard(...);
}
```

We had to reach into the training-agent's internal `state.js` and also would need a parallel path for the idempotency store. No `@adcp/client`-level hook exists.

Every test runner — ours, any other agent's, future CI fixtures — will hit this.

## Proposal

Expose a standard reset hook on the server:

```ts
interface AdcpServer {
  compliance: {
    /** Clear session state store + idempotency cache. Intended for test
     *  harnesses running storyboards sequentially against a shared-state
     *  agent. No-op in production (refuses to run unless the state store
     *  is InMemoryStateStore, or explicit override). */
    reset(): Promise<void>;
  };
}
```

Or as a named export:

```ts
import { resetComplianceState } from '@adcp/client/server';
await resetComplianceState(server);
```

Behavior:

- If state store is `InMemoryStateStore` → call `.clear()`.
- If idempotency store has `clearAll()` → call it.
- If neither exposes a clear method, throw a clear error: "compliance.reset() requires InMemoryStateStore + createIdempotencyStore; set NODE_ENV=test to bypass the guard."
- In production (`NODE_ENV === 'production'` or non-memory store detected), refuse unless an explicit `{ force: true }` flag is passed.

## Related

- Training-agent runner: [adcontextprotocol/adcp#2582](https://github.com/adcontextprotocol/adcp/pull/2582) (has the `clearSessions` workaround)
- Companion: [createExpressAdapter](https://github.com/adcontextprotocol/adcp-client/issues/664) could expose this through its `resetHook` return value

PR #669, opened 2026-04-20
feat: AdCP 3.0 release blockers — adapter, fixtures, schemas, grader, reset

bokelley:
## Summary

Closes six AdCP 3.0 release blockers ([#663](https://github.com/adcontextprotocol/adcp-client/issues/663)–[#668](https://github.com/adcontextprotocol/adcp-client/issues/668)) in one PR. All changes are additive — new subpath exports, a new options parameter on `requireSignatureWhenPresent`, new fields on `RespondUnauthorizedOptions` and `GradeOptions`, and a new `compliance` property on `AdcpServer`. No breaking changes.

**New subpath exports**

- **`@adcp/client/compliance-fixtures`** (#663): canonical `COMPLIANCE_FIXTURES` data for every storyboard-hardcoded ID (`test-product`, `sports_ctv_q2`, `video_30s`, `native_post`, `native_content`, `campaign_hero_video`, `gov_acme_q2_2027`, `mb_acme_q2_2026_auction`, `cpm_guaranteed`, …) plus `seedComplianceFixtures(server)` that writes into the state store under `compliance:*` collections.
- **`@adcp/client/schemas`** (#667): re-exports every generated Zod `*RequestSchema`, plus `TOOL_INPUT_SHAPES` (superset of the framework-registered tools — also covers `creative_approval`, `update_rights`, `comply_test_controller`, `check_governance`, etc.) and a `customToolFor(name, description, shape, handler)` helper for MCP `customTools` registration.

**`@adcp/client/server` additions**

- **`createExpressAdapter({ mountPath, publicUrl, prm, server })`** (#664): returns `rawBodyVerify`, `protectedResourceMiddleware`, `getUrl`, and `resetHook` — the four pieces an Express-mounted agent needs to run behind a router. Removes ~20 lines of hand-wired plumbing.
- **`requireAuthenticatedOrSigned`** / **`requireSignatureWhenPresent({ requiredFor, resolveOperation })`** (#665): presence-gated signature composition now enforces `required_for` on the no-signature path. Unsigned requests with no credentials on a `required_for` operation throw `AuthError` whose cause is `RequestSignatureError('request_signature_required')`. Valid bearer still bypasses (per adcp#2586 consensus).
- **`respondUnauthorized({ signatureError })`** (#665): emits `WWW-Authenticate: Signature error="<code>"`. `serve()` auto-detects via `signatureErrorCodeFromCause(err)` so signature-layer 401s carry the right challenge scheme.
- **`AdcpServer.compliance.reset({ force? })`** (#666): drops session state + idempotency cache between storyboards. `IdempotencyStore.clearAll` is optional on the store — `memoryBackend` implements it, production backends leave it undefined. Refuses to run under `NODE_ENV=production` unless `force: true`.

**Testing (`@adcp/client/testing`)**

- **Grader `agentCapability` option** (#668): vectors whose `verifier_capability.covers_content_digest` or `required_for` doesn't match the agent's declared profile auto-skip with `skip_reason: 'capability_profile_mismatch'`. `skipVectors` stays available for operator overrides.

## Test plan

- [x] 28 new unit tests in `test/lib/adcp-3-0-blockers.test.js` — all 6 features covered, all green
- [x] Full `npm run test:lib` — 3826 tests, 0 failures
- [x] `npm run test` (full suite) — 4449 tests, 0 failures
- [x] `npm run lint` — 0 errors (pre-existing warnings only)
- [x] `npm run format` applied
- [x] Changeset added (`.changeset/adcp-3-0-release-blockers.md`, minor bump)

### DIFF
--- a/.changeset/adcp-3-0-release-blockers.md
+++ b/.changeset/adcp-3-0-release-blockers.md
@@ -1,0 +1,21 @@
+---
+'@adcp/client': minor
+---
+
+AdCP 3.0 release blockers — SDK-level wiring for conformance-runner integration.
+
+**New subpath exports**
+
+- `@adcp/client/compliance-fixtures` — canonical `COMPLIANCE_FIXTURES` data for every hardcoded ID storyboards reference (`test-product`, `sports_ctv_q2`, `video_30s`, `native_post`, `native_content`, `campaign_hero_video`, `gov_acme_q2_2027`, `mb_acme_q2_2026_auction`, `cpm_guaranteed`, etc.) plus a `seedComplianceFixtures(server)` helper that writes fixtures into the state store under well-known `compliance:*` collections. Closes [#663](https://github.com/adcontextprotocol/adcp-client/issues/663).
+- `@adcp/client/schemas` — re-exports every generated Zod request schema plus `TOOL_INPUT_SHAPES` (ready-to-register `inputSchema` map covering non-framework tools like `creative_approval` and `update_rights`) and a `customToolFor(name, description, shape, handler)` helper. Closes [#667](https://github.com/adcontextprotocol/adcp-client/issues/667).
+
+**Server (`@adcp/client/server`)**
+
+- `createExpressAdapter({ mountPath, publicUrl, prm, server })` returns the four pieces an Express-mounted agent needs: `rawBodyVerify` (captures raw bytes for RFC 9421), `protectedResourceMiddleware` (RFC 9728 PRM at the origin root), `getUrl` (mount-aware URL reconstruction for the signature verifier), and `resetHook` (delegates to `server.compliance.reset()`). Closes [#664](https://github.com/adcontextprotocol/adcp-client/issues/664).
+- `requireAuthenticatedOrSigned({ signature, fallback, requiredFor, resolveOperation })` bundles presence-gated signature composition with `required_for` enforcement on the no-signature path. `requireSignatureWhenPresent` grew an options parameter that carries the same `requiredFor` + `resolveOperation` semantics. Unsigned requests with no credentials on a `required_for` operation throw `AuthError` whose cause is `RequestSignatureError('request_signature_required')`; valid bearer bypass stays valid. Closes [#665](https://github.com/adcontextprotocol/adcp-client/issues/665).
+- `respondUnauthorized({ signatureError })` emits a `WWW-Authenticate: Signature error="<code>"` challenge when the rejection comes from the RFC 9421 verifier. `serve()` auto-detects this via `signatureErrorCodeFromCause(err)` — the signed_requests negative-vector grader reads the error code off the challenge, so previously callers had to override the 401 response by hand.
+- `AdcpServer.compliance.reset({ force? })` drops session state and the idempotency cache between storyboards. Refuses to run in production-like deployments unless `force: true` is passed. `IdempotencyStore.clearAll` is now an optional method on the store; `memoryBackend` implements it, production backends leave it undefined. Closes [#666](https://github.com/adcontextprotocol/adcp-client/issues/666).
+
+**Testing (`@adcp/client/testing`)**
+
+- Request-signing grader accepts an `agentCapability` option. When present, vectors whose `verifier_capability` can't coexist with the agent's declared profile (`covers_content_digest` disagreement, vector-asserted `required_for` not in agent's list) auto-skip with `skip_reason: 'capability_profile_mismatch'`. `skipVectors` stays available for operator-driven overrides. Closes [#668](https://github.com/adcontextprotocol/adcp-client/issues/668).
--- a/docs/TYPE-SUMMARY.md
+++ b/docs/TYPE-SUMMARY.md
@@ -1,7 +1,7 @@
 # AdCP Type Summary
 
 > Generated at: 2026-04-20
-> @adcp/client v5.5.0
+> @adcp/client v5.6.0
 
 Curated reference of the types that matter for using the AdCP client. For full generated types see `src/lib/types/tools.generated.ts` and `src/lib/types/core.generated.ts`.
 
--- a/docs/llms.txt
+++ b/docs/llms.txt
@@ -1,7 +1,7 @@
 # Ad Context Protocol (AdCP)
 
 > Generated at: 2026-04-20
-> Library: @adcp/client v5.5.0
+> Library: @adcp/client v5.6.0
 > AdCP major version: 3
 > Canonical URL: https://adcontextprotocol.github.io/adcp-client/llms.txt
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@adcp/client",
-  "version": "5.5.0",
+  "version": "5.6.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@adcp/client",
-      "version": "5.5.0",
+      "version": "5.6.0",
       "license": "Apache-2.0",
       "dependencies": {
         "jose": "^6.2.2",
--- a/package.json
+++ b/package.json
@@ -50,6 +50,16 @@
       "import": "./dist/lib/signing/server.js",
       "require": "./dist/lib/signing/server.js",
       "types": "./dist/lib/signing/server.d.ts"
+    },
+    "./schemas": {
+      "import": "./dist/lib/schemas/index.js",
+      "require": "./dist/lib/schemas/index.js",
+      "types": "./dist/lib/schemas/index.d.ts"
+    },
+    "./compliance-fixtures": {
+      "import": "./dist/lib/compliance-fixtures/index.js",
+      "require": "./dist/lib/compliance-fixtures/index.js",
+      "types": "./dist/lib/compliance-fixtures/index.d.ts"
     }
   },
   "typesVersions": {
@@ -77,6 +87,12 @@
       ],
       "signing/server": [
         "dist/lib/signing/server.d.ts"
+      ],
+      "schemas": [
+        "dist/lib/schemas/index.d.ts"
+      ],
+      "compliance-fixtures": [
+        "dist/lib/compliance-fixtures/index.d.ts"
       ]
     }
   },
--- a/src/lib/compliance-fixtures/index.ts
+++ b/src/lib/compliance-fixtures/index.ts
@@ -51,6 +51,12 @@
 
 import { ADCP_STATE_STORE, type AdcpServer } from '../server/adcp-server';
 import type { AdcpStateStore } from '../server/state-store';
+// Pull canonical unions/shapes from the generated spec types so fixture
+// typings can't drift from the schema. Hand-typing `pricing_model` as
+// `'cpm' | 'cpc' | 'cpv' | 'flat'` is how we shipped `'flat'` (which
+// doesn't exist in the spec) — the generated `PricingModel` has the
+// authoritative union.
+import type { PricingModel } from '../types/tools.generated';
 
 export type ComplianceFixtureCategory =
   | 'products'
@@ -95,9 +101,25 @@ export interface ComplianceFormatFixture {
 
 export interface CompliancePricingOptionFixture {
   pricing_option_id: string;
+  /** ISO 4217 currency code, per the spec's `PricingOption.currency`. */
   currency: string;
-  pricing_model: 'cpm' | 'cpc' | 'cpv' | 'flat';
-  cpm?: number;
+  /**
+   * Pricing model — uses the authoritative generated union. The spec
+   * discriminates `PricingOption` on this field across 9 variants
+   * (`cpm`, `vcpm`, `cpc`, `cpcv`, `cpv`, `cpp`, `cpa`, `flat_rate`,
+   * `time`); mismatches here silently fail schema validation in
+   * seller-side `get_products` responses that merge fixture data in.
+   */
+  pricing_model: PricingModel;
+  /**
+   * Fixed price per unit per the matching `*PricingOption.fixed_price`
+   * field (CPM, CPC, etc.). Present when this is a fixed-price option;
+   * omit for auction-based options that advertise `floor_price` /
+   * `price_guidance` instead.
+   */
+  fixed_price?: number;
+  /** Floor price for auction-based options. */
+  floor_price?: number;
 }
 
 export interface ComplianceCreativeFixture {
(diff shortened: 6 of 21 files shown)


bokelley (commented):
Reviewed against our training-agent PR (adcp#2582) — the workarounds this replaces. Strong API shape overall; would let us delete ~100 lines across `product-factory.ts`, `framework-server.ts`, `index.ts`, and the runner. Three blockers and one nit worth flagging before merge.

## Blockers

### 1. `CompliancePricingOptionFixture.pricing_model` type is wrong

`compliance-fixtures/index.ts:99` hand-types the union as `'cpm' | 'cpc' | 'cpv' | 'flat'`. The spec union (verified in `core.generated.d.ts`) is `'cpm' | 'vcpm' | 'cpc' | 'cpcv' | 'cpv' | 'cpp' | 'cpa' | 'flat_rate' | 'time'`. `'flat'` doesn't exist — the canonical value is `'flat_rate'`. A seller merging `COMPLIANCE_FIXTURES.pricing_options` into a real `get_products` response fails schema validation on the first pricing option.

Fix: derive the union from `schemas.generated` instead of hand-typing (that's the whole point of this package — fixtures that match the spec).

### 2. `requireSignatureWhenPresent({ requiredFor })` pre-check is unreachable from the documented composition

`auth-signature.ts:331–351` fires the `required_for` pre-check only when the fallback returns `null`. With the docstring's example composition — `fallback: anyOf(verifyApiKey, verifyBearer)` — a caller sending a bad bearer triggers `anyOf` to throw an `AuthError` with a `Bearer` challenge. The pre-check never runs. Result: the grader gets `WWW-Authenticate: Bearer ...` instead of `Signature error="request_signature_required"` on the exact vector that motivated #665.

Our training agent works around this by building the fallback *by hand* and throwing `RequestSignatureError` *inside* the fallback when op is in `required_for` + no sig. That sidesteps the SDK's pre-check entirely. If we adopt this PR verbatim, our negative-vector test regresses.

Two fixes, pick one:
- (a) Document that `requiredFor` only triggers on the credentials-absent path (then sellers must still hand-compose for the credentials-present-but-invalid case — doesn't close #665).
- (b) When op ∈ `requiredFor` and no signature header was present, **wrap** the fallback's thrown `AuthError` so its cause is `RequestSignatureError('request_signature_required')` — then `serve()`'s `signatureErrorCodeFromCause` auto-detects and emits the Signature challenge. This actually closes #665 end-to-end.

Recommend (b).

### 3. `createExpressAdapter.resetHook` drops compliance fixtures

`express-adapter.ts:225–231` calls `server.compliance.reset()` but never re-seeds. Fixtures seeded via `seedComplianceFixtures` live in the state store and get wiped on reset. A runner that does `for (sb of storyboards) { await adapter.resetHook(); run(sb) }` silently loses fixtures after storyboard #1, and every subsequent storyboard fails with fixture-not-found.

Fix: accept `seedFixtures: true | SeedComplianceFixturesOptions` on `ExpressAdapterOptions` and have `resetHook` call `reset()` then re-seed when configured. (Or: move `seedComplianceFixtures` into `compliance.reset()` itself so it's always idempotent — cleaner.)

## Nit

### Grader `covers_content_digest: 'either'` asymmetry

`grader.ts:601–608` treats `'either'` as "permissive enough for any vector value", so vectors 007 (`required`) and 018 (`forbidden`) don't auto-skip for an `'either'` agent. But `'either'` means the agent accepts *both* covered and uncovered — it explicitly does NOT enforce a `forbidden` or `required` policy, so vector 018's "agent MUST reject covered" and vector 007's "agent MUST reject uncovered" are structurally incompatible with `'either'`.

In our training agent we keep 007 and 018 in `skipVectors` for exactly this reason. This PR would let us drop `025-jwk-alg-crv-mismatch` but not 007/018 — which defeats the point of the auto-skip.

Fix: when vector asserts `'required'` or `'forbidden'` and agent declares `'either'`, still skip with `skip_reason: 'capability_profile_mismatch'` (agent doesn't commit to the stricter policy the vector grades).

## Compliments

- `TOOL_INPUT_SHAPES` + `customToolFor` shape matches what we hand-rolled in `framework-server.ts` — on adoption we delete the 2 we hand-schema'd for `creative_approval` and `update_rights`. The other 7 (property_delivery, 5× collection_list, comply_test_controller) still need hand-schemas until they're in `TOOL_REQUEST_SCHEMAS` — worth noting in the migration guide.
- `respondUnauthorized({ signatureError })` + `signatureErrorCodeFromCause` auto-detection in `serve()` is exactly the ergonomic we wanted. Our training-agent `requireToken` had a hand-written signature-challenge emitter that goes away entirely.
- Fixture set is a clean superset of what we alias. Good coverage of the storyboard-hardcoded IDs from #2584 / #2585.

## Migration notes (for PR description)

When adopters drop to this SDK:
1. Replace hand-aliased products/pricing in `buildCatalog` → `await seedComplianceFixtures(server)`.
2. Replace hand-zod schemas for `creative_approval` and `update_rights` → `TOOL_INPUT_SHAPES.creative_approval` / `.update_rights`. Keep hand-schemas for the other 7 until upstream.
3. Replace `rawBody` capture + PRM mount + `getUrl` override + `clearSessions` → `createExpressAdapter({ mountPath, publicUrl, prm, server, seedFixtures: true })`.
4. Replace hand-composed fallback auth → `requireSignatureWhenPresent(sig, anyOf(bearer, apiKey), { requiredFor, resolveOperation })` — **only after blocker #2 is fixed**. Until then, keep hand-composition.
5. Drop `007-missing-content-digest` and `018-digest-covered-when-forbidden` from `skipVectors` — **only after blocker-nit #4 is fixed**. `025-jwk-alg-crv-mismatch` stays (SDK-library vector, not agent-graded).

Overall +1 once the three blockers are addressed — this is a substantial quality-of-life win for any new AdCP seller and would close the scar-tissue migration path we just walked. Thanks for the fast turnaround on #663–#668.

bokelley:
Follow-up with concrete line references after pulling the files. Each blocker has a minimal diff-in-prose so the fix is obvious.

## Blocker 1 — `pricing_model: 'flat'` doesn't exist in the spec

`src/lib/compliance-fixtures/index.ts:99` ships:

```ts
pricing_model: 'cpm' | 'cpc' | 'cpv' | 'flat';
```

Verified against `core.generated.d.ts` at these line numbers — the actual spec union is:
```
'cpm' | 'vcpm' | 'cpc' | 'cpcv' | 'cpv' | 'cpp' | 'cpa' | 'flat_rate' | 'time'
```

`'flat'` doesn't exist (should be `'flat_rate'`). Missing: `'vcpm'`, `'cpcv'`, `'cpp'`, `'cpa'`, `'time'`.

The point of this package is fixtures that match the spec. Hand-typing a subset will always drift. Fix: import the generated `PricingModel` (or re-derive via `z.infer<typeof PricingOptionSchema>['pricing_model']`) and use that as the union.

Bonus: the seed pricing bodies at `compliance-fixtures/index.ts:180-199` should similarly be typed from the spec rather than hand-shaped, so they can't drift.

## Blocker 2 — `requireSignatureWhenPresent({ requiredFor })` unreachable from the canonical composition

`src/lib/server/auth-signature.ts:320-352`:

```ts
const combined: Authenticator = async req => {
  if (hasSignatureHeader(req)) { ... }
  const fallbackResult = await fallbackAuth(req);   // ← throws propagate
  if (fallbackResult !== null) return fallbackResult;
  // requiredFor check only reaches here on fallbackResult === null
  if (requiredFor.size > 0 && resolveOperation) { ... throw request_signature_required ... }
  return null;
};
```

The three paths:
| Scenario | fallback behavior | `requiredFor` check fires? |
|---|---|---|
| No creds presented | `null` | ✓ yes |
| Valid bearer | returns principal | no (bearer wins) |
| **Bad bearer** | **throws AuthError** | **no (throw propagates)** |

The third row is the grader's `request_signature_required` vector — an unsigned request, op in `required_for`. In the documented composition (`fallback: anyOf(verifyApiKey({ keys: ... }))`), a caller with a bad bearer triggers `anyOf` to throw. The pre-check never fires. Grader reads `WWW-Authenticate: Bearer` and fails the test.

The current test suite misses this: `test/lib/adcp-3-0-blockers.test.js:47` uses `fakeBearer({})` — empty keys dict, which returns `null` on no bearer rather than throwing on a wrong bearer. The throw path is untested.

Fix (pattern b from the issue):
```ts
let fallbackResult;
let fallbackThrew;
try {
  fallbackResult = await fallbackAuth(req);
} catch (err) {
  fallbackThrew = err;
}
// Check requiredFor BEFORE rethrowing fallback's error
if (requiredFor.size > 0 && resolveOperation) {
  const operation = resolveOperation(req);
  if (operation && requiredFor.has(operation)) {
    throw new AuthError(`Signature required for ${operation}.`, {
      cause: new RequestSignatureError('request_signature_required', 0, ...),
    });
  }
}
if (fallbackThrew) throw fallbackThrew;
return fallbackResult ?? null;
```

And add a test: `fakeBearer({ sk_live: {...} })` + caller sending `Authorization: Bearer wrong-token` + op `create_media_buy` in `requiredFor` → must throw with `RequestSignatureError` cause, not the Bearer `AuthError`.

## Blocker 3 — `resetHook` drops fixtures; `compliance.reset` has no re-seed path

`src/lib/server/express-adapter.ts:225-231`:

```ts
const resetHook = async () => {
  if (options.server) {
    await options.server.compliance.reset({ ... });
  }
};
```

`compliance.reset()` clears the state store. Fixtures seeded via `seedComplianceFixtures(server)` live in the state store under `compliance:*` collections and are wiped.

Runner pattern:
```ts
await seedComplianceFixtures(server);
for (const sb of storyboards) {
  await adapter.resetHook();   // ← wipes fixtures on first call
  await runStoryboard(sb);     // ← every storyboard after the first fails fixture lookups
}
```

Two fixes, pick one:

**(a) Opt-in reseed on adapter:**
```ts
createExpressAdapter({
  ...,
  seedFixtures: true,  // or SeedComplianceFixturesOptions
});
// resetHook: reset, then re-seed if seedFixtures was truthy
```

**(b) Bake into `compliance.reset()`:** reset takes `{ seedFixtures?: boolean | SeedOptions }` and re-seeds after the flush. Cleaner — one call, always idempotent.

Either way: add a test asserting `COMPLIANCE_FIXTURES['test-product']` is readable after a reset cycle.

## Nit 1 — Grader `'either'` asymmetry

`src/lib/testing/storyboard/request-signing/grader.ts:600-609`:

```ts
if (
  vectorCap.covers_content_digest !== 'either' &&
  agentCap.covers_content_digest !== 'either' &&
  vectorCap.covers_content_digest !== agentCap.covers_content_digest
) {
  return (mismatch);
}
```

This short-circuits to "no mismatch" whenever either side is `'either'`. That's right for `vectorCap === 'either'` (vector is permissive — any agent can grade it) but wrong for `agentCap === 'either'`:

| vector | agent | should skip? | current skip? |
|---|---|---|---|
| `required` | `required` | no | no ✓ |
| `required` | `forbidden` | yes | yes ✓ |
| `required` | **`either`** | **yes** | **no ✗** |
| `forbidden` | **`either`** | **yes** | **no ✗** |
| `either` | `required` | no | no ✓ |
| `either` | `either` | no | no ✓ |

Agent declaring `'either'` accepts both covered and uncovered — it explicitly doesn't enforce the `required` or `forbidden` policy vector 007/018 grade. So the vectors can never pass against an `'either'` agent.

Fix: change the condition to `vectorCap.covers_content_digest !== 'either' && vectorCap.covers_content_digest !== agentCap.covers_content_digest`. Vector-side `'either'` stays permissive; agent-side `'either'` is a mismatch with any strict vector.

Test at `test/lib/adcp-3-0-blockers.test.js:565-585` only covers `vectorCap === 'either'` (the permissive case). Add: agent `'either'` + vector `'required'` → skip.

## Nit 2 — `TOOL_INPUT_SHAPES` only covers 2 of 9 custom tools

`src/lib/schemas/index.ts:69-70` hardcodes `creative_approval` and `update_rights`. `validate_property_delivery`, `comply_test_controller`, `create_collection_list`, `get_collection_list`, `update_collection_list`, `list_collection_lists`, `delete_collection_list` are missing.

For our training agent that means we still hand-schema 7 tools, only trim 2. Not wrong per se — #667 explicitly scoped to "non-AdcpToolMap tools the SDK already generates schemas for" — but the changeset implies broader coverage ("ready-to-register `inputSchema` map covering non-framework tools like `creative_approval` and `update_rights`"). Worth either:

- Add the other 7 to the generated schemas (fix upstream).
- Change the changeset + docstring to say "currently covers `creative_approval` and `update_rights`; other custom tools still require hand-authored shapes."

## Nit 3 — `customToolFor` type should infer from `shape`

`src/lib/schemas/index.ts` — `customToolFor(name, description, shape, handler)`. Handler's `args` is typed from `shape`? If not, callers still need `(args: unknown) => ...`. Worth checking — not blocking, but a small ergonomics delta.

---

Happy to open follow-up PRs for any of these if that's helpful. Blocker 2 is the one that would make us keep our hand-rolled composition post-adoption, so it's the highest-leverage fix.

bokelley:
Thanks for the detailed review — all three blockers + both nits addressed in 32e6d70.

**B1 — pricing_model drift**: `CompliancePricingOptionFixture.pricing_model` now imports the generated `PricingModel` union directly. Fixture bodies switched from the hand-made `cpm` field to the spec's `fixed_price` (CPMPricingOption). Canonical unions + spec field names, both derived from the generated types.

**B2 — requiredFor unreachable from canonical composition**: restructured `requireSignatureWhenPresent` to catch the fallback's throw, run the `requiredFor` pre-check against `fallbackResult === null` REGARDLESS of whether it threw, then rethrow the fallback's original error only when op isn't in `requiredFor`. New test covers exactly the bad-bearer-on-required-for path you identified; a second test covers the rethrow-preserved case (bad bearer, op NOT in requiredFor → original `Bearer` challenge survives).

**B3 — resetHook drops fixtures**: added `seedFixtures: boolean | SeedComplianceFixturesOptions` to `ExpressAdapterOptions`. When set, `resetHook` flushes state then re-seeds `COMPLIANCE_FIXTURES`. Test asserts fixtures survive a reset cycle when `seedFixtures: true`.

**N1 — `'either'` asymmetry**: `capabilityMismatch` now only treats vector-side `'either'` as permissive; agent-side `'either'` against a strict vector is a mismatch (matches your truth table). Vector 007 against an `'either'` agent auto-skips as expected.

**N2 — TOOL_INPUT_SHAPES coverage**: `validate_property_delivery` added to `TOOL_REQUEST_SCHEMAS` (was the only tool with a generated schema missing from the map). Everything else you listed — `comply_test_controller`, all 5 collection_list tools — was already covered via `TOOL_REQUEST_SCHEMAS`. Updated the docstring to stop under-advertising coverage.

**N3 — customToolFor type inference**: already inferred via `z.input<z.ZodObject<TShape>>` — handler's `args` picks up the shape's concrete type at call sites.

Test count is 40 now (was 35), full lib suite: 3803 pass / 0 fail. CI green.

Migration path is now fully unblocked — you should be able to drop the hand-rolled composition and `skipVectors: [007, 018]` entries once this lands.

PR #674, opened 2026-04-21
fix(harness): wire --auth so skill-compliant agents don't hit auth_required

### LINKED ISSUE #673: agent-skill-storyboard harness: wire --auth through so skill-compliant agents don't fail auth_required
## Problem

The capstone harness at `scripts/manual-testing/agent-skill-storyboard.ts` doesn't configure authentication on either side of the test:

- The prompt it gives Claude doesn't mention a test API key.
- `runGrader` at `scripts/manual-testing/agent-skill-storyboard.ts:223` calls `adcp storyboard run <url> <id> --json --allow-http` — no `--auth` token.

Result: whether a harness run succeeds or fails is **non-deterministic across Claude instances**:

- If Claude ignores the skill's "Authentication is mandatory" directive and ships an unauthenticated agent, the storyboard runs (and the agent would fail `security_baseline` separately).
- If Claude follows the skill correctly and wires `verifyApiKey` / `verifyBearer` per `serve({ authenticate })`, the grader can't get past auth → `overall_status: "auth_required"`, 0 scenarios graded.

Observed on three runs: `sales-guaranteed` went one way (30/42 scenarios graded, no auth), `sales-broadcast-tv` and `deterministic-testing` went the other (`auth_required`, 0 graded). Same skill, same harness, different outcomes — non-determinism driven by Claude's per-run choice on how to honor the skill.

## Fix

The harness should mirror what a real compliance test does — a registered counterparty with a known static API key. Two small changes:

1. **Update the prompt** at `buildPrompt()` (`scripts/manual-testing/agent-skill-storyboard.ts:82`) to tell Claude:
   - Wire `verifyApiKey({ keys: { 'compliance-runner': { principal: 'compliance-runner' } } })` as the authenticator.
   - Match the `security_baseline` requirement without blocking the grader.
2. **Pass `--auth compliance-runner`** in `runGrader()` at `scripts/manual-testing/agent-skill-storyboard.ts:223`:
   ```ts
   spawnSync('node', [cliPath, 'storyboard', 'run', url, storyboardId,
                      '--json', '--allow-http', '--auth', 'compliance-runner'], ...)
   ```

With both changes, a fully-skill-compliant agent gets graded against the storyboard instead of bouncing on auth.

Alternative: introduce a `--sandbox-unauth` mode on `serve({ authenticate })` that accepts a compliance-runner principal without a credential, so the harness doesn't need the key round-trip at all. Less clean — authentication is a real production concern worth exercising under compliance.

## Discovered via

Running `agent-skill-storyboard.ts` across 5 skill × storyboard pairs after today's skill updates. Two of the runs (`sales-broadcast-tv`, `deterministic-testing`) hit `auth_required` despite the agents being correctly built. See `@adcp/client/.context/harness-runs/*/output.json`.

bokelley:
## Summary

- Adds `verifyApiKey({ keys: { 'compliance-runner': { principal: 'compliance-runner' } } })` to the prompt Claude gets in `agent-skill-storyboard.ts`, so generated agents wire authentication like a real compliance test expects.
- Passes `--auth compliance-runner` to `adcp storyboard run` in `runGrader()` so the grader presents the same static key.

Before: whether a harness run succeeded was non-deterministic — a skill-compliant agent wiring `authenticate` would bounce the grader on `auth_required` (0 scenarios graded); a non-compliant agent skipping auth would run (but fail `security_baseline` separately). Observed across 5 skill × storyboard pairs.

After: both sides share a known static key. The universal `security_baseline` probe still sees authentication wired, and specialism storyboards run authenticated instead of bouncing on 401.

Fixes #673

## Test plan

- [ ] Run `tsx scripts/manual-testing/agent-skill-storyboard.ts --skill skills/build-seller-agent/SKILL.md --storyboard sales-broadcast-tv` and confirm grader progresses past auth (no `overall_status: "auth_required"`).
- [ ] Re-run the same command with `--skill skills/build-seller-agent/SKILL.md --storyboard deterministic-testing` to confirm the second previously-failing pair also gets graded.
- [ ] Verify `sales-guaranteed` (which previously ran without auth) still grades, now authenticated.


bokelley:
Matches #673 exactly — thanks.

Two small observations:

1. **Empty changeset.** `.changeset/upset-things-enjoy.md` has just `---` / `---` with no bump type or body. `npm run changeset status` may flag it. If this is intentionally additive-only (harness-only change, no library consumer impact), `---\n'@adcp/client': patch\n---\n` with a one-line summary keeps the changeset-check CI happy and documents the fix in the release PR. Not blocking — just something to know before merge.

2. **`docs/llms.txt` diff unrelated to the harness fix?** The flow ordering for `media_buy_seller` picked up an extra `list_creative_formats` step before `create_media_buy`, and the generated-at date bumped to today. Looks like a regen sneaked in with the harness change — fine if it's intentional, but worth splitting if you want to keep PRs atomic. If it's incidental, can confirm and squash.

I re-ran the harness against my branch locally — confirmed `sales-broadcast-tv` and `deterministic-testing` previously hit `auth_required` (0 scenarios graded) with the old prompt. With this PR both should grade authenticated. Will re-run on my branch once this merges.

PR #697, opened 2026-04-21
feat(cli): --invariants flag to register storyboard assertion modules

bokelley:
## Summary

Follow-up to #692 (storyboard assertion registry). Adds a CLI surface so operators can dynamic-import assertion modules before the runner resolves `storyboard.invariants: [...]`.

`adcp storyboard run ... --invariants ./my-invariants.js,@adcp/invariants` imports each specifier at CLI entry, triggering each module's `registerAssertion(...)` calls. Without this, storyboards that declare invariants would fail at runner start with \"Storyboard references unregistered assertion\" because nothing is wired to populate the registry.

The framework intentionally keeps concrete assertions out of `@adcp/client` (they ship from `adcontextprotocol/adcp` alongside the specialisms that own them). This flag is the missing hook for that to happen.

## Shape

```bash
adcp storyboard run test-mcp sales-non-guaranteed \
  --invariants ./my-invariants.js,@adcp/invariants
```

- Comma-separated list; relative paths resolve against cwd; bare specifiers pass through as npm packages; `path.isAbsolute` for Windows parity.
- Runs **before** the `--dry-run` gate so bad specifiers surface during preview, not after auth/agent resolution.
- Wired at all three storyboard entry points: `handleStoryboardRun`, `handleMultiInstanceStoryboardRun`, and the `comply` path through `runFullAssessment`.
- Filtered from `positionalArgs` so module specifiers aren't mistaken for agent aliases.

## Test plan

- [x] Single module loads and registers at import time (marker-file verified)
- [x] Comma-separated multi-module load
- [x] Missing value → exit 2 with clear message
- [x] Broken module (throws at import) → exit 2 with error surfaced
- [x] Missing path → exit 2 with error surfaced
- [x] Positional filter doesn't confuse specifier with agent arg
- [x] Whitespace-only value (`\" , , \"`) → exit 2, not silent no-op
- [x] Multi-instance path (`--url ... --url ...`) also loads modules
- [x] 44 neighbor tests (registry, webhook-receiver, file-flag, multi-instance) still green

## Related

- Follows #692 — the registry framework
- Unblocks adcontextprotocol/adcp shipping concrete invariants via npm packages

### DIFF
--- a/.changeset/cli-invariants-flag.md
+++ b/.changeset/cli-invariants-flag.md
@@ -1,0 +1,16 @@
+---
+'@adcp/client': minor
+---
+
+`adcp storyboard run` gains `--invariants <module[,module...]>`. The flag
+dynamic-imports each specifier before the runner resolves
+`storyboard.invariants`, giving operators a way to populate the assertion
+registry (adcp#2639) without editing the CLI. Relative paths resolve against
+the current directory; bare specifiers resolve as npm packages.
+
+Modules are expected to call `registerAssertion(...)` at import time. The
+flag runs before the `--dry-run` gate so bad specifiers surface immediately
+during preview, not after agent resolution and auth.
+
+Applies to `adcp storyboard run`, `adcp comply` (deprecated alias), and
+`adcp storyboard run --url` multi-instance dispatch.
--- a/.changeset/conformance-fuzzer.md
+++ b/.changeset/conformance-fuzzer.md
@@ -1,0 +1,37 @@
+---
+'@adcp/client': minor
+---
+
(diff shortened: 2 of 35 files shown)


bokelley:
CI "Test & Build" failure is pre-existing on main: same 2 failures in `test/lib/storyboard-completeness.test.js` (`expect_substitution_safe has a registered response schema` and `storyboard: sales_catalog_driven`). Both tasks are newly added upstream in the compliance tarball and the SDK hasn't caught up; unrelated to this PR. Verified locally by running the test on main.

PR #717, opened 2026-04-21
fix(storyboard): refs_resolve follow-ups (#710, #711, #712, #714)

### LINKED ISSUE #710: refs_resolve $agent_url should canonicalize transport path before scope compare
## Observation

Running the new `refs_resolve` check (#670 / #671) against `media_buy_seller` on the reference test agent at \`https://test-agent.adcontextprotocol.org/mcp\` produces a clean PASS, but with **every** source ref flagged \`out_of_scope\`.

The refs' \`agent_url\` values are \`https://test-agent.adcontextprotocol.org\` — the correct bare agent URL per AdCP conventions. \`\$agent_url\` expands to the runner's target URL including the \`/mcp\` transport path, and \`normalizeAgentUrl\` preserves the path. So the scope comparison is effectively:

\`\`\`
\"https://test-agent.adcontextprotocol.org\" === \"https://test-agent.adcontextprotocol.org/mcp\"
\`\`\`

Never matches. Net effect: on every MCP agent, the check degrades from \"integrity enforcement\" to \"warning-only observation stream,\" which is the opposite of what the primitive is for.

## Why this matters

The whole point of \`refs_resolve\` on \`media_buy_seller\` is to fail the grade when a seller's \`format_ids\` don't round-trip to their own \`list_creative_formats\`. If the agent under test is the owner of the refs but scope never recognizes it, broken refs silently warn instead of fail.

## Proposed fix

In \`resolveScopeEquals\` (src/lib/testing/storyboard/validations.ts), when the scope \`key\` ends in \`url\` and \`equals\` is \`\$agent_url\`, strip the transport path (\`/mcp\`, \`/a2a\`, \`/sse\`) from the resolved agent URL before normalizing. Two possible shapes:

1. **Explicit list**: strip trailing path segments matching a known transport suffix. Simple, but couples the validator to the transport registry.
2. **Canonical agent URL resolution**: use the agent's declared canonical URL (from \`/.well-known/adagents.json\` or the agent's self-reported identity) rather than the raw runner target URL. Cleaner but requires threading that identity into \`ValidationContext\`.

Option 1 is faster to land; option 2 is more correct long-term.

## Repro

\`\`\`ts
import { runStoryboard } from '@adcp/client/testing';
const sb = await getStoryboardByIdFromCache('media_buy_seller');
const result = await runStoryboard('https://test-agent.adcontextprotocol.org/mcp', sb, { auth: ... });
// Inspect the list_formats step — refs_resolve passes but every ref is an out_of_scope observation.
\`\`\`

## Context

Surfaced while verifying the adcp-side storyboard wiring in adcontextprotocol/adcp#2601. Marking that PR ready — the wiring is correct; this canonicalization bug is orthogonal and should be addressed here so the check actually enforces integrity on MCP agents.

### LINKED ISSUE #711: refs_resolve: warn when scope filter excludes 100% of source refs
## Problem

When a `refs_resolve` validation's `scope` filter excludes every source ref, the check passes silently (or with warning observations that don't fail the grade). This is the opposite of the primitive's purpose — integrity isn't being enforced and nobody knows.

Concrete example from adcontextprotocol/adcp#2601 verification: on the reference MCP test agent, the \`\$agent_url\` canonicalization bug (tracked separately in #710) causes every ref on `media_buy_seller`'s `list_formats` step to fall out-of-scope. The check passes cleanly while enforcing nothing.

## Proposed fix

Independent of #710's canonicalization work, the runner should detect when a scope filter excludes 100% of in-scope refs AND the source ref set is non-empty, and emit a grader-visible warning (not a refs_resolve failure — a meta-observation that the check is misconfigured or broken).

Concretely: in `validateRefsResolve`, after scope partitioning:

```ts
if (scope && sourceRefs.length > 0 && inScope.length === 0) {
  observations.push({ kind: 'scope_excluded_all_refs', count: sourceRefs.length });
}
```

This surfaces silent-no-op checks as grader signal without changing the check's pass/fail semantics. When #710 lands and canonicalization is correct, this observation disappears naturally on well-behaved agents.

## Why this matters

A malicious or misconfigured seller can currently game the refs_resolve check by ensuring their `agent_url` values never canonicalize to match the runner's `\$agent_url`. The check passes. Nothing is enforced. With this meta-warning, the structural smell surfaces even when the underlying comparison is broken.

## Context

Surfaced by security review on adcontextprotocol/adcp#2601. Severity: medium — today's silent degradation on MCP agents means the marquee cross-step check adcp#2597 asked for is effectively a no-op in production.

### LINKED ISSUE #712: refs_resolve: pagination of list_creative_formats target
## Problem

`refs_resolve` compares a source ref set against a target set drawn from a single step's response. On `media_buy_seller`'s `list_formats` step, if the seller paginates `list_creative_formats` and the runner reads only the first page, valid format_ids referenced by products but defined on later pages will grade as missing — a false positive.

## Resolution options

**A. Spec clarification.** Require that when `list_creative_formats` is called during compliance, the seller MUST return every format referenced by any product returned by `get_products` in a single response, regardless of normal pagination policy. Simple; couples compliance-run behavior to a storyboard assertion; probably needs AdCP spec language.

**B. Runner follows pagination.** When target.path evaluates to an incomplete set and the response carries pagination metadata (`next_cursor`, etc.), the runner calls `list_creative_formats` until exhausted, then performs refs_resolve against the aggregated set. Cleaner from the seller's side; adds runner complexity + latency budget.

**C. Document the limitation and degrade gracefully.** If pagination metadata is present, refs_resolve adds an observation noting that the target set may be incomplete and demotes missing refs to observations instead of failures. Simplest; weakest enforcement.

## Recommendation

Option A for v1 (land alongside adcp-client#710 canonicalization fix). List_creative_formats responses are small enough in practice that a compliance-context \"return everything referenced by products\" rule is reasonable and unambiguous. Revisit if real sellers push back.

## Context

Surfaced by protocol review on adcontextprotocol/adcp#2601.

### LINKED ISSUE #714: refs_resolve: observations payload hygiene (userinfo strip, size caps, escaping)
## Problem

`refs_resolve` observations embed full ref objects from agent responses. If compliance reports are published (public dashboards, shared grader output, archived logs), a hostile agent can plant payloads in ref string fields:

- **Credentials leakage**: \`agent_url\` with \`userinfo@\` (\`https://user:token@host/...\`) — normalizer doesn't strip userinfo today.
- **Report bloat / DoS on viewers**: unbounded string fields on refs.
- **Stored XSS / markdown injection** downstream: ref fields rendered as HTML or markdown in a compliance UI without escaping.

Severity: low (impact depends on downstream consumers), but worth fixing in the runner so every consumer gets safe output.

## Proposed fix

In `validateRefsResolve`, before adding a ref to `observations` or `actual.missing`:

1. **Strip userinfo** from any URL-shaped field. Use `normalizeAgentUrl` consistently on observation URLs, not just on the comparison side.
2. **Cap individual ref string fields** — e.g. 512 chars max. Truncate with a trailing \"…\" indicator.
3. **Cap observation array length** — e.g. 50 entries per check. 10,000 fan-out cap already exists, but per-check observation array is still too large for readable reports.
4. **Flag string fields as untrusted** via a convention that compliance UIs can key off (e.g. ref values are always plain strings, never HTML/markdown).

Point (4) is a downstream contract, not a runtime filter. Documenting it in the runner-output-contract is enough.

## Context

Surfaced by security review on adcontextprotocol/adcp#2601.

bokelley:
## Summary

Bundle of four `refs_resolve` validator fixes surfaced by security/protocol review on [adcontextprotocol/adcp#2601](https://github.com/adcontextprotocol/adcp/pull/2601). Each fix closes a dedicated issue and ships with its own changeset.

- **#710** — Canonicalize `\$agent_url` by stripping transport suffixes (`/mcp`, `/a2a`, `/sse`, `/.well-known/agent[-card].json`) while preserving subpath. Mirrors `SingleAgentClient.computeBaseUrl`. Previously, the scope compare degraded to a silent no-op on every MCP/A2A agent because the runner's target URL never equalled the bare agent URL on refs. _Path-preserving_ matters: AdCP `agent_url` can legitimately live under a subpath (`https://publisher.com/.well-known/adcp/sales`) — origin-only would false-positive sibling agents on shared hosts.

- **#711** — Emit `scope_excluded_all_refs` meta-observation when a scope filter partitions 100% of source refs out. Catches the silent-no-op smell even when the underlying comparison is broken. Suppressed under `on_out_of_scope: 'ignore'` (which explicitly opts out of scope warnings).

- **#712** — Detect paginated current-step targets (`pagination.has_more: true`) and demote unresolved refs to `unresolved_with_pagination` observations instead of failing the check. Stops false positives against conformant paginating sellers. `target_paginated` meta-observation surfaces the partition decision. The spec-level fix (compliance mode requires seller returns all product-referenced formats in a single response) should follow in the AdCP repo.

- **#714** — Harden grader-visible payloads: strip userinfo via WHATWG URL parsing plus a regex fallback for credential-shaped substrings in non-URL fields; replace non-`http(s)` schemes (blocks `javascript:` / `data:` / `file:` stored-XSS vectors) with a placeholder; code-point-safe truncate at 512 chars; cap observations at 50 with `observations_truncated` marker; use `hasOwnProperty` throughout to prevent storyboard authors from drawing match keys from `Object.prototype`.

## Expert review notes

Routed through code-reviewer, security-reviewer, and ad-tech-protocol-expert before pushing. Key incorporated feedback:

- **Protocol** — flagged that origin-only canonicalization (my first draft) contradicts the AdCP schema's own `agent_url` examples (`core/format-id.json` shows `https://publisher.com/.well-known/adcp/sales`). Rewrote to path-preserving + transport-suffix-stripping.
- **Protocol** — flagged that `#712` needs to demote failure → observation, not just annotate. Implemented.
- **Security** — non-URL userinfo substrings, `javascript:` scheme survival, `key in ref` prototype access, pre-truncate ordering. All addressed with dedicated tests.
- **Code review** — surrogate-pair-safe truncation; positional assertion on meta-observation ordering; stale comment in `types.ts`. All addressed.

## Observation vocabulary

This PR adds three observation kinds the runner emits on `refs_resolve` results: `scope_excluded_all_refs`, `target_paginated`, `unresolved_with_pagination`, plus `observations_truncated` marker. The runner-output-contract at `compliance/cache/latest/universal/runner-output-contract.yaml` is synced from upstream and doesn't yet enumerate an observation taxonomy — documenting these kinds in the source contract belongs in a follow-up PR against [adcontextprotocol/adcp](https://github.com/adcontextprotocol/adcp).

## Test plan

- [x] `node --test test/lib/storyboard-validations-refs-resolve.test.js` — 39 tests, all pass (18 pre-existing + 21 new covering canonicalization, scope warnings, pagination demotion, and each hardening case)
- [x] `npm test` — full suite 4965 pass / 0 fail / 6 skipped
- [x] `npm run build` clean

## Changesets

Four patch-level changesets, one per closed issue — lets each fix land with its own CHANGELOG entry in the next release PR.

Closes #710, closes #711, closes #712, closes #714.

### DIFF
--- a/.changeset/refs-resolve-canonicalize-agent-url.md
+++ b/.changeset/refs-resolve-canonicalize-agent-url.md
@@ -1,0 +1,21 @@
+---
+'@adcp/client': patch
+---
+
+`refs_resolve` scope: canonicalize `$agent_url` by stripping transport
+suffixes instead of comparing raw target URL to bare agent origins.
+
+Before this fix, storyboards using `scope: { key: 'agent_url', equals:
+'$agent_url' }` silently graded every source ref `out_of_scope` on MCP
+and A2A runners, because `$agent_url` expanded to the runner's target
+URL (with `/mcp`, `/a2a`, or `/.well-known/agent.json` suffixes) while
+refs carried the bare agent URL per AdCP convention. Net effect: the
+check degraded from integrity enforcement to a no-op on every MCP agent.
+
+The scope comparator now mirrors `SingleAgentClient.computeBaseUrl`:
+strip `/mcp`, `/a2a`, `/sse`, and `/.well-known/agent[-card].json`
+suffixes; lowercase scheme and host; drop default ports; strip
+userinfo, query, and fragment. Path below the transport suffix is
+preserved, so sibling agents at different subpaths on a shared host
+(e.g. `https://publisher.com/.well-known/adcp/sales` vs
+`/.well-known/adcp/creative`) remain distinguishable. Closes #710.
--- a/.changeset/refs-resolve-observation-hygiene.md
+++ b/.changeset/refs-resolve-observation-hygiene.md
@@ -1,0 +1,31 @@
+---
+'@adcp/client': patch
+---
+
+`refs_resolve`: harden grader-visible observation and `actual.missing`
+payloads against hostile agent responses.
+
+Compliance reports may be published or forwarded to third parties, so
+every ref field emitted by the runner is now:
+
+- **Userinfo-scrubbed** on URL-keyed fields via WHATWG URL parsing plus
+  a regex fallback that scrubs `scheme://user:pass@` shapes embedded
+  in non-URL fields. Credentials planted in `agent_url` values can no
+  longer leak through compliance output.
+- **Scheme-restricted** on URL-keyed fields: non-`http(s)` schemes
+  (e.g. `javascript:`, `data:`, `file:`) are replaced with a
+  `<non-http scheme: …>` placeholder so downstream UIs rendering
+  `agent_url` as a link cannot inherit a stored-XSS vector.
+- **Length-capped** at 512 code points per string field, with a
+  code-point-boundary truncation that preserves surrogate pairs.
+- **Count-capped** at 50 observations per check, with an
+  `observations_truncated` marker when the cap fires. Meta
+  observations (`scope_excluded_all_refs`, `target_paginated`)
+  precede per-ref entries so the cap never drops primary signal.
+
+Match and dedup behavior is unchanged: the internal projection used
+for ref comparison is kept separate from the sanitized projection used
+for user-facing output, so truncation never false-collapses dedup
+keys. `refsMatch` and `projectRef` also now use `hasOwnProperty` to
+prevent storyboard authors from accidentally drawing match keys from
+`Object.prototype`. Closes #714.
--- a/.changeset/refs-resolve-scope-excluded-warning.md
+++ b/.changeset/refs-resolve-scope-excluded-warning.md
@@ -1,0 +1,10 @@
+---
+'@adcp/client': patch
+---
+
+`refs_resolve`: emit a `scope_excluded_all_refs` meta-observation when
+a scope filter partitions every source ref out. The integrity check
+enforces nothing when no ref falls in-scope; graders previously got a
+silent pass. The meta-observation surfaces the structural smell without
+changing pass/fail semantics. Suppressed under `on_out_of_scope: 'ignore'`
+(which explicitly opts out of scope warnings). Closes #711.
--- a/.changeset/refs-resolve-target-paginated.md
+++ b/.changeset/refs-resolve-target-paginated.md
@@ -1,0 +1,15 @@
+---
+'@adcp/client': patch
+---
+
+`refs_resolve`: detect paginated current-step targets and demote
+unresolved refs to observations instead of failing the check.
+
+Previously, when the target response carried `pagination.has_more:
+true`, any ref legitimately defined on a later page graded as
+`missing` — a false-positive failure against a conformant paginating
+seller. The runner now emits a `target_paginated` meta-observation and
+reports each would-be-missing ref as an `unresolved_with_pagination`
+observation, letting the check pass until the spec-level resolution
+lands (compliance mode requiring sellers to return everything
+referenced by products in a single response). Closes #712.
--- a/docs/TYPE-SUMMARY.md
+++ b/docs/TYPE-SUMMARY.md
@@ -1,7 +1,7 @@
 # AdCP Type Summary
 
-> Generated at: 2026-04-20
-> @adcp/client v5.7.0
+> Generated at: 2026-04-21
+> @adcp/client v5.8.0
 
 Curated reference of the types that matter for using the AdCP client. For full generated types see `src/lib/types/tools.generated.ts` and `src/lib/types/core.generated.ts`.
 
--- a/docs/llms.txt
+++ b/docs/llms.txt
@@ -1,7 +1,7 @@
 # Ad Context Protocol (AdCP)
 
 > Generated at: 2026-04-21
-> Library: @adcp/client v5.7.0
+> Library: @adcp/client v5.8.0
 > AdCP major version: 3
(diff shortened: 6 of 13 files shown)


bokelley:
Downstream review came back as **ship it** with three follow-ups. Filed:

- **#718** — pagination-bypass attack: a malicious seller claiming \`pagination.has_more: true\` indefinitely gets every unresolved ref demoted to an observation, so a grader scoring on \`refs_resolve.passed\` alone sees green. Same shape as #711's silent-no-op smell, different trigger. Substantive — needs a separate scoring signal at the grader level.
- **#719** — factor out the transport-suffix list (currently duplicated between \`SingleAgentClient.computeBaseUrl\` and \`canonicalizeAgentUrlForScope\`). Cleanup; low priority.

Two non-PR items the reviewer is taking themselves:
- Documenting the new observation kinds in upstream \`runner-output-contract.yaml\` against [adcontextprotocol/adcp](https://github.com/adcontextprotocol/adcp) once #717 ships.
- Updating their note in adcp#2601 to reflect landed canonicalization behavior.

Plus one nit deemed not worth fixing here: \`CREDENTIAL_SHAPE_RE\` doesn't catch \`scheme://@host\` (bare \`@\` with empty user). Edge case; mentioned for completeness.

PR #752, opened 2026-04-22
fix(testing): context.no_secret_echo walks structured auth + registerAssertion override

### LINKED ISSUE #751: context.no_secret_echo default is broken for structured auth objects; no way to override SDK defaults
Found while bumping `@adcp/client` 5.8.1 → 5.9.0 in [adcontextprotocol/adcp#2757](https://github.com/adcontextprotocol/adcp/pull/2757). Two related issues in `src/lib/testing/storyboard/default-invariants.ts` and `assertions.ts`.

## Bug 1: `context.no_secret_echo` default's auth extraction doesn't work for structured auth

In `default-invariants.ts:70-82`:

```ts
const optAny = ctx.options as unknown as { auth_token?: string; auth?: string; secrets?: string[] };
if (optAny.auth_token) secrets.add(optAny.auth_token);
if (optAny.auth) secrets.add(optAny.auth);
for (const s of optAny.secrets ?? []) secrets.add(s);
```

The type assertion claims `options.auth` is a `string`, but at runtime `TestOptions.auth` is the structured discriminated union [declared in `src/lib/testing/types.ts`](https://github.com/adcontextprotocol/adcp-client/blob/main/src/lib/testing/types.ts):

```ts
auth?:
  | { type: 'bearer'; token: string }
  | { type: 'basic'; username: string; password: string }
  | { type: 'oauth'; tokens: AgentOAuthTokens; client?: AgentOAuthClient }
  | { type: 'oauth_client_credentials'; credentials: AgentOAuthClientCredentials; tokens?: AgentOAuthTokens };
```

So `secrets.add(optAny.auth)` adds the whole object reference. Then in the scan:

```ts
const dumped = safeStringify(context);
for (const secret of secrets) {
  if (secret && dumped.includes(secret)) { ... }
}
```

`String.prototype.includes(obj)` coerces the object to `[object Object]` and matches nothing. Net effect: for any consumer passing structured `auth` (every consumer, in practice), the default `context.no_secret_echo` assertion is a no-op that always reports \`passed: true\`.

### Reproduction

```ts
import { registerAssertion, getAssertion } from '@adcp/client/testing';
import '@adcp/client/testing'; // triggers default-invariants.ts side-registration

const spec = getAssertion('context.no_secret_echo')!;
const ctx = {
  storyboard: {} as any,
  agentUrl: 'https://agent.example.com',
  options: { auth: { type: 'bearer' as const, token: 'SECRET_TOKEN_ABC' } } as any,
  state: {} as Record<string, unknown>,
};
spec.onStart!(ctx);
const result = spec.onStep!(ctx, {
  step_id: 's1',
  response: { echoed_token: 'SECRET_TOKEN_ABC' },
} as any);
// result is [{ passed: true }] — bug. The secret leaked, but the assertion didn't catch it.
```

### Fix

Walk the `auth` union and extract every leaf string secret:

```ts
function extractAuthSecrets(auth: unknown): string[] {
  if (!auth || typeof auth !== 'object') return [];
  const a = auth as Record<string, unknown>;
  const out: string[] = [];
  if (typeof a.token === 'string') out.push(a.token);               // bearer
  if (typeof a.password === 'string') out.push(a.password);          // basic
  if (a.tokens && typeof a.tokens === 'object') {                    // oauth, oauth_client_credentials
    const t = a.tokens as Record<string, unknown>;
    if (typeof t.access_token === 'string') out.push(t.access_token);
    if (typeof t.refresh_token === 'string') out.push(t.refresh_token);
  }
  if (a.client && typeof a.client === 'object') {                    // oauth confidential client
    const c = a.client as Record<string, unknown>;
    if (typeof c.client_secret === 'string') out.push(c.client_secret);
  }
  if (a.credentials && typeof a.credentials === 'object') {          // oauth_client_credentials
    const c = a.credentials as Record<string, unknown>;
    if (typeof c.client_secret === 'string') out.push(c.client_secret);
  }
  return out;
}
```

And in `onStart`:

```ts
const secrets = new Set<string>();
const optAny = ctx.options as unknown as { auth_token?: string; auth?: unknown; secrets?: string[] };
if (typeof optAny.auth_token === 'string') secrets.add(optAny.auth_token);
for (const s of extractAuthSecrets(optAny.auth)) secrets.add(s);
for (const s of optAny.secrets ?? []) if (typeof s === 'string') secrets.add(s);
ctx.state.secrets = secrets;
```

## Bug 2: No way for consumers to override an SDK default assertion

`registerAssertion(spec)` throws on duplicate id. The new `default-invariants.ts` side-registers three ids at `@adcp/client/testing` import time, so any consumer that also registers one of those ids (e.g. [our stricter `context.no_secret_echo` in adcp-1](https://github.com/adcontextprotocol/adcp/blob/main/server/src/compliance/assertions/context-no-secret-echo.ts)) hits:

```
Error: registerAssertion: "context.no_secret_echo" is already registered
```

The only escape is `clearAssertionRegistry()` — which nukes *all* three SDK defaults, so the consumer then has to re-register `idempotency.conflict_no_payload_leak` and `governance.denial_blocks_mutation` too, or lose them silently. That's fragile (requires knowing what the SDK shipped) and temporally coupled (SDK adds a fourth default → consumer forgets to re-register it).

### Fix

Add an `override` opt-in to `registerAssertion`:

```ts
export interface RegisterAssertionOptions {
  /**
   * Replace an existing registration for the same id instead of throwing.
   * Intended for consumers that want to override a default assertion
   * shipped by this package with a stricter implementation of their own.
   * Default \`false\` preserves throw-on-duplicate behavior.
   */
  override?: boolean;
}

export function registerAssertion(
  spec: AssertionSpec,
  options: RegisterAssertionOptions = {},
): void {
  if (!spec.id) throw new Error('registerAssertion: spec.id is required');
  if (registry.has(spec.id) && !options.override) {
    throw new Error(\`registerAssertion: "\${spec.id}" is already registered\`);
  }
  registry.set(spec.id, spec);
}
```

Backwards-compatible — single-arg call sites behave identically.

## Tests to add

- `context.no_secret_echo` catches a leaked \`auth.token\` (bearer), \`auth.password\` (basic), \`auth.tokens.access_token\` (oauth), \`auth.tokens.refresh_token\` (oauth), \`auth.client.client_secret\` (oauth confidential), \`auth.credentials.client_secret\` (oauth_client_credentials).
- `registerAssertion(spec, { override: true })` replaces an existing registration; the id is looked up via \`getAssertion\` to confirm it's the new spec.
- `registerAssertion(spec)` without the flag still throws on duplicate (regression guard for the default behavior).

## Versioning

Patch release — pure bug fix + additive backwards-compatible parameter.

## Downstream impact

[adcontextprotocol/adcp#2757](https://github.com/adcontextprotocol/adcp/pull/2757) (merged/pending) currently pins to 5.8.1. The bump to 5.9.0 is blocked on either this fix or a workaround (deleting our stricter local assertions and accepting the weaker SDK default — security regression we'd prefer to avoid).

Once this lands as e.g. 5.9.1:
- adcp-1 bumps @adcp/client
- adcp-1 adds \`{ override: true }\` to its three `registerAssertion` calls
- adcp-1 keeps its stricter implementations

[adcontextprotocol/adcp#2761](https://github.com/adcontextprotocol/adcp/issues/2761) (oauth_client_credentials server wiring) also needs this bump to unblock the SDK type, so merging this quickly unlocks two downstream PRs.

bokelley:
## Summary

Closes #751. Two related fixes for `@adcp/client/testing`:

- **`context.no_secret_echo` now catches real leaks.** The default registration in `src/lib/testing/storyboard/default-invariants.ts` typed `options.auth` as `string` and added the whole discriminated-union object to its secret set. `String.includes(obj)` coerces the object to `[object Object]` and matches nothing, so the assertion was a silent no-op for every consumer passing structured auth (bearer, basic, oauth, oauth_client_credentials). It now walks each variant and extracts leaf strings: `auth.token`, `auth.password`, `auth.tokens.{access,refresh}_token`, `auth.client.client_secret`, `auth.credentials.client_secret`.
- **`registerAssertion(spec, { override: true })`** replaces an existing registration instead of throwing. Lets consumers swap in a stricter local `context.no_secret_echo` (or any other SDK default) without `clearAssertionRegistry()` — which would nuke every other default and silently drop any new default the SDK ships later. Default behaviour (no options, or `{ override: false }`) is unchanged.

Patch release — pure bug fix + additive backwards-compatible parameter.

## Test plan

- [x] `node --test test/lib/storyboard-default-invariants.test.js test/lib/storyboard-assertion-registry.test.js` — 17 new secret-extraction cases (bearer/basic/oauth/oauth_client_credentials, including non-string `options.secrets` defensive coercion) and 3 new override cases (replaces existing, `{ override: false }` still throws, override on fresh id registers normally)
- [x] `npm test` — 5129/5129 tests pass, 0 failures
- [x] `npm run build` — TypeScript compiles, type export for `RegisterAssertionOptions` added to `@adcp/client/testing`
- [ ] CI: lint, build, tests, changeset check

## Downstream unblock

- [adcontextprotocol/adcp#2757](https://github.com/adcontextprotocol/adcp/pull/2757) can bump `@adcp/client` to the next patch release and add `{ override: true }` to its three local `registerAssertion` calls — keeping the stricter server-side implementations instead of losing them to the silent SDK default.
- [adcontextprotocol/adcp#2761](https://github.com/adcontextprotocol/adcp/issues/2761) (oauth_client_credentials wiring) needs the same bump.

### DIFF
--- a/.changeset/fix-secret-echo-and-override.md
+++ b/.changeset/fix-secret-echo-and-override.md
@@ -1,0 +1,32 @@
+---
+'@adcp/client': patch
+---
+
+fix(testing): `context.no_secret_echo` walks structured `TestOptions.auth`, and `registerAssertion` accepts `{ override: true }`
+
+- The default `context.no_secret_echo` assertion in `@adcp/client/testing`
+  previously treated `options.auth` as a string and added the whole
+  discriminated-union object to its secret set. `String.includes(obj)`
+  against `[object Object]` matched nothing, so the assertion was
+  effectively a no-op for every consumer passing structured auth (bearer,
+  basic, oauth, oauth_client_credentials). It now extracts the leaf
+  secrets across every variant:
+  - bearer: `token`
+  - basic: `username`, `password`, and the base64 `user:pass` blob an
+    `Authorization: Basic` header would carry
+  - oauth: `tokens.access_token`, `tokens.refresh_token`,
+    `client.client_secret` (confidential clients)
+  - oauth_client_credentials: `credentials.client_id` and
+    `credentials.client_secret` — resolving `$ENV:VAR` references to their
+    runtime values so echoes of the real secret (not the reference string)
+    are caught — plus `tokens.access_token` / `tokens.refresh_token`
+
+  A minimum-length guard (8 chars) skips substring matching on fixture
+  values that would otherwise collide with benign JSON.
+
+- `registerAssertion(spec, { override: true })` now replaces an existing
+  registration instead of throwing. Lets consumers swap in a stricter
+  version of an SDK default (e.g. their own `context.no_secret_echo`)
+  without calling `clearAssertionRegistry()` and re-registering every other
+  default. Default behaviour (`{ override: false }` / no options) is
+  unchanged and still throws on duplicate ids.
--- a/src/lib/testing/index.ts
+++ b/src/lib/testing/index.ts
@@ -200,6 +200,7 @@ export {
   type AssertionSpec,
   type AssertionContext,
   type AssertionResult,
+  type RegisterAssertionOptions,
   // Types
   type Storyboard,
   type StoryboardPhase,
--- a/src/lib/testing/storyboard/assertions.ts
+++ b/src/lib/testing/storyboard/assertions.ts
@@ -77,15 +77,27 @@ export interface AssertionSpec {
 
 const registry = new Map<string, AssertionSpec>();
 
+export interface RegisterAssertionOptions {
+  /**
+   * Replace an existing registration for the same id instead of throwing.
+   * Intended for consumers that want to override a default assertion shipped
+   * by this package (e.g. a stricter `context.no_secret_echo`) without
+   * clearing the whole registry and re-registering every other default.
+   * Defaults to `false` so the throw-on-duplicate behaviour still catches
+   * two modules fighting over the same id unintentionally.
+   */
+  override?: boolean;
+}
+
 /**
- * Register an assertion. Throws on duplicate id — re-registration is almost
- * always a sign of two modules fighting over the same id, not an intent to
- * override. Tests that want to replace a registration should call
- * `clearAssertionRegistry()` first.
+ * Register an assertion. Throws on duplicate id unless `options.override` is
+ * set — re-registration is almost always a sign of two modules fighting over
+ * the same id, not an intent to override. Consumers that want to replace a
+ * default shipped by this package should pass `{ override: true }`.
  */
-export function registerAssertion(spec: AssertionSpec): void {
+export function registerAssertion(spec: AssertionSpec, options: RegisterAssertionOptions = {}): void {
   if (!spec.id) throw new Error('registerAssertion: spec.id is required');
-  if (registry.has(spec.id)) {
+  if (registry.has(spec.id) && !options.override) {
     throw new Error(`registerAssertion: "${spec.id}" is already registered`);
   }
   registry.set(spec.id, spec);
--- a/src/lib/testing/storyboard/default-invariants.ts
+++ b/src/lib/testing/storyboard/default-invariants.ts
@@ -294,11 +294,17 @@ function extractAdcpError(step: import('./types').StoryboardStepResult): AdcpErr
 }
 
 /**
- * Minimum secret length before substring matching runs. Anything shorter is
- * almost certainly a test fixture value that would collide with benign JSON.
- * Bearer / OAuth access tokens in practice are well above this threshold.
+ * Minimum secret length before substring matching runs. Set to 16 because
+ * real OAuth access tokens, refresh tokens, and signing secrets are ≥20 chars
+ * by convention (opaque UUIDs, JWTs, HMAC hex). A shorter floor would
+ * false-positive on benign identifiers (short usernames, environment names,
+ * 8-char hex prefixes, ISO timestamps). Hand-coded fixture keys like
+ * `"test-key"` below this bar are not caught — that's the right tradeoff:
+ * a real agent echoing such a value is an obvious leak a human reviewer
+ * would spot, and the collision cost of matching short strings against
+ * every response context is too high to justify the coverage.
  */
-const SECRET_MIN_LENGTH = 8;
+const SECRET_MIN_LENGTH = 16;
 
 const ENV_REFERENCE_PREFIX = '$ENV:';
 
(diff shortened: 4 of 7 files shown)


bokelley:
Reviewed. The core fix is right — `extractAuthSecrets` walking each variant and pushing leaf strings is exactly what the bug needed, the `$ENV:` resolution (resolved literal wins, reference string doesn't) is the right semantic, `{ override: true }` matches what downstream consumers need, and the test coverage is thorough. Two substantive concerns before merge:

## 1. `client_id` is not a secret (RFC 6749 §2.2) — should-fix

`extractAuthSecrets` treats `credentials.client_id` on `oauth_client_credentials` as a secret:

```ts
pushCredentialValue(out, c.client_id);
pushCredentialValue(out, c.client_secret);
```

Per RFC 6749 §2.2:

> The authorization server issues the registered client a *client identifier* — a unique string representing the registration information provided by the client. The client identifier is not a secret; it is exposed to the resource owner and MUST NOT be used alone for client authentication.

Any legitimate flow can echo `client_id` in logs, redirects, error responses, audience claims, introspection payloads, etc. Treating it as a secret will false-positive in realistic storyboards — particularly against IdPs that echo the requesting client back in token responses. Drop it:

```ts
if (a.credentials && typeof a.credentials === 'object') {
  const c = a.credentials as Record<string, unknown>;
  pushCredentialValue(out, c.client_secret);
  // client_id is a public identifier (RFC 6749 §2.2), not a secret — do
  // not extract. Echoes of client_id in token responses, introspection
  // payloads, audit logs, and error bodies are intentional, not leaks.
}
```

## 2. Basic auth `username` extraction — should-fix

Same pattern for basic auth:

```ts
if (typeof a.username === 'string' && a.username && typeof a.password === 'string' && a.password) {
  out.push(a.username);
  out.push(a.password);
  out.push(Buffer.from(`${a.username}:${a.password}`, 'utf8').toString('base64'));
}
```

The `base64(user:pass)` blob is genuinely sensitive — it IS the `Authorization: Basic` header value. Password alone is the secret. But username alone is a public identifier in every practical sense — welcome messages, audit logs, "last login by <user>" displays all legitimately echo it. A storyboard that exercises a governance call against an agent that echoes `requested_by: "alice"` in its audit trail would false-positive.

Suggest:

```ts
if (typeof a.password === 'string' && a.password) {
  out.push(a.password);
  if (typeof a.username === 'string' && a.username) {
    out.push(Buffer.from(`${a.username}:${a.password}`, 'utf8').toString('base64'));
  }
}
// Note: username is not extracted alone — it's a public identifier. The
// base64 header catches echoes of the full Authorization: Basic value.
```

Tests on the `username` extraction path (`catches a leaked basic auth password` uses `SECRET_PW_456` and doesn't currently assert anything about the username) don't need to change if you drop the extraction — the password case still covers the primary attack surface.

## 3. `SECRET_MIN_LENGTH = 8` — minor

8 is low enough that many non-secret identifiers (8-char usernames, short environment names, timestamps, UUIDs prefixes) can still drive false positives. Real bearer / OAuth access tokens are ≥20 chars by convention; 16 feels like a more honest floor. If you keep it at 8 because you want to catch hand-coded test keys like `"test-key"`, document the choice explicitly in the comment — current comment says 8 "keeps false positives out" which isn't quite true.

Not blocking; just clarify or raise.

## Everything else

- `pushCredentialValue` + `resolveEnvReference` are well-shaped helpers. The "resolve at runtime, match the real value not the reference" semantic on `$ENV:` is exactly right, and the test case asserting the literal reference is NOT flagged nails the contract.
- `{ override: true }` interface is minimal and exactly what the downstream case needs. Type export wired through `testing/index.ts`.
- Test file strategy (`basicAuth(user, pw)` helper to split literals across properties) mirrors what adcp-1 landed in #2757 for the same GitGuardian reason. Good carry-through.
- Changeset is correct (patch).

With (1) and (2) addressed, this is clean. (3) is a judgment call — flag or raise, either works.

bokelley:
Addressed all three points in `3fdfc1cd`:

1. **Dropped `client_id` extraction** on `oauth_client_credentials.credentials` — RFC 6749 §2.2 explicitly: client_id is a public identifier. Added a pinning test that confirms an echoed client_id does NOT flag.
2. **Dropped basic-auth username-alone extraction**. Kept password and the `base64(user:pass)` blob (which IS the `Authorization: Basic` header value). Added a pinning test that confirms username-alone does NOT flag; removed the obsolete test that asserted it did.
3. **Raised `SECRET_MIN_LENGTH` from 8 to 16.** Updated the jsdoc to document the tradeoff honestly — hand-coded short test keys below the floor are not caught, which is the right cost for avoiding false positives against short identifiers, environment names, and hex prefixes.

All fixture secrets extended to ≥16 chars to clear the new floor. Full test suite green (5142 passing, 1 flaky perf test unrelated — passes on rerun).

PR #757, opened 2026-04-22
feat(server): strict response validation default in dev/test (#727)

### LINKED ISSUE #727: Tighten generated types + default response validation to catch schema drift at the right layer
## Problem

PR #725 (response builders for `acquire_rights`, `sync_accounts`, `sync_governance`, `report_usage`) initially added runtime guards for schema-level constraints — `credentials.minLength: 32`, `authentication.schemes.length === 1`, `account_id` required, `accepted` required, `creative_manifest.format_id` object shape. All five are already JSON Schema constraints.

Test-agent-team review (verbatim):

> Builders doing schema validation is putting logic in the wrong layer. JSON Schema in the spec is the source of truth. Generated Zod validates every response on the wire — rejects sparse objects, enforces `minLength: 32`, checks `format_id` shape. One layer, enforced for every tool automatically. Generated TypeScript types are tight enough that sparse `{ rights_id, status: 'acquired' }` fails `tsc`. SDK builders do one job: wrap domain objects in the MCP envelope.

The right architecture is three layers. Today's state:

| Layer | Source of truth | Exists? | Default-on? |
|---|---|---|---|
| JSON Schema | `schemas/cache/latest/**` | ✅ | ✅ |
| Generated Zod | `src/lib/types/schemas.generated.ts` | ✅ | Opt-in via `createAdcpServer({ validation: { responses: 'strict' } })` (PR #694) |
| Generated TS | `src/lib/types/core.generated.ts`, `tools.generated.ts` | ✅ | Loose — sparse `{ rights_id, status: 'acquired' }` passes `tsc` because the generator emits open unions |

PR #725 has been refactored to strip runtime guards from builders — they're now thin envelope wrappers. This issue tracks the real fix.

### A. Default response validation in dev/test

- `createAdcpServer({ validation: { responses: 'warn' | 'strict' } })` currently defaults to `'off'`.
- Change default to `'strict'` when `process.env.NODE_ENV !== 'production'`, or emit a one-time stderr nudge when unset.
- Measure: run the `compliance:skill-matrix` harness pre/post — every schema-drift failure that PR #725 was papering over should surface here, as a single validator error with the offending field path.

### B. Tighten generated discriminated unions

Per test-agent-team: `{ rights_id, status: 'acquired' }` currently passes `tsc` against `AcquireRightsResponse`. Ideally it fails at compile time with a missing-required-field error.

Investigation needed:
- Does the JSON Schema-to-TS generator emit `oneOf` as a true discriminated union with branch-required fields?
- If not, is the fix in the generator, or in the source JSON Schemas (`required` declarations, discriminator hints)?
- Separately, check the Zod generator — does it emit a discriminated-union validator that'd reject the sparse shape on the wire even if `tsc` doesn't?

### C. Remove the last two builder guards (already done in the PR #725 refactor, but track)

- `mediaBuyResponse.assertNoTopLevelSetup` (`responses.ts:91`) predates this review but has the same architectural issue. Candidate for removal once (A) + (B) catch it.

## Why now

The `compliance:skill-matrix` harness keeps surfacing `SERVICE_UNAVAILABLE` errors from Claude-built and hand-built agents shipping responses missing required fields. Every occurrence is a drift between what the agent builds and what the schema demands. Catching it at the validator layer fixes it once for every tool; patching it in builders fixes it for one.

## Related

- PR #725 (response builders, refactored to thin wrappers)
- PR #694 (`validation: { requests, responses }` config, opt-in)

bokelley:
## Summary

- `createAdcpServer({ validation: { responses } })` now defaults to `'strict'` when `NODE_ENV !== 'production'` (was `'warn'`). Handler drift fails with `VALIDATION_ERROR` carrying the field pointer instead of silently logging a warning.
- Production default is unchanged: `'off'`. Explicit per-call config still wins.
- Three internal test files opt out (`validation: { responses: 'off' }`) because they use deliberately sparse fixtures to exercise envelope/state-store/idempotency middleware — schema conformance isn't their unit under test. `test/lib/schema-validation-server.test.js` continues to cover the validator itself.

## Why

The `compliance:skill-matrix` harness keeps surfacing `SERVICE_UNAVAILABLE` from agents that returned spec-incomplete responses (see the test-agent-team review on #725). The dispatcher's response validator is the right layer — it catches drift once for every tool with a clear field pointer. Making it strict by default in dev/test catches drift while the handler is being written instead of in a downstream consumer.

Closes #727(A). Part (B) (tightening generated discriminated unions in the types generator) remains open.

## Test plan

- [x] `npm test` — 5135 pass / 0 fail (baseline before flip: same count)
- [x] `test/lib/schema-validation-server.test.js` still asserts the three modes (`strict` / `warn` / `off`) behave as documented
- [ ] Post-merge: rerun `npm run compliance:skill-matrix` and confirm any residual drift surfaces as a single validator error per tool (the measurement the issue called for)

### DIFF
--- a/.changeset/strict-response-validation-default.md
+++ b/.changeset/strict-response-validation-default.md
@@ -1,0 +1,32 @@
+---
+'@adcp/client': minor
+---
+
+feat(server): response validation defaults to `'strict'` outside production
+
+`createAdcpServer({ validation: { responses } })` previously defaulted to
+`'warn'` when `NODE_ENV !== 'production'`. It now defaults to `'strict'`
+in dev/test/CI so handler-returned schema drift fails with
+`VALIDATION_ERROR` (with the offending field path in `details.issues`)
+instead of logging a warning the caller can silently ignore.
+
+Production behaviour is unchanged: the default stays `'off'` when
+`NODE_ENV === 'production'`, so prod request paths pay no validation
+cost. Pass `validation: { responses: 'warn' }` to restore the previous
+dev-mode behaviour; `validation: { responses: 'off' }` opts out
+entirely.
+
+Why: the `compliance:skill-matrix` harness has repeatedly surfaced
+`SERVICE_UNAVAILABLE` from agents whose responses fail the wire schema.
+The dispatcher's response validator catches this drift with a clear
+field pointer, one layer that every tool inherits automatically. Making
+that the default catches it during handler development rather than in a
+downstream consumer.
+
+Migration: handler tests that use sparse fixtures (e.g.
+`{ products: [{ product_id: 'p1' }] }`) will start returning
+`VALIDATION_ERROR`. Either fill in the missing required fields to match
+the AdCP schema, or set `validation: { responses: 'off' }` on the test
+server to keep the fixture intentionally minimal.
+
+Closes #727 (A).
--- a/.changeset/strict-response-validation-default.md
+++ b/.changeset/strict-response-validation-default.md
@@ -27,6 +27,17 @@ Migration: handler tests that use sparse fixtures (e.g.
 `{ products: [{ product_id: 'p1' }] }`) will start returning
 `VALIDATION_ERROR`. Either fill in the missing required fields to match
 the AdCP schema, or set `validation: { responses: 'off' }` on the test
-server to keep the fixture intentionally minimal.
+server to keep the fixture intentionally minimal. Note that Node's
+test runner does **not** set `NODE_ENV`, so test suites running under
+`node --test` (with `NODE_ENV=undefined`) fall into the dev/test
+bucket and will start validating responses — this is intentional.
+
+Also: the `VALIDATION_ERROR` envelope's `details.issues[].schemaPath`
+is now gated behind `exposeErrorDetails` (same policy as the existing
+`SERVICE_UNAVAILABLE.details.reason` field). Production responses no
+longer leak `#/oneOf/<n>/properties/...` paths that fingerprint the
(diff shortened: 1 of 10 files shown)


bokelley:
Rebased on main and addressed expert review feedback in 6dfcb53b.

**code-reviewer** (PR doc + correctness):
- JSDoc clarified that `warn` still logs; `strict` additionally promotes to `VALIDATION_ERROR`
- `idempotency.release()` catch now logs via `logger.warn` (parity with the sibling post-handler path at `:1938`)
- Three test helpers now shallow-merge `validation` so a per-test override on one side doesn't silently clear the other
- Added four dispatcher-level canary tests in `test/lib/schema-validation-server.test.js` asserting the NODE_ENV-driven default flip and the schemaPath gating behavior
- Changeset now calls out `NODE_ENV=undefined` explicitly (Node's test runner doesn't set it)

**security-reviewer** (details.issues[].schemaPath leak):
- `details.issues[].schemaPath` is now gated behind `exposeErrorDetails` (same policy as `SERVICE_UNAVAILABLE.details.reason`). Production responses no longer leak `#/oneOf/<n>/properties/...` that fingerprints the handler's internal branch selection. Buyers still get `pointer`, `message`, `keyword`.
- M-2 replay-loop finding filed as #758 (pre-existing, affects `warn` mode users too — deserves its own circuit-breaker-vs-cache design).

**debugger** passed the PR clean — no regressions, no timing-sensitive interactions with `NODE_ENV` capture.

`npm test` local: 5158 pass / 0 fail after rebase + fixes.

Issue #793, opened 2026-04-22
Storyboard request-builders emit out-of-spec shapes for log_event / create_media_buy (drops packages[1+], wrong field names)


bokelley:
## Problem

Two of the storyboard `REQUEST_BUILDERS` in `src/lib/testing/storyboard/request-builder.ts` synthesize payloads that don't match the published AdCP schema, and don't look at `step.sample_request` before doing so. Framework-dispatch agents (that run zod at the MCP boundary) reject these with `-32602 invalid_type`. Legacy-dispatch agents let them through only because their handlers are permissive at the transport layer.

This surfaced while diagnosing adcontextprotocol/adcp#2872, which had hypothesized `\$context.*` macro resolution as the culprit. That hypothesis is wrong — the pre-MCP payloads have no unresolved macros. The builders themselves are emitting non-spec shapes.

Full write-up with pre-MCP payload dumps and reproduction harness: adcontextprotocol/adcp#2872 (comment: https://github.com/adcontextprotocol/adcp/issues/2872#issuecomment-4294260975).

## Bug A — `log_event` builder emits `timestamp` + nested `value`

`src/lib/testing/storyboard/request-builder.ts:256-268`:

```ts
log_event(_step, context, _options) {
  return {
    event_source_id: context.event_source_id ?? 'test-source',
    events: [
      {
        event_id: \`evt-\${Date.now()}\`,
        event_type: 'purchase',
        timestamp: new Date().toISOString(),        // ← spec: event_time (REQUIRED)
        value: { amount: 49.99, currency: 'USD' },  // ← spec: custom_data.value (number) + custom_data.currency (string)
      },
    ],
  };
}
```

`media-buy/log-event-request.json` requires `events[].event_time` as a string and places `value`/`currency` under `custom_data`. Under framework dispatch, zod reports:

```
MCP error -32602: Input validation error: Invalid arguments for tool log_event: [
  { \"expected\": \"string\", \"code\": \"invalid_type\", \"path\": [\"events\", 0, \"event_time\"],
    \"message\": \"Invalid input: expected string, received undefined\" }
]
```

The builder also never consults `_step.sample_request`, so storyboards (e.g. `sales_catalog_driven`, `sales_social`) that author a complete spec-conformant `sample_request` with `event_time`, `content_ids`, and siblings never see those bytes go out.

**Affects storyboards:** `sales_catalog_driven`, `sales_social`.

## Bug B — `create_media_buy` builder drops `packages[1+]`

[...]

PR #804, opened 2026-04-22
fix(testing): storyboard builders emit spec shapes for si_* + sync_governance (closes #802)

bokelley:
## Summary
Closes #802. Three storyboard `REQUEST_BUILDERS` emitted payloads the generated Zod schema rejects; strict-validation (framework-dispatch) agents surface this as `-32602 invalid_type`, while legacy-dispatch accepts them permissively and masks the bug.

- `si_get_offering`: drop the string `context` and the out-of-schema `identity`; emit the prose string as optional `intent` (`core/context.json` is an object, not a string).
- `si_initiate_session`: move the prose string from `context` to the required `intent` field. Schema requires `idempotency_key` + `intent` + `identity`; the first is handled by `applyIdempotencyInvariant`.
- `sync_governance`: lengthen default `authentication.credentials` to meet `minLength: 32`; honor `step.sample_request` so fixtures like `signal-marketplace/scenarios/governance_denied.yaml` that author `url: $context.governance_agent_url` flow through.

All three now honor `step.sample_request` first, matching peer builders (`build_creative`, `sync_creatives`, `sync_plans`, `check_governance`, `list_creative_formats`, etc.). Same bug class as #789 and #793.

## Test plan
- [x] `npx tsc --noEmit` clean
- [x] Pre-push validation passed
- [ ] Downstream compliance suite (adcontextprotocol/adcp CI) picks up the fix once published — Group C storyboards touching SI + governance should stop failing the strict-validation dispatch with `invalid_type`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #806, opened 2026-04-22
fix(testing): spec shapes for si_* + sync_governance requests (#802)

### LINKED ISSUE #802: fix(testing): storyboard builders emit spec-violating shapes for si_* + sync_governance
## Summary

Expert review during #793 / #794 identified three additional storyboard `REQUEST_BUILDERS` that emit payloads the generated Zod schema rejects. Framework-dispatch agents running strict validation at the MCP boundary reject these with `-32602 invalid_type`; legacy-dispatch permissively accepts them (same pattern as #793).

### 1. `si_get_offering` passes `context` as a string

`src/lib/testing/storyboard/request-builder.ts:597-606`:

```ts
si_get_offering(_step, _context, options) {
  return {
    offering_id: options.si_offering_id ?? 'e2e-test-offering',
    context: options.si_context ?? 'E2E testing - checking SI offering availability',  // ← spec: object
    identity: { ... },  // ← not in request schema
  };
}
```

Per `schemas/cache/latest/sponsored-intelligence/si-get-offering-request.json`, `context` is a `$ref: /schemas/latest/core/context.json` (object, not string). The human-language string should go in `intent` (optional in this schema). The `identity` field isn't in the schema at all.

### 2. `si_initiate_session` uses wrong field names + missing required `intent`

`src/lib/testing/storyboard/request-builder.ts:594-608`:

```ts
si_initiate_session(_step, context, options) {
  return {
    offering_id: ...,
    offering_token: ...,
    identity: { consent_granted: true, user: { principal: ... } },
    context: options.si_context ?? 'E2E test session',  // ← wrong: context is an object
    placement: ...,
    supported_capabilities: ...,
  };
}
```

Per `schemas/cache/latest/sponsored-intelligence/si-initiate-session-request.json`, required fields are `idempotency_key`, `intent`, `identity`. The builder stuffs the prose string into `context` where `intent` belongs. `identity` shape should follow `si-identity.json`. Missing `idempotency_key` is handled by `applyIdempotencyInvariant`, so that one's fine.

### 3. `sync_governance` credentials too short + ignores sample_request

`src/lib/testing/storyboard/request-builder.ts:432-443`:

```ts
authentication: {
  schemes: ['Bearer'],
  credentials: 'test-governance-token',  // ← 21 chars, schema requires minLength: 32
},
```

Per `schemas/cache/latest/account/sync-governance-request.json`, `authentication.credentials` has `minLength: 32` and `additionalProperties: false` on the enclosing `governance_agents` items. The builder also never consults `step.sample_request`, so `signal-marketplace/scenarios/governance_denied.yaml` fixtures that author `url: \$context.governance_agent_url` silently don't make it through.

## Fix

All three should honor `step.sample_request` first (same convention as other builders). The synthetic fallbacks should match spec:
- `si_get_offering`: drop `context` string + `identity`; use `intent` for the prose string.
- `si_initiate_session`: move prose string from `context` → `intent`.
- `sync_governance`: lengthen default credentials to ≥32 chars; honor sample_request when present.

## Context

- Surfaced during expert review of #794 (adcp-client fix for #793)
- Related: #792 (`get_rights`), #789 (`list_creative_formats`), #793 (`log_event` + `create_media_buy`) — same bug class (builders ignoring sample_request or emitting non-spec shapes)
- Structural fix proposed separately: see companion issue for the runner precedence inversion that would prevent this class of bug

bokelley:
## Summary

Fixes three spec-violating request shapes that framework-dispatch agents running strict validation at the MCP boundary rejected with `-32602 invalid_type`. Legacy-dispatch permissively accepted them (same pattern as #793).

- **`si_get_offering`**: prose string → `intent` (per `si-get-offering-request.json`, `context` is the core context object, not a prose string); drop non-spec `identity` field.
- **`si_initiate_session`**: prose handoff → `intent` (required); `identity` now follows `si-identity.json` — `{ consent_granted, user: { name } }` rather than the non-spec `{ principal, device_id }`.
- **`sync_governance`**: builder now honors `step.sample_request` (so governance storyboards can pin `url: $context.governance_agent_url`); fallback `credentials` padded to ≥32 chars to satisfy the schema's `minLength: 32`.

Same shape bugs existed in the sponsored-intelligence test scenarios path (agent-tester orchestrator) — fixed in both places so both code paths produce spec-compliant requests.

## Test plan

- [x] Builder fallback tests: `npm run test:lib` (4575 pass)
- [x] New regression tests in `test/lib/request-builder.test.js`:
  - `sync_governance` fallback credentials ≥32 chars
  - `sync_governance` honors `$context` placeholders in sample_request
  - `si_get_offering` emits `intent` (not string `context`) and omits `identity`
  - `si_initiate_session` emits `intent` + spec-shaped `identity`
  - Both SI builders honor `step.sample_request` when present

Closes #802.

### DIFF
--- a/.changeset/fix-si-sync-governance-request-shapes.md
+++ b/.changeset/fix-si-sync-governance-request-shapes.md
@@ -1,0 +1,23 @@
+---
+'@adcp/client': patch
+---
+
+Fix spec-violating request shapes for `si_get_offering`, `si_initiate_session`,
+and `sync_governance` in both the storyboard request builders and the
+sponsored-intelligence test scenarios.
+
+- `si_get_offering`: the prose string moves from `context` (which is the core
+  context object per `si-get-offering-request.json`) to `intent`. The `identity`
+  field is dropped — it is not part of the request schema.
+- `si_initiate_session`: the prose handoff moves from `context` to `intent`
+  (required). `identity` now follows `si-identity.json` — `{ consent_granted,
+  user: { name } }` rather than the non-spec `{ principal, device_id }`.
+- `sync_governance`: the builder now honors `step.sample_request` (so
+  storyboards can pin `url: $context.governance_agent_url`) and the fallback
+  `credentials` is padded to ≥32 chars to satisfy the schema's `minLength: 32`.
+
+Framework-dispatch agents running strict validation at the MCP boundary
+previously rejected these with `-32602 invalid_type`; legacy-dispatch
+permissively accepted the wrong shapes.
+
+Closes #802.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@adcp/client",
-  "version": "5.10.0",
+  "version": "5.11.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@adcp/client",
-      "version": "5.10.0",
+      "version": "5.11.0",
       "license": "Apache-2.0",
       "dependencies": {
         "ajv": "^8.18.0",
--- a/src/lib/testing/scenarios/sponsored-intelligence.ts
+++ b/src/lib/testing/scenarios/sponsored-intelligence.ts
@@ -71,11 +71,7 @@ export async function testSISessionLifecycle(
       async () =>
         client.siGetOffering({
           offering_id: offeringId,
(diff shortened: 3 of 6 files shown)


bokelley:
Duplicate of #804 (which landed first). I'll layer the scenarios-path fix and regression tests onto #804.

PR #807, opened 2026-04-22
test(testing): schema round-trip invariant for every request builder (#803)

### LINKED ISSUE #803: test(testing): add schema-driven round-trip invariant for every request builder
## Problem

Four PRs in the last two weeks (#789, #792, #793/#794, and the soon-to-land follow-up for `si_*` + `sync_governance`) have fixed individual storyboard `REQUEST_BUILDERS` that emitted shapes the generated Zod schema rejects. Each bug was discovered in the wild — downstream framework-dispatch agents reported `-32602 invalid_type` or silent test failures, and diagnosis required running the full storyboard harness.

The existing `test/lib/request-builder.test.js` suite asserts field-level shape per-builder, but the assertions are hand-authored per bug. Nothing prevents the next builder (or a regression in an existing one) from drifting out of spec.

## Proposal

Add a schema-driven invariant test: for every task in `REQUEST_BUILDERS`, round-trip the builder's fallback output (empty context, empty sample_request) through the matching Zod schema from `src/lib/types/schemas.generated.ts` and assert `safeParse` succeeds. Do the same for a representative filled-context fallback.

```ts
import { buildRequest, hasRequestBuilder } from '../../dist/lib/testing/storyboard/request-builder.js';
import * as Schemas from '../../dist/lib/types/schemas.generated.js';

const TASK_TO_SCHEMA = {
  create_media_buy: Schemas.CreateMediaBuyRequestSchema,
  log_event: Schemas.LogEventRequestSchema,
  sync_governance: Schemas.SyncGovernanceRequestSchema,
  si_get_offering: Schemas.SiGetOfferingRequestSchema,
  // ... one entry per builder
};

for (const [task, schema] of Object.entries(TASK_TO_SCHEMA)) {
  test(\`\${task} fallback matches request schema\`, () => {
    const req = buildRequest(step(task), {}, DEFAULT_OPTIONS);
    const result = schema.safeParse(req);
    assert.ok(result.success, result.success ? '' : JSON.stringify(result.error.issues, null, 2));
  });
}
```

## Caveats

- Some requests need `idempotency_key` (mutating tasks) to parse. The builder doesn't mint it — `applyIdempotencyInvariant` in the runner does. Either inject a synthetic key before parsing, or use the schema minus the `idempotency_key` constraint for this test.
- Some builders legitimately emit empty `{}` (e.g. `list_creative_formats`, `get_adcp_capabilities`) because all fields are optional. The schema should accept that.
- Fixture-honoring builders (`update_media_buy`, `sync_catalogs`, etc.) should also be tested with a representative `sample_request` to ensure the fixture path round-trips too.

## Why now

- #793 would have been caught pre-PR (zod rejects `timestamp` vs `event_time`).
- The companion `si_*` / `sync_governance` issue (filed separately) would have been caught the same way.
- Prevents the next drift before it ships.

## References

- #789, #792, #793 / #794 — symptoms
- Structural fix to builder/sample_request precedence (companion issue) — preferred long-term fix; this test is the defense-in-depth safety net that catches regressions regardless

bokelley:
## Summary

- Adds `test/lib/request-builder-schema-roundtrip.test.js` — for every task with a storyboard request builder AND a generated Zod request schema, builds the fallback request (empty context + options + synthetic `idempotency_key` where the schema requires one) and asserts it parses cleanly. New builders are picked up automatically by iterating `TOOL_REQUEST_SCHEMAS`.
- Catches the #789 / #792 / #793 / #794 class of drift before it ships — any builder that emits off-spec field names or types now fails CI.
- Fixes eight pre-existing fallbacks the invariant surfaced (full list in the changeset + commit message).

## Why now

Four PRs in the last two weeks each fixed a single builder that had silently drifted out of spec. Each bug was caught downstream, not locally. The invariant makes that entire class local-failable — and brings the current builders back onto spec as a side effect.

## Test plan

- [x] `node --test test/lib/request-builder-schema-roundtrip.test.js` — 41/41 pass
- [x] `node --test test/lib/request-builder.test.js` — 38/38 pass (one assertion updated for `activate_signal` destinations default)
- [x] Full `test/lib/*.test.js` suite — 4588 pass, 0 fail
- [x] `npm run typecheck`, `npm run format:check`, `npm run lint` clean

Closes #803.

### DIFF
--- a/.changeset/fix-si-gov-storyboard-builders.md
+++ b/.changeset/fix-si-gov-storyboard-builders.md
@@ -1,0 +1,30 @@
+---
+'@adcp/client': patch
+---
+
+Storyboard runner: fix spec-violating shapes and `sample_request`
+precedence across the SI + governance request builders. All affected
+builders now honor `step.sample_request` first (matching peer builders),
+and their synthetic fallbacks conform to the generated Zod schemas so
+framework-dispatch agents running strict validation at the MCP boundary
+no longer reject them with `-32602 invalid_type`.
+
+- `si_get_offering`: drop the string `context` and the out-of-schema
+  `identity`; emit the prose string as optional `intent` (per
+  `si-get-offering-request.json`, `context` is a ref to an object).
+- `si_initiate_session`: move prose from `context` (which must be an
+  object) to required `intent`; default the identity fallback to the
+  realistic anonymous handoff shape (`consent_granted: false` +
+  `anonymous_session_id`) instead of `consent_granted: true` with an
+  empty consented user — spec-legal either way, but the anonymous shape
+  is what a host that hasn't obtained PII consent actually sends.
+- `si_send_message` / `si_terminate_session`: honor `sample_request` so
+  storyboards can drive `action_response`, `handoff_transaction`,
+  `termination_context`, and non-default `reason` paths without the
+  fallback stomping the scenario.
+- `sync_governance`: lengthen default `authentication.credentials` to
+  meet `minLength: 32`, and honor `sample_request` so fixtures like
+  `signal-marketplace/scenarios/governance_denied.yaml` that author
+  `url: $context.governance_agent_url` flow through.
+
+Closes #802.
--- a/.changeset/public-wrap-envelope.md
+++ b/.changeset/public-wrap-envelope.md
@@ -1,0 +1,7 @@
+---
+'@adcp/client': minor
+---
+
+Expose `wrapEnvelope` from `@adcp/client/server` — a public helper for attaching AdCP envelope fields (`replayed`, `context`, `operation_id`) to handler responses, with error-code-specific field allowlists (e.g., IDEMPOTENCY_CONFLICT drops `replayed`). Promoted for sellers that wire their own MCP / A2A handlers without the framework.
+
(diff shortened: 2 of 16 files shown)


bokelley:
Addressed review feedback in 9bf70896:

**Must-fixes (nodejs-testing-expert):**
- Replaced local `requiresIdempotencyKey` with imported `MUTATING_TASKS` from `src/lib/utils/idempotency.ts`; the two brand-rights tasks outside `TOOL_REQUEST_SCHEMAS` go through a small `EXTRA_MUTATING` set. No more Zod-internals coupling in the test.
- Replaced the `>= 30` floor with an enumerated `EXPECTED_COVERED_TASKS` list. New builder/schema pairs now force a conscious list update.

**Should-fix (ad-tech-protocol-expert):**
- `si_initiate_session.intent` default changed from `"E2E test session"` (test metadata) to `"Browse available offerings"` (a real user-ask). Storyboards still override via `sample_request` for intent-specific paths.

**Verified, no change needed:**
- `activate_signal` behavior-change risk (code-reviewer): grepped all 9 `activate_signal` storyboards in `compliance/cache/latest/**` — every one declares `sample_request.destinations`, so the new placeholder default only fires for the pure-fallback path, which never existed on wire before.
- `rights_grant_id` → `rights_id` context bridge (ad-tech-protocol-expert): confirmed `src/lib/testing/storyboard/context.ts:235` already extracts both keys from `acquire_rights`, so `context.rights_id` resolves in downstream builders.

**Deferred (testing-expert suggestion):**
- Fixture-path round-trip (iterate `listAllComplianceStoryboards()` and round-trip each real `sample_request`) — stronger invariant but outside the #803 scope. Filing as follow-up.

PR #814, opened 2026-04-22
fix(testing): governance.denial_blocks_mutation skips anchor on expect_error recovery steps (#811)

### LINKED ISSUE #811: governance.denial_blocks_mutation fires on storyboards designed to test recovery
## Summary

The default `governance.denial_blocks_mutation` invariant in `@adcp/client/testing/storyboard/default-invariants.ts` is triggering on storyboards whose **whole point** is to test a buyer recovering from a rejection.

Two first-party storyboards are blocked:

- `media_buy_seller/governance_denied_recovery` — narrative reads *"GOVERNANCE_DENIED is a correctable error, not a dead end. The buyer must be able to [shrink the buy and retry]."* Step `create_media_buy_denied` legitimately returns `GOVERNANCE_DENIED`; the subsequent `create_media_buy_retry` sends a corrected (smaller) payload and SHOULD succeed.
- `media_buy_seller/measurement_terms_rejected` — same shape, `TERMS_REJECTED` on step `create_media_buy_aggressive_terms` then a corrected `create_media_buy_relaxed_terms`.

The invariant observes the denial in `onStep`, anchors it in `state.deniedPlans` / `state.runDenial`, and then flags ANY later mutation in the same storyboard:

```
Mutation acquired a resource after a governance denial — step
"create_media_buy_denied" returned GOVERNANCE_DENIED (run-wide);
subsequent step "create_media_buy_retry" (task=create_media_buy)
acquired media_buy_id=mb_xxx
```

This is a false positive. The invariant exists to catch *silent* post-denial mutations (handler ignores the denial and mints a resource anyway) — it shouldn't fire when the storyboard **explicitly** declares `expect_error: true` on the denial step AND the retry uses a corrected payload.

## Where it happens

```
adcp-client/src/lib/testing/storyboard/default-invariants.ts:252
registerOnce('governance.denial_blocks_mutation', {
  ...
  onStep: (ctx, stepResult) => {
    ...
    const anchor = (planId && state.deniedPlans.get(planId)) ?? state.runDenial;
    if (!anchor) return [];
    return [{ passed: false, description: 'Mutation acquired a resource after a governance denial', ... }];
  },
});
```

The anchor persists for the rest of the run once ANY denial is observed. No mechanism for the storyboard to say *"I expected this denial, the next successful mutation is the recovery path."*

## Proposed fix (one of)

**A. Clear the anchor when a step with `expect_error: true` precedes the acquisition.**
A step authored to expect the error IS acknowledging the denial; the next successful mutation isn't a silent bypass. Easiest read.

**B. Per-step invariant skip.**
Add `invariants: { skip: ['governance.denial_blocks_mutation'] }` at the step level so storyboards can mark a step as an intentional post-denial recovery. More explicit but noisier in YAML.

**C. Scope the anchor to resource identity.**
Block only when the *same* `plan_id` or `media_buy_id` reappears post-denial with an acquired status. The two affected storyboards retry against the same plan but with a corrected payload that legitimately succeeds — under (C) they'd still trip. Ruled out.

(A) seems right to me — every storyboard currently tripping has `expect_error: true` on its denial step.

## Reproduction

Run the training-agent against either storyboard:

```
adcp storyboard run <url> media_buy_seller/governance_denied_recovery --allow-http --auth <token>
adcp storyboard run <url> media_buy_seller/measurement_terms_rejected --allow-http --auth <token>
```

The denial step passes, the retry's tool call succeeds at the wire level, and the invariant fails the step anyway with the message above.

## Willing to send the PR

Happy to send (A). Wanted to confirm the shape before writing it.

## Related

- `default-invariants.ts:203-295` (`governance.denial_blocks_mutation`)
- adcp storyboards: `static/compliance/source/protocols/media-buy/scenarios/governance_denied_recovery.yaml`, `.../measurement_terms_rejected.yaml`

bokelley:
## Summary

- `governance.denial_blocks_mutation` now skips anchoring when the denial-bearing step is authored with `expect_error: true`, treating it as an acknowledged denial rather than a latent anchor
- Unblocks two first-party denial-recovery storyboards that were being false-positived: `media_buy_seller/governance_denied_recovery` and `media_buy_seller/measurement_terms_rejected`
- Silent-bypass detection is preserved for `check_governance` 200 `status: "denied"` and any adcp_error denial code on a step the author did not mark `expect_error`

## Rationale

A step authored with `expect_error: true` is the storyboard declaring "this denial is a planned part of the flow" — the canonical shape for denial-recovery scenarios where the buyer reads the findings, corrects the payload (shrink the buy, relax the terms), and retries. Before this fix, the invariant anchored on the acknowledged denial and then flagged the corrected retry as a post-denial resource acquisition — a false positive on two shipping storyboards.

The narrowed scope is principled: `check_governance` 200 with `status: "denied"` is HTTP-200 on the wire and cannot be marked `expect_error: true` (that flag inverts the pass/fail check for WIRE-level errors). So the silent-bypass shape the invariant actually guards — downstream mutation after an un-acknowledged denial — remains fully covered. Correctness of the recovery PAYLOAD itself (right budget, relaxed terms, etc.) is a storyboard-level concern handled by per-step `validations`.

## Test plan

- [x] `node --test test/lib/storyboard-default-invariants.test.js` — 106 / 106 pass, including:
  - New: acknowledged denial (`expect_error: true`) followed by a retry mutation does NOT fire
  - New: parameterized coverage over all six denial codes on the acknowledged-recovery path
  - New: un-acknowledged `check_governance` 200 status:denied still anchors even when an UNRELATED `expect_error` step sits between it and the mutation (guard against too-broad anchor clearing)
- [x] `npm test` — 5413 / 5413 pass
- [x] `npm run build` + `npm run typecheck` clean
- [x] Independent code-review by the `code-reviewer` agent — "Ship it. No Must-Fix."

Closes #811.

### DIFF
--- a/.changeset/governance-denial-recovery.md
+++ b/.changeset/governance-denial-recovery.md
@@ -1,0 +1,31 @@
+---
+'@adcp/client': patch
+---
+
+Fix `governance.denial_blocks_mutation` false-positives on denial-recovery
+storyboards.
+
+A step authored with `expect_error: true` is the storyboard declaring
+"this denial is a planned part of the flow" — the canonical shape for
+denial-recovery scenarios where the buyer reads the findings, corrects
+the payload (shrink the buy, relax the terms), and retries. The
+subsequent successful mutation is the recovery path, not a silent
+bypass, and the invariant no longer anchors on it.
+
+Unacknowledged denials — e.g. a `check_governance` 200 with
+`status: "denied"`, or an `adcp_error` denial code on a step the author
+did not mark `expect_error` — continue to anchor as before. That is the
+real silent-bypass shape the invariant guards.
+
+Narrowed scope: the invariant no longer catches a buggy mutation that
+fires immediately after an author-acknowledged denial. Correctness of
+the recovery payload itself (right budget, relaxed terms, etc.) is a
+storyboard-level concern — covered by per-step `validations` and status
+checks on the retry, not by this run-wide guard.
+
+Unblocks two first-party storyboards whose whole point is exercising
+denial-recovery:
+- `media_buy_seller/governance_denied_recovery`
+- `media_buy_seller/measurement_terms_rejected`
+
+Closes #811.
--- a/src/lib/testing/storyboard/default-invariants.ts
+++ b/src/lib/testing/storyboard/default-invariants.ts
@@ -267,6 +267,21 @@ registerOnce('governance.denial_blocks_mutation', {
     // Denial observation is never itself a failure — record and return.
     const denial = detectGovernanceDenial(stepResult);
     if (denial) {
+      // A step authored with `expect_error: true` is the storyboard declaring
+      // "this denial is a planned part of the flow." First-party storyboards
+      // use this shape to test denial-recovery — the buyer reads the denial
+      // findings, corrects the payload (shrink the buy, relax the terms), and
+      // retries. The retry SHOULD succeed; it is not a silent bypass. Skip
+      // anchoring so the subsequent mutation is not flagged.
+      //
+      // The invariant's value-add is catching UN-acknowledged denials being
+      // ignored downstream — e.g. `check_governance` 200 with `status:
+      // "denied"` (expect_error is false; the 200 is not a wire error) that a
+      // buggy seller mutates past. Those still anchor below. Post-recovery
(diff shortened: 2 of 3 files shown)


bokelley:
Duplicate of #813, which landed first with the same approach (option A).

PR #819, opened 2026-04-22
fix(testing): honor sample_request in sync_event_sources builder (#818)

bokelley:
## Summary
- Fixes #818: storyboard `sync_event_sources` request builder ignored `step.sample_request` and always sent a generated `test-source-${Date.now()}` id.
- Storyboards that author a specific `event_source_id` (e.g., `sales_catalog_driven` → `amsterdam_website`) synced the generated id while downstream `log_event` / `provide_performance_feedback` steps sent the authored id, producing `EVENT_SOURCE_NOT_FOUND`.
- Builder now delegates to `sample_request` when present, matching the pattern used by `log_event`, `sync_catalogs`, `report_usage`, `list_creative_formats`, `get_rights`, and `sync_governance`.

## Test plan
- [x] New unit test: `sync_event_sources honors step.sample_request so authored event_source_id reaches the wire`
- [x] New unit test: `sync_event_sources injects context into sample_request`
- [x] Existing fallback test still passes
- [x] `test/lib/request-builder.test.js` — 46/46 pass
- [x] `test/lib/request-builder-schema-roundtrip.test.js` — 41/41 pass

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

Issue #818, opened 2026-04-22
Storyboard runner: sync_event_sources ignores step.sample_request


bokelley:
## Summary

`sync_event_sources` in the storyboard request-builder still discards `step.sample_request`, mirroring the pre-#794 `log_event` / `create_media_buy` bug. Storyboards authoring a specific `event_source_id` hit the wire as the hard-coded fallback, so downstream `log_event` / `provide_performance_feedback` steps that reference the authored id fail with `EVENT_SOURCE_NOT_FOUND`.

## Where

`src/lib/testing/storyboard/request-builder.ts` (the `sync_event_sources` builder):

```ts
sync_event_sources(_step, context, options) {     // ← _step ignored
  return {
    account: context.account ?? resolveAccount(options),
    event_sources: [
      {
        event_source_id: `test-source-${Date.now()}`,  // ← generated, not authored
        name: 'E2E Test Event Source',
        event_types: ['purchase', 'add_to_cart'],
      },
    ],
  };
},
```

## Repro

Run `sales_catalog_driven` against any seller. The storyboard authors `event_source_id: "amsterdam_website"` on both `sync_event_sources` and `log_events`. The runner sends a generated id on sync, then the authored id on log_events — handler hasn't seen the authored id, returns `EVENT_SOURCE_NOT_FOUND`.

Traced on the adcp training agent with session-key debug logging:

```
[DBG:log_event] want="amsterdam_website" sessionKey="open:default"
  stores: open:acmeoutdoor.example=[test-source-1776864441375] | open:default=[]
```

## Fix

Honor `step.sample_request` first, same pattern as `log_event` / `create_media_buy` / `list_creative_formats` / `get_rights` / `sync_governance`:

```ts
sync_event_sources(step, context, options) {
  if (step.sample_request) {
    return injectContext({ ...step.sample_request, account: context.account ?? resolveAccount(options) }, context);
  }
  return { /* existing fallback */ };
},
```

Happy to send the PR.

bokelley:
Fixed by #816 (fixture-authoritative runner, pending in 5.14.0).

[...]

Issue #820, opened 2026-04-22
Storyboard runner: demote builder to enricher, hard-fail on missing fixture (extends #801)


bokelley:
## TL;DR

Three independent expert reviews of the #805 fallback fixes converged on the same root cause: the runner's request-construction priority is inverted. Issue #801 already proposed flipping \`sample_request\`/builder precedence. This issue extends that direction into a coherent architectural cleanup.

## The framing the experts gave

> \"Your YAML isn't a fixture, it's a prompt — and your runner isn't a fixture player, it's a code generator. The test that passed against a fabricated payload is worse than a failing test. It's a false green.\"

Today at \`src/lib/testing/storyboard/runner.ts:1145-1156\`:

\`\`\`
1. options.request override
2. expect_error + sample_request → sample_request wins
3. Builder (if registered) ← every task with a builder lands here
4. sample_request (only if no builder)
\`\`\`

The builder-first order means:
- **Authored \`sample_request\` is silently discarded** by builders that didn't opt-in to \`if (step.sample_request) return injectContext(...)\`.
- **Storyboards pass against fabricated payloads** that have nothing to do with what the author wrote. False green.
- **Every few weeks a new fallback drifts out of spec** (#780 / #792 / #793 / #802 / #805) — fabrication is the bug factory.

### 1. Fixture-authoritative precedence (extends #801)

When \`sample_request\` is authored, it is the base payload. The "builder" runs as a post-merge **enricher**, filling only fields the fixture left unset — discovery-derived IDs (\`product_id\`, \`format_id\`, server-assigned \`media_buy_id\`), envelope fields (\`idempotency_key\`), and context-substituted placeholders.

**ERROR** — not warn — if a top-level key the author wrote disappears from the final payload. That's data loss, not divergence.

### 2. Hard-fail at step load for fixture-required tasks

A mutating task (per \`MUTATING_TASKS\`) with no \`sample_request\` and no discovery-derived fixture should **throw at storyboard load**, not fabricate synthetic data and ship it to a real seller. Error message points at the fixture path: *\"Step \\\`update_budget\\\` (task=update_media_buy) requires sample_request. Either author one or add a \\\`create_media_buy\\\` step earlier to populate context.media_buy_id.\"*

Read tasks (\`get_media_buys\`, \`list_creative_formats\`) stay discovery-derived — they legitimately run on context alone.

### 3. Rename \"builder\" → \"enricher\" in code + docs

[...]

PR #829, opened 2026-04-22
refactor(testing): extract delegateSampleRequest helper

### LINKED ISSUE #822: Storyboard runner: generalize sample_request delegation across mutating sync-style builders
## Summary
Following #818 (sync_event_sources) and #821 (sync_audiences), the pattern repeats: roughly half the builders in \`src/lib/testing/storyboard/request-builder.ts\` now begin with an identical

\`\`\`ts
if (step.sample_request) {
  return injectContext({ ...step.sample_request, account: context.account ?? resolveAccount(options) }, context);
}
\`\`\`

prelude — with small variations (some inject \`account\`, some don't; \`sync_audiences\` had carveouts; \`get_rights\` omits the account merge).

Each time a storyboard authors a specific id that downstream steps reference, we discover the builder had been dropping it and we patch one more case. The copy-paste is the vehicle for the class of bug.

## Proposal
Extract a generic delegation helper — e.g. \`delegateSampleRequest(step, context, { withAccount })\` — and route every \"mutating sync-style\" builder through it. The rule we want to enforce globally is:

> If a storyboard authored \`sample_request\`, the runner MUST pass it through (after context injection) instead of overwriting with synthesized data.

Only builders with scenario-specific logic (\`create_media_buy\` package merging, \`get_products\` refine handling, \`comply_test_controller\` sandbox override) should keep custom code paths, and those should explicitly opt out rather than rediscover the pattern.

## Why now
- Two builder-level bug fixes in a week both reduced to the same one-liner.
- \`sync_accounts\`, \`list_accounts\`, \`preview_creative\`, \`get_account_financials\`, and others still ignore \`sample_request\` today — no current storyboard exposure, but the same class of bug will hit us the next time a storyboard authors ids for those tasks.

## Out of scope
- Changing \`sample_request\` semantics itself — this is purely a runner refactor.

## References
- #818 / #819 — sync_event_sources fix
- #821 — sync_audiences fix
- src/lib/testing/storyboard/request-builder.ts

bokelley:
## Summary
Closes #822. Subsumes #819.

The storyboard runner contract "if a storyboard authors \`sample_request\`, pass it through after context injection" was open-coded in 18 builders, split across three variants:

- **Pattern A** (5 builders): delegate **with** account injection
- **Pattern B** (13 builders): delegate **without** account injection
- **Pattern C** (10 builders): silently **ignore** \`sample_request\`

Every time a storyboard authored a specific id for a Pattern C builder, downstream steps broke with \`*_NOT_FOUND\`. #818 (\`sync_event_sources\`) and #821 (\`sync_audiences\` add-shape) were both instances of the same class.

## Change
- New helper \`delegateSampleRequest(step, context, options, { withAccount })\` — the single canonical delegation point.
- Every builder either calls the helper **or** is on the explicit \`INLINE_SAMPLE_REQUEST_BUILDERS\` allowlist (\`create_media_buy\`, \`get_products\`, \`get_brand_identity\`, \`get_signals\`, \`activate_signal\`, \`comply_test_controller\` — builders that consume \`sample_request\` inline for scenario-specific merges).
- New coverage test walks \`listRequestBuilders()\` and asserts every non-allowlisted builder honors a sentinel \`sample_request\`. **The next builder that forgets to delegate fails this test.**

## Side effects
- **Fixes \`sync_event_sources\`** — this was the one remaining Pattern C mutating builder, so this PR subsumes #819.
- Previously Pattern C builders (\`sync_accounts\`, \`list_accounts\`, \`get_media_buys\`, \`get_media_buy_delivery\`, \`list_creatives\`, \`preview_creative\`, \`list_content_standards\`, \`get_content_standards\`, \`get_account_financials\`, \`get_adcp_capabilities\`) now honor \`sample_request\`. No current storyboard authored \`sample_request\` for these, so no downstream behavior change; future storyboards can rely on it working.

## Tests
- [x] New unit test: \`every non-allowlisted builder honors step.sample_request\` (walks all 34 builders, passes a sentinel, asserts it flows through)
- [x] New unit test: \`inline-sample-request allowlist only contains builders with scenario-specific logic\` (catches typos/dead entries)
- [x] All existing request-builder + schema-roundtrip tests preserved (92/92 pass)
- [x] Broader storyboard test surface: 1537/1537 pass

## Public API
New exports from \`src/lib/testing/storyboard/index.ts\`:
- \`listRequestBuilders(): string[]\`
- \`INLINE_SAMPLE_REQUEST_BUILDERS: ReadonlySet<string>\`

Both are for tooling that introspects builders (the coverage test, and any future schema/drift checks).

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.changeset/delegate-sample-request-helper.md
+++ b/.changeset/delegate-sample-request-helper.md
@@ -1,0 +1,27 @@
+---
+'@adcp/client': patch
+---
+
+refactor(testing): extract `delegateSampleRequest` helper and enforce the contract across all storyboard builders
+
+The storyboard runner's rule — **if a storyboard authors `sample_request`, pass it through (after context injection) instead of overwriting with synthesized data** — was previously open-coded in 18 builders, with three separate behaviors:
+
+- Pattern A (5 builders): delegate **with** account injection
+- Pattern B (13 builders): delegate **without** account injection
+- Pattern C (10 builders): silently **ignore** `sample_request`
+
+The inconsistency is the class of bug behind #818 and #821 — every time a storyboard authored a specific id for a builder that ignored `sample_request`, downstream steps broke with `*_NOT_FOUND`.
+
+This change introduces a single `delegateSampleRequest(step, context, options, { withAccount })` helper. Every builder now either:
+
+1. Calls `delegateSampleRequest` (default behavior — honors authored `sample_request`), or
+2. Appears in the new `INLINE_SAMPLE_REQUEST_BUILDERS` allowlist because it consumes `sample_request` inline (`create_media_buy`, `get_products`, `get_brand_identity`, `get_signals`, `activate_signal`, `comply_test_controller`).
+
+A new coverage test walks every builder and asserts the contract, so the next builder that forgets to honor `sample_request` fails the test instead of shipping.
+
+Side effects:
+
+- Fixes `sync_event_sources` (previously Pattern C; subsumes #819).
+- Previously Pattern C builders (`sync_accounts`, `list_accounts`, `get_media_buys`, `get_media_buy_delivery`, `list_creatives`, `preview_creative`, `list_content_standards`, `get_content_standards`, `get_account_financials`, `get_adcp_capabilities`) now honor `sample_request` when authored. No storyboard currently exposes a behavior change on these, but future authoring works correctly.
+
+Exports `INLINE_SAMPLE_REQUEST_BUILDERS` and `listRequestBuilders()` from `@adcp/client/testing/storyboard/request-builder` for tooling that introspects builders.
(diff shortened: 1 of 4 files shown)


bokelley:
Superseded by #820 (landed on main in e2f9ea9d). #820 solves the same problem (#822) with a stronger architecture:

**#820 on main (enrichers + runner-layered fixtures):**
- The runner's `enrichRequest()` layers `sample_request` on top of the enricher's output automatically: `{ ...enriched, ...fixture }`.
- Enrichers don't know about `sample_request` at all — the contract is enforced in one place.
- Only 2 enrichers need fixture-awareness (`FIXTURE_AWARE_ENRICHERS`: `create_media_buy`, `comply_test_controller`) because they splice discovery-derived fields into nested fixture structures.

**This PR (per-builder delegation helper):**
- Every builder manually calls `delegateSampleRequest()`.
- 6-entry allowlist (`INLINE_SAMPLE_REQUEST_BUILDERS`) for scenario-specific builders.
- Coverage test walks every builder to enforce discipline.

Main's approach has a smaller allowlist, no per-builder discipline to forget, and enforces the rule in one place rather than 34. Closing in favor of it.

Rebase attempt surfaced the architectural overlap — every builder I touched had been rewritten on main to produce a default payload that the runner now overlays fixture on top of.

Issue #850, opened 2026-04-23
feat(testing): version-staleness hint when agent is on pre-factory @adcp/client


bokelley:
## Context

PR #848 shipped runtime hints naming \`buildCreativeResponse\` / \`audioAsset\` / \`displayRender\` / \`parameterizedRender\`. If a developer is on \`@adcp/client\` < 5.14 (the version these factories first shipped in — scope3's original path in agentic-adapters#100 was on 4.16.2), the hint is slightly misleading: the helper exists in the SDK, but their pinned version doesn't have it.

DX-expert's review flagged this as a gap — the developer follows the hint, adds the import, gets a "module has no exported member" error, and has to figure out they need to upgrade.

## Proposal

Two places where version staleness could be detected:

1. **Agent-observed case** — the storyboard runner pulls the agent's \`get_adcp_capabilities\` response, which includes \`library_version: '@adcp/client@X.Y.Z'\` when the agent self-reports via the existing capability surface. If the version is present AND older than the minimum release that shipped the referenced helpers, suffix the hint:
   \`\`\`
   ... Use buildCreativeResponse() from @adcp/client/server.
   (Note: your agent reports @adcp/client@4.16.2 — these helpers ship in ≥5.14. Upgrade your SDK dep.)
   \`\`\`

2. **Client-side case** — the CLI startup staleness check (shipped in #835) already hits npm for the latest version. Extend it to note which features the user's pinned version lacks, if we maintain a \`FEATURES_BY_VERSION\` map.

(1) is tighter — it triggers only when the hint actually applies. (2) is broader but less targeted.

## Scope

Start with (1). Add a \`library_version\` field to \`StoryboardContext\` (populated from the capabilities response), thread it into \`detectShapeDriftHint\`, and maintain a small \`HELPER_MIN_VERSION\` map:

\`\`\`ts
const HELPER_MIN_VERSION: Record<string, string> = {
  buildCreativeResponse: '5.14.0',
  audioAsset: '5.14.0',  // shipped earlier, confirm via git log before merging
  displayRender: '5.14.0',
  parameterizedRender: '5.14.0',
};
\`\`\`

Suffix the hint when the agent's reported version is below the min for the helper being recommended.

## Out of scope

- Forcing agents to report \`library_version\`. If it's absent, skip the suffix — don't demand it. Some agents will be hand-rolled and won't carry this field.
[...]

Issue #845, opened 2026-04-23
feat(testing): hint at build_creative shape drift in strict-validation warnings


bokelley:
## Context

Deferred from #844. When a storyboard run hits an agent whose \`build_creative\` response has platform-native fields (\`tag_url\`, \`creative_id\`, \`media_type\`) at the top level instead of \`{ creative_manifest: { format_id, assets } }\`, the AJV rejection message names the missing \`creative_manifest\` field but doesn't point the developer at the typed helpers.

Observed in scope3data/agentic-adapters#100. Docs (#840) and skill (#838) now cover this, but an implementer running \`adcp storyboard run\` today sees a generic schema error, not a targeted hint.

## Proposal

In the strict-validation path (\`src/lib/testing/compliance/\` — strict_validation_summary / ValidationResult.warning plumbing from #820), detect this specific shape:

- Tool is \`build_creative\`
- Response is missing \`creative_manifest\` AND \`creative_manifests\`
- Response has one or more of \`tag_url\`, \`creative_id\`, \`media_type\`, \`tag_type\` at the top level

When detected, prefix the \`warning\` field with:

\`\`\`
This looks like a platform-native response. build_creative must return { creative_manifest: { format_id, assets: { ... } } }. Use buildCreativeResponse() / buildCreativeMultiResponse() from @adcp/client/server to enforce the shape at compile time. See skills/build-creative-agent/ § creative-template.
\`\`\`

## Scope

- Detection: ~10 LOC in the validator error-message composer
- Plumb into \`ValidationResult.warning\` (new field from #820)
- Test: strict-validation integration test with a synthetic non-conforming \`build_creative\` response
- No breaking changes

## Related

- #840 — llms.txt "Watch out" block (landed in #844)
- #838 — SKILL audio example (landed in #844)
- adcp-client#820 — strict validation reporting (where \`warning\` lives)

Issue #841, opened 2026-04-23
feat(cli): adcp storyboard run --no-sandbox to force real-adapter responses


bokelley:
## Context

Agent implementations commonly have two paths for AdCP tools:

1. **Real adapter** — calls the platform API
2. **Sandbox handler** — returns synthetic responses when an env var or account flag is set, so comply can run without real API keys

The sandbox path can mask non-conformance in the real path. Example from scope3data/agentic-adapters#100: real \`buildCreative\` returns \`{ tag_url, tag_type, media_type }\`; sandbox \`handleBuildCreative\` returns \`{ creative_manifest: { format_id, assets } }\`. Compliance passes against the sandbox; buyer agents break against prod.

## Proposal

Add \`--no-sandbox\` flag to \`adcp storyboard run\`:

- When set, the runner adds a header or context hint that tells the agent \"do not route to sandbox mode\" — e.g. \`context.disable_sandbox: true\` or an \`X-AdCP-Sandbox: false\` header, documented as a convention agents should honor.
- In the report, flag when an agent's own sandbox heuristic (env var, brand.domain) is detected and prominently surface \"sandbox-masked: test did not exercise real handler.\"

Also consider:

- Default report header shows \"Sandbox active: yes/no/partial\" for transparency.
- Telemetry on which storyboards passed only via sandbox paths.

## Scope

- Flag parser in \`bin/adcp.js\` (storyboard run)
- Runtime propagation to request builder in \`src/lib/testing/compliance/\`
- Report renderer in \`src/lib/testing/compliance/report.ts\` (or wherever final render lives)
- Document the \`disable_sandbox\`/\`X-AdCP-Sandbox\` convention in \`skills/build-*-agent/\` for implementers

bokelley:
Deferring implementation — lands cleanly only once we pick a convention and announce it to agent builders. Two candidates:

1. **Context field** — \`context.disable_sandbox: true\`. Rides the existing AdCP envelope. Downside: the \`Context\` schema says agents should not parse context. Forcing them to contradicts the spec's own guidance.
2. **Transport header** — \`X-AdCP-Disable-Sandbox: true\`. Out-of-band, HTTP-native, clean conceptual separation. Downside: requires per-transport support in the runner (MCP + A2A), and A2A layered over SSE/streaming may not expose per-message headers.
[...]

PR #853, opened 2026-04-23
feat(testing): extend shape-drift hint to list-shaped tools (bare-array)

### LINKED ISSUE #852: feat(testing): extend detectShapeDriftHint to list_creatives + peer list-shaped tools
## Context

Follow-up from #851's DX review (gap #3). With \`build_creative\` / \`sync_creatives\` / \`preview_creative\` covered, the next tier of drift targets is list-shaped tools. \`list_creatives\` is the obvious next one — handlers sometimes return a bare \`[{...}]\` array or \`{ creatives: [...] }\` without the required \`query_summary\` + \`pagination\` siblings.

## Tools to cover

- \`list_creatives\` — wrapper is \`{ query_summary, creatives, pagination }\`. Drift patterns:
  - bare array at root (\`[{...}]\`)
  - \`{ creatives: [...] }\` without \`query_summary\` or \`pagination\`
- \`get_products\` — wrapper is \`{ products, query_summary? }\`. Drift: bare array or missing query_summary.
- \`list_creative_formats\` — wrapper is \`{ formats: [...] }\`. Drift: bare array at root.
- \`list_authorized_properties\` — wrapper is \`{ properties: [...] }\`. Drift: bare array.

## Scope

Same pattern as #851. One switch branch per tool in \`detectShapeDriftHint\`. For "bare array at root," the check is \`Array.isArray(payload)\` — but note the detector's signature takes \`Record<string, unknown>\`, so bare-array drift would need a wrapper type change OR a separate detection path upstream. Easiest is to handle the bare-array case in the caller (\`validateResponseSchema\`) before the payload is destructured.

For "missing-sibling" drift (e.g. \`{ creatives: [...] }\` without \`query_summary\`), keep it in \`detectShapeDriftHint\` following the existing pattern.

## Priority

Medium. The scope3-style bug (platform-native response) was specifically about build_creative; list-shaped drift is less common in the wild but the framework generalizes cleanly.

## Also consider

DX review suggested a registry refactor (\`Record<string, (payload) => string | undefined>\`) once the detector hits ~5 tools. Bundle with this PR if the count clears 5.

## Related

- #845 — initial framework (build_creative)
- #851 — sync_creatives + preview_creative expansion
- #849 — closed by #851

bokelley:
## Summary

Closes #852. Extends the \`detectShapeDriftHint\` framework from #845/#848/#851 to cover list-shaped tools where handlers return a bare array at the root instead of wrapping it in the required \`{ <key>: [...] }\` envelope.

### What's detected

| Tool | Wrong shape | Required shape | Helper named |
|---|---|---|---|
| \`list_creatives\` | \`[{...}]\` | \`{ creatives, query_summary, pagination }\` | \`listCreativesResponse\` |
| \`list_creative_formats\` | \`[{...}]\` | \`{ formats: [...] }\` | \`listCreativeFormatsResponse\` |
| \`list_accounts\` | \`[{...}]\` | \`{ accounts: [...] }\` | \`listAccountsResponse\` |
| \`get_products\` | \`[{...}]\` | \`{ products: [...] }\` | \`productsResponse\` |

### Why

AJV's error on a bare-array response is "expected object, got array" — accurate but doesn't name the required wrapper. A developer sees the error, doesn't know \`creatives\` vs \`results\` vs \`data\`, and has to dig into the schema. The hint tells them in one line.

The \`LIST_WRAPPER_TOOLS\` map is the single extension point — add a row, ship a helper reference, done.

### Signature change

\`detectShapeDriftHint\` now accepts \`unknown\` rather than \`Record<string, unknown>\` so it can recognize bare-array payloads at the root. Object-branch logic guards with \`typeof payload === 'object' && payload !== null\` at the top, keeping existing behavior intact. The caller in \`validateResponseSchema\` also no longer destructures \`_message\` off array-valued payloads (bare arrays don't carry that field).

### Safety

Bare arrays for **unknown** task names pass through silently. Some APIs legitimately return top-level arrays; the detector only fires on tools in \`LIST_WRAPPER_TOOLS\`. Defensive tests cover null / primitive payloads (detector exits cleanly), empty arrays (still wrong shape, hint fires), and the wrapper-present negative case.

## Test plan

- [x] 29/29 shape-drift-hint tests pass (21 pre-existing + 8 new)
- [x] \`npm run typecheck\` — clean
- [x] \`npm run prettier --check\` — clean

## Still deferred

- **#841** \`--no-sandbox\` — blocked on convention RFC
- **#850** version-staleness hint suffix — next tier

### DIFF
--- a/.changeset/shape-drift-list-tools.md
+++ b/.changeset/shape-drift-list-tools.md
@@ -1,0 +1,16 @@
+---
+'@adcp/client': patch
+---
+
+Extend `detectShapeDriftHint` in the storyboard runner to cover list-shaped tools (closes #852):
+
+- `list_creatives` — handler returns bare `[{...}]` instead of `{ creatives, query_summary, pagination }`
+- `list_creative_formats` — bare array instead of `{ formats: [...] }`
+- `list_accounts` — bare array instead of `{ accounts: [...] }`
+- `get_products` — bare array instead of `{ products: [...] }`
+
+The detector now accepts `unknown` rather than `Record<string, unknown>` so it can recognize bare-array responses at the root — a common drift class where AJV's error ("expected object, got array") doesn't name the required wrapper key. Each known list tool gets a pointed hint naming the wrapper and the response helper (`listCreativesResponse`, `listCreativeFormatsResponse`, `listAccountsResponse`, `productsResponse`) from `@adcp/client/server`.
+
+Bare arrays for unknown task names pass through silently — the detector only fires on registered list tools to avoid false positives on APIs that legitimately return top-level arrays.
+
+8 new tests covering each tool, the wrapper-present negative case, unknown-task pass-through, empty-array handling, and null/primitive defensive cases.
--- a/src/lib/testing/storyboard/validations.ts
+++ b/src/lib/testing/storyboard/validations.ts
@@ -267,9 +267,17 @@ function validateResponseSchema(
     };
   }
 
-  // Strip _message from data before validation — it's added by the response
-  // unwrapper as a text summary and is not part of the AdCP response schema.
-  const { _message, ...dataWithoutMessage } = (taskResult.data ?? {}) as Record<string, unknown>;
+  // Keep the raw payload separately from the object-form one below so the
+  // shape-drift detector can recognize bare-array responses (a common drift
+  // pattern for list tools). Strip _message when it's a top-level property —
+  // bare arrays don't carry that field.
+  const rawData = taskResult.data ?? {};
+  const dataWithoutMessage = Array.isArray(rawData)
+    ? rawData
+    : (() => {
+        const { _message, ...rest } = rawData as Record<string, unknown>;
+        return rest;
+      })();
   const parseResult = schema.safeParse(dataWithoutMessage);
 
   // Strict (AJV) verdict runs alongside the lenient Zod check so the run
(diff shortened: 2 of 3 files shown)


bokelley:
Both expert reviews addressed in db453799.

**DX-expert fixes:**
- **Wording consistency**: "bare array at the root" → "bare array at the top level" so all four tool hints share the same grammatical structure.
- **Gap coverage**: added \`get_media_buys\` and \`get_signals\` to \`LIST_WRAPPER_TOOLS\` — both have existing response helpers (\`getMediaBuysResponse\` / \`getSignalsResponse\`). Table now covers 6 tools.
- **Governance list tools** (\`list_property_lists\`, \`list_collection_lists\`, \`list_content_standards\`): deferred because their response helpers don't exist in \`@adcp/client/server\` yet — the hint would have to name a non-existent helper. Tracked in #854 which covers both (a) adding the helpers and (b) extending the detector once they land.
- **JSDoc**: updated function list, added \`@param\` note for the \`unknown\` signature + the \`typeof payload === 'object'\` guard at the object branches.
- **Registry refactor**: deferred per DX's own "at minimum file the follow-up." Noted in the JSDoc; revisit when branch count justifies it. Current structure (6-row table for list tools + 3 inline branches for object-shape patterns) stays as-is.

**Code-reviewer fix:**
- \`computeStrictVerdict(taskName, payload: Record<string, unknown>)\` was lying about its input type — we were casting arrays through. \`validateResponse()\` underneath already accepts \`unknown\`. Widened the param and dropped the cast at the call site.

Tests: 31 passing (29 pre-existing + 2 new for \`get_media_buys\` and \`get_signals\`). Typecheck + prettier clean.

CI on the updated branch will take ~3min.

PR #876, opened 2026-04-24
feat(testing): field_value_or_absent check matcher (closes #873)

### LINKED ISSUE #873: feat(testing): field_value_or_absent check matcher for envelope-tolerant fields
## Problem

Storyboards sometimes need to positively assert a value on a field that the envelope spec explicitly allows to be **absent**. The current `field_value` check fires when the field is missing, which forces authors to choose between:

1. **Drop the assertion entirely** — loses positive coverage (what we did in adcontextprotocol/adcp#3013 for `universal/idempotency.yaml::create_media_buy_initial`'s fresh-path `replayed`).
2. **Keep it** — penalizes spec-compliant agents that omit the field per `protocol-envelope.json`.

Both options are lossy. The real intent is *"if you report the value, it must be X; if you omit it, that's fine."*

## Proposal

Add a new check matcher `field_value_or_absent`:

```yaml
- check: field_value_or_absent
  path: "replayed"
  allowed_values: [false]
  description: "If replayed is present on fresh execution, it must be false"
```

Semantics: pass when (a) the field is absent or undefined, OR (b) the field is present and its value is in `allowed_values`. Fail only when the field is present with a disallowed value.

## Where it helps today

- Fresh-path `replayed` (envelope spec: MAY omit, MUST NOT be true on fresh)
- Any `additionalProperties: false`-tolerant sibling where the spec wording is "set to X when Y, otherwise omit" — e.g. future envelope fields with similar semantics.
- Generally: any field where silence is spec-compliant but a wrong value isn't.

## Implementation notes

- Mirrors the existing `field_value` resolver path, with an additional "undefined → pass" short-circuit before the `allowed_values` check.
- Keep `field_value` strict (present-and-match) as-is — don't soften its semantics. Authors opt into tolerance explicitly via the new matcher.
- Worth a lint rule: if a storyboard asserts `field_value_or_absent` on a field that is **required** in the response schema, warn — the tolerance is meaningless.

## Context

Surfaced by the ad-tech-protocol-expert review on adcontextprotocol/adcp#3013. Filed per their suggestion ("belt-and-suspenders later, a `field_value_or_absent` check matcher could be added to the runner — but that's an adcp-client ask, not a reason to hold this PR").

Not urgent — the drop-the-assertion workaround covers the one known case. File it so the matcher exists when the next envelope-tolerant field lands.

bokelley:
## Summary

Adds a new storyboard check matcher `field_value_or_absent` that mirrors `field_value` but treats an absent field as a pass. Fails only when the field is present with a disallowed value.

Closes #873.

## Why

Today a storyboard can't positively assert a value on a field the envelope spec explicitly allows to be omitted — `field_value` fires when the field is missing, so authors either drop the assertion (losing positive coverage) or keep it (penalizing spec-compliant agents). This matcher encodes the real intent: *"if you report the value, it must be X; if you omit it, that's fine."*

First target is fresh-path `replayed` (adcontextprotocol/adcp#3013): envelope spec says MAY omit, MUST NOT be `true` on fresh execution.

## Semantics

```yaml
- check: field_value_or_absent
  path: "replayed"
  allowed_values: [false]
  description: "If replayed is present on a fresh execution, it must be false"
```

- **Absent** (field missing, resolved `undefined`) → pass
- **Present + matches `value` / in `allowed_values`** → pass
- **Present + doesn't match** → fail (with `json_pointer`, `expected`, `actual` per the runner-output contract)
- `null` is treated as present, not absent — matches `field_present` precedent

`field_value` stays strict: authors opt into tolerance explicitly via the new matcher, same as the issue suggests.

## Changes

- `src/lib/testing/storyboard/types.ts` — add `field_value_or_absent` to the `StoryboardValidationCheck` union.
- `src/lib/testing/storyboard/validations.ts` — new `validateFieldValueOrAbsent` matcher + switch case. Shares `valuesMatch`, `resolvePath`, `toJsonPointer` with `field_value` — no duplication of path resolution or pointer generation.
- `test/lib/storyboard-runner-contract.test.js` — 5 new contract tests covering absent→pass, present-allowed→pass, present-disallowed→fail (with `json_pointer` / `expected` / `actual`), scalar `value` fallback, and `null` fails (present, not absent).
- `test/lib/storyboard-drift.test.js` — schema-drift path reachability now also walks `field_value_or_absent` paths, so a storyboard using the new matcher against a drifted path fails the same drift gate.
- `.changeset/field-value-or-absent.md` — minor bump.

## Not done (intentionally)

The issue also floats a lint rule: "warn when `field_value_or_absent` is asserted on a field that is required in the response schema." Left as a follow-up — the matcher is useful on its own, and the issue explicitly labels the lint rule as belt-and-suspenders. The upstream storyboard schema in `compliance/cache/3.0.0/universal/storyboard-schema.yaml` is also not updated here because that file is synced from `adcontextprotocol/adcp` via `npm run sync-schemas`; the check list there needs a PR against the spec repo.

## Test plan

- [x] `node --test test/lib/storyboard-runner-contract.test.js` — 24/24 pass (5 new)
- [x] `node --test test/lib/storyboard-drift.test.js` — 389/389 pass, 4 skipped (pre-existing upstream drift allowlist)
- [x] `node --test --test-force-exit test/lib/storyboard-security.test.js` — 85/85 pass
- [x] `npm run build` clean
- [ ] CI green

### DIFF
--- a/.changeset/field-value-or-absent.md
+++ b/.changeset/field-value-or-absent.md
@@ -1,0 +1,5 @@
+---
+"@adcp/client": minor
+---
+
+Added `field_value_or_absent` storyboard check matcher. Passes when the field is absent OR present with a value in `allowed_values` / matching `value`; fails only when present with a disallowed value. Use it for envelope-tolerant assertions (e.g. fresh-path `replayed`) where the spec allows omission but forbids a wrong value. Closes #873.
--- a/src/lib/testing/storyboard/types.ts
+++ b/src/lib/testing/storyboard/types.ts
@@ -371,6 +371,7 @@ export type StoryboardValidationCheck =
   | 'response_schema'
   | 'field_present'
   | 'field_value'
+  | 'field_value_or_absent'
   | 'status_code'
   | 'error_code'
   // HTTP-probe checks (for raw_probe tasks)
--- a/src/lib/testing/storyboard/validations.ts
+++ b/src/lib/testing/storyboard/validations.ts
@@ -87,6 +87,8 @@ function runValidation(validation: StoryboardValidation, ctx: ValidationContext)
       return validateFieldPresent(validation, resolveTarget(ctx));
     case 'field_value':
       return validateFieldValue(validation, resolveTarget(ctx));
+    case 'field_value_or_absent':
+      return validateFieldValueOrAbsent(validation, resolveTarget(ctx));
     case 'status_code':
       return requireTaskResult(ctx, validation, tr => validateStatusCode(validation, tr));
     case 'error_code':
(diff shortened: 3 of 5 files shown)


bokelley:
### Expert review pass

Ran the diff through code-reviewer, security-reviewer, ad-tech-protocol-expert, and nodejs-testing-expert.

**Verdicts**
- Security: ship it, no findings.
- Code review: no blockers, only optional nits.
- Protocol expert: semantics correct; flagged a test gap on the no-`path` error branch.
- Testing expert: flagged four edge cases worth pinning — explicit `undefined`, broken intermediate chain, empty `allowed_values`, missing `path`.

**Addressed in e5a7c96**
- Grouped all `field_value_or_absent` cases under a dedicated nested `describe` block.
- Added 4 tests: explicit `{field: undefined}` (treated as absent), broken-chain mid-path (absent), empty `allowed_values` (falls through to `value`, same as `field_value`), and missing `path` (structured error). 9 tests total; all pass.

**Not addressed (with reasons)**
- **Name change to `field_value_when_present`** (protocol expert): `field_value_or_absent` is the exact name proposed in #873. Renaming at SDK-merge time without spec-author input would fork the naming from the issue. Happy to rename in a follow-up if the storyboard-schema side (adcontextprotocol/adcp) prefers the alternative.
- **Lint rule for `field_value_or_absent` on a schema-required field** (protocol expert): #873 explicitly labels this as "belt-and-suspenders later." Leaving for a follow-up — the matcher is useful on its own.
- **SKILL.md troubleshooting row** (code reviewer): nit. The matcher's error message is self-describing; can add in a docs sweep.
- **Pre-existing `resolvePath` prototype-key handling inconsistency** (security reviewer): flagged as read-only (path is storyboard-authored, fully trusted input) and shared with the whole storyboard runner. Pre-dates this PR; belongs in a separate unification PR across `resolvePath` / `resolvePathAll` / `setPath`.

PR #897, opened 2026-04-24
feat(server): host-aware serve() for one-process multi-host

### LINKED ISSUE #885: serve(): first-class multi-host routing (one process → N hostnames)
## Context

We run 13 seller adapters (Snap, Meta, TikTok, Pinterest, Reddit, LinkedIn, Amazon, Google, Criteo, CitrusAd, Spotify, UniversalAds, FlashTalking) behind one Fly.io process. Each has its own hostname (e.g., `snap.agentic-adapters.scope3.com`), OAuth authorization server, and publisher domain on `get_adcp_capabilities`.

Today's SKILL.md explicitly flags this as out of scope:

> **Multi-host caveat.** `publicUrl` is static per `serve()` call. If a single Node process fronts multiple hostnames ... every host sees the same advertised `resource` ... For multi-host deployments, run one `serve()` per host (separate ports + reverse proxy by Host header) or route hosts to isolated Node processes upstream. A host-aware helper is on the roadmap but not in 5.2.x — until it ships, the single-`publicUrl`-per-process model is the only supported configuration.

The documented escape hatch (`createAdcpServer(...).connect(transport)` from a custom `http.createServer`) works — we've used it — but it means we re-own everything `serve()` does: authentication wiring, `/.well-known/oauth-protected-resource/<path>` publishing per host, request-body buffering for signature verification, and most of the composition plumbing documented in the \"Composing OAuth, signing, and idempotency\" section of SKILL.md. That's ~400 lines of boilerplate that a host-aware `serve()` could collapse.

## Proposal

A host-routing variant of `serve()` that accepts a host → factory mapping plus a per-host `publicUrl` resolver:

\`\`\`ts
serve({
  // Resolve the MCP endpoint URL from the incoming Host / X-Forwarded-Host so
  // /.well-known/oauth-protected-resource is correct per hostname, and the
  // JWT audience the seller advertises matches what buyers actually hit.
  publicUrl: (host) => \`https://\${host}/mcp\`,

  // Factory per host. Cached internally — called once per unique host.
  agent: (host) => createAdcpServer({
    name: hostToAgent.get(host).name,
    version: '1.0.0',
    mediaBuy: hostToAgent.get(host).handlers,
    // ...
  }),

  // Authentication still runs per request (same as single-host).
  authenticate: verifyBearer({ /* ... */ }),

  // protectedResource can also take a function to vary authorization_servers per host.
  protectedResource: (host) => ({
    authorization_servers: [\`https://\${host}/oauth\`],
  }),
})
\`\`\`

## Why it matters for adoption

- **Deployment cost**: running 13 Node processes behind a reverse proxy adds memory / cold-start / ops burden that's disproportionate for small adapter agents. Consolidating to one process is natural.
- **Dev experience**: the documented escape hatch works but the \"what serve() does that you now own\" list in SKILL.md is long and easy to implement subtly wrong (e.g., raw body buffering before signature verification, Fly replay-aware session handling).
- **Spec compliance at the edge**: a host-aware `publicUrl` is the correct behavior for RFC 9728 `/.well-known/oauth-protected-resource` when a single process fronts multiple brands — without it, buyers get audience-mismatched tokens.

## Workaround today (for reference)

We took the escape hatch in \`packages/shared/src/server/http.ts\`:

\`\`\`ts
// One AdcpServer per host, cached.
const servers = new Map<string, AdcpServer>()
function getServer(host: string, adapter: PlatformAdapter) {
  let s = servers.get(host)
  if (!s) {
    s = createAdcpServer({ /* per-adapter config */ })
    servers.set(host, s)
  }
  return s
}

// Streamable HTTP per request.
httpServer.on('request', async (req, res) => {
  const host = req.headers['x-forwarded-host'] ?? req.headers.host
  const adapter = routeHostToAdapter(host)
  const server = getServer(host, adapter)
  const transport = new StreamableHTTPServerTransport({ /* ... */ })
  await server.connect(transport)
  await transport.handleRequest(req, res)
})
\`\`\`

This works but skips the \`/.well-known/oauth-protected-resource\` publishing, the auth composition, and the preTransport signature verifier helpers.

## Prior note

SKILL.md §Alternative Transports already documents the escape hatch and explicitly says \"a host-aware helper is on the roadmap\" — this issue is a concrete ask with a use case.

cc @scope3data — happy to prototype + PR if the API shape lands.

bokelley:
## Summary

Closes #885. Adds first-class multi-host routing to `serve()` so one Node process can front many hostnames (white-label publishers, multi-brand adapters) without re-owning the HTTP plumbing.

## Changes

- **`ServeOptions.publicUrl: string | (host) => string`** — function form resolves per unique host; result cached. Path-match validation runs against each resolved URL. Static string form unchanged.
- **`ServeOptions.protectedResource: PRM | (host) => PRM`** — same pattern. Each host advertises its own RFC 9728 `resource` + authorization servers.
- **`ServeContext.host: string`** — resolved canonical host (lowercased, port preserved) passed into the factory on every request so it can branch on hostname.
- **`ServeOptions.trustForwardedHost: boolean`** — default `false`. Opt-in for `X-Forwarded-Host` when behind a proxy that sanitizes it (Fly, Cloud Run, internal ALB). Without it, an attacker-controlled header can't flip the advertised OAuth `resource`.
- **`verifyBearer({ audience: (req) => string })`** — audience check follows the per-host `publicUrl` so a token minted for `snap.example.com` fails validation at `meta.example.com`. Resolver errors surface as sanitized `AuthError`, not 500.
- **SKILL.md** — §"Multi-Host and Alternative Transports" replaces the old "on the roadmap" caveat with a working multi-host recipe.

## Security properties preserved

- Function-form `publicUrl` still validates each resolved URL's path against the mount path — mismatch fails closed (500) rather than minting audience-mismatched tokens.
- `X-Forwarded-Host` ignored by default; opt-in is explicit.
- Resolver throws are caught and logged, response is 500 — not a silent mis-advertisement.

## Test plan

- [x] 11 new tests in `test/lib/serve-multihost.test.js` — host threading, function-form publicUrl + PRM, caching, `trustForwardedHost` on/off, X-Forwarded-Host chain handling, 404 on missing host, 500 on invalid per-host publicUrl, backward compat for static form.
- [x] 2 new tests in `test/server-auth.test.js` — function-form audience succeeds for matching host / rejects for mismatched host; resolver throw → sanitized `AuthError`.
- [x] Full existing `serve()` + `verifyBearer` suites pass (54/54 targeted, 5627/5629 full suite — the 2 failures are a pre-existing EADDRINUSE flake on `:3112` from a stale process in another Conductor worktree, unrelated).
- [x] `tsc --noEmit` clean.
- [x] CI green.


bokelley (commented):
Strong overall — security posture is right, the scope stayed tight, and the Express-AS composition recipe is the right addition for adapters that run their own OAuth AS. Three things I'd push back on before merge:

**1. `audience: (req) => string` callback doesn't go through `serve()`'s host resolution — footgun in the example.**

SKILL.md multi-host example (`skills/build-seller-agent/SKILL.md`) has:

```ts
audience: req => `https://${req.headers['x-forwarded-host'] ?? req.headers.host}/mcp`
```

`serve()` itself honors `trustForwardedHost` when resolving host. This audience callback doesn't — it reads `X-Forwarded-Host` unconditionally. A user copying this example without `trustForwardedHost: true` advertises one audience (from `publicUrl(host)`) but verifies against another (whatever `X-Forwarded-Host` says). Attacker-controlled header flips the audience check but not the advertised PRM — a token minted for the attacker's claimed host verifies, even though the framework advertised a different `resource`.

Fix options:
- Thread the already-resolved host into the audience callback: `audience: (req, { host }) => ...`
- Better: expose the resolved `publicUrl` directly, so the common case is `audience: (_req, { publicUrl }) => publicUrl` (one line, can't drift from what `serve()` advertised).

Either way, the example should not read `X-Forwarded-Host` directly.

**2. Missing an end-to-end test for the cryptographic backstop.**

The 11 new tests in `test/lib/serve-multihost.test.js` cover host threading, function-form publicUrl/PRM, caching, `trustForwardedHost` on/off, and 404/500 paths. What I don't see (and should exist, since it's *the* security property of multi-host): a named test proving **a token minted for host A, presented on a request landing at host B in the same process, fails audience verification at host B**. The 2 audience-function tests in `test/server-auth.test.js` are unit-level — this wants to be end-to-end: two `publicUrl(host)` mappings, two OAuth audiences, one process, cross-host token, assert 401. If that's covered by one of the existing tests I missed, a rename would make it obvious.

**3. `throw new Error("Unknown host: X")` → 500 is defensible but unkind to ops.**

The factory example throws on unregistered hosts, framework catches and 500s. Fine for security (doesn't leak the adapter table) but confusing when an operator typos a DNS entry. A dedicated `UnknownHostError` mapping to 404 with a clean body would be kinder. Minor — don't block merge on this one.

Security properties I did verify hold: `X-Forwarded-Host` off by default, path-match validation per resolved URL, fail-closed on resolver throws, per-host result caching. Good shape.

bokelley:
reopening to retrigger skipped CI workflows

PR #922, opened 2026-04-24
fix(conformance): runner no longer fails agents on empty-phases storyboards

### LINKED ISSUE #921: Storyboard runner: handle empty-phases storyboards as not_applicable instead of __no_phases__ crash
## Context

Follow-up flagged in [adcp#2973](https://github.com/adcontextprotocol/adcp/pull/2973) (3.0.1 fix for the `governance_aware_seller` storyboard).

## Problem

The runner emits an internal `__no_phases__` structural error when a storyboard has `phases: []` (or when all phases are skipped before any executes). That string is **not** in the storyboard-schema's documented grading vocabulary — it's a runner-internal failure mode that surfaces as an agent compliance failure.

This bit a real seller: an agent passing 68/69 media-buy-lifecycle scenarios failed only on `governance_aware_seller` with `__no_phases__`, even though the seller didn't claim governance capability. The intended outcome was `not_applicable` (the storyboard is scoped to sellers that declare `governance_aware`), but the runner crashed before reaching that check.

adcp#2973 fixes the spec side by transplanting a `capability_discovery` phase into the storyboard so it has at least one phase to enumerate. That unblocks today's regression. But the runner-side defect is independent — any future storyboard that legitimately has no local phases (e.g., pure `requires_scenarios:` composition) will trip the same bug.

## Asks

1. **Runner should skip cleanly when phases are empty.** If `phases: []` and `requires_scenarios:` composes the full surface, treat as `not_applicable` rather than `__no_phases__`. Or surface a documented skip code (`not_applicable`, `capability_not_declared`) per the storyboard-schema vocabulary.

2. **Validation: confirm that with adcp#2973 merged, sellers without `governance_aware` capability now get `not_applicable` for this storyboard (not a pass, not a fail, not `__no_phases__`).**

3. **Replace `__no_phases__` everywhere it surfaces externally** — runners can keep it as an internal assertion, but it should never reach the user-facing report.

## Repro

Run the compliance harness against any seller that does not declare `governance_aware` capability before adcp#2973 lands. Storyboard `governance_aware_seller` reports `__no_phases__`.

## Related

- adcp#2972 (original bug report)
- adcp#2973 (3.0.1 spec fix — transplants `capability_discovery` phase)
- Storyboard schema: `static/compliance/source/universal/storyboard-schema.yaml` (documented grading vocabulary)

bokelley:
Closes #921

## Summary

When a storyboard had `phases: []`, the runner produced a result that made agents appear to **fail** even though the storyboard should have been a neutral skip. A real seller with 68/69 passing media-buy scenarios was failing compliance solely on `governance_aware_seller` with `__no_phases__` — a string not in the storyboard-schema's grading vocabulary.

Three interlinked bugs:

1. **`passed: false` on the synthetic phase** — the phase result was `false` even though its only step had `passed: true, skipped: true`. Per the runner-output contract (`steps.every(s => s.passed || s.skipped)`), a skipped step is neutral; the phase must not fail.
2. **`__no_phases__` leaking externally** — the sentinel strings appeared as `step_id` and `phase_id` in user-facing reports, but `'__no_phases__'` was never in the documented `RunnerSkipReason` vocabulary.
3. **`requiredPhasesPassed` guard not short-circuited** — the guard indexes `storyboard.phases[idx]`, which is `undefined` for an empty-phases storyboard, so `requiredPhasesPassed` stayed `false` and `overall_passed` stayed `false` even after fixing (1).

## What changed (`src/lib/testing/storyboard/runner.ts`)

- Synthetic phase `passed`: `false` → `true`
- Synthetic sentinel IDs: `'__no_phases__'` → `'no_phases'` (matches the documented `RunnerSkipReason` and `SKIP_DETAILS` entry; consistent with `buildNotApplicableStoryboardResult` precedent in `comply.ts`)
- `requiredPhasesPassed`: added `!hasExecutableSteps ||` short-circuit so the index-aligned guard is bypassed when phases is empty, producing `overall_passed: true`
- `requires_scenarios`-composed storyboards get a distinct detail message that explains the structural situation without incorrectly prescribing a call path

## What was tested

- `npm run build:lib` passes (pre-existing `tsx: not found` in env prevents full `ci:quick`; confirmed pre-existing on `main`)
- `npm test`: 61 passing, 548 failing — same pass count as `main`, one fewer failure (the `Compliance result types` test that imports `dist/lib/testing/compliance/index.js` now passes because the fix is reflected in the rebuilt dist)
- Pre-existing test failures are infrastructure-related (no compiled dist, no running agents)

## What is NOT in this PR (follow-up)

Issue #921 also asks about emitting `not_applicable` instead of `no_phases` when `requires_scenarios` is populated. Both experts flagged this as a design decision: `not_applicable` currently means "agent didn't declare the protocol or specialism", which is different from "storyboard delegates to scenario composition". A new `RunnerSkipReason` value (`scenario_composed` or similar) would be the right surgical fix — tracked as a follow-up to avoid conflating the two cases and breaking `storyboard-tracks.ts`'s label disambiguation.

## Known gaps

- No unit tests for the `phases: []` path. The test suite has almost no runner unit coverage (systemic gap, not introduced here). Adding regression tests for this path is tracked.

## Pre-PR review

- **code-reviewer**: approved — no blockers; noted `error: detail` on a skipped step is semantically odd (pre-existing pattern, not introduced here) and the duplicate sentinel comments
- **ad-tech-protocol-expert**: approved — `passed: true` correct per spec formula; `'no_phases'` rename non-breaking (undocumented → documented vocabulary); `requiredPhasesPassed` short-circuit aligns with `buildNotApplicableStoryboardResult` precedent

Session: https://claude.ai/code/session_01N7UFtzFS6q2ACWTfzqBZdG

---
_Generated by [Claude Code](https://claude.ai/code/session_01N7UFtzFS6q2ACWTfzqBZdG)_

### DIFF
--- a/.changeset/runner-no-phases-sentinel-fix.md
+++ b/.changeset/runner-no-phases-sentinel-fix.md
@@ -1,0 +1,14 @@
+---
+'@adcp/client': patch
+---
+
+**Fix: storyboard runner no longer fails agents on empty-phases storyboards.**
+
+When a storyboard had `phases: []` (e.g., a placeholder or a `requires_scenarios:`-composed storyboard), the runner emitted a synthetic phase with `passed: false` even though its only step was `skipped: true`. This caused agents to appear to fail on that storyboard in the compliance report, producing a confusing `__no_phases__` entry in the output — a string not in the storyboard-schema's documented grading vocabulary.
+
+Changes:
+- Synthetic phase `passed` corrected from `false` → `true` (a skipped step is neutral, not a failure).
+- Internal sentinel strings `'__no_phases__'` replaced with `'no_phases'` in `step_id` and `phase_id`, consistent with the documented `RunnerSkipReason` vocabulary.
+- When `storyboard.requires_scenarios` is populated, the detail message now explains the structural reason (scenario composition) rather than the generic placeholder message.
+
+Fixes #921.
--- a/src/lib/testing/storyboard/runner.ts
+++ b/src/lib/testing/storyboard/runner.ts
@@ -490,7 +490,7 @@ async function executeStoryboardPass(
   if (!hasExecutableSteps) {
     const isScenarioComposed = (storyboard.requires_scenarios?.length ?? 0) > 0;
     const detail = isScenarioComposed
-      ? `Storyboard "${storyboard.id}" has no local phases — its test surface is fully composed from the scenarios listed in \`requires_scenarios\`. Run via \`comply()\` to execute the composed scenarios.`
+      ? `Storyboard "${storyboard.id}" has no local phases — its test surface is fully composed from the scenarios listed in \`requires_scenarios\`.`
       : `Storyboard "${storyboard.id}" has no executable phases — populate \`phases[].steps\` or remove the storyboard.`;
     const syntheticStep: StoryboardStepResult = {
       storyboard_id: storyboard.id,
--- a/src/lib/testing/storyboard/runner.ts
+++ b/src/lib/testing/storyboard/runner.ts
@@ -834,11 +834,19 @@ async function executeStoryboardPass(
   // required_tools filtered out everything) would pass vacuously. (c) makes
   // assertions gating — a run with all validations green but a cross-step
   // invariant broken is not conformant.
-  const requiredPhasesPassed = phaseResults.some((p, idx) => {
-    const phaseDef = storyboard.phases[idx];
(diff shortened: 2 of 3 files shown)


bokelley:
I have read the IPR Policy

bokelley:
recheck

bokelley:
recheck

PR #931, opened 2026-04-25
feat(server): idempotency: 'disabled' mode + standalone enum value arrays

bokelley:
## Summary

Two additive surfaces for the @adcp/client SDK aimed at consumers who currently UUID-inject every test payload to satisfy AdCP 3.0's `idempotency_key` requirement, or who duplicate spec enum literals in their own validation code.

### 1. `idempotency: 'disabled'` for `createAdcpServer`

`AdcpServerConfig.idempotency` now accepts `'disabled' | IdempotencyStore`. When set to `'disabled'`:

- **Capability declaration flips to the spec's `IdempotencyUnsupported` branch** — `get_adcp_capabilities` returns `adcp.idempotency: { supported: false }` (no `replay_ttl_seconds`), matching the `oneOf` discriminator in the spec. Buyers reading caps fall back to natural-key dedup before retrying spend-committing operations.
- The replay middleware (`INVALID_REQUEST` / `IDEMPOTENCY_CONFLICT` / `IDEMPOTENCY_EXPIRED`) is skipped.
- Schema validation tolerates a missing `idempotency_key` on mutating tools — surgically, only the `keyword: 'required', pointer: '/idempotency_key'` issue is dropped; every other required field still produces `VALIDATION_ERROR`.
- A pre-middleware shape gate enforces `IDEMPOTENCY_KEY_PATTERN` whenever a key IS supplied, regardless of validation mode — defense-in-depth so malformed keys never reach handler logs.
- Throws at construction under `NODE_ENV=production` (silent double-execution on retry is a money-flow incident waiting to happen). Outside production, a `logger.warn` fires.

Production servers must still wire a real store via `createIdempotencyStore({ backend, ttlSeconds })`. The existing `idempotency: store` and `idempotency: undefined` paths are unchanged.

### 2. Standalone enum value arrays

New generated file `src/lib/types/enums.generated.ts` exports a `${TypeName}Values` const array for every named string-literal union in the AdCP TypeScript types (122 enums: `MediaChannelValues`, `PacingValues`, `MediaBuyStatusValues`, `DeliveryTypeValues`, `AssetContentTypeValues`, …).

```ts
import { MediaChannelValues } from '@adcp/client/types';
const channels = new Set<string>(MediaChannelValues);
if (!channels.has(input)) throw new Error('unknown channel');
```

Codegen wired into `npm run generate-types` so `ci:schema-check` catches drift the same way it catches type drift. Inline anonymous unions (e.g., `formats?: ('jpg' | ...)[]`) are out of scope — they don't have stable names in the generated TypeScript.

## Expert review caught a wire-level blocker

Initial implementation kept `supported: true` in `get_adcp_capabilities` even in disabled mode — three expert reviews (`ad-tech-protocol-expert`, `code-reviewer`, `security-reviewer`) converged on this as a money-flow footgun: a buyer reading caps and retrying after a 504 would double-book real spend. The spec's `IdempotencyUnsupported` branch (`supported: false`, no `replay_ttl_seconds`) is the correct shape and is now what disabled mode emits.

Other expert findings addressed: throw-on-production, pre-middleware shape gate, pinned the pointer-match invariant in a comment, bumped enum codegen floor from `<20` to `<100` to catch partial regex drift.

## Test plan

- [x] Unit tests in `test/server-idempotency.test.js` (+6): warn at construction, throw under `NODE_ENV=production`, schema filter is surgical (other required fields still fail), no replay, capability flips to `supported: false`, shape gate rejects malformed keys
- [x] Wire-level e2e in `test/server-idempotency-disabled-e2e.test.js` (+5): real `serve()` + official `@modelcontextprotocol/sdk` `Client` over `StreamableHTTPClientTransport`; A2A roundtrip via `createA2AAdapter`; both `validation: 'strict'` and `validation: 'off'` paths
- [x] Cross-validation tests in `test/lib/enum-arrays.test.js` (+6): every `Values` array round-trips against its matching Zod `Schema`; `Pacing` and `MediaBuyStatus` literals match the spec; ≥30 enum/schema pairs cross-checked
- [x] `npm run test:lib`: 4941 pass, 0 fail, 6 skipped (pre-existing)
- [x] Typecheck clean (`tsc --noEmit --project tsconfig.lib.json`)
- [ ] Reviewer: confirm storyboard runner auto-skip on `supported: false` is a follow-up (not a blocker) — `compliance/.../universal/idempotency.yaml` says `supported: false` sellers MUST skip, but the runner doesn't auto-detect today

## Follow-ups (not in this PR)

- Storyboard runner auto-skip when agent declares `idempotency.supported: false` (file an issue)
- Inline anonymous enum extraction (image asset formats, video container formats etc. — different extractor needed since they don't have stable names)


bokelley:
Addressed P0 + P2 hygiene; filed P1 as #932.

**P0 — production gate inverted (76db06f).** Allowlist `NODE_ENV` ∈ `{'test', 'development'}`; anything else (unset, `'production'`, `'staging'`, custom names) throws unless `ADCP_IDEMPOTENCY_DISABLED_ACK=1`. Strict ack-value matching — only literal `'1'` acknowledges, not `'true'`/`'yes'` (prevents copy-paste typos).

New tests cover: NODE_ENV unset → throws (the actual Lambda case), `'staging'` → throws, ack escape works, ack only accepts `'1'`.

**P2.1 — generated-at timestamp dropped from `enums.generated.ts`** (76db06f). New file, no in-band timestamp; git log + `.generated.ts` suffix are sufficient signal. Other generated files keep their existing timestamp convention (out-of-scope to refactor).

**P2.2 — `AgeVerificationMethod1Values`**: investigated. Both `AgeVerificationMethod` and `AgeVerificationMethod1` are real types in `core.generated.ts` — json-schema-to-typescript emits the suffix when the same enum is referenced twice from different schema paths. Both are validly importable, and dedup-by-value would arbitrarily drop one and break the corresponding consumer import. Keeping as-is; the cosmetic fix is upstream in json-schema-to-typescript.

**P2.3 — NODE_ENV mutation race**: documented in a comment at the top of the production-gate test block. The `before`/`after`-with-finally pattern is safe under the current node:test serial-within-file default; if `--test-concurrency` ever raises that, we move to a child process.

**P1 — inline anonymous unions** filed as #932. The clean fix is Zod-introspection (named schemas expose `.options` / `.def.options` on union types) rather than regex-on-TS — different tool, different PR. Acceptance criteria + suggested naming convention in the issue.

PR #937, opened 2026-04-25
feat(conformance): add ShapeDriftHint to StoryboardStepHint taxonomy

### LINKED ISSUE #935: Hint taxonomy: extend StoryboardStepHint beyond context_value_rejected
## Context

\`StoryboardStepHint\` is a discriminated union with one member today:

\`\`\`ts
export type StoryboardStepHint = ContextValueRejectedHint;
\`\`\`

The runner produces non-fatal diagnostic prose for several other failure-mode patterns (e.g. shape-drift detection in #848 / #851 / #853, schema validation enrichment in #919, monotonic-status enrichment in #791), but those don't carry the same structured-field contract that lets a downstream renderer (Addie, the CLI, JUnit) build an actionable fix plan instead of paraphrasing prose.

Adopters who want to *act on* the hint — \"name both tools that disagree, suggest the verify call, write the fix plan\" — are blocked on the structured fields. The runner already has the data; surfacing it via the same \`hints[]\` mechanism is the missing step.

## Proposal

Treat \`StoryboardStepHint\` as the canonical surface for *every* runner-side diagnostic that has structured fields a renderer can consume, not just context-value rejections. Candidate next kinds (each one a distinct symptom builders hit during conformance):

| \`kind\` | When it fires | Structured fields a renderer needs |
|---|---|---|
| \`shape_drift\` | Response has the right \`status\` envelope but the payload shape diverges from the expected variant (the failure mode #848 / #851 / #853 already detect inline) | \`expected_variant\`, \`observed_variant\`, \`drift_path\`, \`tool\` |
| \`missing_required_field\` | Strict validation fails on a required field the spec marks as required | \`field_path\`, \`tool\`, \`schema_ref\` |
| \`format_mismatch\` | A field passes lenient Zod but fails strict format (date-time, URI, UUID — the strict/lenient delta in StrictValidationVerdict) | \`field_path\`, \`expected_format\`, \`observed_value\`, \`tool\` |
| \`monotonic_violation\` | The status-monotonic invariant trips (#782 / #791 already enriched the error message) | \`tool\`, \`previous_status\`, \`observed_status\`, \`step_id\` |
| \`auth_misconfiguration\` | The actionable auth hint #766 added inline | \`tool\`, \`auth_kind_expected\`, \`auth_kind_advertised\` |

Each kind would carry a \`message\` field (the existing prose surface) plus the structured fields above, so any renderer that knows only the prose still works, while renderers that consume the structured fields can build per-kind fix plans.

## Why now

Addie just landed a Diagnose / Locate / Fix / Verify formatter for \`context_value_rejected\` (https://github.com/adcontextprotocol/adcp-client/issues/870) — the structured fields were sufficient to produce a deterministic builder playbook with named widen-vs-narrow fix paths and an explicit \`run_storyboard_step\` verify call. That formatter is ~200 lines and the renderer wire-up is ~30. The polish layer pattern works; it just needs more kinds to render.

Each new kind unlocks a similar renderer in Addie / CLI / JUnit consumers without further runner work — they're already reading \`hints[]\`.

## What's not changing

- \`hints[]\` stays non-fatal — pass/fail is unchanged.
- \`message\` field stays the human-readable summary; renderers that don't know a new \`kind\` discriminator can still display it verbatim.
- No protocol-side change. This is a runner internal contract.

## Suggested incremental shape

Don't ship five kinds at once. Pick \`shape_drift\` first — it has the most existing detection logic (#848 / #851 / #853) and the highest builder-confusion volume. Validate the contract holds, then add others.

Happy to write the renderer side per kind once each lands.

bokelley:
Closes #935

## Summary

- Extends `StoryboardStepHint` from a single-member alias to a proper discriminated union: `ContextValueRejectedHint | ShapeDriftHint`
- Adds `ShapeDriftHint` interface with machine-readable fields (`tool`, `observed_variant`, `expected_variant`, `instance_path`) alongside a `message` fallback — renderers that don't know the new `kind` continue to work verbatim
- New pure detection module `shape-drift-hints.ts` (parallel to `rejection-hints.ts`) covers: `bare_array` (9 list tools), `platform_native_fields` (`build_creative`), `per_item_shape` + `wrong_wrapper_key` (`sync_creatives`), `raw_render_fields` (`preview_creative`)
- Runner wires hints unconditionally (same gate as `ValidationResult.warning`), with identical `_message`-strip pre-processing
- Exports `LIST_WRAPPER_TOOLS` from `validations.ts` (was module-private) so the new module avoids duplication
- Exports `ShapeDriftHint` from the public testing index alongside `ContextValueRejectedHint`

## What's not changing

- Pass/fail is unaffected — `hints[]` stays non-fatal
- `ValidationResult.warning` still fires (unchanged) — the hint is additive, not a replacement
- No protocol-side change; this is a runner internal contract
- Existing `detectShapeDriftHint` string function in `validations.ts` is unchanged

## Breaking-change note

Adding a new union member to `StoryboardStepHint` is additive. The JSDoc on the type already says "more kinds may be added over time" — consumers must not assume exhaustive coverage. Runtime backward-compatible via the `message` fallback.

## What was tested

- TypeScript compiles clean (pre-existing env errors only — `@types/node` missing, pre-dates this PR)
- `prettier --check` passes
- Existing `test/lib/shape-drift-hint.test.js` tests the underlying `detectShapeDriftHint` string function (unchanged); those tests continue to pass once node_modules are available
- Detection logic in `shape-drift-hints.ts` mirrors `detectShapeDriftHint` case-for-case, including the `_message`-strip pre-processing applied in the runner before calling the detector

## Nits from pre-PR review (noted, not fixed)

- `expected_variant` for object-shape cases carries a schema fragment rather than a short token (e.g. `"{ creatives: [{ creative_id, action, ... }] }"`); a future cleanup could constrain the domain in JSDoc
- A base `StoryboardStepHintBase` interface with `kind + message` would be a useful constraint to add before the third hint kind lands (see issue #935 for the remaining four candidate kinds)

**Pre-PR review:**
- code-reviewer: approved — fixed RFC 6901 `instance_path` slash and `_message`-strip alignment (two blockers from initial review)
- ad-tech-protocol-expert: approved — non-breaking per runner-output contract extensibility rule; no spec alignment required

Session: https://claude.ai/code/session_01J6Jq6pcjkFrSURhgb5c2y5

---
_Generated by [Claude Code](https://claude.ai/code/session_01J6Jq6pcjkFrSURhgb5c2y5)_


bokelley:
Reviewed against issue #935's stated vision. Three concrete recommendations — addressed all three.

**1. Don't let `Closes #935` orphan the deferred taxonomy.**
Issue #935 enumerates five hint kinds; this PR ships one (per the issue's own incremental-rollout guidance). My triage comment on #935 said the other four were "deferred to follow-up issues" — but those issues didn't exist yet. Filed them so the taxonomy has a home in the tracker after this PR closes #935:

- #946 — `MissingRequiredFieldHint`
- #947 — `FormatMismatchHint`
- #948 — `MonotonicViolationHint`
- #949 — `AuthMisconfigurationHint`
- #950 — `StoryboardStepHintBase` (the base-interface refactor)

Each follow-up cites the parent issue and the in-flight PR.

**2. Tighten `ShapeDriftHint.expected_variant` and `observed_variant` typing.**
The pre-PR review nit called out that `expected_variant` ends up as a free-form schema-fragment string, which weakens the structured-contract value. Addressed in the companion commit (29d87cd7) just pushed:

- `observed_variant` is now a string-literal union of the five detector cases (`bare_array | platform_native_fields | per_item_shape | wrong_wrapper_key | raw_render_fields`) with `(string & {})` for forward compatibility — renderers can `switch` exhaustively, future detectors don't break consumers.
- `expected_variant` JSDoc now states explicitly it's a schema-fragment string for human display, and points renderers that need structured access to dispatch on `observed_variant` and look up the canonical wrapper key / response helper from the SDK rather than parsing this string.

**3. Land `StoryboardStepHintBase` with this PR.**
Same companion commit (29d87cd7). Adds a `StoryboardStepHintBase<K extends string>` interface; both existing kinds (`ContextValueRejectedHint`, `ShapeDriftHint`) `extend` it. The `kind` + `message` invariant every renderer relies on is now compile-time-enforced — the next four kinds can't drop it. Wire format unchanged; pure structural change. tsc clean, all 83 hint tests still pass.

Net: PR is now consistent with the full vision in #935 (one kind shipped + base + four follow-ups filed). The `Closes #935` line still works because the taxonomy is now tracked across the five follow-up issues — but if you'd rather keep #935 open as the umbrella tracker, change to `Refs #935` and close it manually after the four kinds land.

— Reviewed in collaboration with [@bokelley](https://github.com/bokelley).

bokelley:
Superseded by #959, which closes the broader vision in #935: adds the four other proposed hint kinds (`MissingRequiredFieldHint`, `FormatMismatchHint`, `MonotonicViolationHint`, plus `ShapeDriftHint` from this PR), introduces the `StoryboardStepHintBase` constraint, de-duplicates shape-drift between `ValidationResult.warning` and `step.hints[]`, plumbs structured `AssertionResult.hint` from `status.monotonic`, and updates `bin/adcp-step-hints.js` to render per-kind structured fields so the contract is observable in the SDK.

PR #938, opened 2026-04-25
feat(types): inline-union value arrays for anonymous string-literal unions (#932)

### LINKED ISSUE #932: Codegen: extract inline anonymous string-literal unions to const arrays
## Context

PR #931 added `${TypeName}Values` const arrays for every **named** string-literal union in the generated TypeScript (122 enums: `MediaChannelValues`, `PacingValues`, `AssetContentTypeValues`, etc.).

Inline anonymous unions are still uncovered, and that's where consumer-side duplicates live. Examples flagged on downstream PR review:

```ts
// scope3data/agentic-adapters: packages/shared/src/utils/adcp-format-requirements.ts
const VALID_IMAGE_FORMATS = new Set(['jpg','jpeg','png','gif','webp','svg','avif','tiff','pdf','eps'])
const VALID_VIDEO_CONTAINERS = new Set(['mp4','webm','mov','avi','mkv'])
```

These come from inline unions inside the generated TS, which json-schema-to-typescript renders without elevating to a named type:

```ts
// src/lib/types/core.generated.ts (from schemas/cache/3.0.0/core/requirements/image-asset-requirements.json)
formats?: ('jpg' | 'jpeg' | 'png' | 'gif' | 'webp' | 'svg' | 'avif' | 'tiff' | 'pdf' | 'eps')[];
```

The current `generate-enum-arrays.ts` regex skips these because they have no stable name. Adapters can't import them, so they re-declare the literal sets locally — that's the drift bait the original PR was supposed to remove.

## Suggested approach

Switch from regex-on-TypeScript to **Zod-introspection**. The `schemas.generated.ts` codegen already produces Zod schemas where union options expose `.options` (Zod 4) or `.def.options` introspection, and named schemas (e.g., `ImageAssetRequirementsSchema`) preserve the inline-union shapes inside their object schema.

Walk every named schema, recursively descend into object/array properties, and emit named const arrays for every `z.union([z.literal(...), ...])` it finds. Naming convention: `${ParentSchemaName}_${PropertyPath}Values` (e.g., `ImageAssetRequirements_FormatsValues`, `VideoAssetRequirements_ContainersValues`). The names are auto-generated but stable as long as the spec property names are.

Alternative: walk the JSON schemas directly under `schemas/cache/{version}/`, find every `enum: [...]` of strings, and synthesize names from the schema `$id` + JSON path. More authoritative but adds a separate pipeline that could drift from the TS/Zod generation.

## Acceptance

- [ ] At least the user-flagged cases land: `ImageAssetRequirements.formats`, `VideoAssetRequirements.formats` / `.containers`, `AudioAssetRequirements.formats`
- [ ] Cross-validation test (mirror of `test/lib/enum-arrays.test.js`): every emitted inline-array round-trips through the matching parent Zod schema
- [ ] Naming convention documented in the codegen script header so adapters can guess the export name from the property path

## Refs

- PR #931 — initial named-enum extraction
- Downstream review note: https://github.com/adcontextprotocol/adcp-client/pull/931 (see "P1 — inline anonymous unions are still uncovered" in the review thread)

🤖 Filed by Claude Code

bokelley:
## Summary

Closes #932. Companion to the named-enum exports landed in 5.17 (#931). The earlier shipment covered every spec enum that has a stable named type (`MediaChannelValues`, `PacingValues`, etc., 122 total). This adds the inline anonymous unions inside named schemas — exactly the cases where consumers were re-declaring spec literal sets in their own validation code:

```ts
// Before — drift bait, hand-maintained on the consumer side.
const VALID_IMAGE_FORMATS = new Set(['jpg', 'jpeg', 'png', 'gif', 'webp', 'svg', 'avif', 'tiff', 'pdf', 'eps']);
const VALID_VIDEO_CONTAINERS = new Set(['mp4', 'webm', 'mov', 'avi', 'mkv']);

// After — authoritative, drift-detected.
import {
  ImageAssetRequirements_FormatsValues,
  VideoAssetRequirements_ContainersValues,
} from '@adcp/client/types';
```

## Naming convention

Every `z.union([z.literal(...), ...])` (or `z.array(...)`-wrapped variant) inside a named object schema gets `${ParentSchema}_${PropertyName}Values`. Property names PascalCase. Property paths that reference a named enum (e.g. `unit: DimensionUnitSchema.optional()`) are intentionally skipped — those already have `${TypeName}Values` exports from `enums.generated.ts`.

## Coverage

104 inline-union arrays across 51 parent schemas. Includes all user-flagged cases:
- `ImageAssetRequirements_FormatsValues` (jpg, jpeg, png, gif, webp, svg, avif, tiff, pdf, eps)
- `VideoAssetRequirements_FormatsValues`, `_ContainersValues` (mp4, webm, mov, avi, mkv), `_CodecsValues`, `_FrameRateTypeValues`, `_ScanTypeValues`, `_GopTypeValues`, `_MoovAtomPositionValues`, `_AudioCodecsValues`, `_AudioChannelsValues`
- `AudioAssetRequirements_FormatsValues` (mp3, aac, wav, ogg, flac), `_ChannelsValues`
- 90+ more across `Account`, `AppItem`, `BrandResolveRequest`, `CatalogItemError`, etc.

## Implementation

New script `scripts/generate-inline-enum-arrays.ts` walks the compiled Zod schemas via runtime introspection (Zod 4 `_def`) rather than regex on the generated TS — cleaner and future-proofs against codegen output format changes. Output goes to `src/lib/types/inline-enums.generated.ts`. Wired into `generate-zod-schemas` (runs after Zod codegen since it depends on the schemas being current).

The extractor reads source TS (`schemas.generated.ts`) directly via tsx rather than the `dist/` build output, so it can run as part of pre-build codegen without a circular dependency on `build:lib`.

## Test plan

- [x] `npm run build:lib` — clean compilation, including the new generated file
- [x] `node --test test/lib/inline-enum-arrays.test.js` — 7/7 pass
  - User-flagged exports present
  - User-flagged values match the spec exactly (image formats + video containers explicit assertions)
  - Every emitted Values array is non-empty + all-strings
  - Cross-validation: every literal in every Values array round-trips through the parent Zod schema property (catches drift)
  - Dedup gate: named-enum references (e.g. `unit: DimensionUnitSchema`) are NOT re-emitted as inline-union exports
  - Public surface: exports reachable via `@adcp/client/types`
- [x] `npm run test:lib` — 4973 pass, 0 fail, 6 skipped (pre-existing)
- [x] `tsc --noEmit` — clean

## Behavior unchanged

Pure addition. No public-API rename, no breaking change to `enums.generated.ts`, no behavior change in `createAdcpServer` or any other module. Adapters can drop their hand-maintained `VALID_IMAGE_FORMATS`-style constants in a follow-up PR.


bokelley:
Ran three expert reviews in parallel (code-reviewer, dx-expert, ad-tech-protocol-expert). Convergence: ship the convention. Three pre-merge hardenings applied in c2c62830, four follow-up issues filed.

### Pre-merge fixes (c2c62830)

- **Dropped `'pipe'` from UNWRAP_TYPES** (code-reviewer). `z.pipe(in, out)` chains transforms; unwrapping `in` would silently extract the wrong half for `string → enum` style cases. Current `schemas.generated.ts` has zero pipes (verified), so this is latent-footgun removal, not a live-bug fix.
- **Fingerprint fallback on the dedup gate** (code-reviewer). Identity-based check stays as the fast path; sorted-literal-tuple fingerprint catches the regression where Zod ever clones a schema on property access. Side effect: **drops 4 redundant entries** (104 → 100) that were previously double-emitting named-enum values under wrapped names — `ListAccountsRequest_Status` ≡ `AccountStatusValues`, `ReportingWebhook_ReportingFrequency` ≡ `ReportingFrequencyValues`, `VASTAssetRequirements_VastVersion` ≡ `VASTVersionValues`, `WebhookAssetRequirements_Methods` ≡ `HTTPMethodValues`. Each dropped name's values stay reachable via the matching named-enum export.
- **Floor bumped from <20 to <90** (code-reviewer). Catches partial-regression, not just catastrophic collapse.
- **Test cross-validation regex switched to last-underscore split** (code-reviewer). Parents with SCREAMING_SNAKE prefixes (`RATE_LIMITEDDetails`) now get cross-validated instead of silently skipped.
- **Negative-case test added** — pins the invariant that mixed-type unions (string + number, boolean) are skipped.

### Deferred to follow-up issues

- **#939 — Hoist remaining inline duplicates into named enums** (dx-expert). `Account_PaymentTermsValues` and `GetAccountFinancialsSuccess_PaymentTermsValues` are byte-identical; `RightsConstraint`/`RightsPricingOption` `_PeriodValues` ditto; `AudioAsset.channels` vs `VideoAsset.audio_channels`; `Create`/`UpdateMediaBuySuccess._ValidActionsValues`; the IAB-style `_ObjectiveValues` set. ~8–10 of the remaining 100 are duplicates that should be hoisted upstream into named enums (extends #931's convention).
- **#940 — Rename `RATE_LIMITEDDetails` and other SCREAMING_SNAKE schema names** (code-reviewer + dx-expert). The `RATE_LIMITED` parent name is propagating a discriminator value into a public type name (`RATE_LIMITEDDetails_ScopeValues`). Same root cause as the `AgeVerificationMethod1` numbered-dupe in `core.generated.ts` — it's a json-schema-to-typescript artifact, fixable upstream.
- **adcontextprotocol/adcp#TBD — Mark `vast_version` `x-extensible: true`** (protocol-expert). IAB ships VAST 4.3 (and 5.0 working draft); pinning a closed `[2.0..4.2]` set in the SDK risks rejecting conforming traffic on the next spec rev. Concrete OpenRTB-style drift case the protocol expert flagged.
- **adcontextprotocol/adcp#TBD — Document enum stability policy** (protocol-expert). Spec uses `x-extensible: true` to flag append-only taxonomies, but doesn't have a one-paragraph statement saying "absent = closed/breaking-to-extend." Worth pinning explicitly so SDK consumers and adapter authors know the contract these `as const` arrays carry.

### Spot-check verdict

Protocol expert verified all 100 arrays match the spec (image formats, video containers/codecs/audio_codecs, audio formats/channels, vast_version, webhook methods, destination_type, image color_space). One legitimate quirk: `AudioAsset.channels` (`mono, stereo, 5.1, 7.1`) and `AudioAssetRequirements.channels` (`mono, stereo` only) differ deliberately — that's the spec, not an extractor bug.

### Convergence

Three experts, three lenses (correctness, DX, protocol). Zero blockers, all "ship-with-targeted-fixes." The capability-gate footgun pattern from PR #931 didn't repeat here — this PR's contract is "expose values that exist in the spec; consumers handle drift via versioning" — narrower and easier to keep honest.

bokelley:
Correction to issue numbers in my prior comment — the deferred follow-ups landed as **#941** (hoist duplicate inline-union literal sets to named enums) and **#942** (rename SCREAMING_SNAKE schema artifacts). The two adcontextprotocol/adcp spec-side issues (vast_version x-extensible + enum stability policy) still pending.

PR #960, opened 2026-04-25
fix(server): hydrate tools/list inputSchema from bundled request schemas (#954)

### LINKED ISSUE #954: createAdcpServer publishes empty inputSchema on tools/list — breaks buyer introspection and field-stripping
## Symptom

`createAdcpServer` registers each tool's MCP `inputSchema` as `{ type: 'object', properties: {} }` — the tools/list response advertises no fields, even though every AdCP 3.0 tool has a published JSON request schema in `schemas/cache/{version}/<domain>/<tool>-request.json`.

This was surfaced during triage of #940. Two consequences:

### 1. Buyer-side field-stripping silently no-ops

`SingleAgentClient.adaptRequestForServerVersion` (in `src/lib/core/SingleAgentClient.ts:1190-1191`) caches `tools/list` schemas and strips any top-level field not declared in the tool's schema. With `properties: {}` it falls through the empty-properties fail-open path:

```ts
const toolSchema = this.cachedToolSchemas?.get(taskType);
if (!toolSchema || Object.keys(toolSchema).length === 0) return adapted;
```

…which is the right defensive choice given what the framework publishes (you can't strip when you don't know the shape) — but it means the strip-unknown-fields safety net never fires for framework-registered tools. Pre-#909 this didn't matter because request validation was `'warn'`. Post-#909, anything injected upstream of `adaptRequestForServerVersion` reaches AJV strict and fails — which is what blew up #940.

### 2. Buyers cannot introspect tool shapes

A buyer agent inspecting `tools/list` to discover what a server accepts gets back a generic stub. That defeats the point of MCP's tool discovery and forces buyers back to out-of-band schema fetches (which the spec doesn't standardize for framework-registered servers).

## Why this is a server wiring gap, not a spec gap

The schemas exist and ship in the package. `validateRequest` already compiles them at `src/lib/server/create-adcp-server.ts:2203` for request validation. The same loader could populate the MCP `inputSchema` field at registration time.

## Suggested fix

At tool registration in `createAdcpServer`, hydrate `inputSchema` from the loaded request schema for the tool, falling back to `{type: 'object', additionalProperties: true}` if the schema is unavailable. The fallback at minimum communicates "shape unknown, accept anything" rather than the false "shape known to be empty."

## Related

- #940 (root cause of the buyer-side discovery)
- #909 (strict request validation flip that made the empty-properties case visible)
- PR #943 (fixes the buyer-side bug at the source by gating runner injections)

bokelley:
Closes #954

## Summary

`createAdcpServer` was registering every framework tool with `z.object({}).passthrough()` as `inputSchema`, which the MCP SDK serialised to `{ type: 'object', properties: {} }` — advertising no fields. Two consequences:

1. **Buyer introspection broken** — `tools/list` returned stub schemas; agents had to fall back to `get_adcp_capabilities` for field shapes.
2. **`adaptRequestForServerVersion` field-stripping never fired** — `SingleAgentClient` caches `tools/list` schemas and strips unknown fields, but the empty-properties fail-open guard (`Object.keys(toolSchema).length === 0 → return adapted`) was always taken, so the safety net was dead for all framework-registered tools.

### Fix

New exported `getRawRequestSchema(toolName)` in `schema-loader.ts`:
- Returns the pre-resolved JSON schema from `schemas/cache/{version}/bundled/` for bundled tools.
- Relaxes `additionalProperties: false` at root **and** in all direct `oneOf`/`anyOf`/`allOf` branches (reusing a new shared `relaxAdditionalPropertiesFalse` helper, also used by `relaxResponseRoot`) so the MCP SDK does not reject envelope fields (`idempotency_key`, `context`, `caller`, `ext`) if it validates the advertised shape.
- Returns `null` for flat-tree domain schemas (governance, property-lists, brand) that have unresolved `$ref`s — those still fall back to `PASSTHROUGH_INPUT_SCHEMA`.

`createAdcpServer` now passes the bundled schema as `inputSchema` when available. `get_adcp_capabilities` is unchanged (passthrough is appropriate — its `context` field is free-form).

### Scope

Flat-tree tools (governance, property-lists, collection-lists, brand-rights) still advertise `properties: {}`. This is a separate follow-up: the fix for those tools is to pre-resolve their schemas into `bundled/` during `npm run sync-schemas`. The DX gap does not regress from pre-#954 state — those tools already had empty schemas.

## What was tested

- `tsc --project tsconfig.lib.json --noEmit` — no new errors beyond pre-existing env issues (`@types/node` not installed, deprecated `moduleResolution=node10`).
- `npm test` — 72 passing tests both before and after the change; pre-existing failures unrelated (missing `dist/` build artifacts).
- `git diff main...HEAD` reviewed manually — changes confined to `schema-loader.ts`, `create-adcp-server.ts`, `SingleAgentClient.ts` (comment only), and changeset.

## Pre-PR review

- **code-reviewer**: approved — blocker (missing changeset) fixed; branch-relaxation gap fixed via shared `relaxAdditionalPropertiesFalse` helper; cast documented with explanation of why SDK uses raw JSON Schema for `tools/list` only.
- **dx-expert**: approved (no blockers) — notes flat-tree gap is real but does not regress pre-#954; `SingleAgentClient` comment updated; recommends follow-up issue for flat-tree schema bundling.

## Nits noted (not fixed in this PR)

- Flat-tree tools (governance, property-lists, etc.) still show `properties: {}` — needs `bundled/` pre-resolution in `sync-schemas` as a follow-up.
- Silent `catch {}` in `getRawRequestSchema` swallows I/O errors; could log at `debug` level to distinguish "not synced" from "disk error."

## Pre-merge verification needed

⚠️ Confirm that `@modelcontextprotocol/sdk ^1.29.0` uses a raw JSON Schema object passed to `registerTool` for `tools/list` serialisation only and does **not** call `.parse()` on it — this is the assumption behind the `AnySchema` cast. If the SDK compiles and validates raw JSON Schemas in `validateToolInput`, envelope fields would need to be explicitly declared or the `additionalProperties: true` relaxation verified sufficient. The comment in the code at the cast site documents this assumption.

---

Session: https://claude.ai/code/session_014quxUBo2kwYeyrttmnMA8A

---
_Generated by [Claude Code](https://claude.ai/code/session_014quxUBo2kwYeyrttmnMA8A)_

### DIFF
--- a/.changeset/hydrate-tools-list-input-schema.md
+++ b/.changeset/hydrate-tools-list-input-schema.md
@@ -1,0 +1,7 @@
+---
+'@adcp/client': patch
+---
+
+**Fix `createAdcpServer` publishing empty `inputSchema` on `tools/list` (#954).** Framework-registered tools were wired with `z.object({}).passthrough()` as `inputSchema`, which the MCP SDK serialised to `{ type: 'object', properties: {} }` — advertising no fields even though every AdCP 3.0 tool has a published JSON Schema. Two consequences: buyer agents doing `tools/list` introspection could not discover field shapes, and `SingleAgentClient.adaptRequestForServerVersion`'s field-stripping guard silently failed open for all framework-registered tools (empty-properties fail-open path always taken).
+
+`createAdcpServer` now calls the new `getRawRequestSchema(toolName)` from `schema-loader.ts`, which returns the pre-resolved JSON schema from `schemas/cache/{version}/bundled/` for bundled tools. `additionalProperties: false` is relaxed at the root and in all direct `oneOf`/`anyOf`/`allOf` branches so the MCP SDK does not reject protocol envelope fields (`idempotency_key`, `context`, `caller`, `ext`) if it validates the advertised shape. Flat-tree domain schemas (governance, property-lists) contain unresolved `$ref`s and still fall back to the passthrough Zod schema. `get_adcp_capabilities` is unchanged.
--- a/src/lib/core/SingleAgentClient.ts
+++ b/src/lib/core/SingleAgentClient.ts
@@ -1183,9 +1183,11 @@ export class SingleAgentClient {
     // Fails open when no schema is cached OR when the schema declares no
     // properties (JSON Schema semantics: an object with no properties
     // and no `additionalProperties: false` accepts any shape). Post-#909,
-    // framework-registered agents publish `{ type: 'object', properties: {} }`
-    // on tools/list — treating that as "strip everything" would silently
-    // drop every field the buyer sent.
+    // framework-registered agents using bundled schemas publish real field
+    // shapes on tools/list (#954); flat-tree domain tools (governance,
+    // property-lists) still publish `{ type: 'object', properties: {} }` as
+    // a fallback — treating that as "strip everything" would silently drop
+    // every field the buyer sent.
     // MCP-only in practice: A2A agents don't populate cachedToolSchemas.
     const toolSchema = this.cachedToolSchemas?.get(taskType);
     if (!toolSchema || Object.keys(toolSchema).length === 0) return adapted;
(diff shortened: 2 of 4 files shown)


bokelley:
Closing — superseded by #964, which documents that the empty `tools/list` `inputSchema` is intentional (LLM context-window economy). Consumers should read raw schemas from `schemas/cache/` via `schema-loader.ts` rather than re-inlining them into `tools/list`.

PR #968, opened 2026-04-25
fix(protocols): use native A2A tasks/get JSON-RPC for polling (#963)

### LINKED ISSUE #963: fix(protocols): A2A tasks/get polling routes through callTool instead of native JSON-RPC method
## Context

When an AdCP seller returns a \`status: 'submitted'\` arm, the SDK's \`TaskExecutor.pollTaskCompletion\` (\`src/lib/core/TaskExecutor.ts:1152\`) drives a polling loop that calls \`getTaskStatus(agent, taskId)\`. For MCP transports, that routes through the dedicated \`getMCPTaskStatus\` helper (the MCP Tasks-protocol \`tasks/get\` method). For A2A, it falls through to:

\`\`\`ts
// src/lib/core/TaskExecutor.ts:1138-1149
const response = (await ProtocolClient.callTool(
  agent,
  'tasks/get',
  { taskId },
  []
)) as Record<string, unknown>;
\`\`\`

## Bug

\`ProtocolClient.callTool\` for an A2A agent dispatches via \`callA2ATool\` → \`message/send\` with \`{ skill: 'tasks/get', parameters: { taskId } }\`. That treats \`tasks/get\` as an AdCP **tool name**, but A2A \`tasks/get\` is a **native JSON-RPC method** on the A2A protocol — distinct from \`message/send\` and unrelated to AdCP tool dispatch.

A conformant A2A seller running our \`createA2AAdapter\` will:
1. Receive \`message/send { skill: 'tasks/get', input: { taskId } }\`
2. Route to \`AdcpServer.invoke({ toolName: 'tasks/get' })\`
3. Reject with \`A2AInvocationError\` (no AdCP tool by that name registered)
4. Return \`Task.state: 'failed'\`

The SDK's polling loop then sees a failed task and surfaces it as the AdCP task failing — which is wrong. The AdCP task may still be queued; the SDK just couldn't ask the right question.

## Why this didn't surface

The new tests in #952 dodged this because:
- \`test/storyboard-a2a-wire-shape-capture.test.js\` — the seller fixture's \`createMediaBuy\` returns \`submitted\` synchronously and the storyboard never polls
- \`test/storyboard-a2a-async-submitted-yaml.test.js\` — the seller fixture auto-completes the task on the same call (the comment at \`buildSellerHandlers\` calls this out)

Both fixtures short-circuit the polling cycle. A real submitted→completed round-trip over A2A — which is what \`adcontextprotocol/adcp#3083\` plus \`comply_test_controller\` will eventually drive — needs the polling cycle to work natively.

## Proposed fix

\`TaskExecutor.getTaskStatus\` needs an A2A branch analogous to the MCP one. Sketch:

\`\`\`ts
async getTaskStatus(agent: AgentConfig, taskId: string): Promise<TaskInfo> {
  if (agent.protocol === 'mcp') {
    // existing MCP path
  }
  if (agent.protocol === 'a2a') {
    const client = await getOrCreateA2AClient(agent.agent_uri, getAuthToken(agent));
    const task = await client.getTask({ id: taskId });
    return adaptA2aTaskToTaskInfo(task);
  }
  // …
}
\`\`\`

The \`@a2a-js/sdk/client\` exposes \`getTask({ id })\` which dispatches the native A2A \`tasks/get\` JSON-RPC method. Adapting the returned \`Task\` to \`TaskInfo\` requires mapping:
- \`Task.status.state\` (A2A enum) → \`status\` (\`ADCP_STATUS\`)
- \`Task.artifacts[0].parts[0].data\` → \`result\`
- \`Task.id\` → \`taskId\`
- \`Task.history[0].timestamp\` → \`createdAt\`
- (probably) the AdCP \`status\` from \`artifact.parts[0].data.status\` for the submitted→working→completed AdCP-level lifecycle

## Test plan

- Drive a real submitted→completed transition over A2A using \`comply_test_controller\` to flip seller-side state between calls
- Assert \`waitForCompletion()\` resolves with the completion artifact
- Assert the polling cycle issues native A2A \`tasks/get\` (not \`message/send\`) — \`rawA2aProbe\`-style capture will show the wire method
- Negative test: seller that returns \`Task.state: 'failed'\` on \`tasks/get\` surfaces as \`status: 'failed'\` on the resolved \`TaskResult\`

## Why this matters

The pre-#899 ecosystem had A2A-over-MCP-shim sellers where this would have worked by accident. Post-#899 the contract is that A2A sellers expose A2A natively, and the SDK-side polling cycle needs to honor that. The storyboard-side regression guard (#952's \`a2a_submitted_artifact\`) covers the wire shape; this issue covers the round-trip.

## Related

- adcp-client#899 — A2A serve adapter (the contract this fixes against)
- adcp-client#952 — A2A wire-shape regression guard (sibling fix; covers single-call shape, not the polling cycle)
- adcp-client#904 — umbrella tracking issue

bokelley:
## Problem

`TaskExecutor.getTaskStatus` fell through to `ProtocolClient.callTool` for A2A agents, dispatching `message/send { skill: 'tasks/get' }`. Conformant sellers reject this because `tasks/get` is a native A2A JSON-RPC method, not an AdCP tool name. Separately, the polling closures in `setupSubmittedTask` used the client-minted correlation UUID as the task ID, but `tasks/get` requires the server-assigned `Task.id` from the `message/send` response.

Fixes #963.

## Changes

**`src/lib/protocols/a2a.ts`**
- New `getA2ATaskStatus(agentUrl, a2aTaskId, authToken)` — calls `client.getTask({ id })` directly via the `@a2a-js/sdk` transport, bypassing `message/send`
- New `mapA2ATaskToTaskInfo` — maps A2A `Task` to `TaskInfo`; prefers `artifact.parts[last-data].data.status` over `Task.state` (state is always `'completed'` for submitted AdCP arms; real lifecycle lives in the artifact)
- 401 cache eviction evicts both the unsigned cache entry (used by `getA2ATaskStatus`) and any signed entry that may exist from a prior `callA2ATool` call
- `getA2ATaskStatus` exported from `src/lib/protocols/index.ts`

**`src/lib/core/TaskExecutor.ts`**
- `getTaskStatus`: new A2A branch calls `getA2ATaskStatus` with the server-assigned task ID instead of routing through `callTool`
- `setupSubmittedTask`: extracts A2A server Task.id via `responseParser.getTaskId(response)` and threads it through the `track` / `waitForCompletion` closures
- `pollTaskCompletion`: accepts optional `a2aServerTaskId` and threads it into `getTaskStatus`; also adds `REJECTED` as a terminal exit condition (pre-existing gap, now reachable for A2A)

**`test/lib/a2a-task-status.test.js`** (new, 12 tests)
- Verifies status is read from artifact data, not `Task.state`
- Verifies `failed`/`rejected` set `error`; `canceled` does not
- Verifies ISO timestamp → epoch ms conversion
- Verifies `getTask` is called with the server-assigned ID
- Verifies `sendMessage` is never called
- Validates empty/oversized `taskId` input
- Verifies fallback to A2A state when no artifacts are present

**`.changeset/a2a-native-tasks-get-polling.md`** — patch changeset

## Key invariant

A2A `Task.status.state` is always `'completed'` for submitted AdCP arms because the adapter marks the transport call done while AdCP work is queued. The real AdCP lifecycle status (`submitted`, `working`, `completed`, …) lives in `artifact.parts[last-data-part].data.status`. `Task.state` is only used as a fallback when no artifact payload is present yet, and to surface transport-level terminal failures (`failed`, `rejected`).

## Test plan

- [x] `npm run build:lib` — clean build
- [x] `node --test test/lib/a2a-task-status.test.js` — 12/12 pass
- [x] `npx prettier --check` — no formatting issues

https://claude.ai/code/session_015ga1KhzvR7DbWvwsqDVSaW — Tracked at #997

### DIFF
--- a/.changeset/a2a-native-tasks-get-polling.md
+++ b/.changeset/a2a-native-tasks-get-polling.md
@@ -1,0 +1,11 @@
+---
+"@adcp/client": patch
+---
+
+fix(protocols): A2A polling now uses native `tasks/get` JSON-RPC instead of `callTool/message/send`
+
+`TaskExecutor.getTaskStatus` was routing A2A task polling through `ProtocolClient.callTool('tasks/get', …)`, which dispatched `message/send { skill: 'tasks/get' }`. Conformant A2A sellers (running `createA2AAdapter`) reject this because `tasks/get` is a native A2A JSON-RPC method, not an AdCP tool name — causing the polling loop to surface a false task-failed result.
+
+**Fix:** Added `getA2ATaskStatus` to `src/lib/protocols/a2a.ts` (parallel to `getMCPTaskStatus`) that calls `client.getTask({ id })` on the `@a2a-js/sdk` `A2AClient`. `TaskExecutor.getTaskStatus` now dispatches to this for `protocol: 'a2a'` agents. `setupSubmittedTask` extracts the server-assigned A2A `Task.id` from the initial `message/send` response and threads it into the `track`/`waitForCompletion` closures — using it instead of the client-minted correlation UUID that `tasks/get` would not recognize.
+
+**Status mapping:** `Task.status.state` is always `'completed'` for submitted AdCP arms (the transport call finished, not the AdCP work). The real AdCP lifecycle status is now read from `artifact.parts[last-data-part].data.status`, with A2A terminal states (`failed`/`rejected`) as the fallback when no artifact payload is present.
--- a/src/lib/core/TaskExecutor.ts
+++ b/src/lib/core/TaskExecutor.ts
@@ -5,6 +5,7 @@ import { randomUUID } from 'crypto';
 import type { AgentConfig } from '../types';
 import { ProtocolClient } from '../protocols';
 import { getMCPTaskStatus, listMCPTasks } from '../protocols/mcp-tasks';
+import { getA2ATaskStatus } from '../protocols/a2a';
 import { getAuthToken } from '../auth';
 import { is401Error, adcpErrorToTypedError } from '../errors';
 import type { ADCPError } from '../errors';
@@ -912,11 +913,19 @@ export class TaskExecutor {
       await this.config.webhookManager.registerWebhook(agent, taskId, webhookUrl);
     }
 
+    // For A2A agents, `tasks/get` is a native JSON-RPC method keyed on the
+    // server-assigned Task.id — not the client-minted correlation UUID. Extract
+    // it from the initial message/send response so the polling closures below
+    // can pass it to getA2ATaskStatus.
+    const a2aServerTaskId =
+      agent.protocol === 'a2a' ? (this.responseParser.getTaskId(response) ?? undefined) : undefined;
+
     const submitted: SubmittedContinuation<T> = {
       taskId,
       webhookUrl,
-      track: () => this.getTaskStatus(agent, taskId),
-      waitForCompletion: (pollInterval = 60000) => this.pollTaskCompletion<T>(agent, taskId, pollInterval),
+      track: () => this.getTaskStatus(agent, taskId, a2aServerTaskId),
+      waitForCompletion: (pollInterval = 60000) =>
+        this.pollTaskCompletion<T>(agent, taskId, pollInterval, a2aServerTaskId),
     };
 
     return {
@@ -1125,7 +1134,7 @@ export class TaskExecutor {
     }
   }
 
-  async getTaskStatus(agent: AgentConfig, taskId: string): Promise<TaskInfo> {
+  async getTaskStatus(agent: AgentConfig, taskId: string, a2aServerTaskId?: string): Promise<TaskInfo> {
     // Use MCP Tasks protocol method when available
     if (agent.protocol === 'mcp') {
       const authToken = getAuthToken(agent);
@@ -1135,6 +1144,20 @@ export class TaskExecutor {
         if (is401Error(err)) throw err;
         // Fall through to tool call if protocol method is not supported
       }
+    }
+    // Use the native A2A tasks/get JSON-RPC method for A2A agents. The A2A
+    // protocol's tasks/get is distinct from the AdCP tool of the same name —
+    // routing it through callTool/message/send causes conformant sellers to
+    // reject with A2AInvocationError (no such AdCP tool registered).
+    if (agent.protocol === 'a2a') {
+      if (!a2aServerTaskId) {
+        throw new Error(
+          `A2A polling requires the server-assigned Task.id from the initial message/send response. ` +
+            `Ensure the A2A seller returns a Task with a non-empty id on the submitted arm. ` +
+            `(client task id: ${taskId})`
+        );
+      }
+      return await getA2ATaskStatus(agent.agent_uri, a2aServerTaskId, getAuthToken(agent));
     }
     const response = (await ProtocolClient.callTool(
       agent,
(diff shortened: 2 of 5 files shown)


bokelley:
## Expert reviews — approve with non-blocking notes

Two parallel reviews (ad-tech-protocol-expert + code-reviewer) converged on **approve**, with one Should-Fix already addressed in `ca513b27`:

**Applied fix (ca513b27):** runtime `typeof` guard for the AdCP status field in `mapA2ATaskToTaskInfo` — the `as string | undefined` cast over `Record<string, unknown>` would silently accept non-string status values and propagate them through to the `ADCP_STATUS.*` comparison. Now narrows with `typeof rawStatus === 'string'`.

**Protocol expert highlights:**
> `client.getTask({ id })` matches the A2A 0.3.0 SDK signature. The artifact-first / `Task.status.state`-fallback precedence is correct: `Task.status.state` is correctly read as nested (matches `TaskStatus` type), and the AdCP-over-A2A submitted convention does park the lifecycle status in `artifact.parts[].data.status`. `getTaskId` already recognizes `result.kind === 'task'` and returns the server-assigned id.

**Code-reviewer highlights:**
> The fix is correct: `tasks/get` is a native A2A JSON-RPC method, not an AdCP tool, and routing it through `callTool/message/send` is the documented bug. The new path uses `client.getTask({ id })`, threads the server-assigned `Task.id` from the initial `message/send` response, and falls back sensibly when the artifact payload is absent.
> Tests assert wire-level behavior (what `getTask` is called with, that `sendMessage` is never invoked) and the resulting `TaskInfo` shape. Good shape — not testing internals.

**Non-blocking nits filed for follow-up:**
- `taskType: 'unknown'` could persist the originating tool name from `setupSubmittedTask` so downstream consumers don't lose tool context.
- The dual cache-delete in the 401 branch is defensive but only the unsigned-key delete actually matches; the signed-key delete is a no-op outside `signingContextStorage.run`. Worth a comment clarifying.
- One storyboard-level test that exercises the submitted A2A arm → poll → terminal path (in addition to the unit-level `getA2ATaskStatus` tests).

Closing companion PR #972 as superseded — its bug is a subset of what this fixes via the closure capture; the public-API semantic shift (`taskId` = server ID + new `operationId` field) is independent and needs its own discussion.

bokelley:
Closing pending design clarification.

## What the review surfaced

Two parallel reviewers (ad-tech-protocol-expert + code-reviewer) approved the patch with non-blocking notes. The code is well-tested (`test/lib/a2a-task-status.test.js`) and well-mapped to A2A 0.3.0's `client.getTask({ id })` signature.

## Why I'm not merging

Main currently routes `tasks/get` through `ProtocolClient.callTool` for both MCP and A2A by deliberate design (`src/lib/core/TaskExecutor.ts:1293-1340`). The AdCP spec defines `tasks/get` as a buyer-callable AdCP **tool** (`schemas/cache/<v>/bundled/core/tasks-get-{request,response}.json`) that tracks AdCP work lifecycle — not as a transport-native polling method. The comment on main is explicit:

> *"MCP sellers that haven't registered `tasks/get` as an AdCP tool will surface a tool-not-found error rather than silently polling the wrong lifecycle."*

This PR moves A2A polling to native `tasks/get` JSON-RPC, which polls A2A's transport-call lifecycle (always 'completed' for AdCP submitted arms) rather than AdCP work lifecycle. The mapper in this PR partially compensates by reading `artifact.parts[].data.status`, but only when an artifact is present — the path stops being equivalent to what AdCP's tool-based `tasks/get` returns when the seller's submitted-arm hasn't yet emitted its first artifact data part.

## Real interop concern

The reporter premise — *"conformant A2A sellers reject `message/send { skill: 'tasks/get' }` because A2A 0.3.0 reserves `tasks/get` as a native JSON-RPC method"* — is plausible for sellers built on `@a2a-js/sdk`'s `DefaultRequestHandler` (which routes JSON-RPC method names directly). It's a real spec-interop gap, but the fix isn't unilateral SDK-side rerouting.

## Path forward

Filed adcontextprotocol/adcp-client#XXX (will create separately) to track the design question:
- Should AdCP's tool-based `tasks/get` also work at the A2A native JSON-RPC layer (sellers would route both surfaces to the same handler)?
- Or should the SDK probe the seller's `tools/list` / capability surface to determine which path is registered?
- Or is the right answer for sellers to register both?

Until that's resolved upstream, the documented main behavior is the correct one to keep.

The applied Should-Fix on this branch (typeof guard for AdCP status field — `ca513b27`) is a good defensive change worth applying separately to `mapTasksGetResponseToTaskInfo` on main.

Issue #962, opened 2026-04-25
feat(testing): a2a_context_continuity validator for multi-step storyboards


bokelley:
## Context

PR #952 added the \`a2a_submitted_artifact\` storyboard validation, which catches the post-#899 wire-shape regression on a single submitted arm. The protocol expert review on #952 flagged a sibling regression class that needs its own validator.

A2A 0.3.0 §7.1 (\`message/send\`) lets buyers pass \`params.message.contextId\` to bind a send to an existing context; the server MUST echo it on the response Task. The \`@a2a-js/sdk\` \`DefaultRequestHandler\` does this automatically — \`createA2AAdapter\` consumes \`requestContext.contextId\` at \`a2a-adapter.ts:394\` and passes it through to \`publishInitialTask\` at \`:488-499\` — so a passing seller built on the SDK will never trip a single-call check.

The regression class is sellers that bypass the SDK's request handler and stamp their own \`contextId\` on the response. That's only catchable on **multi-step** storyboards: step N captures the response \`Task.contextId\`, step N+1 sends with that contextId in \`params.message.contextId\`, the response \`Task.contextId\` MUST equal the inbound one.

## What's missing

No storyboard validation today asserts cross-step contextId continuity over A2A. Every existing A2A test fires a single \`message/send\` and asserts on the immediate response — sufficient for the wire shape #952 catches, insufficient for the multi-step session model.

## Proposed scope

New storyboard validation: \`a2a_context_continuity\`.

Shape:
- Cross-step (similar pattern to \`refs_resolve\`): runs at step N+1; reads the captured A2A envelope from step N to compare contextIds
- Skips with \`not_applicable\` on:
  - Non-A2A runs (no envelope to compare)
  - First step in the run (nothing to compare against)
  - Steps where the previous A2A response had no \`contextId\` to begin with
- Fails when:
  - Step N+1's outbound \`params.message.contextId\` was set to step N's value but the response Task carries a different \`contextId\`
  - Bonus: when the seller's response Task drops \`contextId\` entirely on a follow-up send (treats undefined/empty as a continuity break)

Implementation notes:
- Cross-step state already accumulated by the runner; the new validator reads \`runState.priorStepResults\` (or a dedicated map of A2A envelopes by step id)
- The runner needs to also track the OUTBOUND \`params.message.contextId\` so the validator can confirm step N+1 actually sent the prior contextId — otherwise we're asserting a tautology
[...]

PR #969, opened 2026-04-25
feat(conformance): a2a_context_continuity validator for multi-step storyboards

bokelley:
## Summary

Refs #962 — adds the `a2a_context_continuity` storyboard validation check, the sibling to `a2a_submitted_artifact` (#952) called for by the protocol expert review on that PR.

- Sellers that bypass the `@a2a-js/sdk` `DefaultRequestHandler` and stamp their own `contextId` on the response Task (instead of echoing the buyer-supplied value) now fail multi-step storyboards
- SDK-based sellers pass silently — no change to conformant agents
- Non-breaking: new additive validator; `not_applicable` skip on non-A2A runs, first step, capture miss, and JSON-RPC error envelopes

## What changed

**`types.ts`** — adds `'a2a_context_continuity'` to `StoryboardValidationCheck`

**`validations.ts`** — adds `outboundA2aContextId?: string` to `ValidationContext` and the `validateA2AContextContinuity()` function

**`runner.ts`** — `ExecutionState.lastA2aContextId` accumulates the response Task.contextId across A2A steps; before each dispatch, the current value is forwarded as `TaskOptions.contextId` (rides on `params.message.contextId` on the wire). `outboundA2aContextId` is snapshotted before dispatch so the validator gets "what was sent", not "what came back"

**`task-map.ts`** — `executeStoryboardTask` accepts `contextId` in its opts and passes it through `TaskOptions`, which `SingleAgentClient` / `TaskExecutor` already forward to `callA2ATool` → `params.message.contextId`

**`test/storyboard-a2a-context-continuity.test.js`** — 9 unit tests

## Validator behavior

| Condition | Result |
|---|---|
| No `outboundA2aContextId` (first A2A step, non-A2A run, prior step had no contextId) | `passed: true`, `not_applicable` observation |
| No `a2aEnvelope` (capture miss) | `passed: true`, `not_applicable` observation |
| JSON-RPC error envelope (no Task returned) | `passed: true`, `not_applicable` observation |
| Response `Task.contextId` equals outbound contextId | `passed: true` |
| Response `Task.contextId` differs from outbound | `passed: false`, `/result/contextId` pointer |
| Response `Task.contextId` absent/empty on follow-up | `passed: false`, `/result/contextId` pointer |

## What I tested

- [x] `tsc --noEmit` clean
- [x] All 9 new unit tests pass
- [x] All 18 existing `a2a_submitted_artifact` tests pass (no regression)
- [x] Pre-existing failures in `storyboard-a2a-wire-shape-capture.test.js` confirmed pre-existing (verified against baseline)

## Pre-PR review

- **code-reviewer**: approved — 1 blocker fixed (JSON-RPC error changed to `not_applicable`; distinct from `a2a_submitted_artifact` because contextId continuity is only verifiable when a Task was returned). 1 nit (in PR body: "self-skip" means `passed: true` with an observation, not a distinct `skipped` state — same pattern as `a2a_submitted_artifact`).
- **ad-tech-protocol-expert**: approved — 1 open question flagged for human review: the docstring cites A2A 0.3.0 §7.1 with "MUST echo" language; the normative keyword in the spec text has not been verified from the canonical `google/A2A` spec doc. @bokelley (issue author) wrote the issue with "MUST echo" language and the existing SDK treats this as mandatory, but if §7.1 says SHOULD, this validator should be downgraded to a warning. Please confirm before merging.

## Nits (not fixed)

- `not_applicable` is surfaced as `passed: true` with an `observations` entry — consistent with `a2a_submitted_artifact`'s pattern, not a separate `skipped` state. Dashboards render as "passed with notes", not "skipped".

Session: https://claude.ai/code/session_01AP7aH782dD99kJamNgiYsd

---
_Generated by [Claude Code](https://claude.ai/code/session_01AP7aH782dD99kJamNgiYsd)_

### DIFF
--- a/.changeset/a2a-context-continuity-validator.md
+++ b/.changeset/a2a-context-continuity-validator.md
@@ -1,0 +1,42 @@
+---
+'@adcp/client': minor
+---
+
+Storyboard runner: add `a2a_context_continuity` validation check for
+multi-step A2A storyboards. Closes the sibling regression class to
+adcp-client#952 — sellers that bypass the `@a2a-js/sdk`
+`DefaultRequestHandler` and stamp their own `contextId` on the response
+Task (instead of echoing the buyer-supplied value) will now fail the
+storyboard suite.
+
+A2A 0.3.0 §7.1 lets buyers pass `params.message.contextId` to bind a
+follow-up send to an existing server-side context; the server must echo
+the same value on the response Task. The SDK handles this automatically
+via `createA2AAdapter`'s `requestContext.contextId` forwarding, so SDK-
+based sellers pass silently. Sellers that bypass the SDK are only
+detectable on multi-step storyboards — this validator is that gate.
+
+Runner plumbing:
+- `ExecutionState.lastA2aContextId` accumulates the contextId from each
+  A2A step's response Task. Before each A2A dispatch, the current value
+  is forwarded as `TaskOptions.contextId` (which rides on
+  `params.message.contextId` on the wire via `TaskExecutor`). After
+  dispatch, the response Task's contextId updates the state for the next
+  step.
+- `outboundA2aContextId` is captured before dispatch and passed to
+  `ValidationContext` so the validator can compare "what was sent" to
+  "what came back" — avoiding the tautology of asserting against a value
+  the runner never actually forwarded.
+- `executeStoryboardTask` accepts a `contextId` in its opts argument and
+  threads it through `TaskOptions`, which `SingleAgentClient` and
+  `TaskExecutor` already forward to `callA2ATool`.
+
+Validator behavior:
+- Self-skips with `not_applicable` when: no `outboundA2aContextId`
+  (first A2A step, non-A2A run, or prior step had no contextId),
+  no `a2aEnvelope` (capture miss), or a JSON-RPC error envelope (no
+  Task returned, so continuity is not verifiable).
+- Fails when response `Task.contextId` ≠ `outboundA2aContextId`, or
+  when response `Task.contextId` is absent/empty on a follow-up send.
+
+9 unit tests added. Refs adcp-client#962.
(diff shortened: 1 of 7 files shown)


bokelley:
## Expert review — approve with caveats

ad-tech-protocol-expert review: sound-with-caveats, approve.

> The validator correctly maps to A2A 0.3.0 contextId-grouping semantics (`Message.contextId` on `message/send` should be reflected on the response `Task.contextId`), the snapshot timing of `outboundA2aContextId` (captured before dispatch, post-dispatch update reserved for the next step) is the right shape, and the skip cases (no prior id, capture miss, JSON-RPC error) are conservative defaults that won't generate false positives. The SDK-vs-bypass framing is accurate — `src/lib/server/a2a-adapter.ts:394` confirms `requestContext.contextId` is forwarded by `DefaultRequestHandler`.

**Non-blocking caveats filed for follow-up:**
- Test coverage gap: tests construct `ValidationContext` directly and bypass the runner's snapshot/update plumbing in `runner.ts:1555-1601`. A two-step integration test (mock client returning contextId X on step 1, asserting step 2's outbound carries X) would close this.
- Spec citation: A2A 0.3.0 lets servers mint a new contextId if the supplied one is unknown; the validator hard-fails on any mismatch. Acceptable for storyboard runs, but worth softening "must echo per §7.1" wording in the error message to "did not preserve" to avoid citing a clause some readers may dispute.
- Adjacent `taskId` resumption (A2A 0.3.0 also carries `taskId` on follow-up sends) — separate gap, not in scope here.

bokelley:
Closing as superseded — the same content landed via **PR #979** (commit `af01482f`).

Verified on `origin/main`:
- `src/lib/testing/storyboard/types.ts:405` — `'a2a_context_continuity'` is in the validation-check union
- The validator function, runner snapshot/update plumbing, and tests all landed via #979

This PR was a duplicate that hadn't yet been rebased. The expert review caveats (test coverage gap for the runner snapshot timing, softer "must echo" wording) apply to the merged code on main and can be addressed in a follow-up if desired.

Closes adcp-client#962 (already closed via #979).

Issue #967, opened 2026-04-25
fix(core): AdCP tasks/get polling — fix param naming, response mapping, drop wrong-lifecycle MCP-experimental primary path


bokelley:
## Bug

\`TaskExecutor.getTaskStatus\` (\`src/lib/core/TaskExecutor.ts:1128-1150\`) has three issues that violate the AdCP 3.0 spec for \`tasks/get\`:

### 1. Wrong primary path for AdCP work polling

Current code tries MCP-native \`experimental.tasks.getTask\` first, falling back to \`callTool('tasks/get')\`:

\`\`\`ts
if (agent.protocol === 'mcp') {
  try { return await getMCPTaskStatus(...); }
  catch (err) { /* fall through */ }
}
const response = await ProtocolClient.callTool(agent, 'tasks/get', { taskId }, ...);
\`\`\`

Per the protocol-expert review (PR #952 follow-up): MCP \`experimental.tasks\` answers transport-call status (the MCP analog of A2A \`Task.state\`), **not AdCP work status**. For polling SUBMITTED arms (which is what \`pollTaskCompletion\` does), the AdCP \`tasks/get\` tool is the authoritative cross-protocol interface.

The two interfaces are not substitutes — they answer different questions:
- MCP \`experimental.tasks\` / A2A native \`tasks/get\` → "did the transport call complete?"
- AdCP \`tasks/get\` tool → "did the AdCP work complete?"

### 2. Wrong request parameter name

SDK passes \`{ taskId }\` (camelCase). AdCP \`tasks/get\` request schema (\`schemas/cache/3.0.0/bundled/core/tasks-get-request.json\`) requires \`{ task_id }\` (snake_case).

### 3. Wrong response shape mapping

\`\`\`ts
return (response.task as TaskInfo) || (response as unknown as TaskInfo);
\`\`\`

Expects the response to either be \`{ task: TaskInfo }\` (legacy nested camelCase) or have camelCase fields directly. AdCP spec response is flat snake_case: \`{ task_id, task_type, protocol, status, created_at, updated_at, ... }\` (\`schemas/cache/3.0.0/bundled/core/tasks-get-response.json\`). Real spec-conformant responses get \`taskId: undefined\` everywhere.

## Why this didn't surface

The mock tests (\`test/lib/task-executor-mocking-strategy.test.js:340-353\`) return the SDK's *expected* shape rather than the AdCP-spec shape. So the SDK never sees a real-shape response in tests.

[...]

PR #971, opened 2026-04-25
fix(core): SubmittedContinuation polls with server-assigned task_id (#966)

### LINKED ISSUE #966: fix(core): pollTaskCompletion uses local UUID instead of server-assigned task ID
## Bug

\`TaskExecutor.pollTaskCompletion\` (\`src/lib/core/TaskExecutor.ts:1152\`) polls the seller using \`getTaskStatus(agent, taskId)\`, where \`taskId\` is the **local UUID** the SDK generated at \`TaskExecutor.ts:279\` for its own bookkeeping. The server has never seen this ID, so any spec-conformant seller responds with NOT_FOUND.

The SDK already extracts the server-assigned ID via \`responseParser.getTaskId(response)\` and stores it on \`metadata.serverTaskId\` (\`TaskExecutor.ts:227-228\`), but never plumbs it to the polling cycle.

## Why this didn't surface

Mock tests at \`test/lib/task-executor-mocking-strategy.test.js:340-353\` ignore the \`taskId\` param when returning the mocked \`tasks/get\` response — the mock just looks at the tool name. So the SDK has been polling with the wrong ID against any real server, but the test suite is green because the mocks don't validate identifier semantics.

## Three IDs, three meanings, all called \`taskId\` today

The architecture review (PR #952 follow-up) flagged this as the bug-pattern source:

- \`TaskState.taskId\` — local UUID, internal bookkeeping key for \`activeTasks\` map. Also drives the \`{operation_id}\` webhook URL macro.
- \`metadata.serverTaskId\` — the server-assigned ID (A2A \`Task.id\` for A2A; AdCP \`task_id\` for AdCP-shim path).
- \`SubmittedContinuation.taskId\` — buyer-facing field. Documented as the seller's task handle but populated from the local UUID. **Source of the bug.**

## Proposed fix

1. Rename \`TaskState.taskId\` → \`operationId\` (matches the existing \`{operation_id}\` webhook macro convention). Stays as the \`activeTasks\` map key.
2. Plumb \`responseParser.getTaskId(response)\` into \`setupSubmittedTask\`'s \`SubmittedContinuation\`. The buyer-facing \`taskId\` field is now actually the server-assigned ID, matching the JSDoc.
3. Rename \`getTaskStatus(agent, taskId)\` parameter to \`serverTaskId\` so the contract is explicit at the call site.
4. Update the broken mock tests so they assert on the ID they receive — the existing mocks have been documenting a non-conformant interface.

## Scope

This is purely the identifier plumbing — pure correctness fix. Transport dispatch path stays unchanged (still MCP-experimental-first / AdCP-tool-fallback). The dispatch path itself has separate bugs (param naming, response shape, wrong primary path for AdCP work-status polling) that are tracked in 967 (filed alongside this).

## Out of scope

- Fixing the AdCP \`tasks/get\` request param name (\`taskId\` → \`task_id\`) and response-shape mapping — separate issue
- Changing which transport path is primary for polling — separate issue
- A2A-native \`tasks/get\` JSON-RPC method — explicitly NOT the answer per protocol-expert review (#899's two-lifecycle contract: A2A \`Task.state\` is 'completed' for AdCP submitted arms; querying it via native \`tasks/get\` returns useless data)

## Test plan

- New test: drive an actual submitted-arm response with a server-assigned ID through the full polling cycle; assert the SDK queries with the server ID, not the local UUID
- Update existing mock tests to validate the \`taskId\` param they receive

## Related

- adcp-client#904 — umbrella tracking issue
- adcp-client#952 — A2A wire-shape capture (sibling fix on the request/response side)
- adcp-client#963 — supersedes (was filed with wrong diagnosis; closed)

bokelley:
## Summary

Closes #966.

`setupSubmittedTask` was plumbing the SDK's runner-side correlation UUID (`TaskState.taskId`, generated at request time for `activeTasks` indexing and the `{operation_id}` webhook URL macro) all the way through to `SubmittedContinuation.taskId`. The polling closures — `track()` and `waitForCompletion()` — then addressed `tasks/get` calls with that local UUID, which the seller has never seen. Any spec-conformant seller would respond NOT_FOUND.

The existing mock tests at `test/lib/task-executor-mocking-strategy.test.js:340-353` masked the bug because they ignored the `taskId` parameter when stubbing the polling response — the SDK has been polling with the wrong ID against any real server, but the test suite was green.

## Fix

Extract the server-assigned handle via `responseParser.getTaskId(response)` (which already walks both the flat AdCP `response.task_id` shape AND the A2A `result.kind === 'task'` → `result.id` shape) and use it for the buyer-facing `SubmittedContinuation.taskId` field plus the polling closures. The local UUID stays internal — `activeTasks` map key and webhook URL macro value, both of which never reach the seller.

**Spec-violation fallback**: when the seller omits the task handle entirely on the submitted-arm response, the SDK falls back to the local UUID so callers still get a non-undefined `taskId` to log. Pollers won't be able to locate the work, but this matches the historical behavior surface and avoids a hard fail at a call site that's been silently wrong.

The buyer-facing `SubmittedContinuation.taskId` JSDoc is updated to document the corrected semantics: it carries the server handle, distinct from the runner-side correlation id.

## Why this is split from #967

The architect review on the polling redesign called out that the identifier-plumbing fix (correctness) and the AdCP `tasks/get` request/response shape fix (architecture hardening) have different review bars and rollback profiles:

- **This PR (#966)**: pure correctness — polls now address the right server task. Doesn't change which transport path is primary.
- **#967 follow-up**: AdCP `tasks/get` becomes the primary work-status path; param naming (`taskId` → `task_id`) and response-shape mapping fixed; MCP-experimental-tasks dropped from polling (wrong lifecycle for AdCP submitted arms per #899).

If #967 has to revert for an SDK quirk, this fix should not go with it.

## What we explicitly did NOT do

- A2A native `client.getTask({id})` — protocol-expert review confirmed this isn't the answer. Per #899's two-lifecycle contract (`a2a-adapter.ts:14-50`), A2A `Task.state` is `'completed'` for AdCP submitted arms, so polling A2A native returns useless data on every poll. The adapter docstring says it explicitly: *"Buyers resume the AdCP work via the `adcp_task_id` on the artifact's metadata, not by re-polling the A2A Task."*
- Renaming `TaskState.taskId` → `operationId`. Architect-suggested but invasive (50+ call sites) and pure refactor; deferred.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `prettier --check` clean (pre-existing nits format-fixed)
- [x] `eslint` 0 errors on changed files
- [x] All 5 new regression tests pass (`test/server-task-id-plumbing.test.js`)
- [x] 76/76 task-executor + error-scenarios tests pass
- [x] 2504/2508 wider regression sweep passes (4 intentional skips, none introduced here)
- [ ] CI green

## Review

- Three rounds of expert review on the broader polling redesign (architect, ad-tech-protocol, javascript-protocol) converged on this scope. Detailed synthesis lives in #966 and #967 issue descriptions.

### DIFF
--- a/.changeset/adcp-tasks-get-hardening.md
+++ b/.changeset/adcp-tasks-get-hardening.md
@@ -1,0 +1,71 @@
+---
+'@adcp/client': patch
+---
+
+Fix `TaskExecutor.getTaskStatus` to dispatch the AdCP `tasks/get` tool
+spec-conformantly. Closes #967.
+
+**Pre-fix bugs**:
+
+1. **Wrong request param**: SDK passed `{ taskId }` (camelCase). AdCP
+   3.0 schema (`schemas/cache/3.0.0/bundled/core/tasks-get-request.json`)
+   requires `{ task_id }` (snake_case). Conformant sellers reject as
+   INVALID_PARAMS.
+2. **Wrong response shape mapping**: SDK read `(response.task as TaskInfo)` —
+   expects a non-spec nested wrapper with camelCase fields. AdCP-spec
+   responses are flat snake_case (`{ task_id, task_type, status,
+created_at, updated_at, ... }`); real spec-conformant responses
+   produced `taskId: undefined` everywhere on the polled `TaskInfo`.
+3. **Wrong primary path**: SDK tried MCP `experimental.tasks.getTask`
+   first for MCP agents and fell through to the AdCP tool on
+   capability-missing. The MCP-experimental path tracks
+   transport-call lifecycle (the MCP analog of A2A `Task.state`),
+   not AdCP work lifecycle. For polling submitted-arm tasks (which
+   is what `pollTaskCompletion` does) we need work status; the two
+   interfaces are not substitutes (per protocol-expert review on
+   #966/#967).
+
+**Fix**:
+
+- Drop the MCP-experimental.tasks first attempt. Always dispatch the
+  AdCP `tasks/get` tool over the agent's transport.
+- Pass the request param as `task_id` (snake_case).
+- Map the response via a new `mapTasksGetResponseToTaskInfo` helper
+  that walks the transport-level wrappers (MCP `structuredContent`,
+  A2A `result.artifacts[0].parts[0].data`, legacy `{ task: ... }`
+  nested wrapper) and the AdCP-spec flat shape, then projects to the
+  internal `TaskInfo`.
+- Bypass `extractResponseData` for `tasks/get` — the generic
+  AdCP-error-arm detection misinterprets the spec's informational
+  `error: { code, message }` block as an error envelope and shreds
+  the response into `{ errors: [...] }`. The new helper handles
+  unwrapping directly.
+- Pass through `result` / `task_data` from `additionalProperties: true`
+  so completion data round-trips when sellers add it. (Note: AdCP
+  3.0 doesn't define a typed completion-payload field on `tasks/get`;
+  see adcp#3123 for the upstream clarification issue. Forward-compat
+  with all three possible spec resolutions.)
+
+**Behavior change**: MCP sellers that supported `experimental.tasks`
+but did NOT register an AdCP `tasks/get` tool will now see polling
+fail rather than silently use the wrong-lifecycle interface. This is
+deliberate — the previous behavior was incorrect (returned transport
+status, not work status). Sellers should register `tasks/get` as an
+AdCP tool to support buyer-side polling.
+
+Adds `test/server-tasks-get-spec-shape.test.js` with six regression
+tests:
+
+- Request param naming (snake_case `task_id`, no camelCase `taskId`)
+- AdCP-spec flat response mapping (incl. ISO 8601 timestamps)
+- Result-data passthrough via additionalProperties
+- Error-block mapping (failed status with `error: { code, message }`)
+- Legacy `{ task: ... }` nested-shape backward compat
+- No MCP-experimental.tasks first attempt
+
+Companion of #966 (server-task-id plumbing). With both PRs landed,
+MCP submitted-arm polling works end-to-end against spec-conformant
+sellers. A2A submitted-arm polling still has additional bugs at the
+parser layer (`getStatus` reads transport state, `getTaskId` extracts
+A2A Task.id instead of `artifact.metadata.adcp_task_id`); tracked in
+adcp-client#973.
--- a/.changeset/server-task-id-plumbing.md
+++ b/.changeset/server-task-id-plumbing.md
@@ -1,0 +1,44 @@
+---
+'@adcp/client': patch
+---
+
+Fix `SubmittedContinuation.taskId` and the polling cycle to use the
+server-assigned task handle instead of the SDK's runner-side
+correlation UUID. Closes #966.
+
+Pre-fix bug: `setupSubmittedTask` plumbed the local UUID generated at
+request time (`TaskState.taskId`, used for the `activeTasks` map and
+the `{operation_id}` webhook URL macro) through to the
+`SubmittedContinuation`. `track()` and `waitForCompletion()` then
+addressed `tasks/get` calls with that local UUID — which the seller
+has never seen, so any spec-conformant seller would respond with
+NOT_FOUND. Existing mock tests masked this because they ignored the
+`taskId` parameter when stubbing the polling response.
+
+Post-fix: `setupSubmittedTask` extracts the server-assigned handle via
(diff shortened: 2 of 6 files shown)


bokelley:
## Round-2 review applied

Three reviews completed (code-reviewer / ad-tech-protocol-expert / nodejs-testing-expert), with one significant finding that reframes the PR scope.

### Critical scope clarification (protocol expert)

I ran a probe that drove a real A2A submitted-arm flow end-to-end against `createA2AAdapter` and traced the response shape reaching `responseParser`. Findings:

1. The A2A protocol client returns the **JSON-RPC envelope** (`{ id, jsonrpc, result: <Task> }`), not the unwrapped Task as the SDK type definitions imply (`SendMessageResult = Message | Task`).

2. `responseParser.getStatus(response)` reads `response.result.status.state`, which is **`'completed'` for A2A submitted arms** per #899's two-lifecycle contract (`Task.state` tracks the HTTP call, not the AdCP work). The SDK never sees the response as `submitted` — it treats it as a completed sync call. `result.submitted` is undefined; the buyer thinks the operation finished.

3. `responseParser.getTaskId` for the wrapped envelope returns `response.result.id` (A2A Task.id) — not `response.result.artifacts[0].metadata.adcp_task_id` (the AdCP work handle the buyer needs).

So **PR #971 fixes MCP submitted-arm polling correctly**, but A2A submitted arms have additional bugs at the parser layer (`getStatus` misclassifies, `getTaskId` extracts wrong handle) that are out of scope here. Both fall under #967, which now expands to take them on.

This PR's stated scope (identifier plumbing) is the right correctness fix for the MCP path. A2A polling will continue to be broken at HEAD until #967 lands; that's documented in this PR's commit message and the #967 issue.

### Test improvements applied (code-reviewer + nodejs-testing-expert)

- A2A wrapped-result test now uses distinct sentinels for `result.id` vs. flat `task_id` (prior version masked the parser-priority branch).
- Added MCP `structuredContent.task_id` extraction path test.
- Added multi-poll (`working → working → completed`) test pinning every poll addresses the server handle.
- Added concurrent-submitted-tasks test pinning `activeTasks` keying doesn't cross-pollinate.
- Added webhook URL macro test pinning the two-IDs invariant (server handle on `submitted.taskId`, runner-side correlation on `{operation_id}`).
- Added `tasks/get` error-path test.

### Diagnosability (ad-tech-protocol-expert)

When a seller violates spec by omitting `task_id` on a submitted-arm response, `setupSubmittedTask` now emits a `debug_logs` warning naming the spec violation. The fallback to runner-side correlation id is preserved (surface stability) but is now visible in logs.

### Test counts

- 10 regression tests in `test/server-task-id-plumbing.test.js` (was 5)
- All pass; 127/127 in the wider task-executor + A2A test sweep
- 0 lint errors, prettier-clean

Latest commit: 92322361.

bokelley:
## Final review applied

Three reviewers (code-reviewer / ad-tech-protocol-expert / nodejs-testing-expert) ran on the rolled-up #966 + #967 state.

### Verdicts

| Reviewer | Verdict |
|---|---|
| code-reviewer | Two issues + 2 nits — **fixed** |
| ad-tech-protocol-expert | Spec-aligned, with one observation: `pollTaskCompletion` doesn't handle `rejected`/`auth-required`/`input-required`. Filed #977 as follow-up (pre-existing gap, not introduced here). |
| nodejs-testing-expert | One blocker + 3 nits — **fixed** |

### Fixes applied (commit 83bf48f1)

- `unwrapTasksGetEnvelope` recursion now bounded at depth 8 (defense against malformed/hostile sellers stack-overflowing the buyer)
- "No experimental.tasks" test replaced with a real spy on `mcp-tasks.getMCPTaskStatus` — prior version passed vacuously even if the deprecated path were reintroduced
- JSDoc on `getTaskStatus` flags the A2A submitted-arm parser gap (#973) so future readers know why the A2A path may still misbehave end-to-end
- 4 new regression tests:
  - `canceled` status (peer terminal state to `failed`)
  - MCP `structuredContent` envelope-unwrap
  - A2A `result.kind:'task'.artifacts[0].parts[0].data` envelope-unwrap
  - `task_data` alias for completion-payload passthrough

### Follow-ups filed

- **#977** — `pollTaskCompletion` doesn't handle `rejected` / `auth-required` / `input-required` (pre-existing terminal/paused state coverage gap; out of scope for #971)

### Test count

20 regression tests across the two new files — all pass; 61/61 in the broader task-executor + spec-shape sweep.

Latest commit: 83bf48f1.

PR #972, opened 2026-04-25
fix(core): poll with server-assigned task ID, not local UUID

bokelley:
Refs #966

## Summary

`TaskExecutor.setupSubmittedTask` was populating `SubmittedContinuation.taskId` — and the `track()`/`waitForCompletion()` polling closures — with the SDK-internal local UUID minted by `randomUUID()`. The server has never seen this ID. Every `submitted.track()` or `submitted.waitForCompletion()` call against a spec-conformant seller returned NOT_FOUND.

The fix is three lines: extract the server-assigned task ID from the initial response via `responseParser.getTaskId()` and use it everywhere the seller needs to identify the task. Falls back to the local UUID for backward compat with non-conformant sellers, with a `console.warn` so developers can identify the gap.

### What changed

- **`TaskExecutor.setupSubmittedTask`** — calls `responseParser.getTaskId(response)` to get the server-assigned task ID; passes it to `SubmittedContinuation.taskId` and the polling closures. Emits `console.warn` when the submitted response carries no task ID.
- **`SubmittedContinuation`** (additive, non-breaking) — new `operationId: string` field exposes the SDK-internal local UUID so buyers have both handles in one place. `taskId` JSDoc updated to make clear it is the server-assigned ID.
- **`pollTaskCompletion`** — optional `operationId?: string` parameter added; used as the `activeTasks` map key in `buildMetadata`/`buildErrorInstance` so idempotency-key lookup continues to work on polled completions.
- **`getTaskStatus` / `pollTaskCompletion`** — parameter renamed `taskId → serverTaskId` for call-site clarity (non-breaking: TS positional params).
- **New test file `test/lib/task-executor-poll-server-id.test.js`** — three tests, no CI skip: asserts `submitted.taskId` is the server-assigned ID, asserts `tasks/get` is called with the server ID, asserts the console.warn fires when the seller omits `task_id`.

### Out of scope (explicitly deferred)

`TaskState.taskId → operationId` rename: `TaskState` is publicly exported (`src/lib/index.ts:131`), so this is a breaking API change requiring a separate major-bump PR. The issue proposes it; this PR skips it and documents why.

## What was tested

- TypeScript: no new errors (`tsc --noEmit`, filtered for pre-existing `node` typedefs environment issue)
- Prettier: clean
- Regression tests in new `test/lib/task-executor-poll-server-id.test.js` verified to detect the bug against the pre-fix dist (all three fail on old dist, confirming the guards work)
- Pre-existing test failure count unchanged (44 failures before/after in `test/*.test.js`, all pre-existing environment issues)

## Pre-PR review

**Round 1 (plan review):**
- code-reviewer: approved non-breaking subset; flagged `TaskState` rename as breaking (deferred per spec)
- dx-expert: confirmed `operationId` companion field improves buyer discoverability without breaking `taskId`

**Round 2 (diff review):**
- code-reviewer: two blockers fixed — `buildMetadata` idempotency-key lookup (now passes `operationId ?? serverTaskId` as `metaTaskId`); regression test moved to non-skipped CI file
- ad-tech-protocol-expert: one blocker fixed — silent fallback now emits `console.warn`; confirmed `getTaskId()` fallback chain covers both A2A and AdCP-shim paths; `tasks/get` polling mechanism is correct for both transports; `operationId` field naming is consistent with existing `{operation_id}` webhook URL macro convention

Nits surfaced (not fixed in this PR, tracked separately):
- MCP Tasks `pollTaskCompletion` returns `data: undefined` on completion because `mapMCPTaskToTaskInfo` doesn't call `getTaskResult()` — pre-existing, separate issue
- `webhookManager.registerWebhook` is passed the local UUID; webhook reconciliation by task ID remains unaddressed — pre-existing

Session: https://claude.ai/code/session_01ThENaCC4Y1KHCio2BStUD6

---
_Generated by [Claude Code](https://claude.ai/code/session_01ThENaCC4Y1KHCio2BStUD6)_

### DIFF
--- a/.changeset/poll-server-task-id-fix.md
+++ b/.changeset/poll-server-task-id-fix.md
@@ -1,0 +1,7 @@
+---
+"@adcp/client": patch
+---
+
+Fixed `pollTaskCompletion` using the client-minted local UUID instead of the server-assigned task ID when polling `tasks/get` (#966).
+
+`setupSubmittedTask` now extracts the server-assigned task ID from the initial response via `responseParser.getTaskId()` and passes it to the polling closures. The buyer-facing `SubmittedContinuation.taskId` field now holds the server-assigned ID (matching its JSDoc), while the new `SubmittedContinuation.operationId` field carries the SDK-internal local UUID used for webhook URL macros and activity events. Before this fix, every `submitted.track()` and `submitted.waitForCompletion()` call against a spec-conformant seller returned NOT_FOUND.
--- a/src/lib/core/ConversationTypes.ts
+++ b/src/lib/core/ConversationTypes.ts
@@ -232,8 +232,19 @@ export interface DeferredContinuation<T> {
  * Continuation for submitted server tasks (server needs time)
  */
 export interface SubmittedContinuation<T> {
-  /** Task ID for tracking */
+  /**
+   * Server-assigned task ID — the handle the seller uses to identify this task.
+   * Pass this to the seller's `tasks/get` endpoint for manual status polling.
+   * Distinct from the SDK-internal operation ID (`operationId`) which is a
+   * local UUID used for webhook URL macros and the `activeTasks` bookkeeping map.
+   */
   taskId: string;
+  /**
+   * SDK-internal operation ID — the local UUID minted by the client for this
+   * call's lifecycle tracking. Used in `{operation_id}` webhook URL macros and
+   * `activity` event payloads. Not meaningful to the seller.
+   */
+  operationId: string;
   /** Webhook URL where server will notify completion */
   webhookUrl?: string;
   /** Get current task status */
--- a/src/lib/core/TaskExecutor.ts
+++ b/src/lib/core/TaskExecutor.ts
@@ -912,11 +912,17 @@ export class TaskExecutor {
       await this.config.webhookManager.registerWebhook(agent, taskId, webhookUrl);
     }
 
+    // Use the server-assigned task ID for polling. The local `taskId` is an
+    // SDK-internal UUID the server has never seen — passing it to tasks/get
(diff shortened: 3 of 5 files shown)


bokelley:
Closing as superseded by #968.

## Rationale

#968's fix (just merged / queued) addresses the underlying bug reported in #966:
- `setupSubmittedTask` now extracts the server-assigned A2A `Task.id` via `responseParser.getTaskId(response)` and threads it into the `track()` / `waitForCompletion()` polling closures and into the new native `getA2ATaskStatus(...)` helper.
- Polling against spec-conformant sellers now succeeds — the original NOT_FOUND symptom is resolved.

## What this PR adds beyond #968

The remaining delta in this PR is a **public-API semantic shift**:
- Change `SubmittedContinuation.taskId` from local UUID → server-assigned ID for all protocols
- Add `operationId: string` field exposing the local UUID

That shift is defensible but carries breakage risk for consumers correlating webhook `operation_id` with `submitted.taskId` (the existing convention per `TaskEventTypes.ts:198-205` is `/{agent_id}/{operation_id}`). It deserves its own discussion + changeset rather than riding along with the routing fix.

## Path forward

- The bug is fixed by #968.
- If we want the broader public-API surface (`taskId` = server ID, `operationId` = local UUID), open a fresh PR with:
  - Explicit `BREAKING:` or behavior-change call-out in the changeset
  - Audit of internal call sites that read `submitted.taskId` (e.g., `pr78-async-patterns-demo.ts`, `REAL-WORLD-EXAMPLES.md`)
  - `logDebug(debugLogs, ...)` for the "missing server task_id" warning instead of `console.warn` (per protocol-expert review)

#966 stays open until that follow-up lands or is explicitly deferred.

PR #974, opened 2026-04-25
ci(skills): post-sync lint catches missing skills + path drift

bokelley:
## Summary

After validating PR #965 against production, two regressions surfaced that aren't caught anywhere in CI today:

1. **Silent-no-op against the pinned release.** `npm run sync-schemas` against the production `3.0.0.tgz` extracts zero skills because the tarball was built before #965's upstream companion (adcontextprotocol/adcp#3097) landed. Tracked as [adcontextprotocol/adcp#3116](https://github.com/adcontextprotocol/adcp/issues/3116).
2. **Path drift in the canonical buyer skill.** `skills/call-adcp-agent/SKILL.md` (now sourced from the protocol tarball) references `dist/schemas/<adcp-version>/bundled/...` — the spec-repo layout — but SDK consumers extract to `schemas/cache/<version>/bundled/`. Coding agents loading the SDK skill get pointed at non-existent paths. Tracked as [adcontextprotocol/adcp#3117](https://github.com/adcontextprotocol/adcp/issues/3117).

This PR adds `scripts/check-skill-sync.ts` (npm script `check:skill-sync`) and wires it into CI right after the existing schema-sync validation step. The lint:

- Asserts the 7 protocol-managed skills (`call-adcp-agent` plus `adcp-{brand,creative,governance,media-buy,si,signals}`) exist with non-empty body content (catches frontmatter-only stubs).
- Scans `skills/call-adcp-agent/SKILL.md` for path-shaped backtick references and asserts each base directory resolves in the SDK layout.

Both known regressions are on a `KNOWN_UPSTREAM_BLOCKS` allow-list keyed to the upstream issue numbers and scoped to specific (file, path-base) tuples — so a future legitimate `dist/schemas/` tree wouldn't get its broken references silently swallowed. Entries are designed to be deleted as upstream fixes land; the lint then either goes silent (fix shipped) or fails red (fix incomplete), either of which is the correct signal.

## Why a lint, not matrix coverage

The existing `compliance:skill-matrix` harness is structured as "build-an-agent then grade it" — it spawns Claude in a scratch workspace, hands it a `build-*-agent` skill + a target storyboard, and grades the resulting server. Buyer-side skills (`call-adcp-agent`, `adcp-*`) are about *calling* an agent, not building one, so they don't fit that mold without a parallel harness. The lint runs in <1s, catches the specific regressions found today, and leaves the door open to a buyer-side harness when there's actual evidence we need one.

## Verification

```
$ npm run check:skill-sync
ℹ️  1 known upstream-blocked issue:
   [known-upstream] call-adcp-agent/SKILL.md references `dist/schemas/` (4 occurrences) but /…/dist/schemas/ does not exist
                    tracked: https://github.com/adcontextprotocol/adcp/issues/3117

✅ check-skill-sync: 7 protocol-managed skills present, all path bases resolve.
```

Failure modes were exercised locally:
- Removing a skill: lint exits 1, message names the missing file.
- Adding a frontmatter-only stub: lint exits 1, message names the empty body.
- Adding a new path drift to `call-adcp-agent`: lint exits 1, message names the unresolved path.

Reviewed by code-reviewer (3 should-fix items addressed: tightened allow-list match, frontmatter-aware empty check, dropped no-op regex flag), dx-expert (added "To fix" trailer + script-path hint, prefixed known-upstream lines, renamed CI step), and docs-expert (upstream issues edited for git-ref clarity, immutability question, table normalization).

## Test plan

- [x] `npm run check:skill-sync` exits 0 with the documented known-upstream output
- [x] Lint exits 1 on a removed skill (verified locally)
- [x] Lint exits 1 on a frontmatter-only stub (verified locally)
- [x] Lint exits 1 on an unresolved new path reference (verified locally)
- [ ] CI step runs green on this PR

## Notes

- No changeset: dev tooling only (script + CI step + npm-only script line).
- No library code touched.
- Allow-list entries should be removed in a follow-up PR when adcontextprotocol/adcp#3116 and #3117 ship.


bokelley:
## Approved

DX-expert review: approve.

> The error block prints kind-tagged violations and three explicit fix paths (`sync-schemas`, edit `SKILL.md`, or add an allow-list entry with tracking issue). The `unresolved-path` violation includes the absolute path that didn't resolve and the occurrence count, which is exactly what a contributor needs to grep for the offending span. Good error legibility.

> Once `sync-schemas` lands the upstream rewrite, the violation no longer matches → `known` array is empty → the "known upstream-blocked" log just stops printing. The allow-list isn't asserting "these should still fire"; a stale entry becomes a no-op rather than masking new drift.

Non-blocking nit: consider a one-line hint in the runtime "known upstream-blocked" log telling maintainers to delete the entry once the issue closes — but file is small and this is fine to land.

Issue #977, opened 2026-04-25
fix(core): pollTaskCompletion doesn't handle rejected | input-required | auth-required terminal/paused states


bokelley:
## Bug

\`TaskExecutor.pollTaskCompletion\` (\`src/lib/core/TaskExecutor.ts:1184-1217\`) recognizes only \`completed\` (success), \`failed\`, and \`canceled\` as loop-exit states. Per the AdCP \`task-status.json\` enum, three more statuses are terminal-or-paused but the polling loop spins until timeout on each:

- **\`rejected\`** — *"Task was rejected by the agent and was not started"*. Definitively terminal. Should exit the polling loop with \`success: false, status: 'failed'\` (analogous to \`canceled\`).
- **\`auth-required\`** — paused state requiring re-auth. Polling alone will never advance it. Should surface to the buyer as a typed error or via a re-auth callback.
- **\`input-required\`** — paused state requiring buyer input. Same — polling alone won't advance.

## Why this didn't surface

Pre-#971/#967, the SDK's polling cycle was misaligned (passed local UUID instead of server task_id, parsed responses with the wrong shape) so end-to-end submitted-arm polling never worked against any spec-conformant seller. The terminal-state coverage gap was hidden by the upstream bugs. Now that the polling cycle works for MCP, the gap surfaces.

Flagged by the ad-tech-protocol-expert review on PR #971.

## Proposed fix

In \`pollTaskCompletion\`:

\`\`\`ts
if (status.status === ADCP_STATUS.FAILED || status.status === ADCP_STATUS.CANCELED || status.status === ADCP_STATUS.REJECTED) {
  // existing failed-branch handling, plus 'rejected'
}
if (status.status === ADCP_STATUS.AUTH_REQUIRED || status.status === ADCP_STATUS.INPUT_REQUIRED) {
  // surface to the buyer — these are paused states polling alone won't advance
  // throw a typed error or return a continuation-style result
}
\`\`\`

The \`rejected\` case is the simplest: collapse it onto the failed-branch code path. The two paused states are bigger — they need either a new continuation type (similar to \`SubmittedContinuation\` for \`submitted\`) or a hard error. Probably worth splitting:

- \`pollTaskCompletion-rejected-as-terminal\`: ~5 line change, low risk.
- \`pollTaskCompletion-paused-state-handling\`: bigger, may need an API surface decision.

## Test plan

[...]

Issue #973, opened 2026-04-25
fix(core): A2A submitted-arm responses misclassified as completed by responseParser


bokelley:
## Bug

For A2A submitted-arm responses, \`ProtocolResponseParser.getStatus\` reads \`response.result.status.state\` (the A2A transport-call lifecycle, always \`'completed'\` per #899) instead of the AdCP work status from \`response.result.artifacts[0].parts[0].data.status\`. The SDK never recognizes A2A responses as \`submitted\` — it treats them as completed sync calls.

\`getTaskId\` has a parallel issue: it reads \`response.result.id\` (A2A Task.id, the transport handle) instead of \`response.result.artifacts[0].metadata.adcp_task_id\` (the AdCP work handle).

## How this surfaces

Probe at HEAD (after #971 / #966 land) — drive a real A2A submitted-arm flow against \`createA2AAdapter\`:

\`\`\`
[probe] getStatus → "completed"  // <-- WRONG, should be "submitted"
[probe] getTaskId → "a657ba42-9f6f-4d30-b908-2b0b0fb69b8f"  // A2A Task.id, not AdCP task_id
[probe] result.status = completed  // SDK never enters the submitted handler
[probe] result.submitted?.taskId = undefined  // no continuation issued
\`\`\`

The seller's \`tasks/get\` AdCP tool would never get called; \`pollTaskCompletion\` never fires; the buyer thinks the operation finished synchronously.

## Why this didn't surface earlier

- All A2A integration tests in \`test/storyboard-a2a-*.test.js\` and \`test/server-a2a-adapter.test.js\` either drive non-submitted arms or pre-#899 wire shapes.
- The polling cycle was already broken before #966 (passing the local UUID to \`tasks/get\`), so end-to-end submitted→completed flows didn't work over A2A regardless. Fixing #966 + this issue + #967 together is what makes the round-trip work.

## Proposed fix

\`responseParser.getStatus\`:
- When \`response.result.kind === 'task'\` AND \`response.result.artifacts[0].parts[0].data.status\` is one of the ADCP_STATUS values, return that status (the AdCP work lifecycle).
- Fall back to existing logic (\`response.result.status.state\`) for non-submitted / non-AdCP-shaped Task responses.

\`responseParser.getTaskId\`:
- When \`response.result.kind === 'task'\` AND \`response.result.artifacts[0].metadata.adcp_task_id\` is set, return that (the AdCP work handle).
[...]

PR #980, opened 2026-04-25
feat(core): pollTaskCompletion handles all non-progressing task statuses (#977)

bokelley:
## Summary

Refs #977

`TaskExecutor.pollTaskCompletion` only exited the polling loop for `completed`, `failed`, and `canceled`. When a server returned `rejected` (task refused before starting), the loop would spin until the caller's timeout — a worse failure mode than a clean error.

- **Add `ADCP_STATUS.REJECTED` to the terminal-state exit condition** — consistent with how `handleAsyncResponse` already handles `rejected` in the synchronous dispatch path (line 627).
- **Fix error-message fallback** — the polling path now checks `status.message` before the generic `"Task rejected"` string, matching the synchronous dispatch path. `TaskInfo` gains `message?: string` for this.
- **Fix `mcp-tasks.ts`** — `mapMCPTaskToTaskInfo` was only mapping `statusMessage → error` for `failed` status; extended to cover `rejected` and `canceled` (pre-existing gap surfaced by this fix).
- **Add `'rejected'` and `'canceled'` to `TaskStatus`** — the type union was missing these, limiting metadata fidelity.
- **New test file** `test/lib/poll-task-completion-terminal-states.test.js` — 4 tests covering error-field path, message-field fallback, single-poll exit, and generic fallback string. Tests use `protocol: 'a2a'` so `getTaskStatus` routes to the `ProtocolClient.callTool` mock directly (not the MCP Tasks protocol path which requires a live server).

## What's deferred (#977 part 2)

`input-required` and `auth-required` paused states are intentionally out of scope. Polling alone can't advance them, but the right surface (typed error vs. continuation) needs a design decision — the two expert reviewers disagreed (code-reviewer: return `TaskResultIntermediate`; dx-expert: throw typed error). Leaving for a follow-up so this fix ships cleanly.

## What was tested

- 4 new unit tests: all pass (`npm run test -- --test test/lib/poll-task-completion-terminal-states.test.js`)
- Pre-push validation: TypeScript typecheck + build pass
- Format: Prettier clean

## Pre-PR review

- **code-reviewer:** approved after two-pass fix — flagged MCP `statusMessage` mapping gap (fixed in commit 2) and test protocol path issue (fixed in commit 2); nits: `require.cache` bust is redundant per-test after first, missing regression tests for `failed`/`canceled` (pre-existing coverage exists in `task-executor-async-patterns.test.js`)
- **ad-tech-protocol-expert:** approved — `rejected` is definitively terminal per AdCP 3.0 GA `task-status.json`; noted `status: 'failed'` collapsing loses `rejected` signal (pre-existing pattern matching `handleAsyncResponse`; `TaskResultFailure.status` extension left as follow-up); confirmed `message?` is defensive tolerance, not a spec contract

## Known follow-ons (not blocking this PR)

- Extend `TaskResultFailure.status` to `'failed' | 'canceled' | 'rejected' | 'governance-denied'` for caller-visible signal fidelity
- `input-required` / `auth-required` handling in `pollTaskCompletion` (design decision needed)

Session: https://claude.ai/code/session_01AzgwdsUqsi7EbD8nWxLHXd

---
_Generated by [Claude Code](https://claude.ai/code/session_01AzgwdsUqsi7EbD8nWxLHXd)_

### DIFF
--- a/.changeset/poll-rejected-terminal.md
+++ b/.changeset/poll-rejected-terminal.md
@@ -1,0 +1,9 @@
+---
+'@adcp/client': patch
+---
+
+fix(core): pollTaskCompletion now exits immediately on `rejected` status instead of spinning until timeout
+
+`TaskExecutor.pollTaskCompletion` previously only exited the polling loop for `completed`, `failed`, and `canceled` statuses. When a server returned `rejected` (task refused before starting), the loop would spin until the caller's timeout. The `rejected` status is now treated as terminal — consistent with how `handleAsyncResponse` handles it — and returns `{ success: false, status: 'failed' }` immediately.
+
+Also fixes the error-message fallback: the polling path now checks `status.message` before falling back to the generic `"Task rejected"` string, matching the behavior of the synchronous dispatch path. `TaskInfo` gains an optional `message` field for this. `TaskStatus` now includes `'rejected'` and `'canceled'` for metadata fidelity.
--- a/.changeset/poll-rejected-terminal.md
+++ b/.changeset/poll-rejected-terminal.md
@@ -49,3 +49,15 @@ covers all three new exit paths plus regressions for `failed` /
 `canceled`. 9 tests; mocks dispatch via `protocol: 'a2a'` so polls
 route directly through `ProtocolClient.callTool` without the MCP
 Tasks protocol fast path.
+
+**adcp#3126 alignment** (typed `tasks/get` result field):
+adcontextprotocol/adcp#3126 closed the spec ambiguity flagged in
+adcp#3123 by adding a typed `result` field on `tasks/get` responses
+(gated by `include_result: true` on the request, populated when
+`status: 'completed'`). The SDK now sets `include_result: true` on
+every polling request so spec-conformant 3.1.0+ sellers populate
+the typed field; pre-3.1.0 sellers ignore the unknown request
+field, and the response mapper continues to read `result` (the
+typed and informal paths share the same field name). Dropped the
+informal `task_data` alias from the mapper — `result` is the
+canonical name.
--- a/.changeset/poll-rejected-terminal.md
+++ b/.changeset/poll-rejected-terminal.md
@@ -2,8 +2,50 @@
 '@adcp/client': patch
 ---
 
-fix(core): pollTaskCompletion now exits immediately on `rejected` status instead of spinning until timeout
+`TaskExecutor.pollTaskCompletion`: handle every non-progressing AdCP
+task status. Closes #977 (both halves).
 
-`TaskExecutor.pollTaskCompletion` previously only exited the polling loop for `completed`, `failed`, and `canceled` statuses. When a server returned `rejected` (task refused before starting), the loop would spin until the caller's timeout. The `rejected` status is now treated as terminal — consistent with how `handleAsyncResponse` handles it — and returns `{ success: false, status: 'failed' }` immediately.
+**Pre-fix**: `pollTaskCompletion` only exited on `completed`, `failed`,
+and `canceled`. Three non-progressing statuses caused the loop to spin
+until the caller's timeout:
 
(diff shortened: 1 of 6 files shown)


bokelley:
## Updated to close all of #977 + reviewer fixes

After protocol-expert and code-reviewer reviews on the original \`rejected\`-only scope, expanded the PR to handle the second half of #977 (paused states) cohesively. Both halves share the same regression class (loop spins until timeout) and the same surface (\`pollTaskCompletion\`'s status-handling block), so a single PR is cleaner than splitting.

### What's now covered

| Status | Pre-fix | Post-fix |
|---|---|---|
| \`completed\` | exits success | unchanged |
| \`failed\` | exits failure | unchanged |
| \`canceled\` | exits failure | unchanged |
| **\`rejected\`** | **spins until timeout** | exits failure |
| **\`input-required\`** | **spins until timeout** | returns paused-state \`TaskResultIntermediate\` |
| **\`auth-required\`** | **spins until timeout** | returns paused-state \`TaskResultIntermediate\` |

Paused states (\`input-required\` / \`auth-required\`) return \`TaskResultIntermediate\` with \`success: true\` — mirrors the synchronous \`handleInputRequired\` no-handler path. Caller branches on \`result.status\` and retries the original tool call after satisfying the paused condition.

### Reviewer fixes applied

**Code-reviewer** (1 blocker + 2 issues + 2 nits, all fixed):

- **[blocker]** \`mcp-tasks.mapMCPTaskToTaskInfo\`: \`statusMessage → error\` projection was checking against the pre-mapping MCP-side status. MCP emits \`'cancelled'\` (British) and never \`'rejected'\` — so MCP-cancelled tasks weren't surfacing the message. Fixed by projecting against the AdCP-mapped status.
- **[issue]** \`onTaskEvents\`: \`'canceled'\` was falling through to \`onTaskUpdated\` instead of \`onTaskFailed\`. Fixed.
- **[issue]** Test coverage gap on \`failed\` / \`canceled\` regressions: added.
- 2 nits resolved by JSDoc clarifications.

**Protocol-expert**: spec-aligned, ship. Two non-blocking caveats noted as follow-ups (rejected-vs-failed semantic distinction; \`unknown\` status spinning behavior). Filed for future work — out of scope for #977.

### CI fix

Pre-rebase the failing test \`preserves message field as error fallback\` was caused by #971's \`mapTasksGetResponseToTaskInfo\` not preserving the top-level \`message\` field. Fixed in commit 3269c0ff (separate from the part-2 work).

### Test count

9 unit tests on \`pollTaskCompletion\` (was 4); 154/154 in the wider task-executor + parser + a2a regression sweep. Two pushes after the original — \`3269c0ff\` (CI message fix) and \`00eed99e\` (part 2 + reviewer fixes).

bokelley:
## adcp#3126 alignment (just pushed)

\`adcontextprotocol/adcp#3126\` was merged — closed the spec ambiguity flagged in adcp#3123 by adding typed \`include_result\` (request) + \`result\` (response) fields to \`tasks/get\` for AdCP 3.1.0.

This PR now speaks the 3.1.0 wire shape:

- \`getTaskStatus\` sends \`include_result: true\` on every poll. Spec-conformant 3.1.0+ sellers populate the typed \`result\`; pre-3.1.0 sellers ignore the unknown request field.
- Dropped the informal \`task_data\` alias from the mapper. \`result\` is the canonical name in 3.1.0; the alias was a speculative passthrough that didn't match the spec.
- JSDoc updated to reference adcp#3126 (closed) instead of adcp#3123 (the original ambiguity ticket).
- Stale \`#973\` known-limitation comment on \`getTaskStatus\` removed (already fixed by PR #981).
- Replaced the \`task_data\` alias test with one that asserts the request includes \`include_result: true\`.

The 3.0.0 schema cache doesn't have the new field yet (3.1.0 hasn't shipped). When it does, follow-up #984 will bump the cache, regenerate types, and narrow \`flat.result: unknown\` to the spec's \`async-response-data\` shape.

Latest commit: a8a6ce9d.

Issue #994, opened 2026-04-25
TaskStore: support caller-supplied task_id and AdCP tasks/get response shape for sandbox controller scenarios


bokelley:
## Context

The compliance-test controller scenarios `force_create_media_buy_arm` and `force_task_completion` (adcp#3104, adcp#3138) let a sandbox runner register *buyer-supplied* task_ids — strings the runner picks for storyboard determinism. The seller is required to honor those IDs verbatim on `create_media_buy` submitted envelopes and on subsequent `tasks/get(task_id, include_result: true)` polls.

Two SDK gaps make this hard for sellers built on `@adcp/client`'s task store today:

## Gap 1: `TaskStore.createTask` doesn't accept caller-supplied `taskId`

```ts
createTask(taskParams: CreateTaskOptions, requestId: RequestId, request: Request, sessionId?: string): Promise<Task>;
```

The InMemoryTaskStore generates `taskId` via `randomBytes(16).toString('hex')` and PostgresTaskStore via the database. Neither exposes a way for a seller (e.g., the AdCP training-agent's compliance controller) to inject a buyer-supplied ID. So when `force_create_media_buy_arm` consumes its directive, the seller emits a `submitted` envelope with the buyer's `task_id`, but no corresponding entry exists in the task store. A subsequent `tasks/get(task_id)` returns `null`.

**Suggested API:**

```ts
createTask(taskParams: CreateTaskOptions & { taskId?: string }, requestId, request, sessionId?): Promise<Task>;
```

Behavior:
- If `taskId` is supplied, use it. If it collides with an existing task, throw `Task with ID ${taskId} already exists` (same as the random-collision branch today).
- If `taskId` is omitted, generate as today.
- Sandbox-only callers (compliance controllers) supply IDs; production handlers omit the option.

This is additive and non-breaking. Makes the AdCP `force_create_media_buy_arm` / `force_task_completion` storyboard pattern viable without sellers wrapping the store with their own virtual-tasks layer.

## Gap 2: AdCP `tasks/get` response shape ≠ MCP SDK's `Task` shape

The MCP SDK auto-registers `tasks/get` when a `taskStore` is provided. Its response shape is the MCP `Task`:

```ts
{ taskId, status, ttl, createdAt, lastUpdatedAt, statusMessage?, pollInterval? }
```

with `status` in `'working' | 'completed' | 'failed' | 'input_required' | 'cancelled'`.

[...]

PR #996, opened 2026-04-25
feat(server): PostgresTaskStore.createTask accepts optional caller-supplied taskId

bokelley:
Refs #994

## Summary

- Adds an optional `taskId?: string` field to `PostgresTaskStore.createTask`'s first parameter. When supplied, the ID is used verbatim; when omitted, a random 32-char hex ID is generated (existing behavior unchanged).
- Collision detection relies on PostgreSQL's `23505` uniqueness constraint caught on INSERT — no TOCTOU check-then-insert race.
- Error message includes recovery guidance consistent with the file's existing style.

**Does not ship Gap 2** (the `tasks/get` response shape mismatch). See the triage comment on #994 for expert synthesis and recommended design. That gap requires architectural discussion about where `task_type` is sourced and whether the `tasks/get` tool should be auto-registered.

## What was tested

- TypeScript: no new errors introduced (pre-existing `@types/node` absent errors are environment-only)
- Prettier: passes `format:check`
- Unit tests: three new cases added to `test/lib/postgres-task-store.test.js` (skipped without `DATABASE_URL`): caller-supplied ID round-trip, duplicate collision throws, auto-generated IDs are distinct

## Nits from pre-PR review (not blocking)

- Empty-string `taskId` accepted by PG as a valid TEXT key; callers should avoid it. A future hardening pass could add explicit validation, but it's not a SQL injection risk (parameterized query).

## Pre-PR review

- code-reviewer: approved — fixed PG error code type guard (`typeof`/`instanceof` double-cast → plain object check); bumped changeset from `patch` to `minor`; renamed one test name for accuracy
- ad-tech-protocol-expert: approved — non-breaking per spec; change is additive on the concrete class, not the interface

Session: https://claude.ai/code/session_01Xwm3Ed1SRCjKFvjQhSzz5M

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xwm3Ed1SRCjKFvjQhSzz5M)_

### DIFF
--- a/.changeset/postgres-task-store-caller-id.md
+++ b/.changeset/postgres-task-store-caller-id.md
@@ -8,6 +8,21 @@ Compliance storyboard controller scenarios (`force_create_media_buy_arm`,
 `force_task_completion`) need to inject buyer-supplied task IDs for storyboard
 determinism. `PostgresTaskStore.createTask` now accepts an optional `taskId`
 field on its first argument: when supplied, the ID is used verbatim; when
-omitted, a random hex ID is generated as before. Throws with a descriptive
-message if the supplied ID already exists (detected via PG uniqueness
-constraint, not a pre-check race).
+omitted, a random hex ID is generated as before. Throws if the supplied ID is
+empty, longer than 128 characters, or already exists (the collision is detected
+via PG uniqueness constraint, not a pre-check race).
+
+**Caveats and follow-ups:**
+- `InMemoryTaskStore` (re-exported from the upstream MCP SDK) does NOT honor
+  caller-supplied `taskId` — sellers running without `DATABASE_URL` (e.g., test
+  paths) get random IDs even when one is supplied. Filing an upstream MCP SDK
+  issue to add `taskId?: string` to `CreateTaskOptions` so both stores can honor
+  it cleanly is the right durable fix; this PR is the Postgres-only shim until
+  upstream lands.
+- The `task_id` namespace on `PostgresTaskStore` is process-global today (no
+  tenant scoping in the schema). Callers using caller-supplied IDs are
+  responsible for namespace isolation. A future migration to a composite
+  `(tenant_id, task_id)` key would close this for production use.
+- The storyboard runner does not yet send caller-supplied IDs through to the
+  controller tool's input schema. That wiring (runner → tool input → task
+  store) is a separate change tracked in the parent issue.
--- a/.changeset/postgres-task-store-caller-id.md
+++ b/.changeset/postgres-task-store-caller-id.md
@@ -1,0 +1,13 @@
+---
+"@adcp/client": minor
+---
+
+feat(server): PostgresTaskStore.createTask accepts optional caller-supplied taskId
+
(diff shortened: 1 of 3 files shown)


bokelley:
## Expert review summary

Three independent reviewers (code-reviewer, security-reviewer, javascript-protocol-expert). No Must Fix. Should-Fix items addressed in `8b194ce`:

**Security (S1+S2):** Added pre-INSERT guard rejecting empty or >128-char `taskId` with descriptive errors. Closes the DoS-bloat path and the "first sandbox caller squats the empty-string slot" footgun. Three new tests: empty, >128, 128-boundary.

**Documentation (code-reviewer N1):** `createTask` JSDoc now flags the global `task_id` namespace and recommends a future composite `(tenant_id, task_id)` migration if production paths ever wire caller-supplied IDs.

**Changeset transparency (code-reviewer + js-protocol-expert):** Caveats section added covering:
1. **InMemoryTaskStore parity gap** — the SDK's re-exported `InMemoryTaskStore` ignores `taskId` because the upstream MCP SDK doesn't yet expose the field on `CreateTaskOptions`. Sellers without `DATABASE_URL` (test/CI paths) silently fall back to random IDs. The right durable fix is an upstream MCP SDK PR adding `taskId?: string` to `CreateTaskOptions` — this PR is the Postgres-only shim until that lands.
2. **Storyboard-runner wiring** — the runner doesn't yet send caller-supplied IDs through to the controller tool's input schema; that's a separate change in the parent issue's scope.

**Other reviewer findings (non-blocking, documented as nits):**
- PG error-code type guard is fine for `pg` driver; if the driver is ever wrapped by Sentry/OpenTelemetry/connection pool middleware, `code` may move to `err.cause.code`. Worth a comment in a future hardening pass.
- AdCP/MCP wire spec is unaffected — `task_id` is opaque on the wire; this is purely a server-internal sourcing decision.

Marking ready for review.

bokelley:
## Expert review — approve after fixes applied

code-reviewer review: no blockers. Two Should-Fix items both applied in `e5e72c7d`:

1. **Dead-code slice removed.** Error message at line 204 used `taskId.slice(0, 256)` but the validator caps inputs at 128 chars. Now uses `taskId` directly.
2. **128-char rationale documented.** JSDoc clarifies the limit is an SDK policy (matches typical request-id / session-id field lengths and keeps the task_id index efficient), not a Postgres or spec constraint.

**Reviewer's other notes (non-blocking):**
> SQL parameterization — clean. All values pass through `$1`–`$5` placeholders; `this.table` is constructor-injected, not user input.
> 23505 scope — currently safe (single unique key on `task_id`). Future-defense: assert `err.constraint === '<task_id_pkey_or_unique_name>'` so a future migration adding a second unique index doesn't silently get swallowed.
> Tests: real PG (skipped without `DATABASE_URL`). Correct.
> JSDoc: accurate, covers verbatim use, generation fallback, validation, and the namespace caveat.

**Note:** ships Gap 1 from #994 only (caller-supplied taskId). Gap 2 (`tasks/get` response shape mismatch) deferred — needs separate design.

PR #1008, opened 2026-04-25
feat(AgentClient): fromMCPClient() factory for in-process MCP transport

### LINKED ISSUE #1003: AgentClient: support transport injection (in-process MCP Client) for in-process compliance test fleets
## Context

Sellers running the storyboard runner (`runAgainstLocalAgent`) or the buyer-side typed surface (`AgentClient.createMediaBuy(...)`) in tests today have to spin up an HTTP loopback server. `AgentConfig.agent_uri: string` is the entry point; the protocol layer at [`protocols/index.ts`](https://github.com/adcontextprotocol/adcp-client/blob/main/src/lib/protocols/index.ts) opens an HTTP transport against the URL.

## Problem

For **in-process** compliance verification (the main reason a seller would use the buyer SDK in tests at all — to pin storyboard pass/fail to a per-PR CI job, not a deployed sandbox), the loopback dance is awkward:

- Each test allocates a free port, calls `serve()`, awaits listening, points `AgentClient` at `http://127.0.0.1:<port>/mcp`, runs assertions, closes the server.
- Test setup/teardown is boilerplate the SDK could absorb.
- Vitest workers running concurrent test files race on `serve()` lifecycle if any helper accidentally shares state.
- The HTTP round-trip adds non-trivial latency for what's logically an in-process call.

The MCP TypeScript SDK already ships `InMemoryTransport.createLinkedPair()` for exactly this case — a server transport and a client transport that pipe to each other, no socket, no port. We use it in our own e2e tests today against the raw MCP `Client`. The buyer-side `AgentClient` could too, but its constructor is URL-only.

## Proposed surface

Add a factory that accepts an already-connected MCP `Client` (or transport pair) instead of an `AgentConfig.agent_uri`:

```ts
import { Client } from '@modelcontextprotocol/sdk/client/index.js';
import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js';
import { AgentClient } from '@adcp/client';

const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair();
await Promise.all([
  client.connect(clientTransport),
  adcpServer.connect(serverTransport),
]);

// NEW:
const agent = AgentClient.fromMcpClient(client, {
  // optional config — auto-generation, validation modes, etc.
  validation: { requests: 'strict' },
  agentName: 'in-process-test-snap',
});

const result = await agent.createMediaBuy({ ... });
// Type-safe response, automatic idempotency_key generation,
// no HTTP, no port allocation.
```

Symmetrically for A2A: `AgentClient.fromA2AAdapter(...)` against `createA2AAdapter`'s output.

## Why this matters

- **Controller-pattern complement.** Issue #1002 (controller for proxy sellers) and this one together let sellers run storyboards against their own framework in-process: stateStore-backed seeding writes via the controller, transport-injected `AgentClient` walks the storyboard, all without HTTP.
- **Drops the `runAgainstLocalAgent` boilerplate.** The current helper exists partly *because* loopback HTTP is annoying; with transport injection, runners can compose `serve()` only when they explicitly want the network exercised (sandbox runs against deployed URLs).
- **Test execution speed.** In-process MCP transport eliminates ~50-200ms per call vs. loopback HTTP + transport handshake. Adds up across a full storyboard run (8-15 steps × N platforms).

## Acceptance criteria

- [ ] `AgentClient.fromMcpClient(client, config?)` (or equivalent factory) accepts an MCP `Client` instance
- [ ] All `AgentClient` typed methods (`createMediaBuy`, `syncCreatives`, etc.) work over the in-process transport with the same idempotency_key auto-injection, retry semantics, and validation hooks as the URL-driven path
- [ ] Test in `test/lib/agent-client.in-process.test.js` covering one happy-path tool call + one error envelope path
- [ ] Docs section in the buyer SDK reference noting when to use which factory (sandbox URL vs in-process transport)

## Optional but useful

A2A-side parity: `AgentClient.fromA2AAdapter(adapter, config?)` against `createA2AAdapter`'s output for sellers running A2A storyboards in-process.

bokelley:
## Summary

Resolves #1003.

Adds `AgentClient.fromMCPClient(client, config?)` — a static factory that accepts a pre-connected `@modelcontextprotocol/sdk` `Client` instead of a URL-based `AgentConfig`. Compliance test fleets can now wire a full `AgentClient` to an `InMemoryTransport.createLinkedPair()` without an HTTP loopback server.

### What changes

- **`AgentClient.fromMCPClient()`** static factory + exported `InProcessAgentClientConfig` type
- **`callMCPToolWithClient()`** in `mcp-tasks.ts` — extracted `callToolOnClient` helper shared between URL and in-process paths (zero behavior change on the URL path)
- **`ProtocolClient.callTool()`** — in-process routing branch before `validateAgentUrl`; injects `adcp_major_version` on v3, skips SSRF guard
- **`SingleAgentClient.normalizeAgentConfig()`** — skips `_needsDiscovery` flag for in-process agents
- **`SingleAgentClient.getAgentInfo()`** — calls `listTools()` directly on the pre-connected client instead of opening a new HTTP connection
- **HTTP-only method guards**: `resolveCanonicalUrl`, `getWebhookUrl`, `registerWebhook`, `unregisterWebhook` all throw descriptive "in-process" errors
- **`AgentConfig._inProcessMcpClient`** `@internal` field (sentinel URI `adcp-in-process://<id>`)

### Behaviors preserved over the in-process path

| Behavior | Verified |
|---|---|
| `adcp_major_version` injected on every tool call | test 2 |
| `idempotency_key` auto-generated for mutating tasks | test 3 |
| `isError` envelopes surface as `TaskResult<{ success: false }>` | test 4 |
| HTTP-only methods throw `in-process` guard errors | tests 7–10 |
| `getAgentName()` / `getAgentId()` / `getProtocol()` reflect factory config | test 5 |

### Test results

```
# tests 9
# suites 2
# pass 9
# fail 0
```

All 9 new integration tests pass. The 318 pre-existing failures are all environment-level (missing `compliance/cache/3.0.0` and `schemas/cache/3.0.0` — requires `npm run sync-schemas`; confirmed identical on `main` before this branch).

---

https://claude.ai/code/session_01S68HAeyFEnxoXoWsp5wcu4

---
_Generated by [Claude Code](https://claude.ai/code/session_01S68HAeyFEnxoXoWsp5wcu4)_

### DIFF
--- a/.changeset/agent-client-in-process-transport.md
+++ b/.changeset/agent-client-in-process-transport.md
@@ -6,6 +6,8 @@ feat(AgentClient): add `AgentClient.fromMCPClient()` factory for in-process MCP 
 
 Adds a new static factory method that accepts a pre-connected `@modelcontextprotocol/sdk` `Client` instance instead of a URL-based agent config. This enables compliance test fleets to wire up a full `AgentClient` against an `InMemoryTransport` pair without an HTTP loopback server.
 
+**MCP only.** This factory wraps an MCP `Client` from `@modelcontextprotocol/sdk`. There is no equivalent in-process bridge for A2A today — for A2A agents, run them on a loopback HTTP server and use the standard `AgentClient` constructor with the agent's `agent_uri`.
+
 Key behaviors preserved over the in-process path:
 - `adcp_major_version` is injected on every tool call
 - `idempotency_key` is auto-generated for mutating tasks
--- a/.changeset/agent-client-in-process-transport.md
+++ b/.changeset/agent-client-in-process-transport.md
@@ -1,0 +1,16 @@
+---
+'@adcp/client': minor
+---
+
+feat(AgentClient): add `AgentClient.fromMCPClient()` factory for in-process MCP transport
+
+Adds a new static factory method that accepts a pre-connected `@modelcontextprotocol/sdk` `Client` instance instead of a URL-based agent config. This enables compliance test fleets to wire up a full `AgentClient` against an `InMemoryTransport` pair without an HTTP loopback server.
+
+Key behaviors preserved over the in-process path:
+- `adcp_major_version` is injected on every tool call
+- `idempotency_key` is auto-generated for mutating tasks
+- `isError` envelopes surface as `TaskResult<{ success: false }>`
+- HTTP-only methods (`resolveCanonicalUrl`, `getWebhookUrl`, `registerWebhook`, `unregisterWebhook`) throw descriptive `in-process` guard errors
+- Endpoint discovery and SSRF validation are bypassed for the sentinel URI
+
+Exports the new `InProcessAgentClientConfig` type for typed factory usage.
--- a/src/lib/core/AgentClient.ts
+++ b/src/lib/core/AgentClient.ts
@@ -1,5 +1,6 @@
 // Per-agent client wrapper with conversation context preservation
 
+import type { Client as MCPClient } from '@modelcontextprotocol/sdk/client/index.js';
 import type { AgentConfig } from '../types';
 import type { MCPWebhookPayload } from '../types/core.generated';
 import type { Task as A2ATask, TaskStatusUpdateEvent } from '@a2a-js/sdk';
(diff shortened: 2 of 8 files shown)


bokelley:
Follow-up commit `c086339` addresses the pre-PR review findings:

- **Blocker fixed**: `isSameAgentResolved` now has an in-process guard — previously it called `resolveCanonicalUrl()` unconditionally, which throws for in-process instances. It now falls back to sentinel-id comparison when either side is an in-process agent.
- `SingleAgentClient.getCapabilities`: added comment explaining why `ensureEndpointDiscovered()` is safe to call for in-process agents (it's a no-op because `_needsDiscovery` is always false for them)
- Fake clients in tests now stub `listTools` so they're safe if `getAgentInfo()` is called
- Added: sentinel-id generation test (`agentId` omitted → `in-process-<random>` format)
- Added: `isSameAgentResolved` in-process comparison test (same id → true, different id → false)

11/11 tests pass.

https://claude.ai/code/session_01S68HAeyFEnxoXoWsp5wcu4

---

PR #1018, opened 2026-04-26
feat(signing): PostgresReplayStore for distributed verifier deployments

### LINKED ISSUE #1015: Distributed ReplayStore adapter (Redis) for multi-instance verifier deployments
## Problem

The verifier's nonce-replay cache is in-process. The `ReplayStore` interface (`src/lib/signing/replay.ts:15`) is async and pluggable, but only `InMemoryReplayStore` ships today. A multi-instance verifier deployment (autoscaled fleet behind a load balancer) loses replay protection across processes — a captured signature can be replayed against a sibling instance whose cache hasn't seen the nonce.

This is a real-deployment gap, not a spec gap. The interface is the right shape; what's missing is a distributed adapter.

## Proposal

Ship a Redis-backed `ReplayStore` adapter alongside the in-memory default. Redis is the obvious starter — `SET key value NX EX <ttl>` is a one-RTT atomic insert-if-absent that maps cleanly onto `ReplayStore.insert`'s `'ok' | 'replayed'` return.

```ts
// examples/redis-replay-store.ts (or src/lib/signing/replay-redis.ts)
export interface RedisReplayStoreOptions {
  client: RedisClientLike;          // ioredis / node-redis / equivalent
  keyPrefix?: string;               // default: 'adcp:nonce:'
  capPerKeyid?: number;             // mirror InMemoryReplayStore default (100k)
}

export class RedisReplayStore implements ReplayStore { ... }
```

Wire-level mapping (notes for the implementation, not a spec):

- `has(keyid, scope, nonce, now)` → `EXISTS adcp:nonce:{keyid}:{scope}:{nonce}`
- `insert(keyid, scope, nonce, ttl, now)` → `SET adcp:nonce:{keyid}:{scope}:{nonce} 1 NX EX <ttl>` — `nil` reply means already present (replay).
- `isCapHit(keyid, scope, now)` → counter on `adcp:cap:{keyid}:{scope}` with TTL matching the longest in-flight nonce window; `INCR` on insert, compare to `capPerKeyid`. `rate_abuse` returns when the counter exceeds the cap.

Composite-key partitioning by `(keyid, scope)` per the existing interface contract (`replay.ts:13`) — must not collapse to keyid alone or a signature captured on one path can be replayed against another.

## Notes

- Async interface stays unchanged. This is purely an additional implementation.
- Decoupled from #1009 (`SigningProvider` / KMS). #1009 is about hiding private keys at sign time; this is about hardening verifier replay protection across a fleet. Different surface, different threat model — but called out together because operators who deploy KMS-backed signing agents typically also run multi-instance verifiers.
- Should ship as an example in `examples/` (no Redis dep added to the published package), matching the convention proposed for the GCP KMS adapter. Authoritative published adapter could come later if community demand justifies it.
- Worth a doc note in the verifier guide explaining the multi-instance failure mode — most operators won't realize their replay protection silently fragments under autoscaling.

## Out of scope

- Other backends (DynamoDB, Cloud Memorystore, KeyDB, etc.). Pattern is identical; document the Redis example, let community mirror.
- Coordinated `capPerKeyid` enforcement across processes when the cap is at exactly the boundary. Best-effort for now — cap-hit is a DoS-resistance signal, not a security boundary.

Filed alongside #1009 — no dependency between them, but they showed up in the same design discussion.

bokelley:
## Summary

Adds a Postgres-backed `ReplayStore` so multi-instance verifier deployments share replay-protection state. Closes #1015.

The default `InMemoryReplayStore` is per-process. On a fleet, an attacker who captures a signed request can replay it against a sibling instance whose cache hasn't seen the nonce — RFC 9421's 5-minute expiry bounds the window but that's plenty of time for an in-flight replay. `PostgresReplayStore` closes the hole using a `(keyid, scope, nonce)` primary key the verifier checks on every signed request.

**This PR's changeset accumulates into Release PR #1014, which currently bumps to 5.20.0 from #1017.** Merging this lands the work on `main`; nothing publishes to npm until you merge the Release PR.

### 1. The single-statement insert handles replay → cap → insert atomically

`src/lib/signing/postgres-replay-store.ts:148` — one CTE that:
- Returns `'replayed'` if an unexpired row already exists for `(keyid, scope, nonce)`
- Returns `'rate_abuse'` if the active count for `(keyid, scope)` is at the cap
- Otherwise inserts (or refreshes an expired row in place) and returns `'ok'`
- Falls through to `'replayed'` on the concurrent-same-nonce loser branch

Result precedence matches `InMemoryReplayStore`: replay > rate_abuse > ok.

### 2. Expired-row recycling — the bug I caught with the integration test

The first version of the SQL used `ON CONFLICT DO NOTHING` and inferred `'replayed'` from an empty `RETURNING`. That broke the "previously-expired same-nonce" case: the row was still on disk (sweeper hadn't run), `ON CONFLICT` fired, `RETURNING` was empty, adapter returned `'replayed'` — but the InMemory store returns `'ok'` for that case (it prunes-then-checks). Switched to `ON CONFLICT DO UPDATE WHERE existing-is-expired`, which refreshes the row in place when its previous TTL has elapsed. Now both stores agree.

### 3. Sweeper is an exported helper, not a background timer

`sweepExpiredReplays(pool, options?)` is exported and callers schedule it themselves (cron, app timer, `pg_cron`, etc.). Deliberately avoiding a constructor-injected timer — those surprise people in tests and serverless. Doc shows three ways to schedule.

### 4. Reuses `PgQueryable` structural type

Imports from `src/lib/server/postgres-task-store.ts` so the SDK stays free of a hard `pg` dependency. Matches the `PostgresTaskStore` / `PostgresStateStore` pattern already in the codebase.

## Test plan

- [x] 13 new integration tests in `test/lib/postgres-replay-store.test.js` (run with `DATABASE_URL=postgres://...`):
  - Default migration + custom table names + SQL-injection-shaped name rejection
  - First insert `'ok'`, second insert same nonce `'replayed'`
  - `(keyid, scope)` partitioning — same nonce on different scopes is not a replay
  - TTL boundary respected by `has()` and by `insert()` (expired same-nonce recycles correctly)
  - Cap precedence: replay > rate_abuse > ok
  - `isCapHit` reflects active count vs cap; expires-out properly
  - Sweeper: full sweep + batched sweep
  - Concurrency: 10 parallel same-nonce inserts produce exactly 1 `'ok'` and 9 `'replayed'`
  - E2E: signed request → verifier wired with `PostgresReplayStore` → first request verifies, second is rejected as `request_signature_replayed`
- [x] All existing 175 signing tests still pass — no regressions

## Deferred follow-ups

- Redis adapter (`@upstash/redis` or `ioredis`-shaped) as a sibling example — pattern is identical, no SDK interface change.
- `pg_cron` schema helper for users who'd rather have Postgres run the sweep — small follow-up doc, no code.

### DIFF
--- a/.changeset/postgres-replay-store.md
+++ b/.changeset/postgres-replay-store.md
@@ -1,0 +1,21 @@
+---
+"@adcp/client": minor
+---
+
+feat(signing): PostgresReplayStore for distributed verifier deployments
+
+Adds a Postgres-backed `ReplayStore` so multi-instance verifier deployments share replay-protection state. The default `InMemoryReplayStore` is per-process; on a fleet, an attacker who captures a signed request can replay it against a sibling whose cache hasn't seen the nonce — RFC 9421's 5-minute expiry bounds the window but that's plenty of time for an in-flight replay. `PostgresReplayStore` closes that hole using a `(keyid, scope, nonce)` primary key the verifier checks on every signed request.
+
+New exports from `@adcp/client/signing/server`:
+
+- `PostgresReplayStore` — `ReplayStore` implementation against the structural `PgQueryable` interface (same pattern as `PostgresTaskStore` and `PostgresStateStore`; the SDK stays free of a hard `pg` dependency).
+- `getReplayStoreMigration(tableName?)` — idempotent DDL for the cache table plus indexes on `expires_at` and `(keyid, scope, expires_at)`.
+- `sweepExpiredReplays(pool, options?)` — exported helper for callers to schedule (cron, app timer, `pg_cron`, etc.); Postgres has no native row-level TTL, so expired rows have to be deleted explicitly.
+
+The insert path is a single CTE statement that handles replay/cap/insert decision atomically. `ON CONFLICT DO UPDATE WHERE existing-is-expired` recycles expired rows in place — a same-nonce insert after the previous registration's TTL elapsed (but before the sweeper ran) correctly returns `'ok'` rather than falsely reporting `'replayed'`. Concurrent same-nonce inserts (10 parallel) consistently produce exactly one `'ok'` and the rest `'replayed'`, matching `InMemoryReplayStore` semantics.
+
+Wire format unchanged. No AdCP version bump.
+
+See [`docs/guides/SIGNING-GUIDE.md` § Verify Inbound Signatures](./guides/SIGNING-GUIDE.md#step-4-verify-inbound-signatures-seller) for the multi-instance failure mode and the wire-up.
+
+Closes #1015.
--- a/docs/guides/SIGNING-GUIDE.md
+++ b/docs/guides/SIGNING-GUIDE.md
@@ -339,7 +339,35 @@ app.post(
 );
 ```
 
-`replayStore` and `revocationStore` default to in-memory implementations — fine for single-process deployments. Replace with a shared store (Redis-backed, etc.) for horizontally scaled fleets so replay detection works across instances.
+`replayStore` and `revocationStore` default to in-memory implementations — fine for single-process deployments.
+
+**For multi-instance verifier deployments, the in-memory default is a real gap.** Each process has its own cache; an attacker who captures a signed request can replay it against a sibling instance whose cache hasn't seen the nonce. The replay-protection invariant is "this `(keyid, scope, nonce)` tuple has not been seen before" — that has to hold across the fleet, not per-process. RFC 9421 expiry bounds the window to 5 minutes, but that's plenty of time for an in-flight replay. Use a shared backend.
+
+The SDK ships `PostgresReplayStore` for this:
+
+```typescript
+import { Pool } from 'pg';
+import { PostgresReplayStore, getReplayStoreMigration, sweepExpiredReplays } from '@adcp/client/signing/server';
+
+const pool = new Pool({ connectionString: process.env.DATABASE_URL });
+await pool.query(getReplayStoreMigration());                    // run once at boot
+
+const replayStore = new PostgresReplayStore(pool);
+
+// Postgres has no native TTL — schedule the sweeper somewhere to delete
+// expired rows. Once a minute is fine for moderate traffic; tune down if
+// the verifier sees thousands of signed requests per second.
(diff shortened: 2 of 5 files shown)


bokelley:
**Local two-instance validation** (`.context/work/two-instance-replay-check.mjs`):

```
# Step 1: Send signed request to verifier instance A (fresh)
  → status=verified keyid=test-ed25519-2026

# Step 2: Replay same headers to verifier instance B (different process)
  → rejected with code=request_signature_replayed step=12
```

Distinct `PostgresReplayStore` instances backed by the same Postgres correctly enforce replay protection across what would be separate verifier processes. Verifier rejects at step 12 (the canonical replay-detection step).

**Three-reviewer pass** (security, code, protocol — same parallel pattern as #1017). No blockers, one consensus blocker from protocol (untested invariant), several worth-fixing items. All addressed in `f759c532`:

| Item | Source | Done |
|---|---|---|
| Cap clears post-expiry without sweeper — untested invariant | protocol blocker | Test added (line 131) |
| `now`/`ttlSeconds` validation (NaN/Infinity → PG `to_timestamp` DoS surface) | security M2 | `assertFiniteSeconds` at all three method boundaries |
| `REPLAY_CACHE_MIGRATION` constant for repo-style parity | code-review nit | Exported alongside `getReplayStoreMigration` |
| JSDoc: cap is best-effort under concurrency | security M1 + code-review | Added |
| Test: concurrent recycle of expired nonce (10 parallel) | security M1 | Test added (line 151) |
| Test: batched sweep with `batchSize` > active expired count | code-review nit | Test added (line 167) |
| Test: rate-abuse rejection through verifier pipeline (vector 020) | protocol note | Test added (line 218) |
| `setCapHitForTesting` hook for white-box conformance grading | protocol note | Documented limitation (per reviewer's option a); white-box grading can use `cap: small` + real inserts |
| `count(*)::bigint` → `>= bigint AS hit` refactor | code-review nit | Skipped — works as-is |

Test totals: **18/18** PostgresReplayStore integration tests (was 13) + **175** existing signing tests. All green.

PR #1019, opened 2026-04-26
feat(testing): adcp grade signer — validate a signer before going live

### LINKED ISSUE #610: signing: signer-side conformance grader (complement to the verifier grader in #585)
Follow-up to #585 and #600.

The verifier-side conformance grader shipped in #600 — it takes an agent advertising `request_signing.supported: true` and drives the 28 RFC 9421 vectors against its verifier pipeline. The PR description explicitly defers the signer-side grader to a separate storyboard.

## What's missing

A grader that drives **the client-side signer**:
- Construct a known-good reference verifier (the one in `test/utils/reference-verifier.js` is a natural starting point — or expose a hosted reference verifier for CI).
- Point a candidate signer (an AdCP client library advertising `request_signing` on its `AgentConfig`) at the reference verifier.
- Observe which operations the signer produces signed requests for, whether Signature-Input covers all mandatory + policy-required components, whether the digest policy is honored, whether the window is within spec, whether the tag matches.
- Report pass/fail per vector, plus capability fingerprint (which ops the signer chose to sign, which it opted to leave unsigned).

## Shape

Parallel structure to #600:
- Vector fixtures: the 28 RFC 9421 vectors, interpreted from the signer perspective (positive vectors = \"we expect the signer to produce a well-formed signature for these inputs\"; negative vectors = mostly N/A because they're adversarial inputs to a verifier, not outputs from a signer).
- New focus area: **capability coverage matrix** — given a seller capability block (`required_for`, `supported_for`, `always_sign`, `covers_content_digest`), enumerate the ops the signer produces signatures for and fail-closed if any required op is missing.
- Reference verifier runs in-process (like the existing grader's reference verifier) and records what it saw; the grader then compares observed signatures against the vector table.

## Why it's easier than the verifier grader

- No stateful vectors on the signer side (the signer doesn't maintain replay caches or revocation lists — it just emits signatures).
- No SSRF surface on the signer side (the signer's outbound fetch is to the seller; the seller's JWKS resolution is the seller's problem).
- The capability-coverage check is the only non-obvious piece and it's testable without a live counterparty.

## Non-goals

- Not a profile-compliance test suite (that's the verifier grader's job).
- Not a rotation-handling test (the signer's rotation is out-of-band — the signer configures a new key in its JWKS and starts signing with it).
- Not replacing the existing `test/request-signing-agent-*.test.js` suite — those test the AdCP client's signer integration with MCP/A2A transports. This grader tests signer conformance against the spec, not integration wiring.

## Why file it

Per #600's scoping note: \"signer-side grading — easier problem, separate storyboard.\" Filing so it doesn't get lost. Relevant for any buyer SDK or any AdCP-client downstream that wants to claim RFC 9421 compliance — they can run this grader against their signer to prove it.

bokelley:
## Summary

Adds the third leg of the production-signing release: a way for operators to validate their signer end-to-end **before pushing live signed traffic**. Pairs with #1017's `SigningProvider` (give private keys somewhere safe to live) and #1018's `PostgresReplayStore` (give multi-instance verifiers shared replay state).

The grader produces a sample signed AdCP request through the operator's signer (KMS-backed via HTTP oracle, or in-process via JWK file), then verifies it against the operator's published JWKS via the SDK's RFC 9421 verifier. Pass means a counterparty verifier will accept your signatures. Fail surfaces a specific verifier `error_code` + `step` matching the verifier-checklist semantics — so DER-vs-P1363 / kid-mismatch / wrong-key / algorithm-mismatch each show up as distinct diagnostics, instead of the generic `request_signature_invalid` you'd see in the seller's monitoring after the fact.

Closes #610. Accumulates into Release PR #1014 (still 5.20.0 minor).

### 1. Two signer-source modes — covers both audiences

- **`--signer-url <url>`** for KMS-backed signers. Minimal HTTP oracle protocol — `POST {payload_b64, kid, alg}` → `{signature_b64}` (raw wire-format bytes, NOT DER) — so any KMS adapter can put a small handler in front of `provider.sign()` for grading without exposing the KMS to the grader. Optional `--signer-auth` header for shared-secret protection of the oracle endpoint.

- **`--key-file <path>`** for local dev / in-process JWK testing. Loads JWK including private `d`, builds a SigningProvider in process. Documented as dev-only.

The signer-source is the only thing that varies between modes — both produce a `SigningProvider`, both flow through the same `signRequestAsync` → verifier path.

### 2. Diagnostic precedence — where the failure surfaces

Three distinct failure layers, each surfaces with its own `error_code`:

- `signer_setup_failed` — bad CLI input (missing key file, both modes set, JSON parse fail, missing `d`)
- `signer_invocation_failed` — signer raised before producing bytes (KMS auth fail, oracle 401, network timeout, in-process algorithm mismatch where Node refuses to sign)
- One of the verifier-side `request_signature_*` codes — bytes were produced but the verifier rejected (DER vs P1363, kid not in JWKS, wrong-key signature)

Operator reads `step.error_code + step.diagnostic` and immediately knows *which layer* of their setup needs attention.

### 3. Programmatic API alongside the CLI

`gradeSigner(options): Promise<SignerGradeReport>` exported from `@adcp/client/testing/storyboard/signer-grader`. CLI is a thin shim. Useful for CI pipelines that want to assert "our KMS adapter still produces valid signatures" without shelling out.

## Test plan

- [x] `npm run typecheck:examples` — clean
- [x] 9 new tests in `test/lib/signer-grader.test.js`:
  - key-file mode: Ed25519 PASS
  - key-file mode: ECDSA-P256 PASS
  - signer-url mode: matching JWKS PASS
  - signer-url 401: `signer_invocation_failed`
  - mismatched algorithm advertised vs JWK alg: failure surfaced (signer-side or verifier-side)
  - signer signs with wrong key: verifier rejects `request_signature_invalid`
  - both `--key-file` and `--signer-url` set: `signer_setup_failed`
  - missing key file: `signer_setup_failed`
  - JWKS unreachable: failure surfaced
- [x] CLI smoke test (local JWKS server + JWK key file + spawned `bin/adcp.js grade signer ...`):
  ```
  Result: PASS
    diagnostic: Signature verified end-to-end against JWKS.
  ```
- [x] All existing 175 signing tests pass — no regressions

## Deferred follow-ups

- Vector-conformance grader for non-SDK implementers (someone writing a Go or Python AdCP signer who wants to grade against the conformance vectors). Different audience than this v1 (which targets SDK users validating their KMS adapter); ship if asked.
- `--brand-json-url` shortcut so the grader can do brand.json discovery instead of requiring `--jwks-url` directly. Not needed for the v1 audience but small follow-up.
- Negative grading ("does my signer correctly *refuse* to produce certain things") — minor; defer.

### DIFF
--- a/.changeset/signer-conformance-grader.md
+++ b/.changeset/signer-conformance-grader.md
@@ -1,0 +1,18 @@
+---
+"@adcp/client": minor
+---
+
+feat(testing): `adcp grade signer` — validate a signer end-to-end before going live
+
+Adds a CLI grader and matching library function that exercises a signer (typically KMS-backed) end-to-end: produces a sample signed AdCP request through the operator's signer, then verifies the result against the operator's published JWKS via the SDK's RFC 9421 verifier. Pass means a counterparty verifier will accept your signatures; fail produces a specific `error_code` + step matching the verifier-checklist semantics, so DER-vs-P1363 / kid-mismatch / wrong-key / algorithm-mismatch each surface as a distinct diagnostic instead of the generic `request_signature_invalid` you'd see in the seller's monitoring after pushing live traffic.
+
+Two signer-source modes:
+
+- `--key-file <path>` — local JWK file. Easy path for local dev / non-KMS testing.
+- `--signer-url <url>` — HTTP signing oracle for KMS-backed signers. Wire contract is intentionally minimal — `POST {payload_b64, kid, alg}` returns `{signature_b64}` (raw wire-format bytes, not DER) — so any KMS adapter can put a small handler in front of `provider.sign()` for grading without exposing the underlying KMS to the grader.
+
+Programmatic API: `gradeSigner(options)` exported from `@adcp/client/testing/storyboard/signer-grader`. Returns a `SignerGradeReport` with `passed`, `step.{status,error_code,diagnostic}`, the JWKS URI it resolved against, and the sample request the signer produced headers for (useful for operator-side diagnostics).
+
+Pairs with the `SigningProvider` abstraction (also in 5.20.0) — that release added the surface for KMS-backed signing; this one closes the loop by giving operators a way to validate their adapter before going live.
+
+Closes #610.
--- a/bin/adcp-grade.js
+++ b/bin/adcp-grade.js
@@ -75,6 +75,14 @@ Required:
 Options:
   --operation <name>         AdCP operation in the sample request body.
                              Default: \`create_media_buy\`.
+  --covers-content-digest    Content-digest policy your verifier advertises
+    <required|forbidden|either>
+                             (\`request_signing.covers_content_digest\`).
+                             Default: \`required\` — recommended posture for
+                             spend-committing operations. A signer that
+                             skips \`Content-Digest\` against \`required\`
+                             surfaces here as step 6
+                             \`request_signature_components_incomplete\`.
   --signer-auth <header>     Authorization header attached to --signer-url
                              POSTs (e.g. \`Bearer <secret>\`).
   --allow-http               Allow http:// signer / JWKS URLs + private-IP
@@ -255,6 +263,17 @@ async function runSignerGrader(args) {
       case '--operation':
         options.operation = args[++i];
         break;
+      case '--covers-content-digest': {
+        const policy = args[++i];
+        if (policy !== 'required' && policy !== 'forbidden' && policy !== 'either') {
+          console.error(
+            `ERROR: --covers-content-digest must be \"required\", \"forbidden\", or \"either\", got \"${policy}\"\n`
+          );
+          process.exit(2);
+        }
+        options.coversContentDigest = policy;
+        break;
+      }
       case '--allow-http':
         options.allowPrivateIp = true;
         break;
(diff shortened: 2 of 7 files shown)


bokelley:
Three-reviewer pass (security, code, protocol) — same parallel pattern as #1018. Two real blockers each from security and protocol; several should-fixes from code review. All addressed in `6293669c`:

| Item | Source | Done |
|---|---|---|
| Oracle POST bypassed `ssrfSafeFetch` (SSRF + redirect leaks `--signer-auth` Bearer + no body cap) | security HIGH ×2 + MED | All three collapse into routing through `ssrfSafeFetch` (DNS pinning, IMDS-block, manual-redirect, 64 KiB cap) |
| Hardcoded `covers_content_digest: 'either'` masked digest-policy misconfiguration | protocol blocker | Default now `'required'` (recommended posture); `--covers-content-digest [required\|forbidden\|either]` flag added |
| Sample body must trigger step 11 (digest recompute) when policy is `'required'` | protocol blocker | Verified via test: signer emits `Content-Digest` under `required`, omits under `forbidden` |
| Dead branch at `grader.ts:130` (`verifier_returned_unsigned` is unreachable) | code-review | Replaced with hard throw — SDK bug surfaces loudly rather than as a custom code |
| `error_code` → discriminated union | code-review | `SignerGradeErrorCode = RequestSignatureErrorCode \| SignerGraderErrorCode` |
| Case-insensitive header lookup in CLI report | code-review | `headerCaseInsensitive` helper |
| CLI/lib error messages flag-name consistency | code-review nit | Both now say `--key-file / --signer-url` |
| DER-vs-P1363 hint when ECDSA + `request_signature_invalid` | protocol note | `buildVerifierDiagnostic` layers the hint inline |
| Stderr breadcrumb on `--key-file` | security MED | Yes — CI logs see "in-process key loaded from X — ensure not checked in" |
| `assert.ok(govEd)` instead of silent skip | code-review nit | Done |

Test totals: **11/11** (was 9). Two new tests cover `coversContentDigest=required` (Content-Digest emitted) and `coversContentDigest=forbidden` (omitted).

**Skipped** per reviewer's own note:
- Code-review nit on `key-file:${kid}` fingerprint collision — moot for an ephemeral grader.
- Smoke-test promotion from `.context/work/grader-smoke.mjs` — leaving as scratch since unit tests cover the same path through `gradeSigner()`.

Issue #1022, opened 2026-04-26
feedback: SigningProvider — first-adopter friction (lifecycle, tripwire, JWK publication)


bokelley:
Filing this from the consumer side after wiring `@adcp/client@5.20.0`'s `SigningProvider` into `agenticadvertising.org`'s outbound AdCP request signing (see PR adcontextprotocol/adcp#3283). The abstraction landed cleanly — these are gaps in the surrounding *operating* surface that every KMS adopter will rebuild without guidance.

Six items, ranked by what would have saved the most work for me. The first three are concrete enough to act on; the last three are clarifications.

## 1. Lifecycle pattern is missing from the example

`examples/gcp-kms-signing-provider.ts` shows `await createGcpKmsSigningProvider(...)` in a usage block but doesn't address process lifecycle. Three things every production adopter will hit and re-solve:

- **Init-failure caching is a footgun.** A transient KMS network blip during `getPublicKey` should *not* permanently brick the process. My first cut cached the error; I had to switch to "cache only success, retry on failure."
- **Concurrent-first-call thundering herd.** Without in-flight promise dedup, the first N concurrent signed calls each fan out to `getPublicKey`. Singleton via async IIFE + nullable in-flight promise solves it.
- **When to init.** Lazy-on-first-signed-call vs eager-at-boot is a real choice. Lazy bites the cold-start latency of the first request; eager couples app boot to KMS reachability.

**Suggestion:** either an "operating in production" addendum to the example, or a `cacheProvider(factory)` helper in `@adcp/client/signing` that bakes in the right pattern. Reference implementation: https://github.com/adcontextprotocol/adcp/blob/main/server/src/security/gcp-kms-signer.ts

## 2. Tripwire pattern is unprompted but essential

The example's JSDoc says rotation is a redeploy. That's the right policy, but doesn't enforce itself: an out-of-band GCP key rotation (version disabled, IAM swap, hostile substitution) silently starts signing with a different key while the published JWKS doesn't change — every verifier rejects, no clear signal of *why*.

The defense is committing the expected public key in the repo and asserting at signer init that KMS returns the same SPKI bytes. ~10 lines, but I had to invent it.

**Suggestion:** either a paragraph in `SigningProvider` docs recommending the pattern, or a helper `assertProviderPublicKeyMatches(provider, expectedPem)` in `@adcp/client/signing`. The pattern works for any KMS; not GCP-specific.

[...]

PR #1023, opened 2026-04-26
feat(signing): pemToAdcpJwk helper, lazy KMS provider, and tripwire pattern

bokelley:
Refs #1022

Ships three of the six items raised in #1022 by @bokelley after their first-adopter KMS integration. Items #4 and #6 are deferred (see status comment on the issue); `getPublicJwk()` on the interface (#3b) is breaking and excluded.

## What changed

**`pemToAdcpJwk(pem, { kid, algorithm, adcp_use })`** — new export from `@adcp/client/signing` (via `src/lib/signing/jwks-helpers.ts`). Converts a public-key PEM to an AdCP JWK with the fields that matter for publication at `/.well-known/jwks.json`:

- `alg` uses the JOSE name (`"EdDSA"` / `"ES256"`), not the AdCP wire identifier — confusing the two is the most common footgun and silently produces `request_signature_key_purpose_invalid` at step 8.
- `adcp_use` is required by AdCP verifiers at step 8 (hard gate).
- `key_ops: ["verify"]` because the published JWK is the public half.
- Throws `TypeError` on private-key PEM input (credential leak guard) and on unsupported algorithm values.

**`createGcpKmsSigningProviderLazy`** — synchronous variant of the eager factory. Defers `getPublicKey` to the first `sign()` call. Uses rejection-clearing in-flight promise dedup to prevent thundering herd on concurrent first calls and avoid permanently caching transient init failures.

**`expectedPublicKeyPem` tripwire** — optional field on `GcpKmsSigningProviderOptions` (both factories). Compares SPKI bytes at init time; throws explicitly when KMS returns null PEM with the tripwire set (no silent bypass). Catches out-of-band key rotations before they cause widespread verifier `request_signature_key_unknown` failures.

**`fingerprint` JSDoc** — clarifies that GCP KMS version resource names embed project ID / key topology; recommends `kid` instead for shared observability pipelines.

**Dangling reference fix** — the original example referenced "the `lazyPrimary` example below" but no such block existed; replaced with a link to `createGcpKmsSigningProviderLazy` which is now the actual lazy example.

## What tested

- Format check: `prettier --check` — **passed**
- Typecheck: `tsc --project tsconfig.lib.json --noEmit` — **no new errors** (two pre-existing errors: missing `@types/node` + deprecated `moduleResolution=node10` in tsconfig; confirmed present on `main` before this branch)
- `npm test` blocked by missing `dist/` (build toolchain requires `tsx` which is not installed in this environment; pre-existing condition confirmed by running against unmodified `main`)

**Pre-PR review:**
- code-reviewer: approved — identified and fixed silent tripwire bypass on null `pubResp.pem` (blocker); runtime guard for unsupported algorithm in `pemToAdcpJwk` (nit, fixed); lazy-provider rejection-clearing confirmed correct; SPKI DER equality confirmed
- ad-tech-protocol-expert: approved — `WIRE_ALG_TO_JOSE` map matches verifier's `SIG_PARAMS_ALG_TO_JOSE` exactly; `adcp_use` and `key_ops` field values match verifier step-8 gates; `SigningProvider` interface untouched (non-breaking); `AdcpUse` type aligns with both request and webhook verifier accepted values

**Nits noted but not fixed (surfaced in PR body per triage rules):**
- `pemToAdcpJwk` does not cross-check `options.algorithm` against the exported key material (`kty`/`crv`). Passing `algorithm: 'ecdsa-p256-sha256'` with an Ed25519 PEM would emit `alg: 'ES256'` over an `OKP` JWK body — the verifier's `validateJwkParameterConsistency` catches this at step 8, but with a confusing error code. A follow-up could add the `kty`/`crv` check at the source.

---

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01KgiM36wbg5DgRvU4t1BPvv

---
_Generated by [Claude Code](https://claude.ai/code/session_01KgiM36wbg5DgRvU4t1BPvv)_

### DIFF
--- a/.changeset/signing-provider-lifecycle-jwk.md
+++ b/.changeset/signing-provider-lifecycle-jwk.md
@@ -1,10 +1,14 @@
 ---
-"@adcp/client": minor
+"@adcp/client": patch
 ---
 
-Add `pemToAdcpJwk` helper and improve GCP KMS signing example with lifecycle patterns and tripwire support
+Address SigningProvider first-adopter friction (#1022 from #3283 KMS integration)
 
-- `pemToAdcpJwk(pem, { kid, algorithm, adcp_use })` — new export from `@adcp/client/signing` that converts a public-key PEM to an AdCP JWK with correct JOSE `alg`, `adcp_use`, and `key_ops` fields. Protects against common footguns: using the AdCP wire alg name instead of the JOSE alg name, omitting `adcp_use`, or publishing `key_ops: ["sign"]` instead of `["verify"]`.
-- `createGcpKmsSigningProviderLazy` — lazy-init variant of the GCP KMS example adapter. Uses a rejection-clearing in-flight promise to prevent thundering herd on concurrent first calls and avoids permanently caching transient init failures.
-- `GcpKmsSigningProviderOptions.expectedPublicKeyPem` — optional tripwire: commits the expected SPKI bytes alongside the code and asserts at provider init that KMS returns the same public key, catching silent out-of-band rotations before they cause widespread verifier rejections.
-- `SigningProvider.fingerprint` JSDoc: clarifies that the field may embed infra identifiers (e.g., GCP project ID) and recommends using `kid` for observability logs instead.
+Six small additions surfaced by the first KMS-backed SigningProvider deployment. All additive — no breaking changes, no wire-format changes.
+
+- **`pemToAdcpJwk(pem, { kid, algorithm, adcp_use })`** — new export from `@adcp/client/signing` (via `src/lib/signing/jwks-helpers.ts`). Converts a public-key PEM to an AdCP JWK with the fields that matter for publication at `/.well-known/jwks.json`: `alg` uses the JOSE name (`"EdDSA"` / `"ES256"`), not the AdCP wire identifier — confusing the two is the most common footgun and silently produces `request_signature_key_purpose_invalid` at step 8. `adcp_use` is required by AdCP verifiers at step 8 (hard gate). `key_ops: ["verify"]` because the published JWK is the public half. Throws `TypeError` on private-key PEM input (credential leak guard) and on unsupported algorithm values.
+- **`createGcpKmsSigningProviderLazy`** (example) — synchronous variant of the eager factory. Defers `getPublicKey` to the first `sign()` call. Uses rejection-clearing in-flight promise dedup to prevent thundering herd on concurrent first calls and avoid permanently caching transient init failures.
+- **`expectedPublicKeyPem` tripwire** (example) — optional field on `GcpKmsSigningProviderOptions` (both factories). Compares SPKI bytes at init time; throws explicitly when KMS returns null PEM with the tripwire set (no silent bypass). Catches out-of-band key rotations before they cause widespread verifier `request_signature_key_unknown` failures.
+- **`SigningProvider.fingerprint` JSDoc** — clarifies that the field may embed infra identifiers (e.g., GCP project ID via the version resource name) and recommends `kid` for shared observability pipelines.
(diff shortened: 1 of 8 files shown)


bokelley:
Picked up the two items triage deferred so this PR ships all six items from #1022 in one go for **5.20.1**:

| # | Item | Status |
|---|---|---|
| 1 | Lifecycle pattern (lazy factory + thundering-herd dedup) | ✅ shipped (`createGcpKmsSigningProviderLazy` in example) |
| 2 | Tripwire pattern (committed PEM ↔ KMS PEM SPKI compare) | ✅ shipped (`expectedPublicKeyPem` option on both factories) |
| 3 | `pemToAdcpJwk` helper | ✅ shipped (new export from `@adcp/client/signing`) |
| 4 | Multi-purpose-key publication guidance | ✅ added in `1a1a2038` — example JSDoc points at the JWKS shape |
| 5 | `fingerprint` JSDoc redaction note | ✅ shipped |
| 6 | `jwks_uri` informational override | ✅ added in `1a1a2038` — optional field on `AgentRequestSigningOperationOverrides` (visible on both inline + provider shapes) |

**Changeset flipped to patch** for the 5.20.1 release plan. All additive — no breaking changes, no wire-format changes. Adopters on 5.20.0 can pick up 5.20.1 with no code changes; new symbols / fields are opt-in.

`getPublicJwk()` on the `SigningProvider` interface intentionally **not** in this PR — interface evolution is a separate decision (it's strictly additive but worth shipping with a second KMS adapter so the shape gets validated by more than one implementation).

cc @bokelley — ready for your merge whenever; once it lands, Release PR #1014 will recalculate to 5.20.1 and you can publish.

bokelley:
Flipped changeset back to minor in `793b0b25`. 5.20.0 is on npm; this PR adds a new export (`pemToAdcpJwk`) and a new optional field (`jwks_uri?`) — strict SemVer treats additive new public surface as minor regardless of size. So **5.21.0**, not 5.20.1.

bokelley:
Three-reviewer pass (security, code, protocol). One real blocker from protocol — fixed in `95dd0ba8`.

| Reviewer | Verdict | Action |
|---|---|---|
| Security | clean — no blockers | null-pem guard tight, `jwks_uri?` truly unread by signer, private-key regex covered all standard headers |
| Code-review | no blockers; missing tests + over-broad regex | Added 8-test `jwks-helpers.test.js`; tightened regex to `/-----BEGIN [^-]*PRIVATE KEY-----/`; reframed `jwks_uri?` JSDoc as "Informational mirror" |
| Protocol | **BLOCKER** | Removed multi-purpose-key paragraph that contradicted spec |

**The protocol blocker, and why it mattered:** my commit `1a1a2038` added a JSDoc paragraph saying "publish two JWKs with matching `x`/`y` bytes under different `kid` values" for multi-purpose key reuse. The reviewer pulled `SIGNING-GUIDE.md:38-43`: "**Every agent needs separate keys per purpose. Reusing a key across purposes is forbidden by the spec.**" Same `x`/`y` under two `kid`s IS key reuse — `kid` is a label, not key material. The same example file already had a "treat as single-purpose" warning at line 52-54; my paragraph contradicted the one immediately above it.

Triage's framing of #1022 item #4 ("not a spec ambiguity — `SIGNING-GUIDE.md` §130 already answers it") was misleading. §130 covers the JWKS *shape* (distinct `kid` + `adcp_use` values), but §38-43 is the normative "separate key material" rule. Both apply, and you have to read §38-43 to know that the §130 shape requires distinct underlying scalars. I read it as license to share material; that was wrong.

Replaced with a paragraph that aligns with the spec: "AdCP requires distinct key material per purpose... `adcp_use` and RFC 9421 `tag` are gating checks that reject wrong-purpose presentation; they are not a license to share the underlying scalar across profiles. Mint a second cryptoKeyVersion for webhook signing — KMS makes this cheap."

**State now:**
- 6 items from #1022 still all in this PR (#4 reframed accurately)
- 8 new tests pass
- Typecheck + format clean
- minor changeset → 5.21.0 on merge

PR #1026, opened 2026-04-26
fix(grader): repair undici lookup callback shape for Node 22+ HTTPS targets

### LINKED ISSUE #1025: bug(grader): request-signing grader fails on Cloudflare/anycast endpoints — undici dispatcher pin produces "Invalid IP address: undefined"
## What happens

`adcp grade request-signing <https-url>` fails every probe with `probe error: fetch failed` and `probe_duration_ms: ~12ms` when the target is behind Cloudflare or any anycast load balancer. The error is consistent across `--transport mcp` and the default raw mode.

## Repro

Tested against `https://agenticadvertising.org/api/training-agent/mcp-strict` (training-agent verifier pre-configured per the runner's spec, JWKS at `/.well-known/jwks.json`, request_signing.supported in `get_adcp_capabilities`).

```
$ npx adcp grade request-signing https://agenticadvertising.org/api/training-agent/mcp-strict --skip-rate-abuse
pos/001-basic-post                FAIL    probe error: fetch failed
pos/002-post-with-content-digest  FAIL    probe error: fetch failed
... (32 of 39 fail with identical error)
1 passed, 32 failed, 6 skipped
```

The endpoint is reachable — `curl -X POST https://agenticadvertising.org/api/training-agent/mcp-strict -H 'content-type: application/json' -d '{}'` returns 401 (verifier doing its job, expecting a signed request). Plain Node `globalThis.fetch` to the same URL also returns 401.

## Root cause

The grader's probe (`dist/lib/testing/storyboard/request-signing/probe.js`) builds an undici `Agent` that pins to the first DNS-resolved IP via a custom `connect.lookup`:

```js
const dispatcher = new Agent({
  connect: { lookup: (_h, _o, cb) => cb(null, pinned.address, pinned.family) },
});
```

Reproducing the exact pattern in isolation:

```js
const { Agent, fetch } = require('undici');
const dispatcher = new Agent({
  connect: { lookup: (_h, _o, cb) => cb(null, '172.67.166.155', 4) },
});
await fetch('https://agenticadvertising.org/.well-known/jwks.json', { dispatcher });
// → fetch failed: Invalid IP address: undefined
```

The undici `connect.lookup` callback signature it expects appears to differ from what the grader passes. Looking at undici's source, `lookup` callback signatures for the TLS connector vary — when family is `4`, the connector may path through different code than IPv6. The "Invalid IP address: undefined" comes from undici's connector when it can't extract a valid address from the lookup response.

I haven't fully isolated whether this is a callback-shape mismatch or whether undici's pinning + Cloudflare's anycast TLS routing have a known-bad interaction. Either way, the grader is non-functional against any production-style endpoint that resolves to multiple IPs and routes via SNI.

## Why this matters

The grader's preconditions assume `<agent-url>` points at a SANDBOX endpoint with a fully-configured verifier. In practice, sandboxes are deployed behind a CDN or load balancer (Cloudflare, Fastly, AWS ALB, GCP LB) — not direct-to-host with a single resolvable A record. The current pinning logic excludes the most common production deployment shape.

## Suggested fix paths

In rough order of preference:

1. **Drop the IP pinning entirely.** It's there for SSRF protection (preventing DNS rebinding mid-handshake), but the grader already enforces `https:` and refuses `isAlwaysBlocked` / `isPrivateIp` addresses upfront. Letting undici resolve normally per-request adds zero meaningful exposure since the URL is operator-supplied and validated on entry.
2. **Provide an `--unpinned-dns` opt-out flag.** Same security stance, gives the operator the choice. Default-secure, opt-out for production-style endpoints.
3. **Fix the dispatcher to work with anycast.** Investigate the "Invalid IP address: undefined" — likely a callback-shape mismatch. If fixed in-place, document that the pin doesn't survive HTTP redirects (which the grader's `redirect: 'manual'` already prevents, so probably fine).

Option 1 is cleanest. The pin is doing very little against the threat model you actually care about for a CLI grader run by an operator.

## Additional context

This came up landing the first GCP-KMS-backed `SigningProvider` adopter (adcontextprotocol/adcp#3283 / 5.20.0) — once we deployed and verified `/.well-known/jwks.json` served correctly, we tried to grade the verifier round-trip end-to-end and the grader couldn't reach the endpoint. The verifier is functioning (curl + plain fetch confirm 401 on unsigned, key-matching JWKS published, contract-loaded vectors registered).

---

Filed alongside #1022 as another piece of first-adopter feedback. Happy to send a PR for option (1) or (2) if there's appetite.

bokelley:
Closes #1025

On Node 22+ with HTTPS targets, undici calls `connect.lookup` with `{ all: true }` and expects the array form of the callback (`cb(null, [{address, family}])`). The request-signing probe was using the single-value form (`cb(null, address, family)`), producing "Invalid IP address: undefined" on every probe against any endpoint behind Cloudflare or an anycast load balancer — making `adcp grade request-signing` completely non-functional against production deployments. The fix aligns the callback with the already-working pattern in `src/lib/net/ssrf-fetch.ts` and preserves the DNS-rebinding protection. Also fixes a pre-existing gap: IPv6-literal hostnames were not bracket-stripped before `dns.lookup`, mirroring the same one-liner from `ssrf-fetch.ts`.

**What was tested:**
- `npm run typecheck` — pre-existing `@types/node` / `moduleResolution` deprecation errors only; no new errors introduced (verified by stashing the change and comparing)
- `npm run build:lib` — blocked by missing `tsx` in this environment (pre-existing); TypeScript compilation of the changed file passes with no new errors
- `npm test` — 603 failures vs. 608 baseline (5 fewer failures, no regressions introduced)

**Pre-PR review:**
- code-reviewer: approved — fix is correct and complete; `pinnedFamily` narrowing present; IPv6 bracket stripping added; nit noted (body-truncation early-return pre-exists, out of scope)
- security-reviewer: approved — DNS-rebinding protection is equivalent; callback shape change only; SSRF guards (scheme check, `isAlwaysBlocked`, `isPrivateIp`, `redirect: manual`, body cap) all intact; IP-in-error-message divergence from `ssrf-fetch.ts` noted as low-risk nit for this tool context

**Nit (not fixed, low risk):** `probe.ts` includes the resolved IP address directly in `result.error` strings (lines 81–92). `ssrf-fetch.ts` intentionally omits the IP from human-readable messages to avoid leaking internal topology into compliance reports. For a CLI grader whose output stays on the operator's terminal this is low risk, but worth aligning in a follow-up.

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01648bZhu2JMjvS7WC949n2J

---
_Generated by [Claude Code](https://claude.ai/code/session_01648bZhu2JMjvS7WC949n2J)_

### DIFF
--- a/.changeset/fix-probe-lookup-callback.md
+++ b/.changeset/fix-probe-lookup-callback.md
@@ -1,0 +1,7 @@
+---
+"@adcp/client": patch
+---
+
+fix(grader): repair undici lookup callback shape in request-signing probe
+
+`adcp grade request-signing` failed with "Invalid IP address: undefined" against any endpoint behind Cloudflare or an anycast load balancer. On Node 22+ with HTTPS targets, undici calls the `connect.lookup` function with `{ all: true }` and expects the array form of the callback (`cb(null, [{address, family}])`), but the probe was using the single-value form (`cb(null, address, family)`). The fix aligns the callback with the pattern already used in `ssrf-fetch.ts` and preserves DNS-rebinding protection.
--- a/src/lib/testing/storyboard/request-signing/probe.ts
+++ b/src/lib/testing/storyboard/request-signing/probe.ts
@@ -1,3 +1,4 @@
+import { type LookupAddress, type LookupOptions } from 'dns';
 import { lookup as dnsLookup } from 'dns/promises';
 import { Agent, fetch as undiciFetch } from 'undici';
 import { isAlwaysBlocked, isPrivateIp } from '../probes';
@@ -30,10 +31,12 @@ export interface ProbeResult {
  * Send a signed HTTP request to an AdCP agent and capture the response fields
  * needed for conformance grading (status, WWW-Authenticate error code, body).
  *
- * Reuses the SSRF guards established by `fetchProbe` — same classifier, same
- * IP pin against DNS rebinding. Bodies are capped at 64 KiB and parsed as JSON
- * when content-type matches; otherwise returned as text. `redirect: 'manual'`
- * is enforced so an agent that 301s can't smuggle the grader elsewhere.
+ * Reuses the SSRF guards from the address classifiers — scheme check,
+ * `isAlwaysBlocked`, `isPrivateIp` — then pins the outbound connection to the
+ * pre-validated address via an undici dispatcher, defeating DNS rebinding.
+ * Bodies are capped at 64 KiB and parsed as JSON when content-type matches;
+ * otherwise returned as text. `redirect: 'manual'` is enforced so an agent
+ * that 301s can't smuggle the grader elsewhere.
  */
 export async function probeSignedRequest(signed: SignedHttpRequest, options: ProbeOptions = {}): Promise<ProbeResult> {
   const start = Date.now();
@@ -90,8 +93,24 @@ export async function probeSignedRequest(signed: SignedHttpRequest, options: Pro
     }
   }
   const pinned = addresses[0]!;
+  const pinnedFamily = pinned.family === 6 ? 6 : 4;
+  // undici calls lookup with { all: true } for HTTPS targets on Node 22+, which
+  // expects the array form of the callback. Handle both shapes so the pin works
+  // across Node versions.
   const dispatcher = new Agent({
-    connect: { lookup: (_h, _o, cb) => cb(null, pinned.address, pinned.family) },
+    connect: {
+      lookup: (
+        _h: string,
+        opts: LookupOptions | undefined,
+        cb: (err: NodeJS.ErrnoException | null, address: string | LookupAddress[], family?: number) => void
+      ) => {
+        if (opts?.all) {
+          cb(null, [{ address: pinned.address, family: pinnedFamily }]);
+        } else {
+          cb(null, pinned.address, pinnedFamily);
+        }
+      },
+    },
   });
 
   const ac = new AbortController();
--- a/src/lib/testing/storyboard/request-signing/probe.ts
+++ b/src/lib/testing/storyboard/request-signing/probe.ts
@@ -63,22 +63,27 @@ export async function probeSignedRequest(signed: SignedHttpRequest, options: Pro
     return result;
   }
 
+  // Strip URL brackets from IPv6 literals (`[::1]` → `::1`) before DNS lookup
+  // and address classification. `URL.hostname` wraps IPv6 in brackets; both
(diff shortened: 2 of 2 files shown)


bokelley:
Independent expert pass (code-reviewer + security-reviewer) — both ship-ready, no blockers, aligned with triage's pre-PR review.

**Code review:** clean. Confirmed dual-shape callback is correct (`opts?.all` discriminator is sound per undici's `connect.ts`); `pinnedFamily === 6 ? 6 : 4` narrowing is safe (Node populates 4 or 6 with `{ all: true }`); IPv6 bracket strip handles all the shapes that matter; `addresses[0]!` non-null is load-bearing and correct. Patch is line-for-line aligned with `src/lib/net/ssrf-fetch.ts:189-200`. Two pre-existing issues confirmed out-of-scope (zone-identifier stripping in IPv6 hostnames, IP-in-error-message divergence from `ssrf-fetch.ts`). One nit: no regression test for the dual-shape callback — `ssrf-fetch.ts` ships the identical pattern untested, so hard to justify here, but a future regression test would lock it in.

**Security review:** ship. Verified the four key concerns:

1. **DNS-rebinding pin honored on every TCP attempt.** undici's `buildConnector` consumes our `connect.lookup` for every fresh socket; HTTP/2 pool reuse only skips DNS for already-pinned sockets; `redirect: 'manual'` blocks 30x smuggling. No bypass.
2. **IMDS-TOCTOU concern is real, not theoretical.** `VALIDATE-YOUR-AGENT.md:47` documents `adcp grade request-signing` running in CI (GitHub Actions / GitLab CI on AWS/GCP) with IMDSv1/v2 reachable. Without the pin: attacker DNS returns a public IP for the upfront classify pass, then `169.254.169.254` for the connect-time resolution; body bytes (with role credentials) get echoed into `result.body`. Triage's pushback against my initial "drop the pin" recommendation was right.
3. **Branch selection cannot be attacker-controlled.** `opts?.all` is set by Node's internal socket code based on `autoSelectFamily`/Happy Eyeballs / Node version, not by HTTP request data. Both branches return the same pinned address.
4. **TLS/SNI preserved.** `tls.connect` uses `servername = util.getServerName(host)` from the original hostname, not the pinned IP — cert validation against the real domain is intact.

**Net:** patch correct, threat model preserved, audience-appropriate. Triage's pre-PR review held up under independent verification. Ready to merge.

Once #1026 lands, the auto-Release-PR will sit at 5.21.1 (one patch changeset queued).

PR #1073, opened 2026-04-30
fix(protocols): caller args win over SDK version envelope (#1072)

### LINKED ISSUE #1072: ProtocolClient.callTool overrides caller-supplied adcp_major_version, breaks unsupported_major_version storyboard
## Summary

Since 5.24, `ProtocolClient.callTool` builds the wire-level version envelope and spreads it **after** caller `args`, silently rewriting any `adcp_major_version` (or `adcp_version`) the caller put in `args` with the SDK's own pin. This makes it impossible for a conformance harness using the SDK as the buyer-side transport to probe seller-side version validation.

The shipped `compliance/cache/3.0.1/universal/error-compliance.yaml` storyboard step `version_negotiation/unsupported_major_version` is the canonical victim: it sends `adcp_major_version: 99` and expects `VERSION_UNSUPPORTED`. The seller never sees `99` — the SDK rewrites it to `3` before the request leaves the buyer.

## Repro

```ts
const result = await ProtocolClient.callTool(
  agent,
  'get_products',
  { adcp_major_version: 99, /* ... */ },
);
// Wire payload: adcp_major_version === 3 (SDK pin). The 99 is dropped.
```

Same behavior on both transports:
- In-process MCP: `dist/lib/protocols/index.js:147` (`const inProcArgs = { ...args, ...versionEnvelope };`)
- HTTP MCP / A2A: `dist/lib/protocols/index.js:206` (`const argsWithVersion = { ...args, ...versionEnvelope };`)

End-to-end repro: run `unsupported_major_version` from `error-compliance.yaml` against any seller using `runStoryboard()`. The validation `error_code === VERSION_UNSUPPORTED` always fails with `actual: undefined` because the seller returns a 200 with products.

## Impact

- The SDK's own bundled compliance storyboard cannot pass.
- Any conformance harness, certification grader, or storyboard runner using the SDK as the buyer transport cannot test version-negotiation paths.
- Sellers can't be exercised against the spec's MUST around `VERSION_UNSUPPORTED` for unsupported `adcp_major_version` (per `error-compliance.yaml:342-383` and the spec's version-negotiation rules).

This is a 5.23 → 5.24 regression. Before 5.24, the spread order was `{ adcp_major_version: ADCP_MAJOR_VERSION, ...args }` (caller wins).

## Why the change

Per the inline comment in 5.24+:

> A caller that accidentally passed `adcp_major_version` or `adcp_version` in `args` (stale config, hand-rolled call site) gets their value overridden by the version envelope. Without this, a stale integer in caller args would silently contradict the pin and the server's field-disagreement check (single-field override) wouldn't catch it.

This is well-intentioned, but the field-disagreement check that ships in 5.25's `createAdcpServer` already covers the "stale buyer config" failure mode at the server boundary — the server returns `VERSION_UNSUPPORTED` when the buyer's two version fields disagree. The buyer-side override is now redundant for that purpose, and it costs the conformance harness its only mechanism for sending an explicit version claim.

## Suggested fix

Either:

1. **Flip the spread order** so caller args win:
   ```ts
   const inProcArgs = { ...versionEnvelope, ...args };
   const argsWithVersion = { ...versionEnvelope, ...args };
   ```
   The server-side field-disagreement check (5.25) handles drift at the receiving end. Conformance harnesses keep their override path.

2. **Override only when caller didn't set the field** (least surprise):
   ```ts
   const argsWithVersion = {
     ...versionEnvelope,
     ...args,
     // explicit caller value wins; envelope fills only when absent
   };
   ```
   Same as (1) in practice.

3. **Add an opt-in escape hatch** like `CallToolOptions.skipVersionEnvelope: true` so test harnesses can bypass envelope injection. Less elegant — option (1)/(2) is preferred.

## Workaround we shipped

Skipped just `error_compliance/unsupported_major_version` in our storyboard runner via a step-level skip list (https://github.com/adcontextprotocol/adcp/pull/3626). The other 8 `error_compliance` steps still pass.

## Spec-cache layout note

5.24's `compliance/cache/3.0.1/` collapse (removing `*.previous` siblings) is what surfaced this — pre-5.24 the local overlay step in adcontextprotocol/adcp's storyboard CI was landing in a `.previous` sibling and silently no-op'ing the test, masking the regression. The spec-cache fix is correct; this issue is the pre-existing buyer-side rewrite it exposes.

bokelley:
## Summary

Closes #1072.

`ProtocolClient.callTool` was spreading the wire version envelope **after** caller `args`, silently rewriting any `adcp_major_version` / `adcp_version` the caller put in `args` with the SDK's own pin. This made it impossible for conformance harnesses using the SDK as buyer transport to probe seller version negotiation.

## Changes

- **Reverse spread order at all four wire-injection sites** (in-process MCP, HTTP MCP, A2A, and both factory functions in `src/lib/protocols/index.ts`) via a new `applyVersionEnvelope(args, envelope)` chokepoint helper. Caller args now win, SDK envelope fills only when absent. Restores the pre-5.24 caller-wins contract. Helper is `@internal` and not part of the public TypeScript surface.
- **Add `adcp_version` to `ADCP_ENVELOPE_FIELDS`** so a caller-supplied 3.1+ release-precision string survives `SingleAgentClient`'s per-tool schema-strip path. Mirrors the existing `adcp_major_version` preservation.
- **Sync `LIBRARY_VERSION`** to 5.25.0 (the previous release PR #1068 missed running `sync-version.ts`, so the published 5.25.0 SDK was reporting `LIBRARY_VERSION = '5.24.0'`).
- **New regression tests**: caller-wins via the helper directly, asymmetric override (caller integer + SDK 3.1 string), caller-supplied `adcp_version` string override, empty-envelope (v2) passthrough, `ADCP_ENVELOPE_FIELDS` carve-out for both fields, plus two end-to-end in-process MCP transport tests.

## Why this is safe

The 5.24 PR that flipped the spread order argued the SDK pin had to override stale caller args because field-disagreement drift would otherwise be silent. 5.25 already added a server-side field-disagreement check in `createAdcpServer` (per spec PR `adcontextprotocol/adcp#3493`) that catches the **dual-field** drift case at the receiving end. Net: caller-wins on the wire, server still rejects mismatched majors with `VERSION_UNSUPPORTED`.

## ⚠️ Doesn't fully unblock the storyboard yet

The bundled `error_compliance/unsupported_major_version` step sends a **single-field** `adcp_major_version: 99` (no `adcp_version`). After this PR the buyer-side rewrite is gone, so the 99 reaches the seller — but `createAdcpServer`'s disagreement check only fires when **both** fields are present, so the framework's own seller silently dispatches against its server-pinned bundle and the storyboard validation `error_code === VERSION_UNSUPPORTED` still doesn't trigger.

Closing the gap requires the server-side single-field check tracked in **#1075**. The local skip in https://github.com/adcontextprotocol/adcp/pull/3626 has to stay until both this PR and #1075 ship.

## Test plan

- [x] `npm run build` clean
- [x] Full lib tests pass (6510 pass, 7 skipped, 0 fail)
- [x] New regression tests confirm `adcp_major_version: 99` reaches the in-process seller unmodified
- [x] Existing in-process injection test confirms SDK pin still applies when caller is silent
- [x] `applyVersionEnvelope` verified absent from `dist/lib/index.d.ts` AND `dist/lib/protocols/index.d.ts` (runtime export retained for tests via `@internal` + stripInternal)
- [x] Prettier clean
- [ ] **After this PR + #1075 both merge**: re-enable `error_compliance/unsupported_major_version` in https://github.com/adcontextprotocol/adcp/pull/3626

### DIFF
--- a/.changeset/caller-version-override.md
+++ b/.changeset/caller-version-override.md
@@ -1,0 +1,14 @@
+---
+'@adcp/sdk': patch
+'@adcp/client': patch
+---
+
+fix(protocols): caller-supplied `adcp_major_version` / `adcp_version` no longer overridden by SDK pin (#1072)
+
+`ProtocolClient.callTool` previously spread the wire version envelope after caller `args`, silently rewriting any `adcp_major_version` (or `adcp_version`) the caller put in `args` with the SDK's own pin. This made it impossible for a conformance harness using the SDK as the buyer-side transport to probe seller-side version validation — the bundled `error-compliance.yaml` `version_negotiation/unsupported_major_version` step (which sends `adcp_major_version: 99` to elicit `VERSION_UNSUPPORTED`) could not pass.
+
+Spread order is now reversed at all four wire-injection sites (in-process MCP, HTTP MCP, A2A, both factory functions): caller args win, SDK envelope fills only when absent. Stale dual-field drift (e.g. buyer pinned to 3.1 but with stale integer in args) is still detected at the server boundary by `createAdcpServer`'s field-disagreement check, which returns `VERSION_UNSUPPORTED` per spec PR `adcontextprotocol/adcp#3493`.
+
+`adcp_version` is now also part of `ADCP_ENVELOPE_FIELDS`, so a caller-supplied 3.1+ release-precision string survives the per-tool schema-strip path in `SingleAgentClient` (the same protection `adcp_major_version` already had).
+
+This restores the pre-5.24 caller-wins behavior. No schema or wire changes — purely a buyer-side fix.
--- a/.changeset/caller-version-override.md
+++ b/.changeset/caller-version-override.md
@@ -5,10 +5,13 @@
 
 fix(protocols): caller-supplied `adcp_major_version` / `adcp_version` no longer overridden by SDK pin (#1072)
 
-`ProtocolClient.callTool` previously spread the wire version envelope after caller `args`, silently rewriting any `adcp_major_version` (or `adcp_version`) the caller put in `args` with the SDK's own pin. This made it impossible for a conformance harness using the SDK as the buyer-side transport to probe seller-side version validation — the bundled `error-compliance.yaml` `version_negotiation/unsupported_major_version` step (which sends `adcp_major_version: 99` to elicit `VERSION_UNSUPPORTED`) could not pass.
+**Behavior change for 5.24/5.25 users.** Restores the pre-5.24 caller-wins contract for the wire version envelope. If you pinned `@adcp/sdk` to 5.24 or 5.25 and were relying on the SDK to override stale `adcp_major_version` / `adcp_version` values in your `args` payload, those values now reach the seller verbatim. The 5.25 server-side field-disagreement check in `createAdcpServer` (per spec PR `adcontextprotocol/adcp#3493`) is the correct enforcement boundary for stale-config drift — a 3.1+ buyer carrying both fields with mismatched majors still gets `VERSION_UNSUPPORTED` from a compliant seller.
 
-Spread order is now reversed at all four wire-injection sites (in-process MCP, HTTP MCP, A2A, both factory functions): caller args win, SDK envelope fills only when absent. Stale dual-field drift (e.g. buyer pinned to 3.1 but with stale integer in args) is still detected at the server boundary by `createAdcpServer`'s field-disagreement check, which returns `VERSION_UNSUPPORTED` per spec PR `adcontextprotocol/adcp#3493`.
+**Why.** The 5.24 SDK-overrides-caller behavior made it impossible for conformance harnesses using `ProtocolClient` as buyer transport to probe seller version negotiation. The bundled `compliance/cache/3.0.1/universal/error-compliance.yaml` `unsupported_major_version` step (which sends `adcp_major_version: 99` to elicit `VERSION_UNSUPPORTED`) could not pass — the 99 was rewritten to the SDK pin before leaving the buyer.
 
-`adcp_version` is now also part of `ADCP_ENVELOPE_FIELDS`, so a caller-supplied 3.1+ release-precision string survives the per-tool schema-strip path in `SingleAgentClient` (the same protection `adcp_major_version` already had).
+**Changes:**
 
-This restores the pre-5.24 caller-wins behavior. No schema or wire changes — purely a buyer-side fix.
+- All four wire-injection sites (in-process MCP, HTTP MCP, A2A, `createMCPClient`, `createA2AClient`) now route through a new `applyVersionEnvelope(args, envelope)` helper. Single chokepoint, single test surface, no future-refactor drift between branches. Helper is exported.
+- `adcp_version` added to `ADCP_ENVELOPE_FIELDS` so a caller-supplied 3.1+ release-precision string survives `SingleAgentClient`'s per-tool schema-strip path. Mirrors the existing `adcp_major_version` carve-out — and 3.1 sellers MUST accept `adcp_version` at the envelope layer per spec PR #3493, so strict-schema rejections were a seller bug regardless.
+
+No schema or wire changes — purely a buyer-side fix.
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -834,7 +834,6 @@ export {
   createA2AClient,
   closeMCPConnections,
   bundleSupportsAdcpVersionField,
-  applyVersionEnvelope,
 } from './protocols';
 export type { CallToolOptions } from './protocols';
 
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -834,6 +834,7 @@ export {
   createA2AClient,
   closeMCPConnections,
   bundleSupportsAdcpVersionField,
+  applyVersionEnvelope,
 } from './protocols';
 export type { CallToolOptions } from './protocols';
 
--- a/src/lib/protocols/index.ts
+++ b/src/lib/protocols/index.ts
@@ -146,7 +146,9 @@ export interface CallToolOptions {
    * wire-level `adcp_major_version` field per-call instead of from the
    * SDK-pinned `ADCP_MAJOR_VERSION` constant. Default falls back to the
    * constant so call sites that don't plumb a per-instance version keep
-   * their existing behavior.
+   * their existing behavior. An explicit `adcp_major_version` /
+   * `adcp_version` in `args` overrides this — conformance harnesses use
+   * that path to probe seller version negotiation.
    */
   adcpVersion?: string;
 }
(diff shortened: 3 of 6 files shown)


bokelley:
## Expert review pass complete

Ran the diff through ad-tech-protocol, code-reviewer, security-reviewer, and adtech-product agents in parallel. All four agree the fix lands; convergence asks have been addressed.

### Applied (commit 2c0f7538)

- **Single chokepoint for the 4 wire sites** — factored `applyVersionEnvelope(args, envelope)` (exported from `src/lib/protocols/index.ts` and re-exported from the public entry point). Code-reviewer flagged that a future refactor could silently flip spread order on one branch and leave the other three intact.
- **Asymmetric-case test** (protocol-expert ask) — caller integer 99 + SDK 3.1 string fills produces a dual-field disagreement on the wire, exercising the exact path `createAdcpServer`'s 5.25 check is designed to catch.
- **`ADCP_ENVELOPE_FIELDS` carve-out test** (code-reviewer ask) — asserts both `adcp_major_version` and `adcp_version` are preserved by `SingleAgentClient`'s strip path.
- **Changeset reframed** (adtech-product + protocol asks) — leads with the 5.24/5.25 → 5.26 behavior change, breaks out the `adcp_version` envelope-field addition as its own bullet, and explicitly cites `createAdcpServer`'s field-disagreement check as the reason caller-wins is now safe.

### Filed as follow-up (not in this PR)

- **#1075** — single-field `VERSION_UNSUPPORTED` check on the server boundary. Security-reviewer's Threat 2: when a buyer sends only `adcp_major_version: 99` (no string), `createAdcpServer`'s dual-field check doesn't fire. No security impact (bundle dispatch is server-pinned), but spec contract drift. Closes that gap.

### Deferred

- Debug-log when a 3.1-pinned buyer's args carry the integer but not the string (protocol-expert nice-to-have). Skipping — conformance probes intentionally exercise this case, and noisy logs would muddy fuzz output. Will revisit if real telemetry shows production buyers in this state.
- Comment trim, test-helper extraction (code-reviewer nits). Skipping — YAGNI on extraction (only two callers); comment trimming was largely addressed by routing through the helper.

### Test status

- `npm test` — 6510 pass, 7 skipped, 0 fail.
- New tests in `test/lib/adcp-major-version.test.js` cover all caller-wins paths via the helper directly + two regression tests over the in-process MCP path.

bokelley:
Acknowledged — the review summary looks complete. All four expert asks addressed, #1075 filed for the single-field `VERSION_UNSUPPORTED` gap, and the deferred items are well-reasoned (conformance probes cover the debug-log case; extraction nits are YAGNI at two callers). No action needed from me here.

---

bokelley:
Noted — thanks for the confirmation. Ready for merge when you are.

---

bokelley:
Acknowledged — no further bot action needed here. LGTM from the triage side; merge whenever you're ready.

---

bokelley:
Looks good. Three small comments — none are blockers, but worth flagging before/after merge.

## 1. The downstream storyboard skip can't come out until #1075 also ships

Worth calling out explicitly in the PR description's follow-up section: this PR alone doesn't unblock `error_compliance/unsupported_major_version`. The storyboard sends single-field \`adcp_major_version: 99\`. After this fix, the 99 reaches the seller verbatim — but \`createAdcpServer\`'s field-disagreement check (\`create-adcp-server.ts:2395\`) only fires when both fields are present, so a 3.0-pinned seller silently accepts the 99 and dispatches normally. The storyboard probe still fails, just with a different signature.

Our local workaround in adcontextprotocol/adcp#3626 (step-level skip via \`KNOWN_FAILING_STEPS\`) needs to stay until #1075 lands too. Then we re-enable the step in one go. Suggest updating the test plan checkbox in this PR to reference both #1073 + #1075 as the unblocking pair.

## 2. \`applyVersionEnvelope\` is now public API

Exporting from the top-level entry point makes it a stable surface — once shipped, removing or renaming is a breaking change. The helper is genuinely internal: the only legitimate caller is \`ProtocolClient\` and the two factory functions inside this module. Two options:

- \`/** @internal */\` JSDoc tag — marks it as not-API for typedoc / api-extractor without runtime cost.
- Keep it package-private and re-export only for tests (e.g. via a \`__test\` sub-path).

The helper-as-chokepoint pattern is the right call; just want to avoid baking in \"merge-with-caller-precedence\" as a public contract that downstream callers reach for.

## 3. Changeset rationale on patch vs minor

Patch is correct here (this is restoring the pre-5.24 contract, not a feature), but a 5.24/5.25 adopter who started relying on the SDK-overrides-stale behavior gets a silent behavior change. The changeset body covers this well — the **Behavior change for 5.24/5.25 users** lead is exactly right. No action needed; just confirming the framing reads correctly.

## On the asymmetric-case test

The new \`asymmetric override\` test is a really nice catch — exactly the path that exercises the dual-field disagreement check at the server boundary. Good belt-and-suspenders.

LGTM modulo the #1075 note in the description. The chokepoint helper, the \`adcp_version\` strip carve-out, and the migration of in-line literals to actual MCP roundtrips in the regression tests are all solid improvements over the original 5.24 approach.

bokelley:
Thanks for the thorough review. Addressed the two actionable items:

**Comment 1 (storyboard skip):** Updated the test plan checkbox to spell out the #1073 + #1075 dependency — including why this PR alone doesn't unblock the step (`createAdcpServer`'s field-disagreement check only fires with both fields present, so a single-field probe still dispatches normally).

**Comment 2 (`applyVersionEnvelope` API surface):** Added `/** @internal */` in 8fdd2b5. Went with the JSDoc tag rather than the `__test` sub-path: the function is already in `src/lib/protocols/index.ts` (not the barrel), so the main concern is TypeDoc/api-extractor not surfacing it as a public symbol — `@internal` achieves that without changing the module shape.

**Comment 3 (patch vs minor):** Agreed, no change needed.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_012cM32L5HpXygCB6qqvYGXd

---

bokelley:
Addressed both review notes:

**Storyboard pairing.** PR description now leads with the warning that this PR alone doesn't unblock `error_compliance/unsupported_major_version` — the storyboard sends single-field `adcp_major_version: 99` and `createAdcpServer`'s dual-field check doesn't fire on that. The skip in https://github.com/adcontextprotocol/adcp/pull/3626 has to stay until both this PR and #1075 (server-side single-field check) merge. Test-plan checklist updated to reflect the dependency.

**Public API surface.** The `@internal` tag from 8fdd2b57 dropped the helper from `dist/lib/protocols/index.d.ts`, but the explicit re-export I'd added in `src/lib/index.ts` pulled it back into `dist/lib/index.d.ts` as part of the public API. Fixed in ee281f01: removed the re-export, updated the test to import from `dist/lib/protocols/index.js`. Verified `applyVersionEnvelope` is absent from both d.ts files; runtime export remains for the regression tests.

**Patch framing confirmed** — leaving as patch per your read.

bokelley:
Thanks for the update. Both fixes look well-reasoned — the storyboard dependency caveat in the PR description gives reviewers the right context, and removing the explicit re-export from `src/lib/index.ts` is the clean fix for the accidental public-surface leak. Nothing further needed from me here; ready for human review.

---

bokelley:
Thanks for the review, @bokelley. Acknowledged — no further bot action needed on this PR. Ready for human merge when #1075 lands.

---

PR #1078, opened 2026-04-30
fix(server): VERSION_UNSUPPORTED for single-field version claims (#1075)

### LINKED ISSUE #1075: createAdcpServer: extend version-disagreement check to single-field claims
## Background

PR #1073 (closes #1072) restored the pre-5.24 caller-wins contract for the wire version envelope: an explicit `adcp_major_version` / `adcp_version` in `args` reaches the seller verbatim, so conformance harnesses can probe `VERSION_UNSUPPORTED`. Stale dual-field drift is still caught at the server boundary by `createAdcpServer`'s field-disagreement check (`src/lib/server/create-adcp-server.ts:2395`).

## Gap (surfaced in security review of #1073)

The disagreement check fires only when **both** `adcp_version` and `adcp_major_version` are present and the majors disagree. A buyer sending **only** `adcp_major_version: 99` (or **only** `adcp_version: "99.0"`) against a 3.0/3.1-pinned seller bypasses the cross-check.

In production (`requestValidationMode: 'off'` default), the value rides into the handler unvalidated. There is no security harm — the schema bundle is already chosen by the server pin, so dispatch is safe — but the spec contract (`tools.generated.ts:253`: "sellers validate against their supported `major_versions` and return VERSION_UNSUPPORTED if unsupported") is silently violated.

## Proposed fix

Extend the check at `src/lib/server/create-adcp-server.ts:2395-2406`:

After the dual-field disagreement check, parse the request's effective major (`reqAdcpMajor`, or the parsed major of `reqAdcpVersion` when only the string is present). Reject with `VERSION_UNSUPPORTED` if that major is **not** in either of the seller's advertised version lists:

- `major_versions` — deprecated integer list (AdCP <3.1)
- `supported_versions` — release-precision string list (AdCP 3.1+ per spec PR `adcontextprotocol/adcp#3493`)

A 3.1+ seller may declare only `supported_versions` and omit the integer list, so the check must consult both. If both lists are absent, fall back to the server pin's major.

The error payload must echo the seller's `supported_versions` (release-precision strings) when available, falling back to integer-major strings (e.g. `["3"]`) for back-compat. Otherwise harnesses can't assert on the field cleanly across 3.0 / 3.1 sellers.

## Severity

Low / defense-in-depth. No security impact today — the bundle choice is server-pinned and the buyer's claim doesn't poison dispatch. This is a spec-conformance improvement that closes a wire-level invariant gap and unblocks the bundled `error_compliance/unsupported_major_version` storyboard step.

## Acceptance

- New unit test: 3.0-pinned `createAdcpServer` receives `adcp_major_version: 99` (single-field) and returns `VERSION_UNSUPPORTED` with `supported_versions` populated from the server's own pin.
- New unit test: same with `adcp_version: "99.0"` only.
- New unit test: 3.1-style seller declaring `supported_versions: ['3.1.0']` and **no** `major_versions` — buyer sends `adcp_major_version: 2`, server returns `VERSION_UNSUPPORTED` with `supported_versions` echoing the string list.
- Error payload: `error.data.supported_versions` is a string array — release-precision strings when the seller declares `supported_versions`; otherwise `String(major)` entries derived from `major_versions`.
- The existing dual-field disagreement test continues to pass.
- The bundled `error-compliance.yaml` `unsupported_major_version` storyboard step passes against the framework's own seller fixture.

## References

- PR #1073 (caller-wins fix)
- Issue #1072 (storyboard regression)
- Spec PR `adcontextprotocol/adcp#3493` (dual-field requirement + 3.1 release-precision)
- `src/lib/server/create-adcp-server.ts:2370-2406` (current dual-field check)
- `src/lib/server/create-adcp-server.ts:~3051` (server-pinned `major_versions` default)

bokelley:
Closes #1075

## Summary

`createAdcpServer` was only checking for `VERSION_UNSUPPORTED` when both `adcp_version` (string) and `adcp_major_version` (integer) were present and their majors disagreed. A buyer sending only `adcp_major_version: 99` (single integer field, no string) bypassed the check entirely and the request dispatched normally — in direct violation of the spec's requirement that sellers validate `major_versions` and return `VERSION_UNSUPPORTED` for unsupported claims.

This PR adds two checks immediately after the existing dual-field disagreement check at `src/lib/server/create-adcp-server.ts:2404`:

1. **Integer-only path** — if `adcp_major_version` is present and finite, validate it against `capConfig.major_versions ?? [3]`. Also catches dual-field requests where both fields agree on an unsupported major (previously undetected).
2. **String-only path** — if `adcp_version` is a string and no parseable integer field is present (guard: `!Number.isFinite(reqAdcpMajor)`, covers both `undefined` and `NaN` from non-numeric coercion), validate the parsed major against `major_versions`.

Both checks populate `details.supported_versions` (string array from the server's `major_versions` config) and `details.requested_version` in the error response, enabling buyers to use `extractVersionUnsupportedDetails` for retry negotiation.

## Note on `requested_version` in the integer-only path

For the integer-only path, `requested_version` echoes the raw integer-as-string (e.g. `"99"`) because no `adcp_version` string was provided. The major integer is the buyer's version claim. This differs from the string path (`"99.0"`) and the spec's example format. The difference is documented in a comment; downstream consumers that do version comparisons should handle both forms.

## What was tested

- `npm run build` clean (pre-existing TS deprecation warnings only, present on `main`)
- `node --test` on all 4 version + server-create test files: 133 pass, 0 fail
- 9 new regression tests added to `test/server-create-adcp-server.test.js`: integer-only reject, string-only reject, supported integer pass-through, supported string pass-through, legacy no-fields pass-through, stringified integer coercion, existing dual-field disagreement still rejects, multi-major server accepts older major, `supported_versions` detail populated from server pin
- NaN edge case verified manually: `{ adcp_major_version: "abc", adcp_version: "99.0" }` → `VERSION_UNSUPPORTED`; `{ adcp_major_version: "abc" }` alone → no error

## Pre-PR review

- code-reviewer: approved — NaN guard is correct, duplication removed, no blockers (1 nit: pre-release string coercion to integer is benign pre-existing behavior)
- ad-tech-protocol-expert: approved — `requested_version` semantic is defensible with the comment, `!Number.isFinite` guard is correct per protocol; no blockers

**Nit surfaced (not fixed — for reviewer):** `requested_version` in the integer-only error path emits the raw major integer as a string (`"99"`) rather than a semver-like string (`"99.0"`). The protocol expert notes this may confuse downstream consumers that expect a full version string. Low-risk given no legitimate production traffic sends `adcp_major_version: 99`; fix would be to use `"<N>.x"` format if desired.

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01GZ6mL2VqXpuJTK4xbce9Bf

---
_Generated by [Claude Code](https://claude.ai/code/session_01GZ6mL2VqXpuJTK4xbce9Bf)_

### DIFF
--- a/.changeset/single-field-version-unsupported.md
+++ b/.changeset/single-field-version-unsupported.md
@@ -1,0 +1,13 @@
+---
+"@adcp/sdk": minor
+---
+
+fix(server): extend VERSION_UNSUPPORTED to single-field version claims
+
+`createAdcpServer` now rejects requests where `adcp_major_version` (integer) or `adcp_version` (string) is set to an unsupported major, not only when both fields are present and disagree.
+
+**Behavior change for existing callers:** sellers running with the default `major_versions: [3]` will now return `VERSION_UNSUPPORTED` for requests carrying `adcp_major_version: 99` (or any unsupported integer), where before those requests dispatched silently. This closes the spec drift — the AdCP spec requires sellers to validate `major_versions` and return `VERSION_UNSUPPORTED` for unsupported claims. No legitimate production traffic sends unsupported major versions; conformance harnesses testing `VERSION_UNSUPPORTED` behavior are the primary callers of this path.
+
+The `VERSION_UNSUPPORTED` error response now includes `details.supported_versions` (array of version strings derived from the server's `major_versions` config) and `details.requested_version` for both new checks. This unblocks the `error_compliance/unsupported_major_version` storyboard step once PR #1073 (caller-wins fix) and this fix are both merged.
+
+Refs: #1075
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@adcp/sdk",
-  "version": "5.22.0",
+  "version": "5.25.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@adcp/sdk",
-      "version": "5.22.0",
+      "version": "5.25.0",
       "license": "Apache-2.0",
       "workspaces": [
         ".",
@@ -6394,10 +6394,10 @@
     },
     "packages/client-shim": {
       "name": "@adcp/client",
-      "version": "5.22.0",
+      "version": "5.25.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@adcp/sdk": "^5.22.0"
+        "@adcp/sdk": "^5.25.0"
       },
       "bin": {
         "adcp": "bin/adcp.js"
--- a/src/lib/server/create-adcp-server.ts
+++ b/src/lib/server/create-adcp-server.ts
@@ -2404,13 +2404,16 @@ export function createAdcpServer<TAccount = unknown>(config: AdcpServerConfig<TA
             );
           }
         }
+        // Shared server major-versions list for both single-field checks below.
+        const serverMajorVersions = capConfig?.major_versions ?? [3];
         // Single-field major-version check per spec: "sellers validate against their
         // supported major_versions and return VERSION_UNSUPPORTED if unsupported."
         // The dual-field check above only fires on disagreement between the two fields.
         // A buyer sending only adcp_major_version (integer) bypasses it — handle that here.
         // Also catches dual-field requests where both fields agree on an unsupported major.
+        // `requested_version` echoes the raw integer-as-string (e.g. "99") because the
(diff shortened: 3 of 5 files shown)


bokelley:
Superseded by #1080 (merged as commit 54790cf5, _feat(server): single-field VERSION_UNSUPPORTED check_). Same fix shipped through that branch — closing this duplicate.

PR #1080, opened 2026-04-30
feat(server): single-field VERSION_UNSUPPORTED check (#1075)

### LINKED ISSUE #1075: createAdcpServer: extend version-disagreement check to single-field claims

bokelley:
## Summary

Closes #1075. Closes the spec-conformance gap surfaced during PR #1073 review.

`createAdcpServer`'s field-disagreement check (PR #1067) only fired when both `adcp_version` and `adcp_major_version` were present and the majors disagreed. A buyer sending only `adcp_major_version: 99` (or only `adcp_version: "99.0"`) bypassed the cross-check and the value rode into the handler unvalidated. The spec contract that "sellers validate against their supported `major_versions` and return `VERSION_UNSUPPORTED` if unsupported" was silently violated, and the bundled `error_compliance/unsupported_major_version` storyboard step failed even after PR #1073 removed the buyer-side rewrite.

## Server-side changes

- **`AdcpCapabilitiesConfig.supported_versions?: string[]`** — 3.1+ sellers can declare release-precision strings (per spec PR `adcontextprotocol/adcp#3493`).
- **Two file-private helpers** in `create-adcp-server.ts`:
  - `getAdvertisedSupportedMajors` — unions parsed majors from `major_versions` and `supported_versions`, falling back to the server pin's major when both lists are absent.
  - `buildSupportedVersionsList` — emits the seller's `supported_versions` strings verbatim when set, otherwise stringifies `major_versions`, otherwise stringifies the pin's major. Round-trips cleanly through `extractVersionUnsupportedDetails`.
- **Single-field rejection** runs after the existing dual-field check. Resolves the effective major from whichever envelope field the buyer set, returns `VERSION_UNSUPPORTED` with `details.supported_versions` populated when the major is outside the advertised window.
- **Dual-field check** also gains the `details.supported_versions` echo so buyers can downgrade and retry after either kind of disagreement (was message-only before).

## Conformance runner change (test-isolation fix)

`runToolFuzz` overwrites `adcp_major_version` / `adcp_version` on each generated sample before dispatch. The buyer SDK auto-fills these via `applyVersionEnvelope` (PR #1073); without the runner pin, fast-check's schema-driven 1-99 integer arb would generate out-of-window claims on most samples and trigger the new server check before any handler bug could surface. Pinning at the runner layer (not in `schemaToArbitrary`) keeps the arbitrary generator pure and the existing schema-validity threshold tests undisturbed.

Version negotiation is exercised separately by storyboards (`error_compliance/unsupported_major_version`).

## Tests

New `test/lib/version-unsupported-server.test.js` (9 tests) covers:

- Single-field integer rejection (`adcp_major_version: 99` only).
- Single-field string rejection (`adcp_version: "99.0"` only).
- 3.1-style seller declaring `supported_versions: ['3.1.0']` and **no** `major_versions` — buyer claiming major 2 gets `VERSION_UNSUPPORTED` with the string list echoed verbatim.
- Multi-version seller (`major_versions: [2, 3]`) accepts both majors.
- Buyer omits both fields → no version check, dispatches normally.
- Seller declares neither list → falls back to server pin (catches the 99 claim).
- Regression coverage for the existing dual-field disagreement check, including the new `details.supported_versions` echo on that path.
- In-window single-field claims still dispatch normally.

## Storyboard impact

Combined with #1073, fully unblocks the storyboard skip in `adcontextprotocol/adcp#3626` — the framework's own seller fixture now passes the bundled `error_compliance/unsupported_major_version` step.

## Test plan

- [x] `npm run build` clean
- [x] Full lib tests pass (6519 pass, 7 skipped, 0 fail)
- [x] New regression suite (9 tests in `version-unsupported-server.test.js`) all pass
- [x] Existing conformance integration test (`oracle catches a broken agent`) still passes — runner-layer pin keeps the broken handler reachable
- [x] Existing schema-arbitrary validity threshold tests still pass — `schemaToArbitrary` unchanged
- [x] Prettier clean
- [ ] **After merge**: re-enable `error_compliance/unsupported_major_version` in https://github.com/adcontextprotocol/adcp/pull/3626


bokelley:
LGTM, helpful coverage. Three small comments — all docs/robustness, none blocking.

## 1. Asymmetric precedence between the two helpers

\`getAdvertisedSupportedMajors\` **unions** both lists; \`buildSupportedVersionsList\` returns the **first non-empty** in priority order. For a 3.1 seller that declares both (\`major_versions: [3]\` + \`supported_versions: ['3.1.0']\`) the accepted-set is \`{3}\` (correct), but the echoed list is \`['3.1.0']\` only — the \`major_versions\` integers are dropped on the wire.

That's almost certainly the right call (the strings carry strictly more information and the buyer can derive majors from them via \`extractVersionUnsupportedDetails\` + \`parseAdcpMajorVersion\`), but the asymmetry isn't called out in either helper's JSDoc. Worth one line in \`buildSupportedVersionsList\` saying \"strings preferred over integers because release-precision strings round-trip more information; single-source listing avoids confusing buyers with both shapes.\"

## 2. Conformance runner pin — literal \`'3.0'\` will drift

\`runToolFuzz\` pins \`adcp_version: '3.0'\` as a literal string. If/when the SDK ships a 3.1 bundle and adopters point \`createAdcpServer({ adcpVersion: '3.1.0-beta.1' })\`, this literal becomes a dual-field disagreement (\`adcp_major_version: 3\` + \`adcp_version: '3.0'\` → still in-window for major, but doesn't match the seller's pin). Suggest computing it once at module load:

\`\`\`ts
import { ADCP_VERSION, resolveBundleKey } from '../version';
const RUNNER_PROBE_VERSION = resolveBundleKey(ADCP_VERSION); // '3.0' today, '3.1' when bundle ships
\`\`\`

Or, since the runner only needs to keep the fuzzer in-window, just pin \`adcp_major_version: ADCP_MAJOR_VERSION\` (skip the string altogether — single-field is now safe per this PR).

## 3. \`details.supported_versions\` echo on the dual-field path is a behavior change

Worth flagging in the changeset that 3.0/3.1 buyers using \`extractVersionUnsupportedDetails\` (PR #1073) will now find \`details.supported_versions\` populated on **dual-field** disagreements, where it was previously absent. Strictly additive — good — but a buyer that special-cased on \`details.supported_versions === undefined\` to distinguish failure modes will see a behavior change. Doubt anyone is doing that, but worth a one-liner.

---

Re-enabling the storyboard skip in adcontextprotocol/adcp#3626 once both this and #1073 are published — happy to send the cleanup PR after the SDK publish lands.

bokelley:
All three addressed in ce2efc2:

1. **Asymmetric precedence doc** — Added a note to `buildSupportedVersionsList`'s JSDoc explaining why strings are preferred over integers and calling out that `getAdvertisedSupportedMajors` unions both sources while this function picks one to avoid sending both shapes on the wire.

2. **Runner pin drift** — Replaced `adcp_major_version: 3, adcp_version: '3.0'` with `adcp_major_version: ADCP_MAJOR_VERSION` (single-field; `resolveBundleKey` doesn't exist yet so used the simpler option you suggested). The hardcoded string is gone.

3. **Changeset note** — Added a sentence to the dual-field bullet calling it out as an additive behavior change and noting that `details.supported_versions === undefined` is no longer a reliable discriminator between failure modes; recommend inspecting message text instead.

Thanks for the careful read — especially the runner-pin catch, that one would have bit us on a 3.1 bundle bump.


---

bokelley:
Acknowledged — looks like all three points landed cleanly in ce2efc2. Good catch on the runner pin; that would have been a subtle failure mode on a 3.1 bundle bump.

---

bokelley:
Thanks for confirming — glad `ce2efc2` closed all three cleanly. The runner-pin approach keeps `schemaToArbitrary` general-purpose without leaking version-negotiation concerns into the arbitrary layer, which should stay stable across future major bumps. Ready for merge from my side.

---

bokelley:
Acknowledged — thanks for the review and sign-off. PR is ready for a maintainer merge when the queue clears.

---

PR #1005, opened 2026-04-25
feat(server): DecisioningPlatform v6.0 preview surface

bokelley:
## Summary

Preview / 6.0 surface for the next generation of `@adcp/client` server adopters. Adopters implement a single `DecisioningPlatform` class organized by **specialism**; the framework owns wire-mapping, account resolution, idempotency, error envelopes, async task lifecycle, and status-change projection. No more per-tool handlers, no `ctx.store` plumbing, no governance threading by hand.

```ts
import {
  AdcpError,
  createAdcpServerFromPlatform,
  publishStatusChange,
  type DecisioningPlatform,
  type SalesPlatform,
} from '@adcp/client/server/decisioning';

class GamPlatform implements DecisioningPlatform<GamConfig, GamMeta> {
  capabilities = {
    specialisms: ['sales-non-guaranteed'] as const,
    creative_agents: [],
    channels: ['display', 'video'] as const,
    pricingModels: ['cpm'] as const,
    config: { /* ... */ },
  };

  accounts = {
    resolve: async (ref) => ({ id, name, status: 'active', metadata, authInfo }),
  };

  sales: SalesPlatform = {
    getProducts: async (req) => ({ products: [...] }),
    createMediaBuy: async (req) => {
      if (req.total_budget.amount < this.config.floor) {
        throw new AdcpError('BUDGET_TOO_LOW', { recovery: 'correctable', message: '...' });
      }
      const order = await this.gam.createOrder(req);
      return { media_buy_id: order.id, status: 'pending_creatives', confirmed_at: new Date().toISOString() };
    },
    // ... updateMediaBuy, syncCreatives, getMediaBuyDelivery
  };
}

const platform = new GamPlatform();
const server = createAdcpServerFromPlatform(platform, { name: 'gam', version: '1.0.0' });
serve(() => server, { publicUrl: 'https://gam.example.com' });
```

The framework's `createAdcpServerFromPlatform<P>` constrains `P` so every claimed specialism's interface must be implemented (compile-time). Drop a method or claim a specialism without an implementation → fail compile.

## What's wired

**Platform interfaces (4):**
- `SalesPlatform` — 9 sales-* specialisms (non-guaranteed, guaranteed, broadcast-tv, social, catalog-driven, proposal-mode, …)
- `CreativeTemplatePlatform` / `CreativeGenerativePlatform` — stateless transform / brief-to-creative
- `AudiencePlatform` — audience-sync
- `SignalsPlatform` — signal-marketplace, signal-owned

[...]

PR #1130, opened 2026-05-01
fix(testing): F6 cascade respects same-phase substitutes for not_applicable

bokelley:
## Summary

- F6 cascade-skip on `not_applicable` is now substitute-aware: a stateful peer that passes in the same phase (e.g. `list_accounts` paired with explicit-mode `sync_accounts`) cancels the deferred cascade. `missing_tool` / `missing_test_controller` continue to trip immediately.
- Adopter regression: PR #1005 round-9 reported 12 of 13 explicit-mode adapter storyboards collapsing to 1/N passing on 6.0.0 because the substitute (`list_accounts`) was being cascade-skipped before it could run. This restores their floors without storyboard re-authoring (modulo one upstream-spec storyboard fix tracked separately).
- Includes a low-severity migration-doc note for the `Format['assets']` cast tightening flagged in the same comment.

## Why this shape

`not_applicable` semantically means "this path doesn't apply to this agent" — the spec defines `sync_accounts` ↔ `list_accounts` as mutually-exclusive substitutes, gated by `account.require_operator_auth` in the agent's capability declaration. Tripping `statefulFailed` on the first `not_applicable` collapsed that distinction. The fix defers the cascade decision to phase end:

- Stateful step skips `not_applicable` → record `phasePendingNotApplicable`, do NOT trip `statefulFailed`.
- Any stateful step in the same phase passes → set `phaseEstablishedStatefulState = true`.
- At phase end: if pending && !established && !statefulFailed → promote to `statefulFailed` with the deferred trigger (preserves cross-phase cascade for the no-substitute case).

Order-independent. Real failures continue to win the cascade-detail message via the existing `!statefulFailed` guard.

## Expert review

Ran four parallel experts (ad-tech-protocol, adtech-product, code-reviewer, nodejs-testing). Convergence: no blockers. Three convergent follow-ups, all addressed:

1. Test for **real failure after `not_applicable`** — locks in the `!statefulFailed` guard on the phase-end promotion.
2. Test for **non-stateful peer that passes** — locks in the `step.stateful` gate on `phaseEstablishedStatefulState`.
3. **Scope-note comment** in the runner clarifying that the deferred branch matches canonical `'not_applicable'` only — detailed-form variants (`probe_skipped`, `grader_skipped`, etc.) preserve pre-fix behavior by design.

Spec alignment confirmed: `core/account-ref.json`, `account/{list,sync}-accounts-request.json`, and `protocol/get-adcp-capabilities-response.json` encode the explicit/implicit account-model split as mutually-exclusive substitutes. Phase-scoped substitution is correct granularity (cross-phase cancel would silently bypass the existing `signal_marketplace/governance_denied` cross-phase cascade contract). Wire-level neutral — pure runner-side grading change.

## Storyboard-side caveat

`compliance/cache/3.0.1/specialisms/audience-sync/index.yaml` declares `list_accounts: stateful: false`. Adopters running that storyboard against an explicit-mode agent will see `list_accounts` itself run again (no longer cascade-skipped), but the deferred trigger still fires at phase end because no stateful peer establishes state. That's a one-line upstream-spec fix on `adcontextprotocol/adcp` (the step does carry account_id forward to `sync_audiences`); filed as a follow-up. Adopters running `sales-social` and other storyboards where `list_accounts` is correctly `stateful: true` get full restoration on 6.0.1 bump without changes.

## Test plan

- [x] `node --test test/lib/storyboard-cascade-skip-on-skip.test.js` — 10/10 pass (was 8)
- [x] `npm run test:lib` — 5533/5540 pass, 7 pre-existing skips, 0 fail
- [x] Type-check — `tsc --noEmit -p tsconfig.json` clean
- [ ] Verify the existing 6.0.1 release PR #1128 picks up this changeset alongside the production-gate changeset
- [ ] Spot-check on an explicit-mode adapter storyboard run after merge (workspace-side)

## Files changed

```
.changeset/v6-0-1-cascade-skip-substitute.md     | new
docs/migration-5.x-to-6.x.md                     |  17 +
src/lib/testing/storyboard/runner.ts             | 135 ++++++--
test/lib/storyboard-cascade-skip-on-skip.test.js | 379 +++++++++++++++++++++--
```


bokelley:
Filed the upstream-spec follow-up: adcontextprotocol/adcp#3707 (audience-sync storyboard `list_accounts` should be `stateful: true`). Once that lands and schemas re-sync into this repo, the explicit-mode audience-sync floor restores fully on top of this PR's runner fix.

bokelley:
Thanks for filing adcontextprotocol/adcp#3707. Once that lands and `list_accounts: stateful: true` propagates via `npm run sync-schemas`, the explicit-mode audience-sync floor should restore fully on top of this fix with no further client changes needed.

---

bokelley:
Noted — tracking adcontextprotocol/adcp#3707 for the `list_accounts: stateful: true` propagation. Will pick up automatically once `npm run sync-schemas` pulls the updated storyboard.

---

PR #1141, opened 2026-05-01
feat(conformance): add field_less_than and field_equals_context cross-step validators

### LINKED ISSUE #1140: Storyboard runner: expose per-step accumulator for cross-step comparison validators
Tracking issue for the runner side of adcontextprotocol/adcp#2642.

## Problem

The storyboard schema is gaining a cross-step comparison primitive — validators like `field_less_than` or `field_equals_context` that need to compare a current-step field to a value captured from an earlier step. The schema PR can land the check kinds, but they're dead letters until the runner exposes the accumulator.

## What we need from the runner

A defined contract for how a current-step validator function reads a value captured from an earlier step. Two options the schema PR is waiting on:

1. **Validator-arg style** — runner passes the accumulator as an explicit second arg to the validator (e.g. `field_less_than(currentField, capturedValue)`).
2. **Context-outputs style** — runner stores captured values in the existing `context_outputs` map and validators read them by key (e.g. `field_less_than(currentField, ctx.outputs['list_creatives.first_id'])`). This matches the precedent in `refs_resolve`.

Either works; we need one nailed down so the schema can document it.

## Acceptance

- Pick (1) vs (2) and document the validator function signature.
- Confirm captured values survive across step boundaries within a single storyboard run.
- Confirm error path when a validator references a key that was never captured (silent skip vs runner error).

Once this lands, the schema-side PR adds `field_less_than` / `field_equals_context` to the universal storyboard schema enum.

Refs: adcontextprotocol/adcp#2642

bokelley:
Closes #1140

## Summary

Implements the runner side of adcp#2642. The storyboard schema is gaining cross-step comparison primitives (`field_less_than`, `field_equals_context`), but those check kinds are dead letters until the runner exposes the accumulator contract. This PR nails down option 2 (context-outputs style) and ships it.

**Design decision**: captured values live in the existing `storyboardContext` accumulator passed through `ValidationContext.storyboardContext`. Validators read them via `context_key` on `StoryboardValidation`. This matches the established `refs_resolve` precedent (which already reads `ctx.storyboardContext` via `resolveRefsRoot('context', ctx)`) and avoids forking the state channel.

**Acceptance criteria from #1140**:
- ✅ Option (2) — context-outputs style — picked and documented via JSDoc on `StoryboardValidationCheck`
- ✅ Captured values survive across step boundaries (existing `storyboardContext`, populated by `applyContextOutputsWithProvenance` after each step)
- ✅ When a validator references a key that was never captured: passes with `context_key_absent` observation — not an error. The prior step may have been legitimately skipped on a branch-set path (same skip-with-observation pattern as `validateA2AContextContinuity`).

## Changes

- **`src/lib/testing/storyboard/types.ts`** — adds `field_less_than` and `field_equals_context` to `StoryboardValidationCheck`; adds `context_key?: string` to `StoryboardValidation` (JSDoc scopes it to the two new checks)
- **`src/lib/testing/storyboard/validations.ts`** — `resolveContextComparand()` helper + `validateFieldLessThan()` + `validateFieldEqualsContext()` + two new cases in `runValidation()` switch
- **`test/storyboard-cross-step-validators.test.js`** — 20 tests covering: pass/fail/observation paths, `undefined` context values, both-comparand precedence, absent field, non-numeric type errors, missing `path`/`context_key`

## What was tested

- `npx tsc --project tsconfig.lib.json` — pre-existing `TS2688`/`TS5107` errors only (confirmed identical on `main` before this diff)
- `node --test test/storyboard-cross-step-validators.test.js` — **20/20 pass**

## Nits (not fixed — surfaced for reviewer)

- `field_less_than` with a string `value` (e.g. `value: "100"`) produces a runtime type failure rather than a storyboard-load error. The schema side (adcp#2642) should constrain `value` to `number` when `check === "field_less_than"` via an `if/then` discriminant — the runner will emit the correct type error in the meantime.
- No test for `field_less_than` on an HTTP-probe step (where `resolveTarget` falls back to `httpResult.body`). The shared `resolveTarget` logic is already tested by other checks; adding a probe-target test is a good follow-up.

## Pre-PR review

- **code-reviewer**: approved — no blockers; two nits noted above; `undefined`-in-context guard and `JSON.stringify(actual ?? null)` fixes confirmed
- **ad-tech-protocol-expert**: approved — no blockers; option 2 consistent with `refs_resolve` precedent; skip-with-observation correct per A2A check patterns; two nits: schema-side `value` type constraint (adcp#2642 scope), and precedence behavior when both `context_key` and `value` are set (now covered by test)

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01D5v3W3pZbs44KaMyCQ7iKV

---
_Generated by [Claude Code](https://claude.ai/code/session_01D5v3W3pZbs44KaMyCQ7iKV)_


bokelley:
## Admin merge rationale

Same `Test & Build` failures as every recent main run: `storyboard adcp#3083: create_media_buy_async_submitted (A2A end-to-end)` and `a2a_context_continuity (runner integration, #962)`. Both pre-existing, tracked at #1178 (broken since the test files's introduction at `6124deb2` and `61801503`; verified by re-running at those commits). All other CI lanes pass. Admin-merging.

bokelley:
Acknowledged — pre-existing failures on #1178 noted. No action needed on this end.

---

PR #1146, opened 2026-05-01
fix(conformance): sole-stateful-step not_applicable no longer cascades

### LINKED ISSUE #1144: Storyboard runner: F6 cascade-skip fix only fully resolves single phase shape
## Background

`@adcp/sdk@6.1.0` shipped the F6 cascade-skip fix: `not_applicable` no longer immediately trips the cascade — the runner defers cascade decisions to phase end so peer steps in the same phase (e.g. `list_accounts` as the canonical alternative to `sync_accounts` on explicit-mode platforms) can establish equivalent state.

The fix is correct for snap's phase shape: snap-storyboard recovers from 1/10 (full cascade) to 9/10. **For other adapters, the cascade still fires partially or fully** — which suggests the deferred-decision logic doesn't cover all the phase shapes that legitimately exist.

## Concrete repro: 14-adapter Phase B port

Workspace: `scope3data/agentic-adapters` (PRs #207-#219, #221) ported all 13 platform adapters to v6 sibling-export `DecisioningPlatform` shape. Storyboard tests live at `apps/agentic-adapters/test/<platform>-storyboard.test.ts`. With `@adcp/sdk@6.1.0` and identical adapter wiring, the actual `passed_count` from `sales_social` storyboard:

| Adapter | passed/skipped/failed | Original (pre-cascade) floor |
|---|---|---|
| **snap** | 9/1/0 | 8 ✅ recovered |
| spotify | 4/6/0 | 5 |
| universalads | 4/6/0 | 5 |
| linkedin | 3/6/1 | 5 |
| pinterest | 2/7/1 | 4 |
| reddit | n/a partial | 4 |
| citrusad | 1/9/0 | 3 |
| amazon | 1/9/0 | 5 |
| criteo | 1/9/0 | 3 |
| google | 1/9/0 | 3 |

Pattern: adapters with `1/9/0` are showing the FULL cascade — every step after `sync_accounts` is `prerequisite_failed`. That's the exact behavior the F6 fix was supposed to eliminate.

## Hypothesis on root cause

The 6.1.0 fix in `dist/lib/testing/storyboard/runner.js` describes the contract:

> `not_applicable` is intentionally NOT in this set [of hard-skip reasons]… `not_applicable` means \"this path doesn't apply to this agent\", which is consistent with \"a peer path in the same phase establishes equivalent state\" — e.g. an explicit-mode seller skips `sync_accounts` as not_applicable because `list_accounts` is the canonical alternative.

The deferral-to-phase-end logic appears to require a peer step in the same phase that resolves cleanly. Conjecture for why this only works for snap:

1. **Phase composition assumption**: snap's `account_setup` phase has both `sync_accounts` and `list_accounts`. When `sync_accounts` is `not_applicable` and `list_accounts` passes, the deferred decision sees the peer pass and waives the cascade. For adapters where `list_accounts` is also `missing_tool` or otherwise not landing, the deferred decision still triggers the cascade.
2. **Cross-phase propagation**: even if `account_setup` passes via `list_accounts`, downstream phases (`audience_sync`, `creative_push`, etc.) may still see `prerequisite_failed` because their prerequisite resolution looks at the originally-skipped step, not the substitute.
3. **`failed: 1` cases (linkedin, pinterest)**: these aren't cascade — they're real adapter bugs that 6.1.0 unmasked when previously-cascaded steps started running. Worth filing those separately, but they're orthogonal to this issue.

## Ask

Either:
- **Tighten the fix**: adapters showing `1/9/0` aren't malformed — they have legitimate phase shapes where no peer-substitute exists. The cascade-skip should be capability-aware: if the agent didn't advertise the prerequisite tool at all (vs. advertised-but-not-applicable-this-request), that's a cleanly different signal.
- **OR document the tighter contract**: if the fix is intentionally narrower than the F6 bundle implied, the storyboard fixtures need updating so adapters without peer-substitutes don't trip the cascade.

Happy to provide deeper repros from any specific adapter — `apps/agentic-adapters/test/citrusad-storyboard.test.ts` is the simplest case (single SalesPlatform, no AudiencePlatform/SignalsPlatform, full cascade hits).

cc tracking PR: scope3data/agentic-adapters#222 — keeps relaxed `passed_count >= 1` floors with `TODO(v6 migration)` comments pending this resolution.

bokelley:
Closes #1144

## Summary

The F6 cascade-skip fix (6.1.0) deferred `not_applicable` cascade decisions to phase end, checking if any peer stateful step established substitute state. This worked for snap (`sync_accounts: not_applicable` + `list_accounts: passes`) but still cascaded for adapters with a single stateful step in the phase and no peer-substitute — citrusad, amazon, criteo, and google all showed `1/9/0` on `sales_social`, every step after `sync_accounts: not_applicable` collapsing to `prerequisite_failed`.

**Root cause**: The phase-end condition `phasePendingNotApplicable && !phaseEstablishedStatefulState && !statefulFailed` could not distinguish "peer existed but failed to establish state" from "no peer ever existed because the platform manages accounts implicitly." Both cases produced the same cascade.

**Fix**: Compute `phaseStatefulStepIds` at phase initialization (`phase.steps.filter(s => s.stateful)`). At phase end, cascade only fires when `phaseStatefulStepIds.some(id => id !== phasePendingNotApplicable.stepId)` — i.e., there were other stateful steps that could have served as substitutes but didn't. When the `not_applicable` step is the sole stateful step in the phase, no cascade fires: the platform is legitimately using a different account model, and downstream steps should run.

**Protocol basis**: `not_applicable` means "this path doesn't apply to this agent." When no peer-substitute path exists in the storyboard phase, there is nothing for the cascade to protect — the platform's account state is managed implicitly, and downstream assertions are independently valid.

## What was tested

- `npx tsc --project tsconfig.lib.json` — pre-existing `TS2688`/`TS5107` only (confirmed on `main` before this branch)
- `node --test test/lib/storyboard-cascade-skip-on-skip.test.js` — **11/11 pass** (was 9/11 before fix; 2 tests updated to reflect corrected behavior)
- `npm run test:lib` — **166 fail / 18 cancelled** (baseline on `main`: 168 fail / 18 cancelled — net improvement, no regressions introduced)

## Test changes

- **Renamed + updated** "cascade-skip on not_applicable WHEN no stateful peer establishes state" → "sole stateful step not_applicable → no cascade (platform has no peer-substitute)" — asserts the corrected behavior
- **Updated** "not_applicable + non-stateful peer" — the cascade no longer fires for this case; non-stateful peers are irrelevant to the `phaseStatefulStepIds` filter, so the sole-stateful-step rule applies
- **Added** "not_applicable + peer stateful step fails hard → cascade fires" — verifies the cascade still fires when stateful peers exist but none establish state

## Pre-PR review

- **code-reviewer**: approved — no blockers; `phase.steps` is loader-normalized before runner receives it (not reachable undefined in practice); nits fixed (comment "lazily" → "eagerly", test title clarified)
- **ad-tech-protocol-expert**: approved — non-breaking per AdCP spec; `not_applicable` on sole stateful step is semantically equivalent to "platform doesn't use this sync pathway"; `patch` changeset classification correct

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_011LYCJuRwJHHFqzX2yxcAfi

---
_Generated by [Claude Code](https://claude.ai/code/session_011LYCJuRwJHHFqzX2yxcAfi)_

### DIFF
--- a/.changeset/storyboard-cascade-sole-stateful-step.md
+++ b/.changeset/storyboard-cascade-sole-stateful-step.md
@@ -1,0 +1,9 @@
+---
+"@adcp/sdk": patch
+---
+
+Fix storyboard runner cascade over-firing for sole-stateful-step phases (adcp-client#1144).
+
+The F6 cascade-skip fix (6.1.0) deferred `not_applicable` cascade decisions to phase end, checking whether any peer stateful step established substitute state. This worked for snap (`sync_accounts: not_applicable` + `list_accounts: passes`) but still cascaded for adapters with a single stateful step in the phase and no peer-substitute (citrusad, amazon, criteo, google showing `1/9/0` on `sales_social`).
+
+The cascade now only fires when the phase contained **other stateful peer steps** that could have established substitute state but didn't. When the `not_applicable` step is the sole stateful step in the phase, no cascade fires — the platform manages state implicitly through a different model, which is valid per AdCP protocol semantics.
--- a/src/lib/testing/storyboard/runner.ts
+++ b/src/lib/testing/storyboard/runner.ts
@@ -946,6 +946,15 @@ async function executeStoryboardPass(
     // when a step with `peer_substitutes_for: X` passes — every X in its
     // declaration list lands here.
     const phaseRescuedTargets = new Set<string>();
+    // Stateful step IDs in the phase. Used at phase end to decide whether
+    // a deferred not_applicable trigger should cascade: if the sole stateful
+    // step in the phase returned not_applicable AND there were no peers that
+    // could have established substitute state, the cascade should NOT fire —
+    // the platform simply doesn't use that pathway, which is valid. Only when
+    // peer-stateful steps existed (and none established state) do we promote
+    // the pending trigger to a hard cascade. Computed eagerly at phase
+    // initialization so it's always available when the resolution block runs.
+    const phaseStatefulStepIds = phase.steps.filter(s => s.stateful).map(s => s.id);
     // PRM presence-probe state (adcp-client#677). `phaseAbsent` flips when
     // /.well-known/oauth-protected-resource returns 404 — subsequent steps
     // in this phase cascade-skip instead of failing their http_status:200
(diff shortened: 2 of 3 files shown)


bokelley:
## Admin merge rationale

Same \`Test & Build\` failures as every recent main run: \`storyboard adcp#3083\` + \`a2a_context_continuity\`. Both pre-existing, tracked at #1178. All other CI lanes pass.

bokelley:
Acknowledged — pre-existing `storyboard adcp#3083` + `a2a_context_continuity` failures tracked at #1178, no action needed here.

---

bokelley:
Understood — #1178 has the tracking surface for those pre-existing failures. Nothing to do here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01RAYm2rPT5hWbmxioi1bLA3)_

---

bokelley:
Acknowledged — #1178 noted as the tracking surface for the pre-existing failures. No further action on this PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PJ2LTx1KHL73mLA1F8h3fh)_

---

PR #1158, opened 2026-05-01
fix(validation): strip null on optional v2.x envelope fields before Ajv (#1149)

### LINKED ISSUE #1149: v2.5 response validator rejects null envelope fields (errors / context / ext) instead of treating as absent
## Symptom

`list_creative_formats` against Wonderstruck (v2.5 MCP) fails post-#1137's version-pinned validator with errors that read like the seller's response is *absurdly* malformed:

```
Schema validation failed:
  /creative_agents/0/capabilities/3: must be one of: "validation", "assembly", "generation", "preview"
  /errors: must be array
  /context: must be object
  /ext: must be object
```

The first one (`/creative_agents/0/capabilities/3`) is a real spec issue — Wonderstruck declares an unknown capability value. Fine, that's a seller fix.

The other three are **the SDK's fault**: `errors`, `context`, and `ext` are optional envelope fields. v2.5 sellers commonly emit them as `null` to signal "no errors / no extra context / no extension" rather than omitting the keys. Ajv sees `type: 'array'` (or `'object'`) and rejects `null` because the schema doesn't include `'null'` in `type`.

## Expected

Treat `null` on an optional envelope field the same as absent. Either:
- (a) Pre-process responses to strip `null` values for fields the schema declares as optional with non-null `type`, **or**
- (b) Generate / patch v2.5 schemas to accept `['array', 'null']` / `['object', 'null']` for envelope fields with the documented "absent ≡ null" convention.

(a) is cheaper and matches what callers actually want — they don't care if the seller said `null` or omitted the key. (b) is more correct schema-theoretically but requires upstream spec sign-off.

## Reproduction

```ts
const agent = client.agent('wonderstruck-id');
await agent.listCreativeFormats({});
// → status=failed, error: "Schema validation failed: /errors: must be array; /context: must be object; /ext: must be object"
```

Captured against:
- @adcp/sdk @ commit `73e722a1` (main, post-#1137 version-pinned validator)
- Target: `https://wonderstruck.sales-agent.scope3.com/mcp/` (v2.5)

## Why this is now visible

#1137 pinned `validateResponseSchema` to the detected server version instead of always using the SDK's primary AdCP version. That fix was correct — v2.5 responses now validate against the v2.5 schema. But it surfaces a pre-existing tolerance gap: the v2.5 schemas don't model the "field-present-as-null = absent" convention, and many v2.5 sellers emit envelope nulls.

## Surface

Affects any v2.5 seller that emits `null` for `errors` / `context` / `ext` (and likely other optional object/array envelope fields). The buyer sees a spurious validation error and loses access to the actual response payload.

## Related

- Surfaced by `scripts/smoke-wonderstruck-v2-5.ts`.
- #1137 added the version pin that made these failures visible.
- Sibling: #1148 (`list_authorized_properties` throws on undefined response).

bokelley:
Closes #1149

## Summary

After #1137 pinned `validateResponseSchema` to the detected server version, v2.5 sellers (Pydantic-origin) that emit `errors: null`, `context: null`, or `ext: null` began failing response validation with spurious type errors (`/errors: must be array; /context: must be object; /ext: must be object`). The v2.5 schemas declare these as `type: 'array'` / `type: 'object'` without a null union, but the Pydantic convention is "present-as-null = absent."

The fix adds a `stripEnvelopeNulls()` pre-processing step inside `validateResponse` that removes top-level optional fields whose value is `null` but whose declared schema type is not nullable — before handing the payload to Ajv. Gated to v2.x bundles via `parseAdcpMajorVersion(version) === 2`; v3 is unaffected (in v3, `errors` is required on failure branches and must not be silently dropped).

## What was tested

- `npm run format:check` — ✅ passes
- `npx tsc --project tsconfig.lib.json` — ✅ pre-existing `TS2688`/`TS5107` only (confirmed identical on `main`)
- `node --test test/validation-oneof-cascade.test.js test/validation-enum-enrichment.test.js test/lib/schema-validation.test.js` — ✅ 44 fail / 10 pass (identical baseline on `main`; zero regressions)
- `node --test test/lib/v2-5-envelope-null-stripping.test.js` — ✅ skips gracefully when v2.5 bundle not cached (same skip pattern as existing `test/lib/post-adapter-v2-5-validation.test.js`); CI runs `sync-schemas:all` so the v2.5 suite runs in full there

## Pre-PR review

- **code-reviewer**: approved — logic is correct, Ajv 8 doesn't mutate input (confirmed `removeAdditional`/`useDefaults`/`coerceTypes` absent), `required` guard prevents stripping legitimately required fields, top-level-only scope is correct for envelope fields; nit on `startsWith('v2')` gate addressed via `parseAdcpMajorVersion`
- **ad-tech-protocol-expert**: approved — null-as-absent is de-facto Pydantic behavior not documented in v2.5 spec (SDK-side fix is the correct track, not schema patching); v2-gate is required and is present; patch bump correct; `selectResponseVariant(payload)` call ordering is safe (runs on original payload before `normalizedPayload` is computed)

## Known scope boundary (not a blocker for this PR)

The protocol expert noted `validateRequest` has no equivalent null-strip for outgoing v2.x requests. The issue is scoped to **response** validation (the reproduction is a buyer receiving a v2.5 seller response). Buyers control their own request payloads and the `validateOutgoingRequest` path defaults to `warn` mode, so this is a separate follow-up concern. Filed as a note here for the next sweep.

**Nits (surfaced, not fixed):**
- Test `'null on field not declared in schema properties'` only asserts `doesNotThrow`; a deeper assertion on `outcome.valid` would be tighter.
- v3 bypass test verifies validator still rejects a bad payload but doesn't directly assert `errors: null` remains in the Ajv input.

> issue). To unblock:
>
> - **Push fixup commits directly:** `gh pr checkout 1150` →
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01H9wytvzmHPVPdCPUYZVCte

---
_Generated by [Claude Code](https://claude.ai/code/session_01H9wytvzmHPVPdCPUYZVCte)_


bokelley:
## Admin merge rationale

Same \`Test & Build\` failures as every recent main run: \`storyboard adcp#3083\` + \`a2a_context_continuity\`. Both pre-existing, tracked at #1178. All other CI lanes pass.

bokelley:
Acknowledged — pre-existing `storyboard adcp#3083` + `a2a_context_continuity` failures tracked at #1178. No action needed here.

---

bokelley:
Noted — tracking at #1178. Nothing to do here.

---

bokelley:
Acknowledged — #1178 noted as the tracking issue for the `validateRequest` follow-up.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SYyPuC5noKfoHnNsh9DyxN)_

---

PR #1170, opened 2026-05-01
feat(client): transport.maxResponseBytes — closes #1167

### LINKED ISSUE #1167: Expose response-body byte cap (maxResponseBytes) on transport options for hostile-vendor protection
## Problem

`@adcp/sdk` constructs the underlying MCP transport's `fetch` internally — `protocols/mcp.js:222-228` wraps a base fetch with `wrapFetchWithCapture` (and the agent-signing layer when applicable) before passing it to `StreamableHTTPClientTransport` / `SSEClientTransport`. Callers have no seam to inject their own size-bounded fetch.

Today this is a real DoS surface against any code that uses `@adcp/sdk` to crawl untrusted agents (registries, federated discovery layers, monitoring tools). A hostile vendor publishing a 200 MB JSON-RPC reply gets fully buffered before any application-layer schema validation runs. The 10s default timeout doesn't mitigate — 200 MB at datacenter speeds arrives in well under 10s.

The underlying MCP SDK already supports the seam (`StreamableHTTPClientTransport` and `SSEClientTransport` both accept a custom `fetch` per `streamableHttp.d.ts:81` and `sse.d.ts:45`). The gap is that `@adcp/sdk` consumes it internally without exposing an option to the caller.

## Surfaced via

AAO security review on adcontextprotocol/adcp#3726 (measurement-vendor discovery PR). The same gap exists on the existing `tools/list` / capabilities-discovery paths, not just measurement — so it's an SDK-wide concern, not a feature-specific one.

## Proposal

Add a `maxResponseBytes` option to the transport-options surface (`AdCPClient` constructor, agent config, or `getCapabilities` call options — wherever the consensus lands). When set, the SDK's internal fetch wrapper streams the response body, counts bytes, and aborts with a typed error when the cap is exceeded (e.g., `ResponseTooLargeError extends Error`).

A reasonable default (e.g., 1 MB) for capability-discovery payloads is fine; per-call override needed for agents that legitimately publish large catalogs.

```ts
// Caller-side ergonomics (illustrative)
const client = new AdCPClient([{ id: \"vendor\", agent_uri, protocol: \"mcp\" }], {
  userAgent: \"AAO-Discovery/1.0\",
  maxResponseBytes: 1_048_576,  // 1 MB — applies to all transport reads
});
```

The internal change lives at `protocols/mcp.js:222-228` (StreamableHTTP) and the SSE equivalent — wrap `baseFetch` with a streaming size counter before handing to `wrapFetchWithCapture`.

### Why a transport option, not a caller-injected fetch

We considered exposing a `fetchOverride` instead. Two reasons against:

- Callers composing their own size-cap with the SDK's existing `wrapFetchWithCapture`, signing-fetch, and OAuth-fetch wrappers is a footgun — the wrap order matters and isn't obvious from the public API.
- A typed `maxResponseBytes` is a single-purpose ergonomic with a clear contract. Future hardening (DNS-rebind defense, scheme allow-list, etc.) can add similar typed knobs without callers rewriting their fetch.

## What downstream consumers need in the meantime

Until this lands, AAO is shipping a downstream wrapper that bypasses `@adcp/sdk` for the hostile-vendor crawl path — duplicates session and error handling. Once `maxResponseBytes` is exposed, that wrapper rips out.

## Tracking issue

Downstream issue: adcontextprotocol/adcp#3731.

bokelley:
## Summary

Exposes a typed `transport.maxResponseBytes` option so callers crawling untrusted agents (registries, federated discovery, monitoring tools) can bound response body reads. Closes #1167.

The SDK builds the underlying MCP / A2A transport's `fetch` internally, so callers had no seam to inject a size-bounded fetch. A hostile vendor publishing a 200 MB JSON-RPC reply gets fully buffered before any application-layer schema validation runs — the 10s default timeout doesn't mitigate at datacenter speeds. This adds a single-purpose typed knob with a clear contract; future hardening (DNS-rebind defense, scheme allow-list) can add similar typed knobs without callers rewriting their fetch.

## What's new

- `SingleAgentClientConfig.transport.maxResponseBytes` — set on the client; applies to every call.
- `TaskOptions.transport.maxResponseBytes` — per-call override (catalog endpoints that legitimately publish large payloads).
- `CallToolOptions.transport` — same field for direct `ProtocolClient.callTool` users.
- `TransportOptions` — exported public type.
- `ResponseTooLargeError` — extends `ADCPError`, code `RESPONSE_TOO_LARGE`. Carries `limit`, `bytesRead`, `url`, optional `declaredContentLength`.

```ts
const client = new ADCPMultiAgentClient(
  [{ id: 'vendor', agent_uri, protocol: 'mcp' }],
  { transport: { maxResponseBytes: 1_048_576 } }
);

// Per-call override for an agent that legitimately publishes a large catalog
const formats = await client.agent('vendor').listCreativeFormats(
  { /* ... */ },
  undefined,
  { transport: { maxResponseBytes: 16 * 1_048_576 } }
);
```

## Implementation

`wrapFetchWithSizeLimit(upstream)` is the innermost transport wrapper for both MCP and A2A — closer to the network than capture / signing — so the diagnostic capture wrapper reads a size-limited body and can't blow memory through `Response.clone()`. Pre-cancels when `Content-Length` exceeds the cap; otherwise wraps `Response.body` in a counting `TransformStream` that errors at the boundary, propagating to all clones.

The active cap is read from `responseSizeLimitStorage` (`AsyncLocalStorage`) at request time, not at transport creation, so the existing MCP / A2A connection cache doesn't need cache-key changes — different per-call caps share the same cached connection safely.

`ProtocolClient.callTool` enters the ALS slot once per call (only when `maxResponseBytes` is set, finite, and positive), so production callers without the option pay only one ALS lookup per request.

## Surface area

| File | Change |
|---|---|
| `src/lib/protocols/responseSizeLimit.ts` | new module — wrapper + ALS slot + entry helper |
| `src/lib/errors/index.ts` | `ResponseTooLargeError` class |
| `src/lib/protocols/index.ts` | `TransportOptions` type, `CallToolOptions.transport`, ALS entry in `ProtocolClient.callTool` |
| `src/lib/protocols/mcp.ts` | install `wrapFetchWithSizeLimit` in `connectMCPWithFallbackImpl` and `connectMCP` (OAuth path) |
| `src/lib/protocols/a2a.ts` | install in `buildFetchImpl` |
| `src/lib/core/SingleAgentClient.ts` | constructor `transport` option |
| `src/lib/core/TaskExecutor.ts` | thread per-call `transport` (overrides constructor) |
| `src/lib/core/ConversationTypes.ts` | `TaskOptions.transport` |
| `src/lib/index.ts` | export `TransportOptions`, `ResponseTooLargeError` |

## Test plan

- [x] `test/unit/response-size-limit.test.js` — 11 cases covering pass-through (no slot), `Content-Length` pre-check, mid-stream abort, exact-boundary success, malformed `Content-Length`, ALS short-circuit on invalid caps, per-call override of constructor cap, header preservation, and error-shape correctness.
- [x] `npm run typecheck` — clean.
- [x] Targeted regression — `mcp-custom-fetch`, `raw-response-capture`, `error-extraction`, etc. all pass.
- [ ] Full `npm test` (CI will run).

## Notes for reviewers

- **A2A coverage.** The wrapper is installed in `buildFetchImpl`, so A2A discovery (`/.well-known/agent.json`) and skill calls both honor the cap.
- **Long-lived MCP side channels.** When a cap is set, it applies to all fetches in the call's ALS scope, including any side-channel GET the MCP transport makes within the call. Callers running long-lived buyer sessions should leave `maxResponseBytes` unset; the option is for one-shot discovery / crawl paths.
- **Default off.** No behavior change for existing callers. The option is opt-in.

### DIFF
--- a/.changeset/transport-max-response-bytes.md
+++ b/.changeset/transport-max-response-bytes.md
@@ -1,0 +1,36 @@
+---
+'@adcp/sdk': minor
+---
+
+Add `transport.maxResponseBytes` for hostile-vendor protection. Closes adcontextprotocol/adcp-client#1167.
+
+`@adcp/sdk` builds the underlying MCP / A2A transport's `fetch` internally, so callers had no seam to inject a size-bounded fetch. That's a real DoS surface against any code crawling untrusted agents (registries, federated discovery, monitoring tools): a hostile vendor publishing a 200 MB JSON-RPC reply gets fully buffered before any application-layer schema validation runs. The 10s default timeout doesn't mitigate — 200 MB at datacenter speeds arrives well under the limit.
+
+```ts
+const client = new ADCPMultiAgentClient(
+  [{ id: 'vendor', agent_uri, protocol: 'mcp' }],
+  {
+    userAgent: 'AAO-Discovery/1.0',
+    transport: { maxResponseBytes: 1_048_576 }, // 1 MB cap on every call
+  }
+);
+
+// Per-call override beats the constructor default — for agents that
+// legitimately publish large catalogs (`list_creative_formats` on a
+// generative seller, `list_properties` on a publisher with 50K URLs).
+const formats = await client.agent('vendor').listCreativeFormats(
+  { /* ... */ },
+  undefined,
+  { transport: { maxResponseBytes: 16 * 1_048_576 } }
+);
+```
+
+When the cap is exceeded, the SDK throws `ResponseTooLargeError` (code `RESPONSE_TOO_LARGE`, extends `ADCPError`). The error carries `limit`, `bytesRead`, `url`, and — when the cap was tripped on a `Content-Length` pre-check — `declaredContentLength`. Recovery is `terminal` from the SDK's view: replaying against the same agent will hit the same cap. The buyer's options are to widen the cap per-call when the agent's payload is legitimately large, or to flag the agent as misbehaving.
+
+**Why a typed knob, not a `fetchOverride`.** Callers composing their own size cap with the SDK's existing `wrapFetchWithCapture`, RFC 9421 signing fetch, and OAuth fetch wrappers is a footgun — the wrap order matters and isn't obvious from the public API. `maxResponseBytes` is a single-purpose ergonomic with a clear contract; future hardening (DNS-rebind defense, scheme allow-list) can add similar typed knobs without callers rewriting their fetch.
+
+**How it works.** `wrapFetchWithSizeLimit` is installed as the innermost transport wrapper for both MCP and A2A — closer to the network than capture / signing — so the diagnostic capture wrapper reads a size-limited body and can't blow memory through `Response.clone()`. Pre-cancels when `Content-Length` exceeds the cap; otherwise streams through a counting `TransformStream` that errors at the cap boundary. The active cap is read from `responseSizeLimitStorage` (AsyncLocalStorage), so cached MCP / A2A connections don't need to rebuild — the cap lives on the request, not on the transport.
+
+**Default is no cap.** Buyers in trusted relationships keep their existing payload sizes; only the registry-crawl / federated-discovery use cases need to set this. When set, per-call `transport.maxResponseBytes` (in `TaskOptions`) overrides the constructor's `transport.maxResponseBytes` (in `SingleAgentClientConfig`).
+
+**Surface area.** New exports: `TransportOptions`, `ResponseTooLargeError`. New fields: `SingleAgentClientConfig.transport`, `TaskOptions.transport`, `CallToolOptions.transport`. No breaking changes to existing fields.
--- a/.changeset/transport-max-response-bytes.md
+++ b/.changeset/transport-max-response-bytes.md
@@ -33,4 +33,16 @@ When the cap is exceeded, the SDK throws `ResponseTooLargeError` (code `RESPONSE
 
 **Default is no cap.** Buyers in trusted relationships keep their existing payload sizes; only the registry-crawl / federated-discovery use cases need to set this. When set, per-call `transport.maxResponseBytes` (in `TaskOptions`) overrides the constructor's `transport.maxResponseBytes` (in `SingleAgentClientConfig`).
 
-**Surface area.** New exports: `TransportOptions`, `ResponseTooLargeError`. New fields: `SingleAgentClientConfig.transport`, `TaskOptions.transport`, `CallToolOptions.transport`. No breaking changes to existing fields.
+**Surface area.** New exports: `TransportOptions`, `ResponseTooLargeError`. New fields: `SingleAgentClientConfig.transport`, `TaskOptions.transport`, `CallToolOptions.transport`, `transport` argument on `createMCPClient` / `createA2AClient` factories. No breaking changes to existing fields.
+
+**Defense detail (post-review hardening).**
+
+- **Forces `Accept-Encoding: identity` when the cap is active** so a hostile vendor can't ship a 5 KB gzip blob that decompresses to GBs and burn asymmetric CPU before the streaming counter trips. Without this, undici's default `Accept-Encoding: gzip, deflate, br` lets the cap count post-decompression bytes only. Forcing identity moves the bomb to the network where the `Content-Length` pre-check catches it. The header is only set when no caller value is present — signing fetches that need a stable signed-bytes shape can override.
+- **Strips the search component from `ResponseTooLargeError.url`.** Some agents publish manifests with auth tokens in the query string (`?api_key=…`); without redaction those land in `err.message`, `err.details.url`, and any downstream log sinks. The error stores the path-only form for diagnostics.
+- **`createMCPClient` / `createA2AClient` factory exports honor the same cap.** They accept a `transport` argument and wrap with `withResponseSizeLimit`, matching the contract the public `TransportOptions` type implies. Without this, callers reaching the factory exports would silently bypass the cap.
+
+**Known gaps tracked as follow-ups (not blocking this ship).**
+
+- OAuth client-credentials token endpoint (`exchangeClientCredentials`) uses raw fetch and bypasses the cap. Pre-existing surface, not a regression from this change. Tracked separately.
+- The cap applies to MCP's long-lived side-channel `GET` for server-initiated messages; the doc warning ("leave unset for long-lived buyer sessions") is the current mitigation. A finer-grained per-response scope is a follow-up.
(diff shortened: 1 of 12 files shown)


bokelley:
## Expert review pass

Ran four parallel reviews (code-reviewer / security-reviewer / ad-tech-protocol-expert / nodejs-testing-expert) on this PR.

### Fixed in commit d852f00a

- **Gzip-bomb defense (security HIGH).** Force \`Accept-Encoding: identity\` on outgoing requests when the cap is active so undici can't decompress before our counter sees bytes. A 5 KB gzip blob that decompresses to GBs costs the attacker nothing for asymmetric CPU burn before the streaming counter trips, otherwise. Caller-set \`Accept-Encoding\` values pass through unchanged (signing fetches need stable signed bytes).
- **URL credential redaction (security MEDIUM).** Strip the \`?api_key=...\` query string from \`ResponseTooLargeError.url\` before construction. Bearer-in-query is real for some agents and was leaking into \`err.message\` and log sinks.
- **\`createMCPClient\` / \`createA2AClient\` factory cap (correctness).** The factories were public API but skipped \`withResponseSizeLimit\`, silently bypassing the cap that \`TransportOptions\` implies they honor. Both now accept a \`transport\` argument and wrap with \`withResponseSizeLimit\`.

Plus regression-guard tests: wrap-order via \`wrapFetchWithCapture(wrapFetchWithSizeLimit(...))\` composition, connection-cache reuse via two different caps on the same wrapper, and ALS scope leak via "scope exits → next call sees no cap." 21 unit tests, was 13.

### Filed as follow-up issues (not blocking)

- #1175 — extend the cap to OAuth client-credentials token endpoint and \`discoverOAuthMetadata\`. Pre-existing surface, not a regression from this change.
- #1176 — per-response (not per-call) ALS scope so MCP's long-lived side-channel \`GET\` and SSE-mode \`tools/call\` responses can coexist with capped one-shots in the same process.
- #1177 — \`TaskExecutor\` secondary call sites (\`listTasksForAgent\` / \`getTaskStatus\` / \`continueTask\`) accept constructor cap but not per-call override; \`declaredContentLength\` → \`contentLengthHeader\` rename; MCP integration test parallel to the A2A discovery test.

Each follow-up issue cites which reviewer surfaced it, so adopters can grep history if they hit a related sharp edge.

bokelley:
Acknowledged. The review-pass summary is noted — fixes from `d852f00a` (gzip-bomb defense, URL credential redaction, factory cap wiring, expanded test suite) look solid, and the follow-up scope is cleanly isolated to #1175, #1176, #1177. No outstanding action items from triage's side; this one is in human-reviewer hands.

---

bokelley:
Noted — agreed, nothing left on triage's side. PR is in human-reviewer hands.

---

bokelley:
Triage note from issue #1176: two follow-ups landed that touch the same `responseSizeLimit.ts` surface this PR introduces — (1) skip the size-limit wrapper when `Content-Type: text/event-stream` is detected in `enforceSizeLimit`, and (2) narrow the `responseSizeLimitStorage` ALS slot to per-response scope (around each `client.callTool`) rather than the outer `withSpan`. Both are non-breaking and the combined diff is small. Worth considering folding before merge to avoid a chained PR landing immediately on top of new code.

---

bokelley:
## CI assessment

\`Test & Build\` reports 2 failing tests:
- \`storyboard-a2a-async-submitted-yaml\` — \`runs the create_media_buy phase against a regressed adapter\`
- \`a2a-context-retention\` — \`fails against a regressed adapter that stamps fresh contextId per send\`

**Both pre-existing on \`main\`.** Identical failures on main runs 25212094217 (chore(scripts): commit v2.5 drift), 25212080923 (default response validation to warn), 25210424003 (decideRetry skill example), 25210219248 (BuyerRetryPolicy), 25209907923 (human-escalate callout). Verified locally by checking out \`origin/main\`'s \`src/lib/protocols/{mcp,a2a,index}.ts\` and re-running the suite — same 2 failures.

Not regressions from this PR. Admin merging.

The 21 unit tests added in this PR (\`test/unit/response-size-limit.test.js\` + \`test/unit/a2a-card-size-limit.test.js\`) all pass.

bokelley:
Acknowledged — confirmed the two failures (`storyboard-a2a-async-submitted-yaml`, `a2a-context-retention`) are pre-existing on `main` and unrelated to this diff. The 21 new unit tests all pass. Admin merge looks correct.

---

bokelley:
Issue #1177 captures three hygiene items from this PR's review that are small enough to fold before merge:

1. **`TaskExecutor` secondary call sites** — `listTasksForAgent`, `getTaskStatus`, `continueTask` only forward `this.config.transport`; `executeTask` already accepts a per-call override. ~6 lines to thread the same pattern through the other three.
2. **Rename `ResponseTooLargeError.declaredContentLength` → `contentLengthHeader`** — makes the header-derived provenance explicit. Since `ResponseTooLargeError` is brand-new in this PR, the rename is a pre-release fix rather than a breaking change for existing users.
3. **MCP integration test for the size cap** — `mcp.ts` lines 319 and 660 wire `wrapFetchWithSizeLimit` but no test proves the cap fires end-to-end through `ProtocolClient.callTool`. A `node:http` test server on the same shape as `test/unit/a2a-card-size-limit.test.js` would close that gap.

All three overlap this PR's diff directly. Suggesting folding here so they ship together rather than as an immediate follow-up.

(Filed as #1177 for tracking if you'd rather defer post-merge.)

---

bokelley:
Both follow-ups from #1176 are now folded in (commit `4d7ccb4`):

1. **SSE pass-through** — `enforceSizeLimit` detects `Content-Type: text/event-stream` (including `; charset=…` variants) and returns the response unchanged. The MCP transport's long-lived `GET /mcp` side channel no longer accumulates toward the cap.

2. **Per-response ALS scope** — `ProtocolClient.callTool` now enters `responseSizeLimitStorage` around each individual protocol call (`callMCPToolWithTasks`, `callMCPToolWithOAuth`, `callA2ATool`) rather than the outer `withSpan`. Each retry gets a fresh counter; long-lived sessions and capped one-shots can coexist in the same process.

The `TransportOptions.maxResponseBytes` TSDoc "leave unset for long-lived sessions" caveat is removed now that both issues are handled structurally. Three new unit tests cover SSE pass-through cases. All 22 tests in the size-limit suite pass.

---

bokelley:
Acknowledged — both structural fixes (SSE pass-through + per-response ALS scope) confirmed folded in via `4d7ccb4`, and the TSDoc caveat is cleaned up. Nothing further needed here.

---

PR #1165, opened 2026-05-01
feat: AccountToolContext + refreshToken (#1145) + peer_substitutes_for (#1144)

bokelley:
## Summary

Two distinct changes, one PR — both surfaced from the 14-adapter v6 port (`scope3data/agentic-adapters` PRs #207–#221).

- **#1145 Gap 1**: `AccountToolContext<TCtxMeta>` threads the resolved Account into `getAccountFinancials` so adopters fronting an upstream platform read tokens / upstream IDs from `ctx.account.ctx_metadata` without re-resolving. Resolves 7-adapter pain point.
- **#1145 Gap 2**: Optional `AccountStore.refreshToken` hook for mid-request token refresh on `AUTH_REQUIRED`. Reactive only, retry-once, in-flight scope. Resolves UA's v5-OAuth-provider workaround.
- **#1144**: Storyboard runner `peer_substitutes_for` opt-in — declared substitution rather than implicit "phase membership = substitutability." Contract hygiene + diagnostic improvement.

### Honest framing on #1144

This is **not** a fix for the 1/9/0 adopter cluster. Citrusad/amazon/criteo/google show 1/9/0 because (a) they're often running the wrong storyboard for their specialism shape and (b) the cascade is over-broad across independent phases — see follow-up at adcp-client#1161. The `peer_substitutes_for` change tightens the implicit substitutability contract and improves cascade-detail messages, but the headline 1/9/0 → 9/10 lift requires phase-independence (#1161) and storyboard-selection alignment.

## Companion upstream issues

- `adcontextprotocol/adcp#3734` — storyboard schema field + `sales-social/index.yaml` edit (required before any storyboard exercises the new path).
- `adcontextprotocol/adcp#3735` — bundled wire enum gaps (`creative-builder`, `audio` channel, `cpv`/`vcpm`).
- `adcp-client#1161` — phase-independence cascade scoping (the actual 1/9/0 lever).
- Status comment on `adcp-client#1145` with design decisions for Gaps 1, 2, 4.
- Status comment on `adcp-client#1144` referencing the SDK contract.

## Test plan

- [x] All 15 storyboard cascade tests pass (10 existing + 5 new for `peer_substitutes_for`).
- [x] All 101 from-platform tests pass (98 existing + 3 new for refresh hook + ctx.account threading).
- [x] Wider sweep: 282/282 in `test/server-decisioning*.test.js` + `test/lib/storyboard-cascade*.test.js`.
- [x] Typecheck clean.
[...]

PR #1171, opened 2026-05-01
fix(conformance): capability-aware cascade — missing_tool over prerequisite_failed for un-advertised steps

### LINKED ISSUE #1169: Storyboard runner: cascade should be capability-aware — `prerequisite_failed` over-applied to phases independently `not_applicable`
## Background

Filed alongside #1144 (peer-substitution). Different mechanism, different fix surface, distinct motivating data — so filing separately.

When an upstream phase trips, the runner cascade-skips every downstream stateful step with `skip_reason: 'prerequisite_failed'`. The cascade doesn't check whether each downstream step is *independently* `not_applicable` or `missing_tool` based on the agent's declared specialism set — it just stamps `prerequisite_failed` on everything downstream. That's the wrong diagnostic when the downstream phase was never going to apply regardless of the upstream tripping.

## Concrete data — social platforms running `sales_social`

Workspace: scope3data/agentic-adapters, after #222 lands SDK 6.1.0 + setup_snippet fix. Social-platform storyboard runs (per-platform `<adapter>-storyboard.test.ts` against mocked clients, `sandboxMode: 'off'`). Each adapter declares the right specialism set for `sales_social`, so the storyboard IS applicable — they're not edge cases.

| Adapter | passed/skipped/failed | Specialisms claimed |
| --- | --- | --- |
| snap | 9/1/0 ✅ | sales-social, audience-sync |
| linkedin | 3/6/1 | sales-social, audience-sync |
| pinterest | 2/7/1 | sales-social, audience-sync |
| spotify | 4/6/0 | sales-social, audience-sync |

The `failed: 1` cases (linkedin, pinterest) are real adapter bugs that 6.1.0 unmasked when previously-cascaded steps started running — orthogonal to this issue.

The `skipped` counts are where the cascade overreach lives. Concretely:

**linkedin** declares `audience-sync` specialism but DOES NOT support `sync_creatives` (returns `NOT_SUPPORTED` — creatives go inline through `create_media_buy.packages[].creatives[]`). The runner skips `creative_push/sync_creatives` with `prerequisite_failed`. But it shouldn't — sync_creatives is independently `missing_tool` on linkedin's tool advertisement. The cascade is rolling up "tool not advertised" under "previous phase tripped" and producing a misleading diagnostic.

**spotify** declares the right specialisms but doesn't expose `sync_catalogs`, `log_event`, or `get_account_financials` (they're optional methods on `SalesPlatform` and spotify omits them). Those steps get `prerequisite_failed` instead of `missing_tool` / `not_applicable`.

**snap (the F6-fixed case)** still has 1 skip — `account_setup/sync_accounts` itself, correctly marked `not_applicable`. The downstream phases pass because snap genuinely supports them. So snap is what the runner should look like for *all* applicable agents: cascade only fires when there's a real prerequisite dependency.

## Hypothesis on root cause

When a phase trips, the runner enters cascade mode and stamps every subsequent stateful step with `prerequisite_failed` — without first checking the step's intrinsic skip-eligibility against the agent's advertised tools / declared specialisms.

## Proposed fix shape

Reorder the evaluation: evaluate each step's intrinsic skip-eligibility BEFORE applying upstream-failure cascade. Pseudo-code:

```ts
when phase trips:
  for each downstream stateful step:
    if step.specialism not in agent.declared_specialisms:
      mark as 'not_applicable'        // independently inapplicable
    else if step.tool not in agent.advertised_tools:
      mark as 'missing_tool'           // independently missing
    else:
      mark as 'prerequisite_failed'    // genuine cascade
```

Equivalent restatement: a step's skip reason should reflect why *that specific step* didn't run — not why the upstream phase tripped. `prerequisite_failed` is the right signal only when the step would have run if upstream had succeeded.

## Operational impact

For agents with reduced specialism surfaces (most non-canonical adopters), the storyboard report becomes honest:
- `prerequisite_failed` count → only what genuinely depends on upstream state
- `missing_tool` count → tools the agent doesn't advertise
- `not_applicable` count → specialisms the agent didn't claim

This lets adopters distinguish "this agent has a real setup bug" from "this agent doesn't claim this surface, by design." Today both produce identical-looking 9-skip cascades.

## Relationship to #1144

#1144 (peer-substitution) handles: "step is `not_applicable` BUT a peer in the same phase establishes equivalent state." Substitution-as-state-establishment.

This issue handles: "step is independently `not_applicable` because the agent doesn't claim the relevant specialism/tool." Capability-as-skip-reason.

They're orthogonal and stack: the substitution fix waives the cascade for snap's phase shape; this fix waives the cascade for linkedin/spotify-style reduced surfaces. Both can ship independently.

bokelley:
Closes #1169

## Summary

When `statefulFailed` fires, the runner previously stamped every downstream stateful step with `prerequisite_failed` (`passed: false`) without checking whether the step's tool was even in the agent's advertised tool set. For agents with reduced specialism surfaces (LinkedIn, Spotify in the issue's concrete data), this produced misleading cascades: `sync_creatives`, `sync_catalogs`, `log_event`, and `get_account_financials` all showed `prerequisite_failed` when the honest diagnostic was `missing_tool` — the agent never claimed those surfaces.

The fix: before applying the cascade reason, check intrinsic skip-eligibility via `resolveTaskName(step, options)` against `options.agentTools`. If the task is not advertised, emit `missing_tool / passed: true`. If it is advertised (state genuinely never materialized), emit `prerequisite_failed / passed: false` as before.

Uses `resolveTaskName()` (not `step.task`) so `$test_kit.*` steps are checked against the concrete resolved name — the template string would never be in `agentTools` and would incorrectly classify every test-kit step as `missing_tool`.

Also sets `priorStepResults.set(step.id, result)` in the new arm for symmetry with the `phaseAbsent` cascade arm (defensive: future `priorStepResults` consumers see cascade-skipped `missing_tool` steps in the map).

This fix is orthogonal to #1144 / PR #1146 (sole-stateful-step `not_applicable` → no cascade) and PR #1165 (`peer_substitutes_for` + `AccountToolContext`). Both address different code paths in the cascade; both can ship independently.

**Note on `overall_passed` when all stateful steps are `missing_tool`:** The protocol expert flagged that an agent advertising no applicable storyboard tools could get `overall_passed: true` after this fix. This is a pre-existing design property of the runner — the non-cascade `executeStep` path already emits `missing_tool / passed: true` for unadvertised steps — and is gated by the empty-agentTools abort guard at runner init. No new exposure is introduced here; the `overall_passed: true` case is the honest score for an agent that claims a specialism but doesn't implement any of the storyboard's tools (same behavior as the storyboard-level `required_tools` check).

## What was tested

- `npx tsc --project tsconfig.lib.json` — pre-existing TS2688/TS5107 only (confirmed on `main` before this branch)
- `node --test test/lib/storyboard-cascade-skip-on-skip.test.js` — **13/13 pass** (10 pre-existing + 3 new: primary reclassification case, fall-through guard for advertised tasks, `$test_kit.*` regression guard)
- Full `node --test test/lib/storyboard-*.test.js` sweep — same 10 suite failures as `main` baseline (numbers shifted +1 by the new describe block); no regressions introduced

## Pre-PR review

- **code-reviewer**: approved — no blockers; fixed `priorStepResults.set` gap identified as an issue; nit on `task: resolvedTask` vs `task: step.task` inconsistency in the `prerequisite_failed` fall-through noted in PR body
- **ad-tech-protocol-expert**: approved — `missing_tool / passed: true` is semantically correct per AdCP spec and consistent with `SKIP_DETAILS`, `RunnerSkipReason` enum, and `isHardMissingStateSkipReason`; `patch` changeset classification is correct

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01Ya6Gy5FekjG4pxxyvGE2yE

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ya6Gy5FekjG4pxxyvGE2yE)_


bokelley:
## Admin merge rationale

Same `Test & Build` failures as every recent main run: `storyboard adcp#3083: create_media_buy_async_submitted (A2A end-to-end)` and `a2a_context_continuity (runner integration, #962)`. Both pre-existing, tracked at #1178 (broken since the test files's introduction at `6124deb2` and `61801503`; verified by re-running at those commits). All other CI lanes pass. Admin-merging.

bokelley:
Acknowledged — pre-existing CI failures noted and tracked at #1178. No action needed here.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_01GDuc3yuRhgFqktZdqvzBmk_

---

PR #1181, opened 2026-05-01
fix(validation): skip missing-schema bundle in warn mode instead of throwing

### LINKED ISSUE #1178: Pre-existing failure on main: A2A storyboard regression-adapter tests fail on every PR
**Two storyboard tests fail on every CI run against main**, including the post-merge run for every recent PR. The failures are not introduced by the PRs they appear on — verified by checking out \`origin/main\` directly and running the test files. Recent main runs all show the same failures: 25212094217, 25212080923, 25210424003, 25210219248, 25209907923.

### Failing tests

\`test/storyboard-a2a-async-submitted-yaml.test.js\`:
\`\`\`
not ok 3 - runs the create_media_buy phase against a regressed adapter (a2a_submitted_artifact fails specifically)
  error: 'a2a_submitted_artifact ran'
  expected: true
  operator: '=='
\`\`\`

\`test/lib/a2a-context-retention.test.js\`:
\`\`\`
not ok 2 - fails against a regressed adapter that stamps fresh contextId per send
\`\`\`

### What these tests assert

Both tests deliberately install a regressed adapter and assert that the regression is caught. The current state appears to be: the regressed adapter is no longer caught (the assertion that the adapter "ran" or that the validator "fails" is itself failing), suggesting the surrounding harness changed in a way that bypasses the regression rather than detecting it.

### Why it's been silent

\`Test & Build\` is failing on every main run, but no one's been blocked because:
1. Other CI lanes are passing (CodeQL, GitGuardian, format check, typecheck).
2. PR authors who notice assume their PR caused it — and it didn't, so they admin-merge.
3. The pattern compounds: every subsequent PR sees the same red status and assumes the same.

### Provenance

Tests were added in:
- \`6124deb2 feat(testing): A2A wire-shape capture + a2a_submitted_artifact validation (#904) (#952)\`
- \`61801503 fix(a2a): retain contextId + pendingTaskId across multi-turn calls (#729)\`

Suggested triage: \`git bisect\` between #952 landing (when these tests existed and presumably passed) and the most recent main commit to find the introducing change.

Surfaced while admin-merging #1170 — the rationale is documented in that PR's comment thread.

bokelley:
$(cat <<'EOF'
## Summary

Fixes #1178 — two test files failed on every CI run because `df91b27` changed the client-side validation default from `'off'` to `'warn'`, but the `warn` path didn't handle `SchemaBundleNotFoundError` (the error `resolveSchemaRoot` throws when `npm run sync-schemas` has not been run). The unhandled throw propagated through `SingleAgentClient.createMediaBuy` (and every other typed method), crashing the A2A storyboard tests and context-retention tests before they could even reach the agent.

Three code changes + one test fix:

- **`schema-loader.ts`**: Introduce `SchemaBundleNotFoundError` (typed, `code: 'SCHEMA_BUNDLE_NOT_FOUND'`) so callers can distinguish "schemas not populated" (infrastructure gap, safe to skip in `warn` mode) from `ConfigurationError` (bad version string, always re-throw). These two error classes are mutually exclusive on any given call — `resolveBundleKey` throws `ConfigurationError` before `resolveSchemaRoot` is reached.

- **`client-hooks.ts`**: Catch `SchemaBundleNotFoundError` in `validateOutgoingRequest` and `validateIncomingResponse` when mode is `'warn'` — log a `type: 'warning'` debug entry and skip rather than throwing. In `strict` mode, re-throw so hard-stop callers still get the error they opted into. Callers who need a pre-flight check before entering strict mode can use the existing `hasSchemaBundle(version)` export from `schema-loader.ts` — no need to catch `SchemaBundleNotFoundError` directly.

- **`validations.ts`**: Wrap the `validateResponse` call in `computeStrictVerdict` in a try-catch that returns `undefined` on `SchemaBundleNotFoundError`. The function's existing contract is already "no AJV schema available → return undefined" (via `variant: 'skipped'`); this extends that contract to cover the missing-bundle case.

- **`test/storyboard-a2a-async-submitted-yaml.test.js`**: The regressed-adapter fixture used `supported_protocols: ['media-buy']` (hyphen) instead of `['media_buy']` (underscore). The underscore form is the correct `AdcpProtocol` wire value — the hyphen form caused the `create_media_buy` capability feature gate to skip the step entirely, so the `a2a_submitted_artifact` wire-shape assertion never ran. This was the second reason test 3 was returning `validations: []`.

## Test plan

- [x] `test/storyboard-a2a-async-submitted-yaml.test.js` — all 3 subtests pass (was: tests 2 and 3 failing)
- [x] `test/lib/a2a-context-retention.test.js` — all 10 subtests pass (was: 7 subtests in second describe block failing)
- [x] `npm run ci:quick` passes format, typecheck, and build; test failure count drops from 219 (baseline) to 186 (remaining failures are pre-existing, unrelated to this PR)
- [x] Changeset created (`patch` bump — bug fix, no API additions)

https://claude.ai/code/session_012oTudtWvJZboMNcBNV1EbE
EOF
)

---
_Generated by [Claude Code](https://claude.ai/code/session_012oTudtWvJZboMNcBNV1EbE)_

### DIFF
--- a/.changeset/schema-bundle-warn-skip.md
+++ b/.changeset/schema-bundle-warn-skip.md
@@ -1,0 +1,17 @@
+---
+"@adcp/sdk": patch
+---
+
+Fix warn-mode validation crashing when schema bundle is not populated
+
+Commit `df91b27` changed the client-side validation default from `off` to `warn`, but the `warn` path in `validateOutgoingRequest` and `validateIncomingResponse` did not handle the `SchemaBundleNotFoundError` that `resolveSchemaRoot` throws when `npm run sync-schemas` has not been run. The unhandled error propagated through `SingleAgentClient.createMediaBuy` (and every other typed method), causing storyboard A2A tests and context-retention tests to fail on every CI run.
+
+Three fixes:
+
+1. `schema-loader.ts`: Introduce `SchemaBundleNotFoundError` (typed, `code: 'SCHEMA_BUNDLE_NOT_FOUND'`) so callers can distinguish "schemas not populated" (infrastructure gap, safe to skip in `warn` mode) from `ConfigurationError` (bad version string, always re-throw).
+
+2. `client-hooks.ts`: Catch `SchemaBundleNotFoundError` in `validateOutgoingRequest` and `validateIncomingResponse` when mode is `warn` — log a `type: 'warning'` debug entry and skip validation rather than throwing. In `strict` mode, re-throw so hard-stop callers still get the error they opted into.
+
+3. `validations.ts`: Wrap the `validateResponse` call in `computeStrictVerdict` in a try-catch that returns `undefined` on `SchemaBundleNotFoundError` — the function's contract is already "return undefined when no AJV schema is available."
+
+Test fix: the regressed-adapter fixture in `storyboard-a2a-async-submitted-yaml.test.js` used `supported_protocols: ['media-buy']` (hyphen) instead of `['media_buy']` (underscore), causing the capability feature gate to skip the `create_media_buy` step before the wire-shape assertion could run.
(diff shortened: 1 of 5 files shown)


bokelley:

bokelley:
Acknowledged — noted that #1188 resolved the underlying regression and this branch is now stale. No further action needed here.

---

PR #1188, opened 2026-05-01
feat(storyboard): per-phase cascade scoping via phase.depends_on (#1161)

### LINKED ISSUE #1161: Storyboard runner: cascade is over-broad — phases default to dependent on every prior phase
## Background

`statefulFailed` lives at storyboard scope (one bool per run). Once a stateful step in any phase fails or skips for a missing-state reason, every downstream stateful step in every downstream phase is cascade-skipped with `prerequisite_failed`. This is correct for cross-phase consumption patterns (e.g., `sync_governance` in phase 1 → `activate_signal_denied` in phase 3) but wrong for phases that are functionally independent.

Concrete example surfaced during the 14-adapter port (#1144 follow-up): citrusad implements only `SalesPlatform` (no `AudiencePlatform`, no `SignalsPlatform`). When `account_setup` cascades, every downstream phase including `audience_sync` and `creative_push` cascade-skips — but `audience_sync` has nothing to do with account state, and on a citrusad-shaped agent it would have legitimately skipped `not_applicable` anyway because the platform doesn't sync audiences.

The cascade conflates "this phase needed account state to produce a real test" with "every subsequent phase depends on every prior phase by transitivity." Substitution support (#1144) fixes the "peer establishes state" case; it doesn't fix the "this phase was always independent" case.

## Pattern across the port

| Adapter | passed/skipped/failed against `sales_social` | Cause |
|---|---|---|
| snap | 9/1/0 | sync_accounts not_applicable + list_accounts substitute → no cascade |
| citrusad | 1/9/0 | account_setup cascade propagates to audience_sync (which is independent) + creative_push + financials |
| amazon | 1/9/0 | same |
| criteo | 1/9/0 | same |
| google | 1/9/0 | same |

The 1/9/0 reports tell adopters their adapter is fundamentally broken when in fact most downstream phases were never going to run cleanly anyway (no AudiencePlatform / no SignalsPlatform), and the ones that COULD run (creative_push, event_setup) are skipped only because account_setup tripped the storyboard-scope cascade.

## Possible approaches

1. **Per-phase cascade scoping with explicit deps.** New `phase.depends_on: [phase_id]` field. Phases default to *independent*; a phase that genuinely consumes state from a prior phase declares it explicitly. The runner only propagates cascade across declared dependencies. Most explicit, requires storyboard authoring.
2. **Per-stateful-step deps.** Each stateful step declares which prior step(s) it consumes state from (`consumes_state_from: <step_id>`). More granular but more authoring overhead.
3. **Heuristic: cascade only within a phase by default; explicit `cascade_to: [phase_id]` to extend.** Inverts the current default.

Option (1) feels right — phases are the natural unit of "thing that should run as a coherent block" and declaring deps once per phase is low-cost relative to the diagnostic improvement.

## Coupling with #1144

Substitution declaration (`peer_substitutes_for`) is shipping in #1144. Phase-independence is the *next* layer of cascade tightening. They compose: `peer_substitutes_for` waives the cascade within a phase; phase-deps scope which phases the cascade can travel across.

## Storyboard YAML touches

If we go with (1), upstream (`adcontextprotocol/adcp`) needs:
- Schema: add `phase.depends_on: [phase_id]` (optional).
- Audit existing storyboards: most will be `depends_on: []` (phases are independent — `capability_discovery`, `account_setup`, `audience_sync`, etc.). The genuine dep cases (`signal_marketplace/governance_denied` setup → consume) get explicit declarations.

cc #1144 (substitution) — adjacent fix that doesn't subsume this.

bokelley:
## Summary

Closes adcontextprotocol/adcp-client#1161 (the actual lever for the citrusad-shape 1/9/0 adapter reports surfaced during the 14-adapter v6 port).

Stateful cascade now scoped per-phase rather than storyboard-wide. Phases declare which prior phases they actually depend on for state via the new `depends_on?: string[]` field; only those phases tripping their cascade gates the current phase's stateful steps. Independent phases (`depends_on: []`) run normally even if other phases tripped.

## Why

Today's runner trips one storyboard-scope `statefulFailed` boolean. Once any stateful step in any phase fails or skips for a missing-state reason, every downstream stateful step in every downstream phase cascade-skips with `prerequisite_failed`. Correct for the F6 round-2 cross-phase pattern (`signal_marketplace/governance_denied`), wrong for phases that are functionally independent.

Concrete example from the 14-adapter port: citrusad implements only `SalesPlatform` (no `AudiencePlatform`, no `SignalsPlatform`). When `account_setup` cascades, every downstream phase including `audience_sync` and `creative_push` cascade-skips — but `audience_sync` doesn't actually depend on account state, and on a citrusad-shape agent it would have legitimately skipped `not_applicable` anyway because the adapter doesn't implement audiences. The cascade conflates "this phase needed account state" with "every subsequent phase depends on every prior phase by transitivity."

## Field shape

`phase.depends_on?: string[]` on `StoryboardPhase`. Three semantic modes:

| Declaration | Meaning |
|---|---|
| Field absent (default) | Implicit "depends on all prior phases" — backward-compat with storyboard-scope behavior |
| `depends_on: []` | Independent phase. Runs even if every prior phase tripped. |
| `depends_on: ['phase_id', ...]` | Targeted dependencies only. |

Within-phase cascade is preserved unconditionally — stateful steps later in a phase still cascade when an earlier stateful step in the same phase trips. Intra-phase state dependency is authoring intent that `depends_on` (which scopes inter-phase cascade) doesn't override.

## Implementation

Replaces storyboard-scope `statefulFailed: boolean` + `statefulSkipTrigger` pair with `phaseStatefulCascades: Map<phase_id, CascadeTrigger | null>`. Cascade-skip check at each stateful step consults `phase.depends_on` (default: all prior phases) plus the current phase's own within-phase state. Cascade-detail messages now reference the trigger from the specific gating phase rather than the first-trip-anywhere.

## Loader validation

Forward references, self-references, and unknown phase IDs in `depends_on` fail loud at parse time. Empty list (`[]`) is legal — it's the "independent phase" declaration.

## Companion spec change

Will file at `adcontextprotocol/adcp` to add the field to the storyboard schema and audit each specialism storyboard for which phases are functionally independent. For `sales_social` specifically: `audience_sync`, `creative_push`, `catalog_driven_dynamic_ads`, `event_setup`, `event_logging`, `financials` are arguably independent of `account_setup` for explicit-mode platforms (each carries its own account ref in the request body; resolution doesn't require `sync_accounts` having run). Once those phases gain `depends_on: []`, citrusad-shape adapters move from 1/9/0 to a more honest report.

## Test plan

- [x] All 19 existing storyboard cascade tests pass (legacy "all prior phases" default preserved including F6 round-2 cross-phase cascade).
- [x] 7 new tests for `phase.depends_on`:
  - `depends_on: []` runs phase even when prior tripped
  - `depends_on: ['specific_phase']` cascades only when that named phase tripped, not other tripped phases
  - Default (undefined) preserves all-prior cascade
  - Loader rejects forward references at parse time
  - Loader rejects self-references at parse time
  - Loader rejects unknown phase IDs at parse time
  - Within-phase cascade still fires regardless of `depends_on`
- [x] Wider sweep: 2708/2714 passing on `test/lib/storyboard-*.test.js` (1 pre-existing flake on `context_value_rejected hints via real MCP transport` — unrelated, not introduced by this PR).
- [x] Typecheck clean.
- [x] Prettier check clean.


bokelley:
## Expert review pass — code-reviewer + protocol-expert

Fired both in parallel. Skipped security-reviewer (no auth surface).

### Convergence: zero code blockers

Both reviewers verdict: ready to push. All flagged items are spec-issue clarifications (not SDK code) or sentinel-test additions.

### Code-reviewer findings (all addressed)

State-machine refactor walked through all four tricky orderings — within-phase failure-after-skip, transitive A→B→C, merge integration with #1171's capability-aware fast-path, `priorPhaseIds` accumulator across early-continue paths. Verdict: behavior matches spec. Nits filed:

- ✅ **Three-phase transitive sentinel** added: A trips, B `depends_on: []` runs clean, C `depends_on: ['B']` runs clean despite A. Locks in that `depends_on` is direct, not transitive.
- ✅ **`peer_substitutes_for` × `depends_on` composition sentinel** added: phase declaring both still rescues within-phase regardless of the inter-phase dep. Confirms #1144 (intra-phase) and #1161 (inter-phase) occupy independent layers.
- ✅ **Loader non-array rejection test** added: `depends_on: 'string'` now hits the parse-time error path.

Nits not addressed (cosmetic, follow-up):
- `cascadeForPhase` cast at runner.ts:1117 could disappear with discriminated-return narrowing. Cosmetic.
- Edge case where a phase has BOTH deferred not_applicable AND deferred missing_tool with declared substitutes — first resolution wins via `.has(phase.id)` guard, losing the `substitution_chain` diagnostic. Pre-existing pattern, not a regression.
- `depends_on: null` (explicit null vs undefined) currently throws "must be array." YAML authors don't write explicit nulls; not worth a special case.

29 → 32 cascade tests passing.

### Protocol-expert findings (pinned on adcp#3745)

Verdict: sound, with two pinnable spec-side items:

1. **Direction-grammar rule**: `depends_on` is direction-consistent with `contributes_to`, `peer_substitutes_for`, `requires_capability`, etc. — all point from the dependent thing to its dependency. Worth anchoring as an explicit AdCP grammar rule so future fields don't reinvent the convention. Also the deciding context for the open `peer_substitutes_for` direction debate at adcp#3734.

2. **Empty-list discoverability**: `depends_on: []` is the right call (composes with targeted form, matches OpenRTB pattern, industry-canonical for CI/build DAGs), but YAML authors will reach for `independent: true` first. Mitigate with a 3-mode example block in the spec (absent / `[]` / `['phase_id']`).

Both items posted to adcp#3745. SDK ships fine — these are authoring-docs annotations.

Protocol-expert also flagged additional specialism storyboards that should be audited for `depends_on` annotations (signal_marketplace as the contrast retain-default case; sales-retail-media, creative-template, governance-spend-authority). Added to adcp#3745's scope for the per-storyboard YAML pass.

### Ready for admin merge once CI greens

Same admin-merge rationale as #1165 if the pre-existing flakes resurface — but #1178 (the a2a regression-detector tests) is now FIXED in this PR (commit `a00ea973`), so Test & Build should run clean for the first time in weeks.

bokelley:
Thanks for the thorough review pass, @bokelley. All ✅ items are already in the diff, nits are deferred, and spec-side annotations are tracked at adcp#3745. No further action needed from triage — watching for CI green.

---

bokelley:
## Round-2 expert review pass — merge integration

Re-reviewed the post-merge state (commit `dc1499c2`) integrating main's #1146 sole-stateful-step exception with this PR's per-phase cascade scoping.

### Code-reviewer: ready to push

All 8 correctness concerns walked and verified clean against the merged code:

- Sole-stateful-step gate composes with `phaseStatefulCascades` map — phase doesn't get an entry, downstream phases (default deps OR explicit `depends_on`) don't see it as tripped.
- `phaseStatefulStepIds` accumulator at runner.ts:994 is correctly captured eagerly per-phase; the `hadStatefulPeers` check at 1382 asks "did the storyboard *declare* peers" not "did peers run."
- Real-failure-wins-within-phase (`phaseStatefulCascades.set(phase.id, null)`) preempts the deferred `not_applicable` resolution via the `.has(phase.id)` guard. Sole-step exception did NOT undermine the failure-wins precedent.
- Three sentinel tests from `f8a2d28c` (transitive, `peer_substitutes_for × depends_on`, non-array rejection) all pass against merged runner.
- 33/33 cascade tests pass (32 mine + 1 new from #1146).

No new nits.

### Protocol-expert: sound-with-caveats

Verdict: zero SDK changes needed; the merge composes correctly. Spec-side recommendations posted to adcp#3745:

- Document the **trio of cascade-authoring affordances** (#1144 explicit substitution, #1161 explicit dependency, #1146 implicit pathway-disclaimer) as a single "Cascade authoring model" section in the storyboard spec rather than three scattered notes.
- Frame `depends_on` as "scope cascade visibility" not "wait for completion" — adopters who read it as build-system "needs:" can misinterpret the edge case where `depends_on: ['phase_with_sole_step_not_applicable]` runs without waiting.
- Add a "Phases that don't trip cascade" subsection adjacent to the 3-mode `depends_on` example so adopters don't add redundant `depends_on: []` to phases that already qualify for the sole-step exception.
- Naming: "pathway-disclaimer phase" (semantic) for authoring guides, "non-tripping not_applicable" (mechanical) for runner-behavior tables.
- Pin the precedent: *implicit cascade rules require a structural trigger AND a single unambiguous semantic; anything ambiguous gets an explicit YAML field.*

### Ready for admin merge once Test & Build greens

Both reviewers cleared the merge; the runner refactor + #1146 + #1171 + #1161 compose cleanly across the cascade state machine.

bokelley:
Acknowledged — round-2 review summary noted. Both reviewers cleared; waiting on Test & Build green before admin merge. Spec-side recommendations tracked at adcontextprotocol/adcp#3745.

---

bokelley:
Acknowledged. No further action needed from this side — CI gate + admin merge is the right call here.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_01StEWaWLrmtNZahKQdCjcjz_

---

PR #1194, opened 2026-05-01
fix(validation): swallow schema-bundle-not-found in warn mode (#1178 context-continuity leg)

bokelley:
## Problem

`test/storyboard-a2a-context-continuity-integration.test.js` (both tests) was silently swallowing the `a2a_context_continuity` validator because the storyboard runner never saw a `taskResult` — the step always resolved as `{ success: false }` with an unhandled error thrown from `validateOutgoingRequest`.

Root cause chain:

1. `validateOutgoingRequest` (warn mode, the default) calls `validateRequest(taskName, params, version)`.
2. `validateRequest` → `resolveSchemaRoot` throws when `schemas/cache/` is unpopulated (CI doesn't run `sync-schemas`).
3. The throw escapes _before_ the `if (mode === 'warn')` guard — warn mode never catches it.
4. The exception propagates up through `SingleAgentClient.performTask` → `dispatch()` → `runStep`, which sets `taskResult = undefined`.
5. The validation gate in `runner.ts` (`if (step.validations?.length && (taskResult || httpResult))`) never opens, so `validations = []` for every step.
6. `continuityCheck` is `undefined` → `assert.ok(continuityCheck, 'continuity validator ran')` fails.

### 1. `src/lib/validation/client-hooks.ts`

Wrap `validateRequest` in a try/catch inside `validateOutgoingRequest`:

- **warn mode** — swallow the error and return `undefined` (skip validation). Schema absence is a dev-env gap, not a protocol error. Same behaviour as `off` for missing bundles, but without crashing the call.
- **strict mode** — re-throw, so misconfigured CI environments surface the problem explicitly.

### 2. `test/storyboard-a2a-context-continuity-integration.test.js`

Fix `supported_protocols: ['media-buy']` → `['media_buy']` in the regressed fixture's `get_adcp_capabilities` response. The hyphenated form (`'media-buy'`) doesn't match the underscore-separated `FeatureName` constant (`'media_buy'`), so `resolveFeature` returns false → `FeatureUnsupportedError` → `skip_reason: 'not_applicable'` → `second_send` cascade-skipped → `validations = []`. This secondary bug would have re-broken the test even after fix #1.

## Verification

```
node --test test/storyboard-a2a-context-continuity-integration.test.js

# tests 2
# pass  2
# fail  0
```

## Relationship to sibling PRs

| PR | Test file fixed |
|----|----------------|
| #1181 | `storyboard-a2a-async-submitted-yaml.test.js` |
| #1183 | `storyboard-a2a-regression-adapter.test.js` |
| **this PR** | `storyboard-a2a-context-continuity-integration.test.js` |

All three failures share the same root cause (schema-bundle throw in warn mode); this PR is the third leg of the fix for #1178.

https://claude.ai/code/session_019BHrpGSqf9tgrhmi75MaWH

---
_Generated by [Claude Code](https://claude.ai/code/session_019BHrpGSqf9tgrhmi75MaWH)_

### DIFF
--- a/.changeset/fix-schema-bundle-warn-mode.md
+++ b/.changeset/fix-schema-bundle-warn-mode.md
@@ -1,0 +1,9 @@
+---
+"@adcp/sdk": patch
+---
+
+fix: `validateOutgoingRequest` in warn mode no longer throws when `schemas/cache/` is unpopulated
+
+Previously, calling any AdCP tool with `validation.requests: 'warn'` (the default) would throw an unhandled `SchemaBundleNotFoundError` if the schema bundle had not been populated via `npm run sync-schemas`. The throw escaped before the warn-mode guard, causing `taskResult` to be `undefined` in the storyboard runner and silently skipping all step validations — including `a2a_context_continuity`.
+
+The fix wraps the `validateRequest` call in a try/catch: in warn mode the error is swallowed and validation is skipped (same behaviour as `off` for missing bundles); in strict mode it re-throws so CI catches misconfigured environments. Closes #1178 (context-continuity integration test leg).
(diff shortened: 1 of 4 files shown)


bokelley:

PR #1198, opened 2026-05-01
fix(server): default account.supported_billing to ['agent'] in v6

### LINKED ISSUE #1186: fix(server): default supported_billing: [] in createAdcpServerFromPlatform (match v5 behavior)
## Problem

\`get_adcp_capabilities\` responses from v6 agents fail schema validation when the platform doesn't explicitly set \`supportedBillings\` in capabilities config. The schema requires \`account.supported_billing\` unconditionally, but the v6 framework spreads the field conditionally:

\`\`\`ts
// src/lib/server/decisioning/runtime/from-platform.ts:767
...(supportedBillings?.length && { supported_billing: [...supportedBillings] }),
\`\`\`

If \`supportedBillings\` is undefined or empty, the framework omits the field entirely. Schema validation then fails with:

\`\`\`
/account/supported_billing: must have required property 'supported_billing'
\`\`\`

The v5 path (\`createAdcpServer\`) is safer — it defaults to \`[]\`:

\`\`\`ts
// src/lib/server/create-adcp-server.ts:3431
supported_billing: capConfig.account.supportedBilling ?? [],
\`\`\`

So this is a **regression in v6 ergonomics**. Adopters wiring up signals/creative/governance/etc. specialisms (which don't have media-buy billing concerns) hit this immediately.

## Reproduction

Surfaced via the matrix v2 mock-server run on PR #1155 / branch \`bokelley/mock-server-signals\`. The signals SKILL.md doesn't mention \`supported_billing\` (no hits in \`grep -i\`) — adopters have no signal that they need to set it. The v6 framework should make this safe by default.

## Cascade impact

When capabilities fail validation, the storyboard runner auto-downgrades the agent to \"v2 synthetic capabilities\" mode, which then asks for v2.5 schemas (not shipped), which cascades into every subsequent step erroring with \"AdCP schema data for version v2.5 not found.\" One missing field → 8 cascade-skipped steps. Filing the runner brittleness separately.

## Fix

Match v5 behavior in \`from-platform.ts:767\`. Always emit the field with a sensible default:

\`\`\`ts
supported_billing: supportedBillings ?? [],
\`\`\`

Drop the conditional spread.

Belt-and-suspenders option: add a runtime invariant test in \`test/lib/server/\` asserting that any \`createAdcpServerFromPlatform\` invocation produces a capabilities response that passes schema validation, even with minimal config. Catches the next regression of this class.

## Related

Real spec-level fix is to make \`supported_billing\` conditional on \`supported_protocols.includes('media_buy')\` upstream. Filed separately at adcontextprotocol/adcp. Until that lands, this SDK fix is the right defensive default.

## Refs

- PR #1155 (matrix v2 + first mock-server)
- Skill silence: \`skills/build-signals-agent/SKILL.md\` — 0 hits for \`supported_billing\`
- v5 reference: \`src/lib/server/create-adcp-server.ts:3431\`
- v6 culprit: \`src/lib/server/decisioning/runtime/from-platform.ts:767\`

🤖 Filed by Claude Code (matrix v2 first-run signal)

bokelley:
Closes #1186.

## Summary

Schema requires \`account.supported_billing\` (\`minItems: 1\`) whenever the \`account\` block is emitted in \`get_adcp_capabilities\`. The v6 framework was spreading the field conditionally:

\`\`\`ts
// src/lib/server/decisioning/runtime/from-platform.ts:767 (before)
...(supportedBillings?.length && { supported_billing: [...supportedBillings] }),
\`\`\`

When adopters didn't declare \`supportedBillings\`, the field was dropped entirely and the capabilities response failed schema validation. The storyboard runner's v2-fallback heuristic then compounded that single missing field into \"AdCP schema data for version v2.5 not found\" errors across every subsequent step (filed separately as #1189 — that brittleness should be tightened independently).

## Fix

Always emit \`supported_billing\` when the account block is emitted, defaulting to \`['agent']\`:

\`\`\`ts
// after
...(hasAccountProjection && {
  supported_billing: supportedBillings?.length ? [...supportedBillings] : ['agent'],
}),
\`\`\`

\`'agent'\` (agent consolidates billing) matches the platform-interface contract documented at \`src/lib/server/decisioning/capabilities.ts:130\` (\"Defaults to \`['agent']\` when omitted\") and is the least-surprising default for non-media-buy specialisms.

## Why \`['agent']\` and not \`[]\`

The schema requires \`minItems: 1\`, so \`[]\` would also fail validation. v5 had \`?? []\` which was equally broken on schema-strict callers — this fix is what v5 should have done.

## Backwards compatibility

- Adopters who declared \`supportedBillings\` explicitly are unaffected — projection still flows their declared values.
- Adopters who omitted the field were the only ones exposed to the regression; they now get the spec-correct \`['agent']\` default.
- No new types, no signature changes — pure runtime behavior fix.

## Test plan

- [x] \`npm run build\` clean
- [x] Two regression tests added in \`test/server-decisioning-capability-projections.test.js\`:
  - explicit \`accounts.resolution: 'explicit'\` without \`supportedBillings\` → emits \`['agent']\`
  - explicit \`requireOperatorAuth: true\` without \`supportedBillings\` → emits \`['agent']\`
- [x] All 14 capability-projection tests pass
- [x] Existing supportedBillings projection tests still pass (declared values flow unchanged)

## Refs

- #1186 — issue this fixes
- #1185 — matrix v2 PR that empirically surfaced this
- #1189 — companion runner brittleness (v2-fallback should be tightened so a single validation failure doesn't downgrade a v3 agent)
- adcontextprotocol/adcp#3746 — upstream spec fix (make field conditional on \`media_buy\` protocol)

### DIFF
--- a/.changeset/fix-supported-billing-default.md
+++ b/.changeset/fix-supported-billing-default.md
@@ -1,0 +1,17 @@
+---
+'@adcp/sdk': patch
+---
+
+fix(server): default `account.supported_billing` to `['agent']` in `createAdcpServerFromPlatform`
+
+When the v6 framework emits the `account` block in `get_adcp_capabilities` (because the platform declares `requireOperatorAuth: true` or `accounts.resolution: 'explicit'`), the schema requires `account.supported_billing` to be present with `minItems: 1`. The framework was spreading the field conditionally (`...(supportedBillings?.length && { supported_billing: [...] })`), which dropped the field entirely when adopters didn't declare `supportedBillings`. The capabilities response then failed schema validation.
+
+Worse, downstream tooling (storyboard runner) auto-downgrades agents to "v2 synthetic capabilities" when capabilities probe fails, which cascades into every subsequent step erroring with "AdCP schema data for version v2.5 not found." One missing field collapsed entire storyboard runs.
+
+Fix: when emitting the `account` block, always include `supported_billing`, defaulting to `['agent']` when adopters don't declare. `'agent'` (agent consolidates billing) matches the platform-interface contract documented at `src/lib/server/decisioning/capabilities.ts:130` ("Defaults to `['agent']` when omitted") and is the least-surprising default for non-media-buy specialisms (signals/creative/governance/etc.).
+
+Surfaced empirically by the matrix v2 mock-server run on adcontextprotocol/adcp-client#1185. Closes adcontextprotocol/adcp-client#1186.
+
+Adopters declaring `supportedBillings: ['operator', 'agent']` (or any non-empty subset) are unaffected — the projection still flows their declared values. Only adopters who omitted the field were exposed to the schema-validation regression; those will now pass with the spec-correct `['agent']` default.
+
+Related upstream issue: adcontextprotocol/adcp#3746 (make `supported_billing` conditional on `supported_protocols.includes('media_buy')` at the schema level — the principled fix). This SDK change is the defensive default until the spec change lands.
--- a/src/lib/server/decisioning/runtime/from-platform.ts
+++ b/src/lib/server/decisioning/runtime/from-platform.ts
@@ -762,9 +762,18 @@ export function createAdcpServerFromPlatform<P extends DecisioningPlatform<any, 
   const requireOperatorAuth = platform.capabilities.requireOperatorAuth ?? platform.accounts.resolution === 'explicit';
   const supportedBillings = platform.capabilities.supportedBillings;
   const hasAccountProjection = requireOperatorAuth === true || (supportedBillings?.length ?? 0) > 0;
+  // Schema requires `supported_billing` (minItems: 1) whenever the account
+  // block is emitted. Default to ['agent'] when adopters don't declare —
+  // matches the documented platform interface default at
+  // `capabilities.ts:130` ("Defaults to ['agent'] when omitted"). v6 was
+  // dropping the field on undefined which failed schema validation; v5's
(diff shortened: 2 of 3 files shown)


bokelley:
Triage note: issue #1199 (widen `DecisioningCapabilities.supportedBillings` to include `'advertiser'`, backed by the three-value `billing-party.json` enum) was surfaced by this PR's protocol review and shares `src/lib/server/decisioning/` file scope with your diff. It's a one-line union change + doc comment + type-check fixture update — small enough to fold before merge and avoids a second PR on the same type surface.

---

bokelley:
Folded #1199 into this branch per @bokelley's suggestion. Commit `2d8598c`.

**Changes added:**
- `src/lib/server/decisioning/capabilities.ts:132` — widened `supportedBillings` from `'operator' | 'agent'` to `'operator' | 'agent' | 'advertiser'` to match the three-value `billing-party.json` enum; updated doc comment to describe the `'advertiser'` model.
- `src/lib/server/decisioning/decisioning.type-checks.ts` — added `_capabilities_supported_billings_advertiser` and `_capabilities_supported_billings_all_three` compile-time fixtures; updated the `@ts-expect-error` comment to name all three valid values.
- `.changeset/fix-supported-billing-default.md` — changeset updated to cover both fixes under a single patch entry; adds `Closes #1199`.

All 14 capability-projection tests pass. Pure additive type widening — no breaking change for existing adopters.

---

PR #1238, opened 2026-05-01
fix(client): getAgentInfo() uses SSE fallback for SSE-only MCP servers

### LINKED ISSUE #1233: getAgentInfo() lacks SSE fallback that every other code path has
`SingleAgentClient.getAgentInfo()` opens its `listTools()` connection via `connectMCP()` (`src/lib/protocols/mcp.ts:611`). Every other production path (`callMCPTool`, `mcp-tasks.ts`) goes through `withCachedConnection` → `connectMCPWithFallback`, which:

1. Tries StreamableHTTP
2. Retries StreamableHTTP once on `StreamableHTTPError` (covers stale sessions per SDK #1852)
3. Falls back to `SSEClientTransport` for any other non-401 failure

`getAgentInfo()` does none of this — first error is fatal. Discovery (`discoverMCPEndpoint`) does use the fallback path, so the URL it picks may have been validated via SSE; the immediately-following `getAgentInfo()` connection then re-tries StreamableHTTP-only at the same URL and fails.

### Repro

SSE-only FastMCP server at a public address: `discoverMCPEndpoint` succeeds, `getAgentInfo()` fails. Vanilla StreamableHTTP FastMCP works fine because StreamableHTTP just succeeds — the gap only shows when fallback would matter.

### Fix shape

Either:
- Extend `connectMCPWithFallback` to accept an optional `authProvider`, then have `getAgentInfo()` use it (and delete the OAuth-aware-but-fallback-less branch of `connectMCP` if no other caller needs it), **or**
- Have `connectMCP` itself implement the same fallback as `connectMCPWithFallback`. OAuth flows still work — both `StreamableHTTPClientTransport` and `SSEClientTransport` accept an `authProvider`.

First option is cleaner. Add a regression test that uses an SSE-only server and verifies `getAgentInfo()` succeeds.

Discovered while investigating #1231; not the cause of that issue but a real asymmetry surfaced along the way.

bokelley:
Closes #1233

## Summary

`SingleAgentClient.getAgentInfo()` was calling `connectMCP()` directly — StreamableHTTP only, first error fatal. Every other production MCP path (`callMCPTool`, `mcp-tasks.ts`) routes through `connectMCPWithFallback`, which retries StreamableHTTP on session errors and falls back to `SSEClientTransport` for any other non-401 failure. The asymmetry caused `getAgentInfo()` to fail for SSE-only servers (e.g. FastMCP/Python) even when `discoverMCPEndpoint` had just succeeded via SSE at the same URL. It also silently broke the AAO dashboard probe and any compliance run that calls `discoverAgentProfile()`, since both delegate to `getAgentInfo()` as their first step.

**Fix:** `getAgentInfo()` now routes through `connectMCPWithFallback`. `connectMCPWithFallback` (and its impl) gain an optional `authProvider?: OAuthClientProvider` param so OAuth tokens work through the fallback path — `SSEClientTransport` accepts `authProvider` natively. SSE error logging is added so failures on both transports are visible in debug output (was previously silent on SSE failure).

## What was tested

- `npm run typecheck` — clean
- `npm run build:lib` — clean
- New regression test: `test/unit/get-agent-info-sse-fallback.test.js` — 4 tests, all pass
  - StreamableHTTP server with no auth
  - StreamableHTTP server with static bearer token
  - url/protocol field mapping
  - inputSchema → parameters[] mapping
- Baseline comparison: pre-fix test suite had 222 failures; post-fix has 221 (one fewer — likely noise reduction). No new failures introduced.

## Pre-PR review

- **code-reviewer:** approved — no blockers on final diff; `Parameters<>[4]` idiom replaced with direct import before merge; pre-existing `agentHeaders` gap (not merged in `getAgentInfo`) predates this PR and is unchanged.
- **dx-expert:** approved after two blockers fixed — explicit `OAuthClientProvider` type import added; OAuth+SSE-only gap documented in comment.

### Nits (not fixed — surfaced for reviewers)

- `getOrCreateConnection` does not forward the new `authProvider` param — if a future caller routes through the connection cache with OAuth, `authProvider` would be silently dropped. Cache path intentionally headers-only today (mirrors `callMCPToolWithOAuth` which bypasses cache). A comment should be added to `getOrCreateConnection` before the cache is extended.
- `getAgentInfo()` opens a fresh connection each call (bypasses `withCachedConnection`) — expected given it is not a hot path, but a comment would prevent a well-intentioned reviewer from routing it through the cache.
- `debugLogs = []` in the `connectMCPWithFallback` call is a throwaway; transport diagnostics are not surfaced on failure. Pre-existing pattern (matches `discoverMCPEndpoint` at line 659).
- The `agentHeaders` (agent-level custom headers, e.g. `User-Agent`) are not merged into `authHeaders` in `getAgentInfo()`. Pre-existing gap — the original `connectMCP()` call also omitted them. Not a regression.

---

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_011qouP7zRYkWiz8fdnp24jW

---
_Generated by [Claude Code](https://claude.ai/code/session_011qouP7zRYkWiz8fdnp24jW)_

### DIFF
--- a/.changeset/fix-get-agent-info-sse-fallback.md
+++ b/.changeset/fix-get-agent-info-sse-fallback.md
@@ -1,0 +1,17 @@
+---
+"@adcp/sdk": patch
+---
+
+fix(client): getAgentInfo() now uses SSE fallback for SSE-only MCP servers
+
+Previously, `SingleAgentClient.getAgentInfo()` called `connectMCP()` directly,
+which is StreamableHTTP-only. Every other production path (`callMCPTool`,
+`mcp-tasks.ts`) goes through `connectMCPWithFallback`, which retries
+StreamableHTTP on session errors and falls back to SSE for non-401 failures.
+
+The asymmetry caused `getAgentInfo()` to fail for SSE-only servers (e.g.
+FastMCP/Python) even when `discoverMCPEndpoint` succeeded via SSE at the same
+URL. `getAgentInfo()` now routes through `connectMCPWithFallback`, matching
+every other MCP code path. `connectMCPWithFallback` gains an optional
+`authProvider` parameter so OAuth-issued tokens work through the fallback path
+alongside static bearer tokens.
--- a/src/lib/core/SingleAgentClient.ts
+++ b/src/lib/core/SingleAgentClient.ts
@@ -2662,20 +2662,24 @@ export class SingleAgentClient {
       // Discover endpoint if needed
       const agent = await this.ensureEndpointDiscovered();
 
-      // Use the shared connectMCP path so both static bearer AND saved OAuth
-      // tokens work. OAuth takes the refresh-capable authProvider branch.
-      const { connectMCP } = await import('../protocols/mcp');
-      const connectOptions: Parameters<typeof connectMCP>[0] = { agentUrl: agent.agent_uri };
+      // Route through connectMCPWithFallback so SSE-only servers work.
+      // Both static bearer and saved OAuth tokens flow through the
+      // fallback-capable path — SSEClientTransport accepts authProvider too.
+      const { connectMCPWithFallback } = await import('../protocols/mcp');
+      const mcpUrl = new URL(agent.agent_uri);
+      let authProvider: Parameters<typeof connectMCPWithFallback>[4] = undefined;
+      let authHeaders: Record<string, string> = {};
+
       if (this.normalizedAgent.oauth_tokens) {
         const { createNonInteractiveOAuthProvider } = await import('../auth/oauth');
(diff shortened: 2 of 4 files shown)


bokelley:
Closing as superseded — issue #1233 closed by author. Related discovery surface covered by 6.7's `discover-mcp-endpoint-hint` + `widen-mcp-retry-predicate` plus the storyboard runner's hard-failure on discovery error (5.x SSE-fallback fix). If the SSE-only-server path still doesn't work for a specific FastMCP variant, please file a fresh issue with a reproducer.

Issue #1231, opened 2026-05-01
AdCPClient.getAgentInfo() fails against FastMCP StreamableHTTP agents (no session init)


bokelley:
## Summary

`AdCPClient.getAgentInfo()` does not negotiate the StreamableHTTP session-init handshake that FastMCP requires. Calls against a FastMCP-based MCP agent using StreamableHTTP transport fail with HTTP 400 ("Missing session ID") even though the agent is healthy and serves compliance traffic correctly through other clients.

This blocks AAO's dashboard probe, which delegates to `getAgentInfo()` via `health.ts`. See cross-link to adcontextprotocol/adcp#3066 for the downstream impact and the in-progress `/health`-fallback stopgap (probe-side, no SDK dependency).

## Repro

1. Stand up a FastMCP server with StreamableHTTP transport (e.g. https://adcp.bidmachine.io)
2. Construct an `AdCPClient` with `{ protocol: "mcp", agent_uri: <url> }`
3. Call `client.getAgentInfo()`
4. Observed: `MCP connection failed: ...` — underlying response is 400 Bad Request because no `Mcp-Session-Id` was negotiated before the tool call
5. Expected: client performs the StreamableHTTP `initialize` → `Mcp-Session-Id` → tool call sequence, returns agent info

## Why it matters

FastMCP is a common Python framework for building MCP agents. Any seller using FastMCP + StreamableHTTP fails the AAO listing probe and shows as offline / unknown type / Tools (0) — even though their agent is fully functional. The compliance suite hits the same endpoint successfully, so the gap is specifically in `AdCPClient`'s transport handling.

## Fix shape (suggestions, not prescriptive)

- **Detect transport**: probe response/headers on initial request and switch to StreamableHTTP session flow when the server signals it (e.g. 400 + `Missing session ID` body, or response `Content-Type: text/event-stream`)
- **Or accept an explicit hint**: allow `transport: \"sse\" | \"streamable_http\"` in the per-agent config and let the caller declare it. AAO has a parallel discussion (Option B in adcp#3066) about adding `mcp_transport` to the registered agent type — that hint is only useful if `AdCPClient` can act on it
- Either path should ensure `getAgentInfo()` and the rest of the MCP surface (tool calls, etc.) work uniformly against StreamableHTTP servers

## Cross-links

[...]

PR #1244, opened 2026-05-01
fix(client): route getAgentInfo() through connectMCPWithFallback (#1233, #1234)

### LINKED ISSUE #1233: getAgentInfo() lacks SSE fallback that every other code path has

### LINKED ISSUE #1234: discoverMCPEndpoint error should hint at wrong-path registration
When `discoverMCPEndpoint` exhausts all candidates with no 401 from any, it throws:

```
Failed to discover MCP endpoint. Tried:
  1. https://example.com
  2. https://example.com/
  3. https://example.com/mcp
  4. https://example.com/mcp/
None responded to MCP protocol.
```

In practice, the most common cause is that the registered `agent_uri` is the host root, but the actual MCP endpoint lives at a non-standard path (e.g. `/adcp/mcp`, `/api/mcp/v1`). The current error gives no hint that this is the likely cause, leading users (and dashboards) to report "agent offline" when the agent is fine and just registered with the wrong path.

### Suggested change

When all candidates 404 (no 401, no successful StreamableHTTP/SSE response anywhere), append:

> Hint: this usually means `agent_uri` does not include the MCP endpoint path. The SDK only probes `/`, `/mcp`, and `/mcp/` automatically. If your server exposes MCP at a different path (e.g. `/api/mcp`, `/adcp/mcp`), register that exact path as `agent_uri`.

No behavior change, just a more actionable error. Helps downstream tools (AAO dashboard probe, etc.) surface the actual issue to operators.

Discovered while investigating #1231.

bokelley:
## Summary

- Closes #1233: `SingleAgentClient.getAgentInfo()` now uses `connectMCPWithFallback` instead of `connectMCP`, gaining the same StreamableHTTP-retry / SSE-fallback semantics that every other code path (`callMCPTool`, `mcp-tasks`) already has. `connectMCPWithFallback` learns an optional `authProvider` parameter, forwarded to both StreamableHTTP and SSE transports so OAuth-protected agents still work through the fallback.
- Closes #1234: `discoverMCPEndpoint`'s "Failed to discover MCP endpoint" error message now appends a hint that the most common cause is `agent_uri` pointing at the host root while the MCP endpoint lives at a non-standard path (e.g. `/api/mcp`, `/adcp/mcp`).

## Background

Discovered while investigating #1231 (closed as not-a-bug). Root cause at `adcp.bidmachine.io` was registration mismatch — the MCP endpoint is at `/adcp/mcp` and requires bearer auth, but the AAO probe registered the host root. Two real-but-separate gaps split out of that triage:

1. `getAgentInfo()` was the only normal-path call without fallback — measurably more brittle than every other code path against the same server.
2. The discovery error message doesn't surface the most common operator mistake (wrong-path registration).

## Test plan

- [x] `npm run test:lib` — 5681/5688 pass, 7 skipped, 0 failed
- [x] New regression test (`test/lib/get-agent-info-fallback.test.js`) pins both fixes:
  - StreamableHTTP retry path: server fails the second `initialize` with 400; without the fix `getAgentInfo` propagates the error, with the fix it retries and succeeds (verified by stash/restore round-trip).
  - Discovery hint: server 404s every path; error message contains the wrong-path hint with `/api/mcp` / `/adcp/mcp` examples.

### DIFF
--- a/.changeset/get-agent-info-fallback.md
+++ b/.changeset/get-agent-info-fallback.md
@@ -1,0 +1,15 @@
+---
+"@adcp/sdk": patch
+---
+
+fix(client): getAgentInfo() now uses the same StreamableHTTP-then-SSE fallback as every other code path
+
+`SingleAgentClient.getAgentInfo()` previously routed its post-discovery `listTools` connection through `connectMCP`, which has no retry on `StreamableHTTPError` and no SSE fallback. Every other production path (`callMCPTool`, `mcp-tasks`) goes through `connectMCPWithFallback`, which retries once on transient session errors and falls back to `SSEClientTransport` for non-401 failures.
+
+The asymmetry surfaced as `getAgentInfo()` failing on flaky StreamableHTTP servers where the comply suite kept working. After this change, both code paths behave identically.
+
+- Extends `connectMCPWithFallback` with an optional `authProvider` parameter, forwarded to both StreamableHTTP and SSE transports so OAuth-protected agents still work through the fallback.
+- Rewires `getAgentInfo()` to use `connectMCPWithFallback` for both bearer-token and saved-OAuth-token cases.
+- `discoverMCPEndpoint`'s "Failed to discover MCP endpoint" error now includes a hint that the most common cause is `agent_uri` pointing at the host root when the MCP endpoint lives at a non-standard path (e.g. `/api/mcp`, `/adcp/mcp`).
+
+Closes #1233, #1234.
--- a/.changeset/get-agent-info-fallback.md
+++ b/.changeset/get-agent-info-fallback.md
@@ -4,7 +4,7 @@
 
 fix(client): getAgentInfo() now uses the same StreamableHTTP-then-SSE fallback as every other code path
 
-`SingleAgentClient.getAgentInfo()` previously routed its post-discovery `listTools` connection through `connectMCP`, which has no retry on `StreamableHTTPError` and no SSE fallback. Every other production path (`callMCPTool`, `mcp-tasks`) goes through `connectMCPWithFallback`, which retries once on transient session errors and falls back to `SSEClientTransport` for non-401 failures.
+`SingleAgentClient.getAgentInfo()` previously routed its post-discovery `listTools` connection through `connectMCP`, which has no retry on `StreamableHTTPError` and no SSE fallback. Every other production path (`callMCPTool`, `mcp-tasks`) goes through `connectMCPWithFallback`, which retries once on `StreamableHTTPError` (covers stale-session 400s per MCP SDK #1852) and falls back to `SSEClientTransport` for non-401 failures on public addresses.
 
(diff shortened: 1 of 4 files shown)


bokelley:
Closing without merge after stepping back. The PR addresses a theoretical asymmetry (getAgentInfo lacks fallback) that no user has reported biting them, and the SSE-fallback test ended up mocked enough that it pins the wiring rather than real behavior.

Splitting the actually-useful pieces:

- **#1234** (wrong-path discovery hint) — pulling into a smaller PR. Five lines, zero risk, real operator value when discovery fails on a host-root URL.
- Retry-predicate widening (caught by the protocol reviewer here) — filing as a separate issue. `instanceof StreamableHTTPError` is too narrow; widening it would help every caller, not just getAgentInfo. That's a real reliability gap.
- The asymmetry argument itself — closing #1233 as won't-fix. `connectMCP` failing fast on transient StreamableHTTP errors is arguably correct behavior; "symmetry" alone doesn't justify adding fallback for a failure mode I couldn't reproduce.

Per @adcp project guidance: don't add error handling for scenarios that can't happen.

bokelley:
Acknowledged — clean decomposition. The discovery hint (#1234) and retry-predicate widening (separate issue) are both trackable independently now. Nothing further needed from triage here.

---

PR #1250, opened 2026-05-02
fix(harness): convert runGrader to async spawn — resolves #1237 same-process deadlock

### LINKED ISSUE #1237: Matrix harness: 'storyboard run --webhook-receiver' subprocess hangs to 120s when invoked via spawnSync
## Symptom

Direct CLI invocation completes in ~1.3s with full JSON output:

```bash
node bin/adcp.js storyboard run http://127.0.0.1:4900/mcp sales_social --json --allow-http \
  --auth sk_harness_do_not_use_in_prod --webhook-receiver
```

Same args via `spawnSync` from the matrix harness (`agent-skill-storyboard.ts:runGrader`) hits the 120s timeout:

```
[harness] grader: status=null signal=SIGTERM error=spawnSync node ETIMEDOUT stdout=0b stderr=0b
```

Reproducible across multiple verify-mode runs against a clean workspace (no zombie agents, ports `4900`/`4950` confirmed clear via `lsof`).

## Hypothesis

`--webhook-receiver` boots a local HTTP listener. The grader's exit path may not always close it cleanly, so on subsequent harness-driven runs the bind leaks. From outside the harness it works because each direct invocation gets a fresh address. The harness's `spawnSync` may be hanging on the listener's open socket.

Other possibilities:
- The receiver waits for an inbound webhook that the agent never sends (storyboard has a webhook step, agent's response is async, harness's grader timeout is hit before the webhook arrives).
- stdio buffering in spawnSync that the direct shell flushes differently.

## Why this matters now

The matrix harness's verify step relies on `runGrader` returning a parseable JSON result. Empty `stdout` + ETIMEDOUT makes the harness misreport `storyboard=fail` regardless of actual storyboard outcome. Concretely: the matrix-blind-fixtures validation run reports `storyboard=fail, traffic=façade` when running via the harness; running the grader directly against the same agent produces `overall_status: passing` with a full traffic profile.

The traffic-check signal is independent and still works (which is the whole point of the redesign). But the harness's secondary signal — storyboard pass/fail — is unreliable until this is fixed.

## Workaround

`runGrader` could fall back to spawning without `--webhook-receiver` when the storyboard track doesn't require webhook conformance. Or the receiver's port could be passed in explicitly so the harness can guarantee a clean port per run.

## Evidence

- `.context/matrix-sales-social-verify-{3,4,5}.log` from PR #20 branch.
- `agent-skill-storyboard.ts:runGrader` call site (now logs `grader: status= signal= error=` on every call to make this debuggable).

bokelley:
## Summary

PR #1241 closed the stdin-pipe stall but the grader still hung 120s on storyboards that exercise upstream calls. Root cause: the harness boots the mock upstream HTTP server in the **same Node process**, then called `spawnSync` for the grader. `spawnSync` blocks the event loop — while it runs, the in-process upstream can't respond. Agent calls upstream, upstream's HTTP handler never gets a tick, agent blocks on the response, grader blocks on the agent, full deadlock at the configured timeout.

This is the residual hang flagged in #1237 that #1241 didn't address.

## Diagnosis

A standalone reproducer (upstream in a separate child process) ran identical `spawnSync` invocations to completion in ~1s across three stdio configurations (`pipe/pipe`, `pipe/inherit`, `inherit/inherit`). That ruled out:

- stdio buffering / pipe-buffer exhaustion
- webhook-receiver port collision
- the stdin-pipe stall already addressed by #1241

The hang was specific to the harness's combination of (a) in-process upstream + (b) synchronous child wait. Same-process deadlock.

## Fix

Convert `runGrader` to async: `spawn` + readable-stream chunk collection + `Promise` on `'close'`, with `setTimeout` → `child.kill('SIGTERM')` for the 120s cap. The parent event loop stays live, so the in-process upstream serves alongside the grader.

Call site (`main()`) gets `await`.

## Verification

`signal_marketplace` verify-mode against a kept blind-fixture workspace:

| | log size | wall | traffic |
|---|---|---|---|
| before | 22 lines | 120s | `_lookup/operator: 1` only |
| after | 632 lines (full grader JSON) | 1.5s | `_lookup/operator: 2, /v2/cohorts: 2` |

`overall_status: partial` matches direct-grader invocation. The harness's traffic check still correctly flags `POST /v2/activations: 0` (separate storyboard-cascade issue, adcontextprotocol/adcp#3796).

## Refs

- #1237 (the original hang)
- #1241 (closed the stdin-stall; this addresses the residual same-process deadlock)
- adcontextprotocol/adcp#3796 (storyboard cascade — separate, unaffected)

## Test plan

- [x] Async runGrader compiles (strict tsc)
- [x] prettier format clean
- [x] Verify-mode on signal_marketplace returns full grader JSON in 1.5s (was 120s timeout)
- [x] Traffic counters reflect actual storyboard execution
- [ ] CI

### DIFF
--- a/.changeset/grader-async-spawn-fix.md
+++ b/.changeset/grader-async-spawn-fix.md
@@ -1,0 +1,13 @@
+---
+'@adcp/sdk': patch
+---
+
+fix(harness): convert matrix harness `runGrader` from `spawnSync` to async `spawn`
+
+The matrix harness (`scripts/manual-testing/agent-skill-storyboard.ts`) is dev-only and not shipped in the npm package. Tagging this `patch` to satisfy the changeset gate; no published code changes.
+
+Background: the harness boots the mock upstream HTTP server in the same Node process. `spawnSync` blocked the event loop while the grader child waited on agent responses → in-process upstream couldn't tick → agent's upstream calls hung → 120s timeout. Same-process deadlock that #1241 didn't catch (its stdin-close fix was correct but only addressed half the failure path).
+
+Fix: async `spawn` + Promise on whichever of `'error'` or `'close'` fires first. Event loop stays live, upstream serves alongside the grader. Verified: signal_marketplace verify-mode log went from 22 lines (timeout) to 632 lines (full grader JSON) at 1.5s wall.
+
+Resolves the residual hang from #1237.
--- a/scripts/manual-testing/agent-skill-storyboard.ts
+++ b/scripts/manual-testing/agent-skill-storyboard.ts
@@ -461,26 +461,52 @@ async function runGrader(url: string, storyboardId: string): Promise<{ passed: b
   child.stderr.on('data', c => stderrChunks.push(c));
 
   let timedOut = false;
+  // Wrap in an object so TS doesn't narrow the closure-assigned `let` to
+  // `null` at the read site (control-flow analysis can't see across async
+  // boundaries on `let` reassignment).
+  const spawnFailure: { error: NodeJS.ErrnoException | null } = { error: null };
   const timer = setTimeout(() => {
     timedOut = true;
     child.kill('SIGTERM');
   }, 120_000);
 
-  const exitCode: number | null = await new Promise(resolveFn => {
-    child.on('close', code => resolveFn(code));
+  // Wire `'error'` (spawn-fail: ENOENT, EACCES, …) and `'close'` (normal
+  // exit, after stdio drain) together. Without an `'error'` listener Node
+  // throws on emit; without resolving the awaiter on `'error'`, a failed
+  // spawn could hang the harness. `'close'` always fires after `'error'`
+  // for processes that started, so we settle on whichever comes first.
+  const exitCode: number | null = await new Promise<number | null>(resolveFn => {
+    let settled = false;
+    const settle = (code: number | null): void => {
+      if (settled) return;
+      settled = true;
+      resolveFn(code);
+    };
+    child.on('error', err => {
+      spawnFailure.error = err;
+      settle(null);
+    });
+    child.on('close', code => settle(code));
   });
   clearTimeout(timer);
 
   const stdout = Buffer.concat(stdoutChunks).toString('utf8');
   const stderr = Buffer.concat(stderrChunks).toString('utf8');
-  const raw = stdout + stderr;
+  const raw = stdout + stderr + (spawnFailure.error ? `\n[spawn error] ${spawnFailure.error.message}` : '');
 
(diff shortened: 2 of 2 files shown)


bokelley:
Expert review pass: 3 reviewers in parallel (code-reviewer, nodejs-testing-expert, dx-expert). Convergent finding: **one blocker** — missing `'error'` listener on the child. Addressed in `9b8ebb60`:

**Fixed:**
- Wired `'error'` (ENOENT/EACCES) listener; settle Promise on whichever of `'error'` or `'close'` fires first. Without this, a missing `node` binary would either crash the harness with an uncaught exception or hang the awaiter forever.
- New diagnostic line on spawn failure: `grader: spawn failed — ENOENT: ...`
- Spawn error message appended to `raw` so it lands in the matrix log via the existing stderr-dump.
- Per DX expert: rebuild the timeout debug-hint from the actual `args.join(' ')` so the suggested command can't drift from the actual spawn call. Hint also points operators at the `--mode build --keep + --mode verify` recovery loop.

**Deferred to follow-ups (non-blockers):**
- Regression test for in-process-upstream + child-process-grader (testing expert suggested `test/scripts/run-grader.test.ts` — server with `setImmediate` response forces a live event loop)
- Extract `runWithStdoutCapture(cmd, args, { timeoutMs })` helper (DX expert: this is the second runGrader bug in two PRs; the plumbing is begging to be isolated from grading logic)
- SIGKILL escalation 5s after SIGTERM if the child ignores the signal (code-reviewer nit)

Verified spawn-error semantics: `spawn('nonexistent', [])` emits `error` with `code === 'ENOENT'` as expected. Typecheck + prettier clean.

bokelley:
Thanks for the thorough review pass. The `'error'` listener fix in `9b8ebb60` looks like the right call — a missing binary silently hanging the awaiter would be rough to diagnose in CI. The deferred items (regression test, `runWithStdoutCapture` helper, SIGKILL escalation) are well-scoped follow-ups; the helper extraction in particular seems worth doing before the next grader bug surfaces.

---

bokelley:
Agreed — the `'error'` listener is the right move to surface missing-binary failures quickly rather than letting the awaiter hang silently. The three deferred items are tracked in the PR body; the `runWithStdoutCapture` helper in particular will pay for itself the first time someone adds another grader-adjacent command. Ready for merge when CI clears.

---

bokelley:
Acknowledged — agree the current state looks good. Waiting on CI; will keep an eye on the run.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_019o91h3KWyxbAgSFZvejewE_

---

PR #1274, opened 2026-05-02
examples: hello_seller_adapter_signal_marketplace — worked passing reference

bokelley:
## Summary

Adds `examples/hello_seller_adapter_signal_marketplace.ts` — the first **worked, passing** reference adapter for any specialism in the matrix-blind-fixtures lineage. A complete v6 typed-platform AdCP signals agent that wraps an upstream signal-marketplace platform via HTTP, designed for adopters to fork-and-modify rather than build-from-scratch.

## Why this exists

Every seller agent is an adapter — even ones whose "upstream" is internal infrastructure are translating AdCP wire calls into backend calls. Reading the spec + skill and synthesizing a correct agent from scratch is hard; empirically (from blind LLM-built adapters during SDK dogfood) most first-pass attempts had at least one missing field, missing error code, or misshapen response that cascaded through the storyboard. Giving adopters a working starting point removes most of that friction.

## Validated end-to-end

Against the published `adcp mock-server signal-marketplace`:

```
steps_passed:  7
steps_failed:  0
steps_skipped: 4   (all missing_tool — tools this specialism doesn't claim)
```

Traffic counters confirm real upstream integration on every expected route:

```
GET  /_lookup/operator:    6
GET  /v2/cohorts:          4
GET  /v2/cohorts/{id}:     2
GET  /v2/destinations:     2
POST /v2/activations:      2
```

This is the **first clean traffic-verified pass** on any specialism in the matrix-blind-fixtures lineage (#1239 origin). Documents what "this is what a clean E2E pass looks like" actually means — not as theory, but as runnable code with passing tests.

## Spec/schema gotchas surfaced while building

Each is documented inline in the code so the next reader doesn't re-discover them:

- **`signal_ids` is `signal_id[]`** (array of provenance objects `{source, data_provider_domain, id}`), NOT array of strings — schema says so explicitly. Comparing against bare strings silently returns no matches and cascades-skips downstream steps.
- **`ActivationKey` is a oneOf** — `type: 'key_value'` puts `key` and `value` at the TOP level of the activation_key object, NOT nested under a `key_value` sub-field. `value` MUST be `string`.
- **The mock's `agent_activation_key` is `{ agent_segment: <string> }`** (an object), not a bare string — extract `.agent_segment` for the `ActivationKey.value` field.
- **`signal_spec` is a semantic brief**, not a substring-match query. Passing it to the upstream's literal-text `q` filter breaks against real-platform-style mocks. In production with a real semantic search backend, pass `signal_spec` to that index; against a literal-match mock, fetch the full catalog and filter client-side via `signal_ids`.

These four gotchas are the kind that schema validators surface at runtime but type checkers don't catch at compile time. Adopters who fork this example skip all four.

## Where to swap

Search for `// SWAP:` markers — eight of them, each marking a place that assumes the published mock as upstream. Adopters replace each with calls to their real backend's HTTP/SDK client; the AdCP-facing platform methods (everything below the SWAP markers) stay the same.

## Run

```bash
# Demo — boot mock + agent
npx @adcp/sdk@latest mock-server signal-marketplace --port 4150
UPSTREAM_URL=http://127.0.0.1:4150 \
  npx tsx examples/hello_seller_adapter_signal_marketplace.ts

# Validate
adcp storyboard run http://127.0.0.1:3001/mcp signal_marketplace \
  --auth sk_harness_do_not_use_in_prod
curl http://127.0.0.1:4150/_debug/traffic
```

## Refs

- adcontextprotocol/adcp#3826 (validate-with-mock-fixtures spec guide — references this as the worked example)
- #1239 lineage (matrix-blind-fixtures, façade detection)
- #1241, #1250 (grader-hang fixes that made `partial`-with-no-failures detectable as the actual signal)

## Test plan

- [x] `npm run build` clean
- [x] `tsc --noEmit -p tsconfig.examples.json` clean
- [x] End-to-end run produces 7 passed / 0 failed / 4 skipped, with traffic gate showing all 5 upstream routes hit
- [x] Single-step calls verify response shapes match `get-signals-response.json` and `activate-signal-response.json` schemas
- [x] Both `platform` and `agent` destination branches pass storyboard validation
- [ ] CI


bokelley:
Cleanup pass after review concerns about size + typecheck strictness. Pushed \`dde8920d\`.

**Before** (initial commit \`8a1d7e0a\`):
- 435 lines (284 code, 117 comment, 34 blank)
- Failed under \`--strict --noUncheckedIndexedAccess --exactOptionalPropertyTypes --noPropertyAccessFromIndexSignature\`: 8 errors including 2 real bugs

**After** (\`dde8920d\`):
- **360 lines** (down 75)
- **0 errors** under the strictest realistic adopter config
- Runtime gates unchanged: 7/11 passed, 0 failed, 4 skipped (all \`missing_tool\`); traffic counters verify all 5 upstream routes hit

## Real bugs found and fixed

1. **\`Account.authInfo\` missing** — the field is required (framework strips before wire emission so adapters can authorize against the principal in resource handlers). My initial \`accounts.resolve\` returned objects without it. Now: \`authInfo: ctx?.authInfo ?? { principal: 'anonymous' }\`.

2. **\`DecisioningCapabilities.{creative_agents, channels, pricingModels}\` required** — even for signals-only platforms that don't sell media. Empty arrays are correct. Filed adcontextprotocol/adcp-client#1278 for the underlying DX issue (these fields should be optional or specialism-conditional).

3. **\`SignalID\` discriminated-union narrowing** — \`signal_id.data_provider_domain\` only exists on the \`source: 'catalog'\` variant; the \`source: 'agent'\` variant has \`agent_url\` instead. The filter now narrows on \`sid.source === 'catalog'\` before reading provenance. Documented inline.

The first two would have shipped as latent bugs that surface only when an adopter forks the example and runs against a different storyboard or stricter validator. Catching them is exactly the point of the strictness pass.

## Trim

Replaced 5 \`fetch + check + parse\` UpstreamClient methods with one \`httpJson<T>(method, path, opts)\` helper plus 5 thin typed entry points. Each entry point keeps its \`// SWAP:\` marker — adopters still see exactly what to replace when wiring to a real backend, just without the boilerplate noise.

Comments tightened: kept the spec-gotcha learnings (the four "things schema validators catch but type checkers don't"), dropped paragraph-length prose where one line works.

Ready to merge.

bokelley:
Pushed 5a9338dc — added the three-gate CI test you suggested. Now self-policing.

\`test/examples/hello-seller-adapter-signal-marketplace.test.js\` runs on every commit (\`test:examples\` npm script, wired into \`ci.yml\` after \`npm test\`):

1. **tsc strict** — \`--strict --noUncheckedIndexedAccess --exactOptionalPropertyTypes --noPropertyAccessFromIndexSignature\` + 2 other hardening flags. Catches latent type errors that default \`--strict\` misses.
2. **Storyboard** — boot mock + agent, run \`adcp storyboard run signal_marketplace\`, assert \`summary.steps_failed === 0\`. Catches AdCP wire-shape regressions.
3. **Traffic** — GET \`/_debug/traffic\`, assert each headline route hit ≥1. Catches façade regressions.

Adversarially validated by sabotaging the example (\`const cohorts: UpstreamCohort[] = []\` instead of the upstream call):
- Gate 1 still passes (regression is runtime-only — correct, types not affected)
- Gate 2 fails with \`storyboard reported N failed steps\` + per-step diagnostics
- Gate 3 fails with \`These upstream routes had zero hits — the adapter is a façade for them\`

Each gate fires for a distinct failure mode. Together they make the example self-policing: a contributor (human or LLM) who modifies the example or the SDK in a way that breaks any one of the three fails CI, not "looks fine" review.

Future \`hello_seller_adapter_<specialism>\` additions each get their own \`test/examples/*.test.js\` with the same shape — three gates, ~20s each, separate file per specialism so they can run in parallel.

CI wall locally: 18s for this one. Acceptable per-commit gate.

This is also the pattern the validate-with-mock-fixtures spec guide (adcontextprotocol/adcp#3826) describes — applied here to our own reference adapter so what we tell adopters to do is what we ourselves do.

bokelley:
Three-gate structure looks solid. The adversarial sabotage validation (Gate 1 passes, Gates 2+3 fail on the façade regression) is exactly the right way to prove each gate is independently load-bearing. 18s wall time is well within acceptable for a per-commit gate. Approved pattern for future `hello_seller_adapter_<specialism>` additions.

---

bokelley:
Thanks for the sign-off. Noted that the three-gate / adversarial-sabotage pattern is approved for future `hello_seller_adapter_<specialism>` additions — will carry that forward as the template when the next specialism example lands.

---

bokelley:
Noted — three-gate / adversarial-sabotage pattern is locked in as the template for future `hello_seller_adapter_<specialism>` additions.

---

bokelley:
Acknowledged — logged as the canonical template for future `hello_seller_adapter_<specialism>` additions.

---

bokelley:
Pushed `e5ee7fe4` — cleanup pass using the SDK ergonomics that landed in this round (#1259, #1278/#1282, #1286/#1295, #1287/#1294, #1284/#1296). Demonstrates the friction-removal in concrete terms.

**Before** (when the example was first written):
- `capabilities` declared empty `creative_agents`, `channels`, `pricingModels` arrays — required by type even for signals-only platforms.
- `accounts.resolve` returned `authInfo: ctx?.authInfo ?? { principal: 'anonymous' }` — required field on `Account`.
- Inline ~5-line gotcha comments duplicated knowledge about `signal_ids` and `ActivationKey`.
- Test imported from `../../dist/lib/mock-server/index.js` — brittle path because the public sub-export didn't exist.

**After** (this commit):
```diff
 capabilities = {
   specialisms: ['signal-marketplace'] as const,
-  creative_agents: [] as const,
-  channels: [] as const,
-  pricingModels: ['cpm'] as const,
   config: {},
 };

 resolve: async ref => {
   ...
   return {
     id: operatorId,
     name: adcpOperator,
     status: 'active',
     operator: adcpOperator,
     ctx_metadata: { operator_id: operatorId },
-    authInfo: ctx?.authInfo ?? { principal: 'anonymous' },
   };
 }
```

Inline comments now reference `skills/SHAPE-GOTCHAS.md` instead of duplicating the catalog. Test imports `@adcp/sdk/mock-server` (the public path adopters will use).

**Result**: 348 lines (was 360), strict tsc clean, all 3 CI gates pass.

The example is now both shorter AND more honest about what's required vs optional. Adopters who fork this don't have to wonder "do I really need empty arrays here" — those types are actually optional, and the example doesn't lie about it.

Ready to merge.

bokelley:
Acknowledged — cleanup pass noted. The before/after diff looks like a clean friction-removal demonstration: optional fields dropped, `authInfo` removed, import path updated to the public sub-export, inline gotcha comments pointed at `skills/SHAPE-GOTCHAS.md`. No action needed from triage; ready for human merge review.

---

PR #1297, opened 2026-05-02
feat(server): BuyerAgentRegistry Phase 1 Stage 2 — resolve seam + ctx threading (#1269)

bokelley:
## Summary

Phase 1 Stage 2 of #1269 — wires the `BuyerAgentRegistry` from Stage 1 (#1293) into the dispatcher. The resolved `BuyerAgent` is threaded through `ctx.agent` to `resolveAccount`, `resolveAccountFromAuth`, `resolveSessionKey`, the v6 `RequestContext`, and the v6 `AccountStore.resolve` ctx.

## What's in

- `agentRegistry?: BuyerAgentRegistry` added to `AdcpServerConfig<TAccount>` (v5 surface) and `DecisioningPlatform` (v6 surface). v6 platforms thread through to the v5 config automatically via `createAdcpServerFromPlatform`.
- `agent?: BuyerAgent` field added to: `HandlerContext`, `ResolveAccountContext`, `SessionKeyContext`, `ResolveContext` (v6 `accounts.resolve`), `AccountToolContext`, and `RequestContext`.
- Dispatcher seam in `create-adcp-server.ts` runs after `ctx.authInfo` is populated and before account resolution. Calls `agentRegistry.resolve({ extra })`; on non-null result, freezes the record and sets `ctx.agent`. On throw → `SERVICE_UNAVAILABLE`. On `null` → continues (Stage 4 / Phase 2 add the rejection gate).
- `to-context.ts` `buildRequestContext` propagates `handlerCtx.agent` into the v6 `RequestContext.agent`.

## Strict opt-in

When `agentRegistry` is unset on either surface, `ctx.agent` stays `undefined` and the framework's request flow is unchanged. First test case in the new file asserts this: with no registry configured, the handler sees `ctx.agent === undefined` and `accounts.resolve` sees no `agent` in its ctx.

## Note on `Object.freeze` and `Set`

`Object.freeze` on a `Set` locks the Set's own enumerable properties but does NOT protect the internal `[[SetData]]` slot — `.add()` / `.delete()` / `.clear()` still mutate. `ReadonlySet` is a TypeScript-only contract. Adopters MUST NOT rely on freeze preventing membership changes at runtime. Phase 2's billing-capability check reads via `.has()` only, so adopter mis-mutation would only affect the same request's own logic — no cross-request leak. Documented inline at the freeze site.

## What's NOT in (later Stages)

[...]

PR #1293, opened 2026-05-02
feat(server): BuyerAgentRegistry Phase 1 Stage 1 — types + factories (#1269)

bokelley:
## Summary

Phase 1 Stage 1 of #1269 — durable buyer-agent identity surface, types and factory functions only. Pure additive, no runtime wiring; later stages thread the registry through the dispatcher.

## What's in

- `BuyerAgent` interface with `readonly` fields, set-valued `billing_capabilities`, optional `default_account_terms`, `allowed_brands`, `aliases` (rotation reservation).
- `AdcpCredential` discriminated union — `kind: 'api_key' | 'oauth' | 'http_sig'`. The `http_sig` variant carries the cryptographically-verified `agent_url` (per adcontextprotocol/adcp#3831's derivation rule); the bearer variants don't.
- `BuyerAgentStatus` (`'active' | 'suspended' | 'blocked'`) and `BuyerAgentBillingMode` (re-export of wire `BillingParty`).
- `BuyerAgentRegistry` Protocol + three factory functions:
  - `signingOnly({ resolveByAgentUrl })` — bearer kinds resolve to `null`; only `http_sig` routes to the resolver.
  - `bearerOnly({ resolveByCredential })` — all credential kinds route through the adopter's mapping.
  - `mixed({ resolveByAgentUrl, resolveByCredential })` — `http_sig` routes to the signed path, bearer kinds to the credential path; signed preferred when both present.
- `BuyerAgentResolveInput` — minimal input shape for `resolve()`, decoupled from `ResolvedAuthInfo` to keep the registry surface stable across the migration cycle.

## What's NOT in (later Stage)

- Resolve seam wiring in the dispatcher → Stage 2.
- `ResolvedAuthInfo.credential` field with two-minor deprecation → Stage 3.
- Status enforcement (`suspended`/`blocked` → 403), multi-credential conflict resolution, credential redaction → Stage 4.
- `CachingBuyerAgentRegistry` decorator + conformance fixtures → Stage 5.
- Framework-level `billing_capability` enforcement + AdCP-3.1 error-code emission → Phase 2 (#1292), gated on SDK 3.1 cutover.

## Why types-first

Three reasons to land Stage 1 alone:
1. The shape is up for review — adopters and the Python team see the API before any wiring locks it in.
2. Stage 2 (the dispatcher seam) is a more sensitive change; landing the types separately keeps the diffs reviewable.
[...]

Issue #1292, opened 2026-05-02
feat(server): BuyerAgentRegistry Phase 2 — wire billing-capability enforcement (AdCP 3.1)


bokelley:
## Background

Phase 2 of #1269 — wires the `BuyerAgent.billing_capabilities` set into framework-level enforcement on `sync_accounts` and emits the AdCP-3.1 error codes registered in spec PR adcontextprotocol/adcp#3831 (merged on `main`, lands in 3.1).

**Gated on the SDK's 3.1 cutover** — the new error codes (`BILLING_NOT_PERMITTED_FOR_AGENT`, `BILLING_NOT_SUPPORTED`, `PAYMENT_TERMS_NOT_SUPPORTED`, `BRAND_REQUIRED`) and the `error-details/billing-not-permitted-for-agent.json` clamped recovery shape only exist from 3.1 forward. Phase 1 (#1269) ships in 3.0.x with the registry/identity surface but no new wire emission.

### 1. Framework-level `billing_capability` enforcement on `sync_accounts`

When the buyer's `sync_accounts` request carries a `billing` value not in the resolved `BuyerAgent.billing_capabilities` set, framework rejects with `BILLING_NOT_PERMITTED_FOR_AGENT` per adcp#3831.

- Set-membership check (composes naturally with the set-valued capability shape from Phase 1).
- When `billing_capabilities` contains an alternative the framework would have accepted, populate `error.details.suggested_billing` with that single value (per the clamped recovery shape).
- When `billing_capabilities` is the empty set (improbable but legal), no `suggested_billing` populated; buyer surfaces to a human.

### 2. Cross-tenant onboarding oracle clamp

Per adcp#3831's normative rule, framework MUST emit `BILLING_NOT_PERMITTED_FOR_AGENT` ONLY when buyer-agent identity has been established via signed-request derivation (`credential.kind === 'http_sig'`) OR a credential-to-agent mapping (`BuyerAgentRegistry.resolveByCredential` returned a non-null `BuyerAgent`).

- Bearer-credential-without-mapping → emit `BILLING_NOT_SUPPORTED` instead, with `error.details.scope` **omitted**. Same uniform-response posture as `*_NOT_FOUND`.
- Phase 1's structural distinction (`credential.kind` discriminator + `BuyerAgent` resolution) already provides the bit; Phase 2 wires it into the emission path.

### 3. Clamped error-details schema

[...]

Issue #1269, opened 2026-05-02
design: BuyerAgentRegistry — Phase 1 (3.0.x): identity surface + factory pattern + ResolvedAuthInfo migration


bokelley:
## Motivation

Today the SDK has zero first-class concept of a buyer agent. `accounts.resolve(ref, ctx)` receives a credential (`ctx.authInfo`) and the adopter's DB does whatever it wants — there's no SDK-modeled answer to:

1. **Is this agent authenticated?** — i.e., is the caller a buyer agent we recognize at all?
2. **Can this agent be billed?** — i.e., do we have a payments relationship with the agent itself, or are accounts under this agent always operator-billed (passthrough)?

Question 2 has wire-level consequences (the spec PR adcontextprotocol/adcp#3831 — landing in 3.1 — registers `BILLING_NOT_PERMITTED_FOR_AGENT` for this gate), but the durable identity infrastructure (registry, factory pattern, migration shim) doesn't depend on the spec rev. It can ship in 3.0.x today.

Both questions are durable infrastructure, not pre-trust-beta crutches. Even with full v3 identity (HTTP-Sig + brand.json `authorized_operators` verification), the seller still needs:

- An allowlist (\"we accept these counterparties\").
- Onboarding state (rate cards, payment terms, ToS attestation, billing entity, suspension flags).
- A stable handle for reporting / UI / ops.
- A place to record \"this agent is passthrough-only\" vs \"this agent has a payments account.\"

HTTP-Sig + brand.json answer \"who signed?\" and \"are they brand-authorized?\" — not \"who are they to us commercially?\" The OpenRTB / DSP analogy lands here: an SSP doesn't just verify a DSP's auth and call it done — it has a `buyer_id` row with rate cards, credit, payment status, suspension flags. Token proves identity; row drives commercial behavior.

## Phased delivery

This issue is **Phase 1** — the durable identity/registry surface that ships in 3.0.x, with NO new wire-emission. Adopters get the registry, factory pattern, and migration shim today. Framework-level billing enforcement and the new error codes from adcp#3831 land in **Phase 2** (gated on the SDK's 3.1 cutover) and are tracked in a follow-up issue.

| Phase | AdCP version | Scope | Status |
|---|---|---|---|
[...]

PR #1305, opened 2026-05-02
feat(server): BuyerAgentRegistry Phase 1 Stage 3 — credential synthesis + ResolvedAuthInfo migration (#1269)

bokelley:
## Summary

Phase 1 Stage 3 of #1269. Wires the kind-discriminated `AdcpCredential` through the auth pipeline so the `BuyerAgentRegistry` factories from Stage 2 actually route on real credentials.

## What's in

**Authenticators stamp `credential` on `AuthPrincipal`:**
- `verifyApiKey` → `{ kind: 'api_key', key_id: token }` (preserves an adopter-provided credential when the matched principal carries one)
- `verifyBearer` → `{ kind: 'oauth', client_id: payload.sub, scopes, expires_at }`
- `verifySignatureAsAuthenticator` → `{ kind: 'http_sig', keyid, agent_url, verified_at }` when the verifier's `agentUrlForKeyid` callback resolved an `agent_url`. Omits the credential when the verifier has no keyid→URL mapping (Stage 1's runtime guard would reject the malformed http_sig anyway).

**`serve.ts:attachAuthInfo` propagates `credential`** into `info.extra.credential` so it round-trips through MCP's `AuthInfo` shape without a wire-shape change.

**Dispatcher hoists `extra.credential`** onto top-level `ctx.authInfo.credential` and passes it to `agentRegistry.resolve`. Stage 2's seam was structurally present but functionally inert without a credential; Stage 3 lights it up.

**`ResolvedAuthInfo` migration — additive, two-minor cycle:**
- N (this release): adds `credential?: AdcpCredential`, `agent_url?: string` (informational, post-resolution), `operator?: string` alongside legacy `token` / `clientId` / `scopes`. Legacy fields tagged `@deprecated`; framework continues to populate them for adopter compatibility.
- N+1: warn-once when adopter `authenticate` returns the legacy shape.
- N+2: legacy fields removed.

When `BuyerAgentRegistry` resolves an agent, framework stamps `ctx.authInfo.agent_url` informationally so adopters reading the top-level field see the registry's view. Security-relevant decisions MUST still read `credential.agent_url` (verified, `http_sig`-only) per adcontextprotocol/adcp#3831.

## Strict opt-in preserved

Custom `authenticate` callbacks that don't stamp `credential` see `BuyerAgentRegistry.resolve` return `null` — the framework's request flow is unchanged for unmigrated adopters. Test #6 below locks this contract.

## What's NOT in (later Stages)

- Status enforcement (`suspended`/`blocked` → 403), multi-credential conflict resolution, credential redaction in error paths → Stage 4.
- `CachingBuyerAgentRegistry` decorator + conformance fixtures → Stage 5.
- Framework-level `billing_capability` enforcement + AdCP-3.1 error-code emission → Phase 2 (#1292), gated on SDK 3.1 cutover.

## Test plan

- [x] `test/server-buyer-agent-credential-synthesis.test.js` — 9/9 pass:
  - `verifyApiKey` static-key match populates `credential.key_id` from the bearer token.
  - `verifyApiKey` does not overwrite an adopter-provided credential on the matched principal.
  - `verifyApiKey` dynamic-`verify` path stamps credential when the verifier didn't.
  - Dispatcher hoists `extra.credential` to `ctx.authInfo.credential` reachable from `accounts.resolve`.
  - `signingOnly` resolves an `http_sig` credential via `resolveByAgentUrl`.
  - `signingOnly` returns null for an `api_key` credential — bearer traffic refused.
  - `mixed` routes signed → `resolveByAgentUrl`, api-key → `resolveByCredential`.
  - Legacy `authInfo` with no `credential` → registry returns null, dispatch continues.
  - `signingOnly` resolves and stamps informational `ctx.authInfo.agent_url`.
- [x] Full suite: 7299 pass, 0 fail.

## Cross-links

- #1269 (Phase 1 parent issue)
- #1293 (Stage 1 — types + factories, merged)
- #1297 (Stage 2 — resolve seam + ctx threading, merged)
- adcontextprotocol/adcp#3831 (spec PR for the new error codes Phase 2 emits)

### DIFF
--- a/.changeset/buyer-agent-registry-stage-3.md
+++ b/.changeset/buyer-agent-registry-stage-3.md
@@ -1,0 +1,31 @@
+---
+'@adcp/sdk': minor
+---
+
+feat(server): BuyerAgentRegistry — Phase 1 Stage 3 (credential synthesis + ResolvedAuthInfo migration)
+
+Phase 1 Stage 3 of #1269 — wires the kind-discriminated `AdcpCredential` from Stage 1 through the auth pipeline so `BuyerAgentRegistry` factory functions actually route on real credentials.
+
+**Authenticators stamp `credential` on `AuthPrincipal`:**
+
+- `verifyApiKey` → `{ kind: 'api_key', key_id: token }` (preserves an adopter-provided credential when present)
+- `verifyBearer` → `{ kind: 'oauth', client_id, scopes, expires_at }`
+- `verifySignatureAsAuthenticator` → `{ kind: 'http_sig', keyid, agent_url, verified_at }` when the verifier resolved an `agent_url` from the keyid; otherwise omitted
+
+`serve.ts:attachAuthInfo` propagates `credential` into `info.extra.credential` so it round-trips through MCP's `AuthInfo` shape. The dispatcher hoists `extra.credential` onto top-level `ctx.authInfo.credential` and passes it to `agentRegistry.resolve` — `signingOnly`, `bearerOnly`, and `mixed` factories now route on actual credentials instead of returning `null` for every request.
+
+**`ResolvedAuthInfo` migration shim** — additive, two-minor cycle:
+
+- N (this release): `credential?: AdcpCredential`, `agent_url?: string` (informational, post-resolution), `operator?: string` added alongside legacy `token` / `clientId` / `scopes`. Legacy fields tagged `@deprecated` in JSDoc; framework continues to populate them for adopter compatibility.
+- N+1: framework warns once per process when adopter `authenticate` returns the legacy shape without `credential`.
+- N+2: legacy fields removed.
+
+Adopters with custom `authenticate` callbacks can stamp `credential` directly on the returned `AuthPrincipal` to opt in. Custom callbacks that don't migrate see `BuyerAgentRegistry.resolve` return `null` (no credential = no known agent) — preserving the strict-opt-in invariant from Stage 2.
+
+When `BuyerAgentRegistry` resolves an agent, the framework also stamps the informational top-level `ctx.authInfo.agent_url` so adopters reading it see the registry's canonical view. Security-relevant decisions MUST still read from `credential.agent_url` (verified, `http_sig`-only) — the top-level field is informational per the spec semantics in adcontextprotocol/adcp#3831.
+
+`HandlerContext.authInfo` (v5 surface) and `ResolveContext.authInfo` (v6) both extended with the new fields. v6 `RequestContext` does not expose `authInfo` directly — handlers see `ctx.account` and `ctx.agent`; per-request credential identity flows to `accounts.resolve` and registry resolution.
+
+9 new tests cover: each authenticator's credential synthesis (verifyApiKey static + dynamic + adopter-credential preservation); dispatcher hoist; live registry routing for signingOnly / bearerOnly / mixed; legacy shape continues to work without `credential`. Full suite: 7299 pass, 0 fail.
+
+Phase 2 (#1292) — framework-level billing-capability enforcement and AdCP-3.1 error-code emission — is still gated on the SDK's 3.1 cutover.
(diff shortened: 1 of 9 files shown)


bokelley:
## Security review fixes pushed (aeeecaba)

Three-expert review surfaced **two security blockers (H1, H2)** plus **four medium findings** with strong cross-reviewer convergence. All addressed in this commit.

### Security blockers fixed

**H1 — `http_sig` credential forgery via custom authenticate.** A custom `authenticate` callback could return a literal-shape `{ kind: 'http_sig', agent_url: 'attacker.com', ... }` and `signingOnly` / `mixed` would route it as verified.

Fix: introduced a module-private `VERIFIED_HTTP_SIG_BRAND` symbol in `decisioning/buyer-agent.ts`, stamped only by the new `markVerifiedHttpSig()` helper which is called exclusively from `verifySignatureAsAuthenticator`. Registry factories check `isVerifiedHttpSig(credential)` before treating http_sig as authentic. The brand is non-enumerable so JSON serialization / structured-clone strip it — a relay or replay loses the attestation, which is correct for the security model.

**H2 — `verifyApiKey` credential-kind smuggling.** The `keys: Record<string, AuthPrincipal>` map allowed a static api-key entry to claim `credential: { kind: 'http_sig', ... }`. Combined with H1 this was an escalation path.

Fix: `verifyApiKey` now ALWAYS stamps `kind: 'api_key'` regardless of `matched.credential`. There's no legitimate use case for an api-key holder presenting a non-api-key credential variant. Same posture for the dynamic `verify` path.

### Medium fixes

- **M1** — `api_key.key_id` is now a sha256 hash prefix of the bearer token, not the raw token. The credential surface flows broadly through handlers, `accounts.resolve`, registry resolution, and structured logs — Stage 4's redaction wouldn't reach all those paths. Raw token still appears on `principal.token` for legacy compat.
- **M2** — Dropped `ResolvedAuthInfo.agent_url` stamping. A top-level field invited handlers to gate on registry-derived URLs without verification. Security checks MUST read `credential.agent_url` (http_sig only); adopters reading the registry's view use `ctx.agent.agent_url`.
- **M3** — `verifyBearer` now prefers `payload.client_id` over `payload.sub` per RFC 9068 §2.2 (sub is the resource owner; client_id is the OAuth client). When neither is a usable string, omits the credential entirely — no `'unknown'` bucketing.
- **M5** — `token` / `clientId` / `scopes` on `ResolvedAuthInfo` are now optional through the deprecation cycle. Custom `authenticate` callbacks that only populate `credential` no longer get a TS error.

### Code quality

- Replaced the `HandlerContext.authInfo` inline shape with a direct `ResolvedAuthInfo` import — closes the drift risk code-reviewer flagged.

### Tests extended from 9 to 14 cases

- **H1**: literal-shape http_sig refused; branded http_sig accepted; mixed-rejection behavior matches signingOnly; JSON round-trip strips brand and the result is refused.
- **H2**: `matched.credential` overwrite; dynamic-verify overwrite.
- **M1**: sha256 hash format + stable correlator across calls.
- **M2**: no top-level `agent_url` stamp.

Stage 1 tests (`test/lib/buyer-agent-registry.test.js`) updated to use `markVerifiedHttpSig` so existing http_sig fixtures route correctly through the new brand check.

### Validation

- `npm run typecheck` clean.
- `npm run format:check` clean.
- New tests: 14/14 pass.
- Stage 1 tests: 22/22 pass (after fixture update).
- Full suite: **7294 pass, 0 fail**.

### Confirmed non-issues from review

- `http_sig` omit-on-no-`agent_url` (verifier without `agentUrlForKeyid` callback): correct — Stage 1's runtime guard would reject a partial credential anyway. Decided against runtime warn-once per protocol-expert M4 — the JSDoc on `agentUrlForKeyid` already documents the contract.
- Two-minor deprecation cycle: kept (protocol-expert M6). Cross-language coordination with Python is documented but the cycle length is acceptable.
- MCP `extra.credential` round-trip: verified in Stage 2 review.

Issue #1314, opened 2026-05-02
Composition primitive: wrap platform tool handlers (pre/post hooks) for response-level enrichment + early-return


bokelley:
## Problem

`createAdcpServerFromPlatform` requires adopters to **implement** tool methods (`getMediaBuyDelivery`, `getProducts`, etc.) on the platform object, but provides no first-class way to **compose around** those methods. Two real adopter needs surface in production today:

1. **Pre-handler short-circuit** — Scope3 wants `getMediaBuyDelivery` to early-return a price-optimization payload without invoking the underlying platform method.
2. **Post-handler enrichment** — Scope3 wants to decorate `getMediaBuyDelivery` responses with carbon emissions data, and wants `getProducts` responses to merge brand-manifest data fetched from an external service (24h TTL cache).

Today the workaround is to wrap the entire platform object before calling `createAdcpServerFromPlatform`, re-typing every method. Workable but painful, and it doesn't compose with the framework's resolve/account threading.

## Repro

`scope3data/agentic-adapters#237` defers two gaps for this exact reason:

> - \`getMediaBuyDelivery\` price-optimization early-return + Scope3 carbon enrichment.
> - \`getProducts\` brand-manifest cache (24h TTL, fetched from \`agenticadvertising.org/api/brands\`).

Both are response-shaping concerns that v5's `createServerFromAdapter` allowed via middleware-style composition; v6 has no equivalent.

## Proposal

Add a typed wrapper helper, e.g.:

\`\`\`ts
const wrapped = wrapPlatform(basePlatform, {
  getMediaBuyDelivery: {
    before: async (params, ctx) => {
      if (params.optimization === 'price') return { /* short-circuit */ };
      return undefined; // fall through to base impl
    },
    after: async (result, params, ctx) => enrichWithCarbon(result),
  },
  getProducts: {
    after: async (result) => mergeBrandManifest(result, await brandCache.get()),
  },
});
const server = createAdcpServerFromPlatform(wrapped, opts);
\`\`\`

Keeps the platform-method contract intact (typed, single-responsibility) while letting adopters compose around it.

Open design questions:

- Does `before` returning `undefined` mean \"fall through\" or \"empty result\"? Sentinel token (e.g., `wrapPlatform.PASS`) is more explicit.
- Should the wrapper be per-tool or chain-of-responsibility? Per-tool keeps types tractable.
- Does this generalize to all DecisioningPlatform methods, or just the read-side ones (delivery / products / list)?

[...]

Issue #1312, opened 2026-05-02
createAdcpServerFromPlatform: thread `instructions` to underlying createAdcpServer


bokelley:
## Problem

`createAdcpServer` accepts an `instructions: string` option (`src/lib/server/create-adcp-server.ts:1332`) that flows to the MCP/A2A surface as the server-level instructions field — the canonical place for sellers to publish platform facts, decision policy, and trends to buying agents.

`createAdcpServerFromPlatform` does **not** thread this option through. Adopters using the typed-platform path (`DecisioningPlatform` + `createAdcpServerFromPlatform`) cannot publish instructions today; they have to drop down to the raw `createAdcpServer` path or wrap the returned server.

## Repro

The storefront-platform port in `scope3data/agentic-adapters#237` lists this as one of four deferred gaps:

> Server \`instructions\` (platform facts / decision policy / trends) thread-through to v6 server.

The v5 path published instructions; the v6 port has to leave them off.

## Proposal

Add `instructions?: string` to the `RequiredOptsFor<P>` shape on `createAdcpServerFromPlatform` (or read it from `platform.instructions` if we want it co-located with the platform), and pass it through on the inner `createAdcpServer` call in `from-platform.ts`.

Trivial surface — one field, no behavior change for adopters who don't set it.

## Why now

Cited in #237 alongside two other v6-effectiveness gaps. Smallest of the four to close. Worth folding into the next release so the storefront port can drop the deferred-gap callout.

bokelley:
Done in #1319 — `instructions` is now threaded through `createAdcpServerFromPlatform`. Closing.

Issue #1310, opened 2026-05-02
AccountStore.upsert / list: forward request ctx (authInfo, principal) for principal-based gates


bokelley:
## Gap

`AccountStore.upsert(refs)` and `AccountStore.list(filter)` (in `decisioning/account.d.ts`) don't accept the request `ctx` that `AccountStore.resolve(ref, ctx)` does. The framework's wiring at `decisioning/runtime/from-platform.js` ~line 2291 captures the framework `_ctx` but doesn't forward it:

\`\`\`js
if (accounts.upsert) {
    handlers.syncAccounts = async (params, _ctx) => {
        const refs = (params.accounts ?? []);
        return projectSync(() => accounts.upsert(refs), rows => ({ accounts: rows }));
    };
}
\`\`\`

So a platform implementing `accounts.upsert` cannot access the calling principal — same authInfo that's already available to `accounts.resolve`.

## Why this matters — concrete blocker

Lands on top of merged adcp spec PR [adcontextprotocol/adcp#3851](https://github.com/adcontextprotocol/adcp/pull/3851), which implements two billing gates on `sync_accounts`:

- **`BILLING_NOT_SUPPORTED`** — seller-wide capability gate. No principal needed.
- **`BILLING_NOT_PERMITTED_FOR_AGENT`** — per-buyer-agent commercial-relationship gate. **Principal-keyed** by spec contract: the gate fires when the calling agent's onboarded commercial state with the seller doesn't permit the requested billing value (e.g., passthrough-only agents may only invoice via `operator`).

The training-agent reference impl wires both gates on the legacy `/mcp` route via the v5 `handleSyncAccounts` handler, where principal flows through the request context naturally. The v6 per-tenant routes (`/api/training-agent/<tenant>/mcp`) can't expose `sync_accounts` with the per-agent gate because there's no way to read the principal inside `accounts.upsert`.

Without ctx forwarded, wiring `accounts.upsert` on v6 would silently NO-OP the per-agent gate — which is worse than not wiring it (where the seller correctly returns "tool not registered" and conformant callers fall back to a route that does enforce the gate).

## Proposed fix

Add an optional second parameter mirroring `resolve`'s shape:

\`\`\`ts
interface AccountStore<TCtxMeta = Record<string, unknown>> {
  resolve(ref: AccountReference | undefined, ctx?: ResolveContext): Promise<Account<TCtxMeta> | null>;
[...]

Issue #1272, opened 2026-05-02
AccountStore.upsert() and .list() should receive ResolveContext


bokelley:
## Problem

The `AccountStore` interface in `@adcp/sdk@6.6.0` is asymmetric on auth context:

```ts
resolve(ref: AccountReference | undefined, ctx?: ResolveContext): Promise<Account<TCtxMeta> | null>
upsert?(refs: AccountReference[]): Promise<SyncAccountsResultRow[]>          // ← no ctx
list?(filter: AccountFilter & CursorRequest): Promise<CursorPage<Account>>   // ← no ctx
reportUsage?(req: ReportUsageRequest, ctx?: ResolveContext): Promise<...>
```

`resolve()` and `reportUsage()` receive the buyer's `authInfo` via `ResolveContext`. `upsert()` and `list()` don't.

## Impact

Multi-tenant adopters (where the auth principal determines tenant) need `authInfo` inside `upsert()` and `list()`:

- **`upsert()`**: extract the buyer's `org_id` (or equivalent) from the JWT and write it onto the new account's tenant key (e.g. `externalIds.scope3 = org_id`). Without this, an adopter can't enforce tenant scoping at write time and can't decide which tenant a sync_accounts call should land in.
- **`list()`**: filter the account roster to those belonging to the calling buyer's organization. Without this, `list_accounts` either returns all accounts (over-disclosure) or returns none (under-disclosure).

We hit this on the storefront-platform port in `scope3data/agentic-adapters` (PR scope3data/agentic-adapters#237 + follow-up #242). The storefront's v5 path threaded auth through `ManagedContext`; on the v6 port we had to leave `accounts.upsert` undefined so buyers see UNSUPPORTED_FEATURE on `sync_accounts`.

## Why every workaround fails

- **Stash auth from `resolve()` in a class instance field, read in `upsert()`:** `AccountStore` is shared across concurrent requests, so the stashed value races between buyers.
- **AsyncLocalStorage:** the framework calls `resolve()` and `upsert()` on separate await chains, so ALS context set in `resolve()` doesn't propagate.
- **Per-request `AccountStore`:** the SDK contract is that the store is long-lived (constructed at platform-build time).

## Where the fix lives

`from-platform.js:2290-2295` — the wrapper has `_ctx` available locally but doesn't forward it:

```js
if (accounts.upsert) {
    handlers.syncAccounts = async (params, _ctx) => {
        const refs = (params.accounts ?? []);
[...]

PR #1319, opened 2026-05-02
feat(server): close v6 gaps from scope3 PR#237 — instructions (#1312) + composeMethod (#1314)

bokelley:
## Summary

Closes two of the four v6-effectiveness gaps surfaced by the storefront-platform port in [scope3data/agentic-adapters#237](https://github.com/scope3data/agentic-adapters/pull/237). The other two (#1272 / #1310 — `AccountStore.upsert`/`list` ctx threading) are tracked separately and being handled in a parallel branch.

### #1312 — `DecisioningPlatform.instructions` thread-through

`createAdcpServer` already accepted server-level `instructions` (surfaced on the MCP `initialize` response, the canonical place to publish platform facts / decision policy / trends). `createAdcpServerFromPlatform` did not thread it through.

This adds `DecisioningPlatform.instructions?: string` as the v6 surface, colocated with the rest of the platform declaration. Same precedence pattern as `agentRegistry`: platform declaration wins when both `platform.instructions` and `opts.instructions` are set; the latter remains as the v5 escape hatch.

```ts
const platform: DecisioningPlatform = {
  capabilities: { specialisms: ['sales-non-guaranteed'], /* ... */ },
  accounts: { /* ... */ },
  sales: { /* ... */ },
  instructions:
    'Publisher-wide brand safety: alcohol disallowed. ' +
    'Carbon-aware pricing applies to display impressions only.',
};
```

### #1314 — `composeMethod` + `PASS` sentinel

Adopters need to wrap individual platform methods with pre/post hooks for short-circuit + enrichment patterns. Examples from #237: `getMediaBuyDelivery` price-optimization early-return + carbon emissions enrichment, `getProducts` brand-manifest cache merging.

`composeMethod(inner, { before, after })` is a type-preserving HOF — the wrapped function has the same `(params, ctx) => Promise<TResult>` signature and slots into a typed `DecisioningPlatform` shape without `as` casts.

```ts
import { composeMethod, PASS } from '@adcp/sdk/server';

const wrapped = {
  ...basePlatform,
  sales: {
    ...basePlatform.sales,
    getMediaBuyDelivery: composeMethod(basePlatform.sales.getMediaBuyDelivery, {
      before: async (params) =>
        params.optimization === 'price' ? cachedPriceOpt : PASS,
      after: async (result) => ({
        ...result,
        ext: { ...result.ext, carbon_grams_per_impression: await score(result) },
      }),
    }),
    getProducts: composeMethod(basePlatform.sales.getProducts, {
      after: async (result) => mergeBrandManifest(result, await brandCache.get()),
    }),
  },
};
```

Semantics:
- `before` returning `PASS` falls through to the wrapped method. Any other value short-circuits.
- `after` runs on the result whether it came from the inner method or from a `before` short-circuit.
- `PASS` is registered via `Symbol.for(...)` so reference-equality survives CJS/ESM dual-package boundaries.

## Test plan

- [x] `npm run build` clean
- [x] New: `test/server-decisioning-instructions.test.js` — 4/4 pass (platform.instructions, opts.instructions, precedence when both, omitted-when-neither)
- [x] New: `test/server-decisioning-compose.test.js` — 8/8 pass (passthrough, short-circuit, PASS fall-through, enrichment, params/ctx forwarding, sentinel identity, realistic end-to-end shape)
- [x] Existing: `test/server-decisioning-from-platform.test.js` — 107/107 pass (no regressions)
- [x] Two changesets included

## Out of scope

Gap 1 from #237 (AccountStore.upsert/list ctx threading) is tracked in #1272 + #1310 and being handled in a parallel branch. That signature widening lines up with BuyerAgentRegistry Phase-1 work and should fold into the same release.

### DIFF
--- a/.changeset/compose-method-helper.md
+++ b/.changeset/compose-method-helper.md
@@ -1,0 +1,37 @@
+---
+"@adcp/sdk": minor
+---
+
+Added `composeMethod` and the `PASS` sentinel for wrapping individual `DecisioningPlatform` methods with `before` / `after` hooks. Closes #1314.
+
+Use to layer short-circuit + enrichment patterns on a typed platform without re-typing every method by hand:
+
+```ts
+import { composeMethod, PASS } from '@adcp/sdk/server';
+
+const wrapped = {
+  ...basePlatform,
+  sales: {
+    ...basePlatform.sales,
+    getMediaBuyDelivery: composeMethod(basePlatform.sales.getMediaBuyDelivery, {
+      // Short-circuit: skip the inner platform call on price-optimization paths.
+      before: async (params) =>
+        params.optimization === 'price' ? cachedPriceOpt : PASS,
+      // Enrich: decorate the response with carbon emissions data.
+      after: async (result) => ({ ...result, ext: { ...result.ext, carbon: await score(result) } }),
+    }),
+    getProducts: composeMethod(basePlatform.sales.getProducts, {
+      after: async (result) => mergeBrandManifest(result, await brandCache.get()),
+    }),
+  },
+};
+createAdcpServerFromPlatform(wrapped, opts);
+```
+
+Semantics:
+
+- `before` returning `PASS` falls through to the wrapped method. Returning any other value short-circuits the inner call.
+- `after` runs on the result whether it came from the inner method or from a `before` short-circuit, with the original `params` and `ctx` available.
+- `PASS` is registered via `Symbol.for(...)` so reference-equality holds across CJS/ESM dual-package boundaries.
+
+Surfaced from the storefront-platform port in `scope3data/agentic-adapters#237`, which had to defer `getMediaBuyDelivery` price-optimization early-return + carbon enrichment and `getProducts` brand-manifest cache merging on the v6 path.
--- a/.changeset/compose-method-helper.md
+++ b/.changeset/compose-method-helper.md
@@ -2,12 +2,12 @@
 "@adcp/sdk": minor
 ---
 
-Added `composeMethod` and the `PASS` sentinel for wrapping individual `DecisioningPlatform` methods with `before` / `after` hooks. Closes #1314.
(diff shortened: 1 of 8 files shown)


bokelley:
## Expert review pass

Ran code-reviewer + ad-tech-protocol-expert + dx-expert in parallel. Three pre-merge fixes addressed in `a3a3ce1f`:

**DX expert (blocker-flavored):** the `PASS`-vs-`undefined` semantic is a silent footgun — adopter forgets the `PASS` import, returns bare `undefined`, the hook silently short-circuits with `undefined` as the result. Switched to a discriminated wrapper:

```ts
before?: (params, ctx) => Promise<{ shortCircuit: T } | undefined>
```

`undefined` (or no return) falls through. `{ shortCircuit: value }` short-circuits. TypeScript catches the omit-the-wrapper case at compile time. `PASS` is gone from the public API.

**DX expert:** `composeMethod(undefined, hooks)` was throwing at first call, not at wrap time. Added eager `typeof inner === 'function'` check with an actionable error message that names the most likely cause (referencing an optional method that wasn't implemented on the platform).

**Protocol expert:** `platform.ts` JSDoc claimed `instructions` surfaces on "the MCP `initialize` response and the A2A agent card." No A2A wiring exists — that would have been silent cross-protocol drift. Dropped the A2A claim and noted that the `AgentCard.description` analog is tracked separately.

**Code reviewer:** clean. No blockers. Confirmed the precedence ordering in `from-platform.ts:993` matches the `agentRegistry` pattern at `:987`. tsc --noEmit clean.

Tests now cover 9 cases including eager-validation and explicit `{ shortCircuit: undefined }` disambiguation. 121/121 (compose 9 + instructions 4 + existing from-platform 107 + 1 suite-summary) pass; format check clean.

bokelley:
Acknowledged — expert review pass noted. All three fixups (discriminated `{ shortCircuit }` wrapper, eager `composeMethod` validation, JSDoc A2A-claim drop) look like the right calls, and 121/121 green is a solid gate. No further action from triage side.

---

PR #1317, opened 2026-05-02
feat(server): BuyerAgentRegistry Phase 1 Stage 5 — caching decorator (#1269)

bokelley:
## Summary

Phase 1 Stage 5 of #1269 — final stage. Adds `BuyerAgentRegistry.cached(inner, options)`, a TTL-based caching decorator with concurrent-resolve coalescing and LRU eviction.

## Usage

\`\`\`ts
const registry = BuyerAgentRegistry.cached(
  BuyerAgentRegistry.signingOnly({
    resolveByAgentUrl: async url => db.findBuyerAgent(url),
  }),
  { ttlSeconds: 60 }
);
\`\`\`

## Properties

- **TTL-bounded.** Successful resolutions cached for `ttlSeconds` (default 60). Negative cache (null returns) is OPT-IN via `cacheNullsTtlSeconds` (default 0).
- **LRU-evicted.** Bounded to `maxSize` (default 10000). Map iteration order gives LRU-on-access for free; touch on hit re-inserts as most-recent.
- **Concurrent-resolve coalesced.** N parallel `resolve()` calls on the same credential produce ONE upstream invocation. Subsequent callers await the in-flight promise.
- **Per-kind cache keys** (`http_sig:`, `api_key:`, `oauth:`) prevent cross-kind string collisions from leaking an agent across resolution paths.
- **Skips uncacheable inputs.** When `credential === undefined`, falls through to the inner registry — adopters with custom auth that hasn't migrated to credential synthesis (Stage 3) see no behavior change.

Mirrors the `AsyncCachingJwksResolver` over `JwksResolver` pattern already in `src/lib/signing/`.

## What's NOT in (Phase 2)

- Framework-level `billing_capability` enforcement + AdCP-3.1 error-code emission → Phase 2 (#1292), gated on SDK 3.1 cutover.

## Test plan

- [x] `npm run typecheck` clean.
- [x] `test/lib/buyer-agent-cache.test.js` — 16/16 pass:
  - Basic semantics: cache hit within TTL, expiration, no-cache-nulls default, opt-in null caching, pass-through on undefined credential.
  - Per-kind cache keys: cross-kind collision prevented; same-kind different-value cached separately.
  - Concurrent-resolve coalescing: 10 parallel resolves → 1 upstream call; coalesced resolves see the same result reference; in-flight entries released after completion.
  - LRU eviction: oldest entry evicts on overflow; touch-on-hit delays eviction.
[...]

PR #1311, opened 2026-05-02
feat(server): BuyerAgentRegistry Phase 1 Stage 4 — status enforcement + credential redaction (#1269)

bokelley:
## Summary

Phase 1 Stage 4 of #1269. Wires durable identity (Stages 1-3) into request gating + closes the credential-leak surface in error projections.

## What's in

**Status enforcement:**
- Resolved `BuyerAgent.status === 'suspended' | 'blocked'` → `PERMISSION_DENIED` envelope with `error.details.scope: 'agent'` and `error.details.status: <status>`.
- Runs after registry resolution, BEFORE `accounts.resolve` — no tenant lookup wasted on rejected requests.
- `'active'` agents pass through unchanged.
- `null` registry returns do NOT trigger status enforcement (per-agent billing rejection is Phase 2's gate).
- In-flight protection: seam runs once per synchronous request, not on `tasks_get` polls or background webhook deliveries — a task started under an `'active'` agent and continued after suspension is NOT retroactively cancelled.

**Credential pattern redaction:**
- New `redactCredentialPatterns(message)` helper (`src/lib/server/redact.ts`, internal).
- Applied to all 6 dispatcher sites where `err.message → details.reason`: registry resolution, `resolveAccount`, `resolveAccountFromAuth`, `resolveSessionKey`, idempotency principal resolution, tool handler throws.
- Scrubs: Bearer headers, JSON-quoted credential fields (`"token":"..."`), unquoted forms (`token=...`, `key=...`), long token-shaped strings (≥ 32 chars of base64/hex/url-safe).
- Conservative — false positives get redacted; false negatives minimized.

## What's NOT in (Stage 5 / Phase 2)

- `CachingBuyerAgentRegistry` decorator + conformance fixtures → Stage 5.
- Framework-level `billing_capability` enforcement + AdCP-3.1 error-code emission → Phase 2 (#1292), gated on SDK 3.1 cutover.
- `billing_capabilities` Set hardening (`Object.freeze` doesn't lock `[[SetData]]`) → deferred per Stage 2 review; Phase 2's check is `.has()`-only so the leak surface is bounded.

## Test plan

- [x] `npm run typecheck` clean.
- [x] `test/server-buyer-agent-status-and-redaction.test.js` — 13/13 pass:
  - Active agent dispatches normally.
  - Suspended → 403 with `details.scope: 'agent'`, `details.status: 'suspended'`; handler not invoked.
  - Blocked → 403 with `details.status: 'blocked'`.
[...]

PR #1320, opened 2026-05-02
docs(example): wire BuyerAgentRegistry into hello_seller_adapter_signal_marketplace

bokelley:
## Summary

Every seller needs a buyer-agent registry. This PR updates the worked reference adapter at `examples/hello_seller_adapter_signal_marketplace.ts` to demonstrate the Phase 1 surface (#1269) end-to-end so adopters cloning the example get the full identity story by default rather than a partial scaffold.

## What's added

**In `examples/hello_seller_adapter_signal_marketplace.ts`:**

- **`ONBOARDING_LEDGER`** — an in-memory `Map<key_id, BuyerAgent>` (production sellers replace with a Postgres query). Keyed by the credential's `key_id` (sha256 hash prefix that `verifyApiKey` stamps), not the raw bearer token — so a leak of the ledger doesn't disclose usable credentials.
- **`hashApiKey(token)`** helper showing the same hash the framework uses, so the seller's onboarding flow can pre-compute the cache key when issuing a bearer.
- **`BuyerAgentRegistry.cached(BuyerAgentRegistry.bearerOnly({...}))`** wired to the platform's `agentRegistry` field. Framework runs `resolve()` once per request; status enforcement (`suspended` / `blocked` → 403) fires before any handler.
- **Demonstration site in `accounts.resolve`** where adopters route tenant resolution against `ctx.agent` (operator-vs-agent gating, allowed_brands cross-checks). Currently a `void buyerAgent;` placeholder so adopters see the seam without imposing a specific check.

**In `test/examples/hello-seller-adapter-signal-marketplace.test.js`:**

- **New Gate 4** — asserts unknown api-key tokens are rejected at the auth layer (`401`) before the registry runs. Locks the auth-before-registry order and prevents future refactors from silently dropping the wiring.

## Why this matters

The existing example was the canonical reference for a real-world signal-marketplace adapter, but it didn't include the buyer-agent identity surface that every seller in production needs. Adopters cloning it would either:
1. Skip the registry entirely (no gate on which buyer agents are accepted, no status enforcement, no commercial-relationship surface).
2. Add it ad-hoc later, missing patterns the framework already encodes.

Now the example carries the full Phase 1 surface — registry + caching + status enforcement — as the default starting point. The skill / docs treat it as table-stakes.

## Test plan

- [x] Strict typecheck under the 4-flag adopter-realism config (Gate 1 of the example test): clean.
- [x] Storyboard runs against the updated example: zero failed steps (Gate 2).
- [x] Every expected upstream route hit ≥1 time (Gate 3).
- [x] Unknown api-key rejected at auth layer (new Gate 4).
- [x] Full suite: 7448 pass, 0 fail.

## Cross-links

- #1269 (Phase 1 parent — completed)
- #1293 (Stage 1 — types + factories)
- #1297 (Stage 2 — resolve seam)
- #1305 (Stage 3 — credential synthesis)
- #1311 (Stage 4 — status enforcement + redaction)
- #1317 (Stage 5 — caching decorator)

### DIFF
--- a/.changeset/hello-seller-buyer-agent-registry.md
+++ b/.changeset/hello-seller-buyer-agent-registry.md
@@ -1,0 +1,17 @@
+---
+'@adcp/sdk': patch
+---
+
+docs(example): wire BuyerAgentRegistry into hello_seller_adapter_signal_marketplace
+
+Every seller needs a buyer-agent registry. Updated the worked reference adapter at `examples/hello_seller_adapter_signal_marketplace.ts` to demonstrate the Phase 1 surface (#1269) end-to-end so adopters cloning the example get the full identity story by default rather than a partial scaffold.
+
+Added:
+- An in-memory `ONBOARDING_LEDGER` (replaces with a Postgres-backed query in production) keyed by `credential.key_id` (the sha256 hash prefix `verifyApiKey` stamps).
+- `BuyerAgentRegistry.cached(BuyerAgentRegistry.bearerOnly({...}))` wired to the platform's `agentRegistry` field — framework runs `resolve()` once per request, status enforcement (`suspended` / `blocked` → 403) fires before any handler.
+- Demonstration site in `accounts.resolve` showing where adopters route tenant resolution against `ctx.agent` (operator-vs-agent gating, allowed_brands cross-checks).
+
+Test additions to `test/examples/hello-seller-adapter-signal-marketplace.test.js`:
+- Gate 4 asserts unknown api-key tokens are rejected at the auth layer before the registry runs (locks the auth-before-registry order).
+
+The existing three gates (strict typecheck, storyboard pass, façade traffic) still pass — the registry wiring is implicitly exercised by every storyboard step that authenticates.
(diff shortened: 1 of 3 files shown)


bokelley:
## Review fixes pushed (edc677c8)

Rebased on top of #1322's review fixes; two changes for #1320 itself:

1. **Dropped Gate 5** — both code-reviewer and security flagged the source-string match as brittle (regex matches comments alone, so deleting the load-bearing code lines would still pass). Behavioral coverage is provided by Gate 2 (storyboard runs end-to-end through the gate composition) and `test/server-buyer-agent-sandbox-only.test.js` (framework gate behavior in isolation). Replaced with a one-line comment pointing at those.

2. **Adopter `FIXME` inline** at the resolver's `sandbox: true` line. The comment block above already explained the demo-vs-production trade-off, but the FIXME surfaces during copy-paste rather than only on careful reading — adopters cloning this who skip the comment block won't accidentally mark all production accounts as sandbox.

Test count for the example: 5 → 4 (Gate 5 dropped). Behavioral coverage unchanged.

Ready to land. Merge order: #1322 first, then #1320 (which is currently rebased on #1322; will rebase against main cleanly once #1322 lands).

Issue #1256, opened 2026-05-02
feat(server): Account v3 wire alignment — billing_entity, reporting_bucket, write-only bank strip


bokelley:
## Background

The internal `Account<TCtxMeta>` interface at `src/lib/server/decisioning/account.ts:39` lags the AdCP 3.0.1 wire schema (`schemas/cache/latest/core/account.json`). Wire fields are codegen'd correctly into `src/lib/types/core.generated.ts:411` (`billing_entity`, `reporting_bucket`, `rate_card`, `payment_terms`, `credit_limit`, `setup`, `account_scope`, `governance_agents`), but the framework's *adopter-facing* `Account` type carries only `id` / `name` / `status` / `brand` / `operator` / `advertiser` / `billing` / `ctx_metadata` / `authInfo`.

Result: adopters returning `Account<TCtxMeta>` from `accounts.resolve` / `accounts.list` cannot populate spec fields the wire requires. `toWireAccount` (line 399) silently drops anything it doesn't know about.

This is a wire-correctness bug, not a parity-with-Python ask. Cross-language adopters notice because their Python sellers emit `billing_entity` and their JS sellers don't.

Spawned out of #1249 (closed); item 3 of that issue.

## Scope

1. **Extend `Account<TCtxMeta>`** at `src/lib/server/decisioning/account.ts` with the missing optional fields, typed against the codegen'd wire types where possible:
   - `billing_entity?: BusinessEntity`
   - `reporting_bucket?: { protocol: '...'; ... }`  (shape per schema)
   - `rate_card?: string`
   - `payment_terms?: PaymentTerms`
   - `credit_limit?: { amount: number; currency: string }`
   - `setup?: { url?: string; message: string; expires_at?: string }`
   - `account_scope?: AccountScope`
   - `governance_agents?: ...`

2. **Project through `toWireAccount`** (line 399). Each field passes through unchanged except `billing_entity` (see #3).

3. **Strip `billing_entity.bank` on emit.** The schema marks bank details write-only: \"included in requests to provide payment coordinates, but MUST NOT be echoed in responses\" (`core.generated.ts:581`). `toWireAccount` returns response payloads, so it must remove `bank` from `billing_entity` before returning.

4. **Runtime guard (optional, inexpensive).** Add a `dev`-mode assertion in `toWireAccount` that throws if `billing_entity.bank` is present after stripping — catches refactors that bypass the projection.

5. **Test.** Conformance fixture + unit test:
[...]

PR #1322, opened 2026-05-02
feat(server): BuyerAgent.sandbox_only — defense-in-depth for test-only agents

bokelley:
## Summary

Phase 1.5 of #1269 — adds `BuyerAgent.sandbox_only?: boolean` and a framework-level gate. Proposed during review on PR #1320 (Hello Seller adapter integration): every seller has test agents (CI runners, internal QA agents, partner pre-prod environments) that should ONLY operate against sandbox accounts. Without this field, sellers either accept that test agents have production reach or hand-roll the gate per platform.

## What's added

**`BuyerAgent.sandbox_only?: boolean`** — when `true`, the framework rejects any request whose resolved `Account.sandbox !== true` with:

\`\`\`json
{
  "code": "PERMISSION_DENIED",
  "message": "Buyer agent is sandbox-only; this request targets a non-sandbox account.",
  "recovery": "terminal",
  "details": { "scope": "agent", "reason": "sandbox-only" }
}
\`\`\`

Default `false` / undefined — production-allowed.

## Gate placement

Runs **after** `accounts.resolve` so the framework can compare `agent.sandbox_only` against `account.sandbox`. Runs **after** Stage 4's status enforcement, so a suspended sandbox-only agent fails with status (not sandbox) — predictable error ordering.

**Account-less tools pass the gate.** `provide_performance_feedback`, `list_creative_formats`, etc. don't operate on a specific account; the sandbox/production axis doesn't apply.

## Composes with `Account.sandbox` (#1256)

Same axis at the agent level. Adopters who set `Account.sandbox: true` on their resolver get the gate for free; those who don't see no behavior change.

## What's NOT in (deferred)

- Per-row sandbox enforcement on `sync_accounts` request body (where each account in `accounts[]` has its own sandbox flag) — adopter-side gating in `accounts.upsert` for now; Phase 2 candidate.
- Spec issue for `AGENT_SANDBOX_ONLY_VIOLATION` standardized error code — using `PERMISSION_DENIED + scope:'agent' + reason:'sandbox-only'` as the structured signal until then. File upstream alongside adcp#3871 if the pattern broadens.

## Test plan

- [x] `test/server-buyer-agent-sandbox-only.test.js` — 7/7 pass:
  - Sandbox-only agent → sandbox account: succeeds.
  - Sandbox-only agent → non-sandbox account: PERMISSION_DENIED with structured envelope (`scope: 'agent'`, `reason: 'sandbox-only'`, `recovery: 'terminal'`); handler not invoked.
  - Production agent (sandbox_only undefined): both account types succeed.
  - Production agent (sandbox_only: false): explicit false matches default.
  - Account-less tool: gate skipped.
  - Status enforcement fires BEFORE sandbox-only check (suspended sandbox-only agent → status error, not sandbox error).
  - Null registry result: gate skipped (no agent to check).
- [x] Full suite: 7459 pass, 0 fail.

## Cross-links

- #1269 (Phase 1 parent — completed; Phase 1.5 is this PR)
- #1320 (review on this PR's parent example proposed the field)
- #1256 (Account v3 wire alignment — `Account.sandbox` field this composes with)

### DIFF
--- a/.changeset/buyer-agent-sandbox-only.md
+++ b/.changeset/buyer-agent-sandbox-only.md
@@ -1,0 +1,17 @@
+---
+'@adcp/sdk': minor
+---
+
+feat(server): BuyerAgent.sandbox_only — defense-in-depth for test-only agents
+
+Phase 1.5 of #1269 (proposed during review on PR #1320). Adds a `sandbox_only?: boolean` field to `BuyerAgent` plus a framework-level gate.
+
+When `sandbox_only === true`, the framework rejects any request whose resolved `Account.sandbox !== true` with `PERMISSION_DENIED`, `error.details.scope: 'agent'`, `error.details.reason: 'sandbox-only'`, `recovery: 'terminal'`. Composes naturally with `Account.sandbox` from #1256 — same axis at the agent level.
+
+**When to use it.** CI runners, internal QA agents, partner pre-prod environments. If the test agent's credential leaks, blast radius is bounded to sandbox accounts. Production agents leave the field unset (or `false`).
+
+**Gate placement.** Runs after `accounts.resolve` so the framework can compare `agent.sandbox_only` against `account.sandbox`. Runs after Stage 4's status enforcement, so a suspended sandbox-only agent fails with status (not sandbox). Account-less tools (`provide_performance_feedback`, `list_creative_formats`, etc.) pass the gate — the sandbox/production axis doesn't apply when there's no account in scope.
+
+7 new tests cover: sandbox→sandbox succeeds, sandbox→prod rejected with structured envelope, undefined-or-false sandbox_only is permissive on both account types, account-less tools pass, status enforcement fires before sandbox-only enforcement, null registry result skips the gate. Full suite: 7459 pass, 0 fail.
+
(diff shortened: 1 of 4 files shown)


bokelley:
## Review fixes pushed (cd09abec)

Coordinated review on PRs #1322 + #1320 — **no blockers**, four fixes for #1322:

1. **JSDoc on `BuyerAgent.sandbox_only`** now explicitly names the load-bearing pair contract: for sandbox-only agents to admit any traffic, the resolver MUST return `sandbox: true` on the matching accounts. Mismatches 403 every request.
2. **Strict-by-default inline comment** in `create-adcp-server.ts` explaining the type cast: legacy/non-Account shapes lacking `sandbox` read as `undefined !== true → reject`.
3. **+2 tests** (7 → 9): production-agent + production-account explicitly invokes handler; explicit `sandbox: false` (vs unset) on sandbox-only agent rejects.
4. **Stripped absolute test count** from changeset (rots).

Full suite green. Ready to land alongside #1320.

bokelley:
Acknowledged — fixes noted (JSDoc contract wording, strict-by-default cast comment, two additional test cases, changeset cleanup). No action needed from triage; this is ready for human review to land alongside #1320.

---

PR #1356, opened 2026-05-02
fix(validation): surface success-variant missing fields in oneOf near-miss errors

### LINKED ISSUE #1337: DX: undefined ctx_metadata fields silently produce schema-invalid responses with unactionable validator errors
## Summary

When an adopter's `ctx_metadata` carries an `undefined` field that's spread into a response, the response ships with the field missing entirely (JSON-serialization drops `undefined`). The framework's response validator then fires \`must match exactly one schema in oneOf\` with the schema's `not` clause referenced — pointing at the wrong variant — without naming the actually-missing required field. Adopters spend an iteration figuring out the validator's diagnostic doesn't match the actual fix.

## Empirical observation

Building `examples/hello_seller_adapter_sales_social.ts` for the `sales-social` specialism. The flow:

1. `accounts.resolve` runs against the mock's `/_lookup/advertiser` endpoint which returns only `{adcp_advertiser, advertiser_id, display_name}` — no currency / timezone.
2. My TypeScript interface (loosely matching the mock's seed data type) declared `currency` + `timezone` on the lookup response, both required.
3. The compile pass — strict tsc on the example — said nothing. The `unknown`-narrowing produced `currency: undefined` at runtime; TS sees `string | undefined` from the spread.
4. Account resolved with `ctx_metadata: { advertiser_id, currency: undefined, timezone: undefined }`.
5. `getAccountFinancials` builds a response from `ctx.account.ctx_metadata` plus extra fields — but the `currency` and `timezone` slots are undefined, so `JSON.stringify` drops them.
6. Wire response: `{ account, period, spend }` — missing `currency` and `timezone`, both required by `GetAccountFinancialsSuccess`.
7. Framework's response validator fires:

   ```
   VALIDATION_ERROR: get_account_financials response failed schema validation at /: must NOT be valid

   Issues:
   - `/` must NOT be valid (schemaPath #/oneOf/1/not)
   - `/errors` must have required property 'errors' (schemaPath #/oneOf/1/required)
   - `/` must match exactly one schema in oneOf
   ```

The validator error references the **Error variant's `not` clause** (which forbids `account`/`currency`/etc.) and the **Error variant's `errors` required field** — neither of which is the actual problem. The actual problem is that the Success variant is failing because `currency` and `timezone` are missing — but those failures aren't surfaced.

## Why this happens

`json-schema` `oneOf` validators show issues for the closest-matching variant by default. When a response is malformed enough to fail BOTH variants, the validator picks one to report against — and Ajv typically picks the variant that fails fewer fields. The Error variant fails on `not` (because we have `account`) but only one field is missing (`errors`); the Success variant fails on TWO missing required fields (`currency`, `timezone`). Ajv reports against Error because it has fewer failures. The result: the diagnostic points the adopter to the wrong variant.

The right diagnostic would be: "Success variant: missing required fields `currency`, `timezone`. Error variant: rejected (you have `account` which is forbidden when reporting errors)."

## Proposed fix

Two paths, in order of preference:

### 1. Hint engine for oneOf misses with discriminator-like patterns

When a response fails a `oneOf` AND one variant has a clear "happy path" indicator (more required fields satisfied than the others), include a hint in the validator output:

```
VALIDATION_ERROR: response shape doesn't match any oneOf variant.
  Closest match: GetAccountFinancialsSuccess (variant 0). Missing required fields:
    - /currency
    - /timezone
  (Variant 1 'GetAccountFinancialsError' would require dropping `account`,
  `currency`, `period`, `timezone` and adding `errors[]`.)
```

#1284 (curated hints for known shape gotchas) is the precedent — the hint registry already exists; this issue is "extend it for `oneOf` near-miss diagnostics on disjoint sums where one variant is success-shaped and the other is error-shaped."

### 2. Strip `undefined` keys before serialization, with a typed warning

The response wrapper currently relies on `JSON.stringify`'s drop-undefined behavior. Add a step before stringify that walks the object, flags `undefined` keys, and either:
- Logs a warning naming the path that disappeared, OR
- Re-runs validation with `{ verbose: true }` to surface the missing-field diagnostic.

This makes the diagnostic mechanically clearer at runtime even when the type system can't catch the upstream drift.

## Acceptance

- A response that drops `currency: undefined` triggers a validator error that names `/currency` as missing in the Success variant, not just the `not` clause failure on the Error variant.
- The `examples/hello_seller_adapter_sales_social.ts` example doesn't need to change — but the comment-trail in its commit history (where I narrowed down the bug from "must match exactly one in oneOf" to "the `currency` field is undefined upstream") becomes self-evident from the validator output.

## Cross-reference

- #1284 (existing curated hints registry — extend for `oneOf` near-miss patterns)
- #1327 (related — typed handler signatures vs. runtime undefined)
- #1274 / #1288 (worked-reference adapter lineage)

---

bokelley:
Closes #1337

## Summary

When a handler response contains `undefined` fields that `JSON.stringify` silently drops, the response validator was firing a misleading `must NOT be valid` error pointing to the Error variant's `not`-clause — instead of the Success variant's missing required fields (`/currency`, `/timezone`). Adopters spent 1–2 iterations realising the diagnostic was a red herring.

Two fixes:

**Path 1 — `compactUnionErrors` variant-selection priority** (`src/lib/validation/schema-validator.ts`): Adds `onlyNotAtRoot` as the highest-priority selection criterion. Variants whose *only* residuals are `not`-keyword errors at the root instance path are deprioritised over variants with at least one non-`not` residual. An Error variant with a single root-level `not` error (1 residual) no longer beats the Success variant's `required` failures (2 residuals). Scoped to root-level `not` to avoid penalising `not` discriminators embedded deeper in the schema.

**Path 2 — pre-serialization undefined-field warning** (`src/lib/server/create-adcp-server.ts`): Unconditionally walks the handler response payload before schema validation and emits `logger.warn` for any `undefined`-valued fields (JSON pointer paths). Fires regardless of `responseValidationMode` (including `off`/production) because undefined fields are always a handler mistake and the walk has no meaningful runtime cost.

## What tested

- `npm run typecheck` — pre-existing TS2688/TS5107 errors only (unrelated to this diff)
- `npm run build:lib` — clean
- `node --test test/lib/oneof-variant-selection.test.js` — **4/4 pass** (new tests; schema-dependent tests in the broader suite require `npm run sync-schemas` which is unavailable in this environment — pre-existing gap)

**Pre-PR review:**
- code-reviewer: approved — `onlyNotAtRoot` logic is correct (rootInstance scoping is the union root, not document root); test 4 fixed to avoid insertion-order tie-breaking; depth-cap guard/comment consistency tightened. One nit: `collectUndefinedPaths` silently misses sparse arrays (not a realistic AdCP payload shape; acknowledged in code comments).
- dx-expert: approved — warning message updated to "will be missing from the wire response"; production-mode gap fixed (warning now unconditional). Remaining nits: `@internal` JSDoc tag on test export, array-slot undefined test case (out of scope for this PR).

**Nits surfaced (not fixed in this PR):**
- `_compactUnionErrors` export could carry `@internal` JSDoc (matches existing `_hintRuleCount` convention)
- `collectUndefinedPaths` doesn't detect `undefined` in sparse arrays (not realistic for handler returns)

> **Triage-managed PR.** This bot does not currently iterate on review comments or PR conversation threads (only on the source issue). To unblock:
>
> - **Push fixup commits directly:** `gh pr checkout <num>` → fix → push.
> - **Or re-trigger:** comment `/triage execute` on the source issue.
>
> See [adcp#3121](https://github.com/adcontextprotocol/adcp/issues/3121) for context.

Session: https://claude.ai/code/session_01LGN8jNXxDdFzYtoZyTiuLp

---
_Generated by [Claude Code](https://claude.ai/code/session_01LGN8jNXxDdFzYtoZyTiuLp)_

### DIFF
--- a/.changeset/oneof-near-miss-validator-hints.md
+++ b/.changeset/oneof-near-miss-validator-hints.md
@@ -1,0 +1,27 @@
+---
+"@adcp/sdk": patch
+---
+
+fix(validation): oneOf near-miss diagnostics now surface the correct variant
+
+When a handler returns a response with `undefined` fields that JSON.stringify
+silently drops, the response validator now surfaces the *success* variant's
+missing required fields instead of the error variant's `not`-clause rejection.
+
+Previously, `compactUnionErrors` chose the "best" variant by raw residual
+error count. An error variant with a single root-level `not` error (1 residual)
+beat the success variant with two `required` failures (2 residuals), producing
+a misleading `must NOT be valid` diagnostic that pointed adopters to the wrong
+variant entirely.
+
+The fix adds a priority rule: variants whose *only* residuals are `not`-keyword
+errors at the root instance path are deprioritised over variants that have at
+least one non-`not` residual. The `not`-penalty is scoped to root-level `not`
+errors to avoid affecting schemas that use `not` as a discriminator deeper in
+the tree.
+
+Additionally, the server's response-validation block now logs a developer-
+facing warning for any `undefined`-valued fields in the handler response before
+the schema validator runs. This intercepts the root cause (undefined keys that
+stringify will drop) and names the affected JSON Pointer paths in the same log
+flush as the schema error.
--- a/src/lib/server/create-adcp-server.ts
+++ b/src/lib/server/create-adcp-server.ts
@@ -2059,6 +2059,43 @@ function sanitizeAdcpErrorEnvelope(response: McpToolResponse): void {
   }
 }
 
+/**
+ * Walk `value` (a handler response payload) and return JSON-pointer paths
+ * of any object keys whose value is `undefined`. JSON.stringify silently
+ * drops these, producing missing-field validation errors whose messages
+ * point to the wrong oneOf variant (issue #1337). Collecting them before
+ * serialization lets the server emit a developer-facing warning that names
+ * the culprit fields.
+ *
+ * Recursion depth is capped at 8 — deeper nesting is not a realistic
+ * AdCP response shape, and the cap keeps the hot path O(fields).
+ */
+function collectUndefinedPaths(value: unknown, pointer = '', depth = 0): string[] {
+  if (depth > 8 || value == null || typeof value !== 'object') return [];
+  const paths: string[] = [];
+  if (Array.isArray(value)) {
+    for (let i = 0; i < value.length; i++) {
+      if ((value as unknown[])[i] === undefined) {
+        paths.push(`${pointer}/${i}`);
+      } else {
+        paths.push(...collectUndefinedPaths((value as unknown[])[i], `${pointer}/${i}`, depth + 1));
+      }
+    }
+  } else {
+    for (const key of Object.keys(value as object)) {
+      const child = (value as Record<string, unknown>)[key];
+      const encodedKey = key.replace(/~/g, '~0').replace(/\//g, '~1');
+      const childPointer = `${pointer}/${encodedKey}`;
+      if (child === undefined) {
+        paths.push(childPointer);
+      } else {
+        paths.push(...collectUndefinedPaths(child, childPointer, depth + 1));
+      }
+    }
+  }
+  return paths;
+}
+
 // Echo the request context into a formatted MCP tool response so buyers can
 // trace correlation_id across both success and error responses. Only plain
 // objects are echoed: `si_get_offering` and `si_initiate_session` override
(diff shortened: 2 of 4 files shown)


bokelley:
Closing as superseded — `oneof-near-miss-not-rejection` shipped in 6.7 (commit `e7f8e22`, closes #1337). `compactUnionErrors` excludes `oneOf` variants whose only residual is a `not`-keyword failure when picking the 'best surviving variant.' For Success-vs-Error oneOf where the payload populates Success-only fields but is missing required, the validator now points at the Success variant residuals (`#/oneOf/0/required`) rather than the unactionable `not` clause. This is a simpler approach than the `onlyNotAtRoot`-priority + pre-serialization undefined-walk this PR proposed; the shipped fix lands the diagnostic flip without needing the second walk.

PR #1424, opened 2026-05-03
feat(server): createMediaBuyStore — opt-in targeting_overlay echo (#1415)

bokelley:
## Summary

Closes #1415.

`createAdcpServerFromPlatform` gains a `mediaBuyStore` opt-in. When wired, the framework satisfies the seller spec contract for `targeting_overlay` echo on `get_media_buys` without each adapter persisting + merging by hand.

```ts
import { createAdcpServerFromPlatform, createMediaBuyStore, InMemoryStateStore } from '@adcp/sdk/server';

createAdcpServerFromPlatform(myPlatform, {
  mediaBuyStore: createMediaBuyStore({ store: new InMemoryStateStore() }),
});
```

The framework then:

- Persists `packages[].targeting_overlay` from `create_media_buy` requests, joined with seller-assigned `package_id` (or `buyer_ref` when supplied). Runs in the createMediaBuy projection so it covers both the sync arm AND the HITL completion arm.
- On `get_media_buys`, backfills missing `packages[].targeting_overlay` from the store. Packages the seller already echoed are left untouched.
- On `update_media_buy`, deep-merges the patched overlay against prior persisted: omitted keys keep prior, non-null values replace, explicit `null` clears. `new_packages[]` are persisted as fresh entries.

Backed by any `AdcpStateStore` (`InMemoryStateStore` for dev, `PostgresStateStore` for production). Account-scoped per-tenant via `scopedStore`. Failures logged + swallowed — auto-echo never breaks a successful seller response.

`hello_seller_adapter_guaranteed.ts` now wires `mediaBuyStore` by default — every fork inherits the contract for free.

## Triage resolution: drops four stale suppressors

Triage flagged that `test/lib/storyboard-drift.test.js:291–299` suppressed the four `media_buy_seller/inventory_list_targeting/get_after_*` assertions as `UPSTREAM_SCHEMA_DRIFT` per `adcontextprotocol/adcp#2488`. Verified directly: `schemas/cache/3.0.5/media-buy/get-media-buys-response.json` includes `targeting_overlay` on `PackageStatus` with prose mandating echo for sellers claiming property-lists/collection-lists. adcp#2488 is resolved by the 3.0 GA schema update; the suppressors are stale and dropped in this PR.

## Test plan

- [x] `npx vitest run src/lib/server/media-buy-store.test.ts` — 9/9 pass
- [x] `node --test test/lib/storyboard-drift.test.js` — 498/499 pass (1 unrelated skip), 0 fail
[...]

PR #1423, opened 2026-05-03
feat(server): export canonical state-machine transition maps and assert helpers

bokelley:
Closes #1416

## Summary

Exports the canonical AdCP state-machine transition maps from `@adcp/sdk/server` so sellers don't copy-paste the graph (three examples already had diverged copies). The conformance runner's `status.monotonic` invariant uses the same maps, so sellers who guard handlers with `assertMediaBuyTransition` are guaranteed to agree with the storyboard runner.

**New public exports (available from both `@adcp/sdk` and `@adcp/sdk/server`):**

| Symbol | Purpose |
|---|---|
| `MEDIA_BUY_TRANSITIONS` | `ReadonlyMap<MediaBuyStatus, ReadonlySet<MediaBuyStatus>>` |
| `isLegalMediaBuyTransition(from, to)` | boolean predicate |
| `assertMediaBuyTransition(from, to)` | throws `NOT_CANCELLABLE` (re-cancel) or `INVALID_STATE` |
| `CREATIVE_ASSET_TRANSITIONS` | `ReadonlyMap<CreativeStatus, ReadonlySet<CreativeStatus>>` |
| `isLegalCreativeTransition(from, to)` | boolean predicate |
| `assertCreativeTransition(from, to)` | throws `INVALID_STATE` |
| `CREATIVE_APPROVAL_TRANSITIONS` | `ReadonlyMap<CreativeApprovalStatus, ReadonlySet<CreativeApprovalStatus>>` |
| `isLegalCreativeApprovalTransition(from, to)` | boolean predicate |
| `assertCreativeApprovalTransition(from, to)` | throws `INVALID_STATE` |

`CreativeApprovalStatus` is also newly exported from the root barrel.

**Error code semantics for `assertMediaBuyTransition`:**
- `NOT_CANCELLABLE` (recovery: correctable) — only for `canceled → canceled` re-cancel; matches storyboard `media_buy_seller/invalid_transitions/second_cancel`
- `INVALID_STATE` — all other illegal transitions including `completed → canceled` and `rejected → canceled` (per manifest example: "updating a completed or canceled media buy")

**Structural change:** `src/lib/testing/storyboard/default-invariants.ts` now imports the three exported maps from `server/state-machine.ts` instead of maintaining its own copies. Import direction is `testing → server`, which already existed via `server/envelope-allowlist`. No circular dependency.

[...]

Issue #1416, opened 2026-05-03
SDK: export MEDIA_BUY_TRANSITIONS + assertMediaBuyTransition for runtime use


bokelley:
# SDK opportunity: export `MEDIA_BUY_TRANSITIONS` + `assertMediaBuyTransition` for runtime use

## Problem

The canonical AdCP `MediaBuy` state-machine **already exists in the SDK**, at [`src/lib/testing/storyboard/default-invariants.ts:539-558`](adcp-client/main/src/lib/testing/storyboard/default-invariants.ts):

```ts
const MEDIA_BUY_TRANSITIONS: TransitionGraph = {
  transitions: new Map<string, ReadonlySet<string>>([
    ['pending_creatives', new Set(['pending_start', 'active', 'paused', 'canceled', 'rejected'])],
    ['pending_start', new Set(['active', 'paused', 'canceled', 'rejected'])],
    ['active', new Set(['paused', 'completed', 'canceled'])],
    ['paused', new Set(['active', 'completed', 'canceled'])],
    ['completed', new Set()],
    ['rejected', new Set()],
    ['canceled', new Set()],
  ]),
  enumFile: 'media-buy-status.json',
};
```

The conformance runner uses this graph for cross-step `status.monotonic` validation. It's the source of truth.

But it's **not exported** for sellers. Every adapter that needs to enforce transitions has to copy-paste the graph:

- [`examples/seller-test-controller.ts:137-145`](adcp-client/main/examples/seller-test-controller.ts) — copies the graph + ships `assertMediaBuyTransition`
- [`examples/comply-controller-seller.ts`](adcp-client/main/examples/comply-controller-seller.ts) — same pattern
- `examples/hello_seller_adapter_guaranteed.ts` — doesn't enforce at all (cascade storyboard `media_buy_seller/invalid_transitions/second_cancel` fails: expects `NOT_CANCELLABLE` on re-cancel of canceled buy, gets nothing)

## Reproduction

`bokelley/issue-1375` branch:

```bash
# (boot mock + adapter as in sdk-issue-1)
node bin/adcp.js storyboard run http://127.0.0.1:35199/mcp media_buy_seller/invalid_transitions \
  --json --allow-http --auth sk_storyboard_probe --webhook-receiver
```

Failure:
```
media_buy_seller/invalid_transitions / second_cancel
  step: Re-cancel the canceled buy
  check: error_code
[...]

Issue #1415, opened 2026-05-03
SDK: opinionated MediaBuyStore + auto-echo for targeting_overlay


bokelley:
# SDK opportunity: opinionated `MediaBuyStore` + auto-echo for `targeting_overlay`

## Problem

The `sales-guaranteed` skill explicitly mandates ([`skills/build-seller-agent/specialisms/sales-guaranteed.md`](adcp-client/main/skills/build-seller-agent/specialisms/sales-guaranteed.md)):

> `get_media_buys` must echo `packages[].targeting_overlay.property_list` / `.collection_list`. Per the AdCP types, `property_list` and `collection_list` live inside `TargetingOverlay`, not directly on `Package`. Persist the full `targeting_overlay` verbatim; echo verbatim. `update_media_buy` should merge new targeting overlays without dropping prior refs.

Today the SDK provides only a generic `ctx.store` key-value bag. Every adapter has to:

1. Manually call `ctx.store.put('media_buys', id, {...buy, packages: [{...pkg, targeting_overlay}]})` from `createMediaBuy`
2. Manually re-hydrate it on `getMediaBuys`
3. Manually merge on `updateMediaBuy` (preserving prior `property_list`/`collection_list` refs)

The `hello_seller_adapter_guaranteed.ts` worked example shipped without ANY targeting-overlay persistence. Cascade storyboard `media_buy_seller/inventory_list_targeting/get_after_create` fails on the assertion `property_list.list_id persisted after create` — because the echo never happens. Every adopter who forks the worked example inherits this gap and silently fails the storyboard.

## Reproduction

`bokelley/issue-1375` branch:

```bash
node bin/adcp.js mock-server sales-guaranteed --port 4504 &
NODE_ENV=test ADCP_SANDBOX=1 UPSTREAM_URL=http://127.0.0.1:4504 PORT=35199 \
  ADCP_AUTH_TOKEN=sk_storyboard_probe \
  UPSTREAM_API_KEY=mock_sales_guaranteed_key_do_not_use_in_prod \
  npx tsx examples/hello_seller_adapter_guaranteed.ts &
node bin/adcp.js storyboard run http://127.0.0.1:35199/mcp sales_guaranteed \
  --json --allow-http --auth sk_storyboard_probe --webhook-receiver
```

Failure surface includes:
```
media_buy_seller/inventory_list_targeting / get_after_create
  check: field_value
  description: property_list.list_id persisted after create
[...]

PR #1426, opened 2026-05-03
feat(runner): task_completion. path prefix on context_outputs (#1417)

### LINKED ISSUE #1417: Storyboard runner: can't capture context_outputs.path from HITL task completion artifact
# SDK opportunity: storyboard runner can't capture `context_outputs.path` from HITL task completion artifact

## Problem

The base `sales_guaranteed` storyboard's `create_media_buy` step declares:

```yaml
context_outputs:
  - name: media_buy_id
    path: "media_buy_id"
```

This step explicitly tests the IO-signing async path — the seller returns a task envelope:

```json
{
  "status": "submitted",
  "task_id": "task_async_signed_io_q2",
  "message": "Awaiting IO signature; typical turnaround 2-4 hours"
}
```

There's no `media_buy_id` on the immediate response — only on the **task completion artifact** when IO signing finishes. But the runner's `context_outputs.path` extraction reads ONLY the immediate response (`taskResult.data` per [`src/lib/testing/storyboard/runner.ts:2645-2682`](adcp-client/main/src/lib/testing/storyboard/runner.ts)). It doesn't poll `tasks/get` or fold webhook payloads back into the captured context.

A code comment elsewhere in the runner makes this explicit ([`runner.ts:2956-2957`](adcp-client/main/src/lib/testing/storyboard/runner.ts)):

> "the SDK doesn't poll synchronously after a Task with terminal state"

So the storyboard fails with `capture_path_not_resolvable` on the very `media_buy_id` field that the storyboard is testing the seller correctly produces (just on the artifact, not the immediate response).

## Reproduction

`bokelley/issue-1375` branch:

```bash
# (boot mock + adapter)
node bin/adcp.js storyboard run http://127.0.0.1:35199/mcp sales_guaranteed \
  --json --allow-http --auth sk_storyboard_probe --webhook-receiver
```

Failure (the only `sales_guaranteed`-base failure after wiring `comply_test_controller`):

```
sales_guaranteed / create_media_buy
  step: Create a guaranteed media buy
  check: capture_path_not_resolvable
  description: context_outputs path "media_buy_id" (key "media_buy_id") did not resolve to a usable value
  expected: media_buy_id
```

The seller IS producing the right shape — `submitted` task envelope on the immediate response, `media_buy_id` on the eventual artifact. The runner just doesn't connect them.

### Option 1 — Storyboard-spec change

Extend `context_outputs.path` syntax to support task-artifact paths:

```yaml
context_outputs:
  - name: media_buy_id
    path: "task_completion.media_buy_id"  # poll until task is `completed`, capture from artifact.parts[0].data
```

When the response is a `submitted` task envelope, the runner polls `tasks/get` (or reads webhook payload) until terminal, then captures the artifact path.

**Pros:** explicit, no surprise side effects.
**Cons:** every HITL storyboard needs the new syntax; spec PR.

### Option 2 — Runner auto-poll on capture failure

If `context_outputs.path` doesn't resolve in the immediate response AND the response is a `submitted` task envelope, the runner automatically polls `tasks/get` (bounded short-default, configurable per-step) and retries the capture against the completed artifact's `data`.

**Pros:** zero spec changes; existing storyboards "just work" against HITL flows.
**Cons:** polling logic in the runner; adds latency to negative paths.

Bounded scope:

- **Default timeout: 5s.** Negative-path latency cap. Storyboards that need longer override per-step (`auto_poll_timeout_ms: 30000` or similar). 5s covers the controller-driven submitted→completed path in every existing cascade scenario.
- **Trigger gate: response shape, not transport.** Auto-poll fires only when (a) the response payload's `status === 'submitted'` AND (b) `task_id` is present and non-empty. MCP and A2A both normalize to this shape at the AdCP-payload layer, so a single check covers both transports — but the auto-poll path itself uses `tasks/get` over MCP-style JSON-RPC; A2A flows also work because the same envelope conventions apply.
- **Storyboard opt-out.** A step can declare `auto_poll: false` to bypass — useful for storyboards that want to assert the immediate task-envelope shape before completion (e.g. the existing `a2a_submitted_artifact` invariant).

### Option 3 — Webhook receiver folds payloads into context

The runner already has `--webhook-receiver` and the seller is expected to webhook the completion artifact. If the runner correlates webhook payloads back to the originating step (via `task_id` ↔ webhook payload's `task_id`), the captured context can fill in from the webhook.

**Pros:** uses the webhook path already in scope; no polling.
**Cons:** more state in the runner; correlation logic; webhook timing must beat downstream-step `task_inputs` lookups.

## Recommended path

**Option 2** — runner auto-poll bounded by submitted-task shape detection. Cleanest user experience (storyboards work as written), small runner-side change, no spec churn. Filed as adcp-client repo issue.

If task polling in the runner is undesirable for any reason (deployment isolation, webhook-only test environments), fallback to **Option 3** + **Option 1** in tandem.

## Why this matters

- The base `sales_guaranteed` storyboard explicitly tests the IO-signing path. With current runner, the storyboard's own context_outputs declaration is incompatible with the seller's correct HITL response shape
- Worked example `hello_seller_adapter_guaranteed.ts` returns the spec-correct task envelope and is graded `failed` for it — misleading signal
- This is a runner-side surface; sellers shouldn't have to invent workarounds (e.g. emitting a synthetic `media_buy_id` on the submitted envelope, which would fail the cross-transport `a2a_submitted_artifact` check at [`src/lib/testing/storyboard/validations.ts:1376`](adcp-client/main/src/lib/testing/storyboard/validations.ts))

## Scope

~80-150 lines in the runner: detect submitted-arm response, bounded `tasks/get` poll loop, retry `context_outputs.path` extraction against the completed artifact's `data`. Plus a feature gate so unaffected storyboards don't pay the cost.

bokelley:
## Summary

Closes #1417. Implements **Option 1** from the issue's three options — the path forward both the protocol-expert and code-reviewer triage notes converged on.

Storyboard authors can now declare `path: "task_completion.<inner>"` on `context_outputs`. When the immediate response is a submitted-arm task envelope (HITL / async-signed-IO flows), the runner polls `tasks/get` until terminal and resolves the rest of the path against the artifact's `data` — instead of failing with `capture_path_not_resolvable` for a value the seller correctly produces on the completion artifact.

```yaml
context_outputs:
  - name: media_buy_id
    path: "task_completion.media_buy_id"
```

## Design choices that fell out of triage

- **Explicit prefix, never auto-poll.** Code-reviewer's blocker on Option 2 was that shape-inferred polling risks flipping currently-passing storyboards with intentional submitted envelopes. The prefix IS the explicit opt-in — no risk to existing storyboards.
- **Bounded poll, env-overridable.** Default 30s timeout (`STORYBOARD_TASK_POLL_TIMEOUT_MS`) and 1.5s cadence (`STORYBOARD_TASK_POLL_INTERVAL_MS`). Stuck tasks surface the failure on the step that authored the dependency rather than the storyboard wall-clock budget.
- **Distinct failure code.** `capture_poll_timeout` (not the recycled `capture_path_not_resolvable`) so the failure-class is unambiguous in compliance reports.
- **task_id validation before SDK call.** Length cap (256) + control-char rejection. AdCP doesn't constrain task_id shape on the wire, but unbounded strings are an SSRF / log-injection lever — defensive validation is cheap.

## What this does NOT change

- The runner still doesn't auto-poll based on response shape. Storyboards with intentional submitted envelopes (testing the submitted-arm contract directly) keep working as before.
- No spec change. The prefix lives entirely in the SDK runner.
- The cached `sales_guaranteed` storyboard at `compliance/cache/3.0.5/specialisms/sales-guaranteed/index.yaml` is unchanged — upstream authors can adopt the prefix when they re-author the assertion.

## Test plan

- [x] `node --test test/lib/storyboard-runner-task-completion-capture.test.js` — 5/5 pass:
  - polled artifact captures `media_buy_id` from completion data
  - poll timeout emits `capture_poll_timeout` (not recycled code)
  - sync-arm response on `task_completion.` path falls through to plain extraction (no poll)
  - control-character `task_id` rejected before polling
  - plain `path` (no prefix) regression-free
- [x] `node --test test/lib/storyboard-runner-output-contract-v2-runner.test.js` — 8/8 pass (no regression in existing context_outputs handling)
- [x] `npx tsc --noEmit -p tsconfig.lib.json` clean
- [ ] End-to-end against `hello_seller_adapter_guaranteed` once an upstream storyboard adopts the prefix (deferred until storyboard author updates)

## Companion follow-ups

- #1415 — `createMediaBuyStore` opt-in for sellers (PR #1424; closes the targeting_overlay echo gap on the OTHER side of the storyboard).
- #1416 — `MEDIA_BUY_TRANSITIONS` + `assertMediaBuyTransition` exports (PR #1423; closes the seller's transition-enforcement gap).

### DIFF
--- a/.changeset/storyboard-runner-task-completion-capture.md
+++ b/.changeset/storyboard-runner-task-completion-capture.md
@@ -1,5 +1,5 @@
 ---
-"@adcp/sdk": minor
+'@adcp/sdk': minor
 ---
 
 Storyboard runner — `task_completion.<path>` prefix on `context_outputs.path`.
@@ -11,13 +11,19 @@ The `task_completion.<path>` prefix is an explicit author opt-in: when the runne
 ```yaml
 context_outputs:
   - name: media_buy_id
-    path: "task_completion.media_buy_id"
+    path: 'task_completion.media_buy_id'
 ```
 
-Polling is bounded (default 30s, override with `STORYBOARD_TASK_POLL_TIMEOUT_MS`); per-poll cadence default 1.5s (override with `STORYBOARD_TASK_POLL_INTERVAL_MS`). The runner never auto-polls based on response shape — the prefix is the only opt-in, so existing storyboards with intentional submitted-arm responses don't change behavior.
+The prefix triggers polling on any non-terminal status that carries a `task_id` — `submitted`, `working`, and `input-required`, per the AdCP `tasks-get-response.json` enum. Storyboards with intentional non-terminal-arm assertions are unaffected because the prefix is the only opt-in (no shape inference).
 
-`task_id` is validated (length cap + control-char rejection) before reaching the SDK's `tasks/get` call.
+Polling is bounded (default 30s, override with `STORYBOARD_TASK_POLL_TIMEOUT_MS`); per-poll cadence default 1.5s (override with `STORYBOARD_TASK_POLL_INTERVAL_MS`). `task_id` is validated (length cap + control-char rejection) before reaching the SDK's `tasks/get` call.
 
-Failures emit a distinct `capture_poll_timeout` validation result (not the recycled `capture_path_not_resolvable`) so the failure-class is unambiguous.
+Failures map to three distinct validation checks so compliance reports surface the right diagnostic:
 
-Closes #1417 (Option 1 from the issue's three options; the protocol-expert and code-reviewer triage notes converged on this surface as the cleanest fix).
+- `capture_poll_timeout` — task didn't reach terminal state within the timeout
+- `capture_task_failed` — task reached terminal `failed` / `canceled` / `rejected`
+- `capture_path_not_resolvable` — task succeeded but the artifact's data didn't carry the requested field
+
+This applies to any HITL flow where artifact-only fields land in `context_outputs`: `create_media_buy` (the originally-reported case), `sync_creatives` async approval (`creative_id`), `acquire_rights` (`rights_grant_id`), `si_initiate_session` / `si_send_message`, and any future tool routed through async-signed-IO.
+
+Closes #1417 (Option 1 from the issue's three options; protocol-expert and code-reviewer triage notes converged on this surface as the cleanest fix).
--- a/.changeset/storyboard-runner-task-completion-capture.md
+++ b/.changeset/storyboard-runner-task-completion-capture.md
@@ -1,0 +1,23 @@
+---
+"@adcp/sdk": minor
+---
+
+Storyboard runner — `task_completion.<path>` prefix on `context_outputs.path`.
+
+When a step's immediate response is a submitted-arm task envelope (HITL / async-signed-IO flows), the seller-assigned IDs only exist on the eventual completion artifact, not the immediate response. Pre-this-release, storyboard authors who tried to capture from the artifact got `capture_path_not_resolvable` for a value the seller correctly produced — just on a later message. The workaround was to author a redundant follow-up `get_*` step purely to expose the field.
+
+The `task_completion.<path>` prefix is an explicit author opt-in: when the runner sees this prefix, it polls `tasks/get` until terminal and resolves the rest of the path against the artifact's `data`.
+
+```yaml
+context_outputs:
+  - name: media_buy_id
+    path: "task_completion.media_buy_id"
+```
+
+Polling is bounded (default 30s, override with `STORYBOARD_TASK_POLL_TIMEOUT_MS`); per-poll cadence default 1.5s (override with `STORYBOARD_TASK_POLL_INTERVAL_MS`). The runner never auto-polls based on response shape — the prefix is the only opt-in, so existing storyboards with intentional submitted-arm responses don't change behavior.
+
+`task_id` is validated (length cap + control-char rejection) before reaching the SDK's `tasks/get` call.
+
+Failures emit a distinct `capture_poll_timeout` validation result (not the recycled `capture_path_not_resolvable`) so the failure-class is unambiguous.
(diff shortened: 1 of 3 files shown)


bokelley:
**Triage note — fold candidate from #1431:** The protocol-expert review on this PR surfaced a spec gap that #1431 proposes to address: sellers exercising HITL purely over webhooks currently fail-closed with `capture_poll_timeout` even though the artifact arrived via webhook, because `resolveTaskCompletionOutputs` only polls.

Since the fix targets the same function (`resolveTaskCompletionOutputs`) in the same file, by the same author, and this PR isn't yet approved, triage flagged it as a natural fold-before-merge candidate: extend the resolution to "whichever arrives first — poll OR webhook" when the runner has `--webhook-receiver` active and can correlate the webhook payload's `task_id` back to the originating step.

If the scope would materially delay approval here, merging Option 1 (poll-only) now is fine — #1431 carries the `Blocked-on: #1426` audit trail and will resurface on merge.

---

bokelley:
Acknowledged the fold-candidate note from #1431. My read:

- **Option A — fold before merge:** extend `resolveTaskCompletionOutputs` to "poll OR webhook" in this PR. Same file, same function — diff would grow but reviewers get one coherent change.
- **Option B — merge as-is (poll-only):** #1431 carries the `Blocked-on: #1426` audit trail and resurfaces on merge.

No directive in the comment, so I'm not making the call. @bokelley — which do you prefer? If Option A, comment `/triage execute` on #1431 and I'll fold the webhook leg into a follow-up commit here before merge.

---

Issue #1054, opened 2026-04-29
connectMCPWithFallback: StreamableHTTP first attempt sometimes fails → SSE fallback gets 405 → silent runner skip-everything


bokelley:
**Layer 2 of the v6 training-agent migration spike's "4/4 clean with 100% skipped" finding.** Layer 1 (storyboard runner fail-loud on discovery error) landed in #1005's branch via commit \`6c9e7a04\`. This issue tracks Layer 2 — the underlying transport-selection bug that triggered it.

## Symptom

Storyboard runner against a localhost agent reports clean storyboards with 100% skipped steps when run under \`NODE_ENV\` unset / \`production\`:

\`\`\`
signals_baseline   ✓ 0P / 4S / 0N/A
[skip] ? — missing_tool :: Agent did not advertise tool "get_signals"; agent tools: [].
storyboards: 1/1 clean
steps: 0 passed | 0 failed | 4 skipped | 0 not applicable
\`\`\`

Setting \`NODE_ENV=test\` for the same command produces the expected output. vitest tests against the same agent also work (vitest sets \`NODE_ENV=test\` by default).

After the Layer 1 fix at \`runner.ts:489-498\`, the same scenario now surfaces a hard storyboard failure — but the underlying transport bug remains. Investigation note from the spike:

> The error message that surfaces (when patched in for visibility) is:
> \`SSE error: Non-200 status code (405)\`
> coming from the SSE fallback transport hitting the agent's \`GET /mcp\` which (correctly) returns 405. The StreamableHTTP attempt that should have succeeded (POST works fine; verified via vitest with NODE_ENV=test) fails first for reasons not yet pinned down.

## Where the SSE fallback kicks in

\`src/lib/protocols/mcp.ts:386-404\` — \`connectMCPWithFallback\` falls through to \`SSEClientTransport\` when StreamableHTTP throws AND the URL isn't in \`knownStreamableHTTPUrls\` cache. Cache is populated only on prior successful StreamableHTTP connects in the same process; manual runner runs hit empty cache → SSE fallback → 405.

The asymmetry: \`NODE_ENV=test\` apparently has the cache pre-warmed (or the StreamableHTTP attempt succeeds for a different reason). Under unset / production NODE_ENV against the same agent, first attempt fails. Couldn't pin down WHY in the spike.

## Suspected vectors (untested)

- vitest globals or \`NODE_TLS_REJECT_UNAUTHORIZED\` differing
[...]

PR #1442, opened 2026-05-03
fix(mcp): restore SSE fallback for loopback/private-address MCP servers

### LINKED ISSUE #1438: Regression: 6.7.0 storyboard runner MCP discovery fails against valid endpoints (broke all training-agent storyboards in adcp#3962)
## Summary

Bumping \`@adcp/sdk\` from 6.0.0 to 6.7.0 in adcontextprotocol/adcp#3962 broke MCP discovery for the storyboard runner across every per-tenant training-agent endpoint. All 6 tenants (\`signals\`, \`sales\`, \`governance\`, \`creative\`, \`creative-builder\`, \`brand\`) report \`discovery_failed\` for every storyboard, with 0 clean and 0 passed steps. The same matrix runs cleanly on main with 6.0.0.

CI run: https://github.com/adcontextprotocol/adcp/actions/runs/25277610332

## Reproduction

In adcp@main:
1. \`npm install @adcp/sdk@6.7.0 --save\`
2. \`TENANT_PATH=sales PUBLIC_TEST_AGENT_TOKEN=storyboard-ci-token npx tsx server/tests/manual/run-storyboards.ts\`

Server boots at \`http://127.0.0.1:<port>/api/training-agent/sales/mcp\` (correct). Discovery probe fails:

\`\`\`
× discovery_failed: Discovery failure: Failed to discover MCP endpoint. Tried:
  1. http://127.0.0.1:<port>/api/training-agent/sales/mcp
  2. http://127.0.0.1:<port>/api/training-agent/sales/mcp/
None responded to MCP protocol.
\`\`\`

## What changed

- **6.0.0** → 6.6.0: discovery works against same server. \`upstream_traffic\` and \`capture_path_not_resolvable\` not yet implemented (validations.js \`case 'upstream_traffic'\` absent).
- **6.7.0**: \`upstream_traffic\` implemented (validations.js \`case 'upstream_traffic': return validateUpstreamTraffic(...)\`); discovery broken.

So the regression is in 6.7.0 specifically. Both the runner-side spec adoption AND the discovery break landed in the same release.

## Pinpoint

\`SingleAgentClient.discoverMCPEndpoint\` in \`dist/lib/core/SingleAgentClient.js\` builds a \`urlsToTry\` array (the providedUri + trailing-slash variant + \`/mcp\` permutations) and calls \`connectMCPWithFallback(new URL(url), authHeaders)\` against each. None succeed and none return 401, so the generic \`Failed to discover MCP endpoint\` is thrown.

Likely cause: a change in \`connectMCPWithFallback\` between 6.6 and 6.7 added a stricter MCP-protocol-version probe or initialization sequence that the training agent's server doesn't satisfy. The server is the same in both runs (no AdCP-side changes); only the SDK changed.

## Impact

- adcp#3962 is blocked: storyboard adoption of \`upstream_traffic\` (the contract from adcp#3816 / closed adcp-client#1253) requires 6.7.0 for runner support, but 6.7.0 can't even reach the agent.
- Adopters running 6.7.0 against any training-agent-style server will hit the same wall.

## Workaround

Pin to 6.6.0 in adcp's package.json. Loses upstream_traffic runner support but keeps discovery working.

## Asks

1. Patch the discovery probe so 6.7.0 connects to MCP servers that worked under 6.6.0.
2. Ship 6.7.1 (or 6.8.0) with the fix.
3. Once shipped, adcp#3962 can re-bump and re-adopt \`upstream_traffic\`.

bokelley:
Closes #1438

Commit `587177f` (6.7.0) introduced an `isPrivateAddress()` gate in `connectMCPWithFallbackImpl` that skipped SSE fallback for loopback and RFC-1918 addresses, assuming private-network servers always support StreamableHTTP. This assumption is false for locally-running SSE-only servers (e.g. training-agent endpoints at `127.0.0.1`). The result: every storyboard in the training-agent matrix reported `discovery_failed` after bumping `@adcp/sdk` from 6.0.0 to 6.7.0. The existing `knownStreamableHTTPUrls` guard already handles the "this URL previously answered StreamableHTTP — skip SSE on reconnect" case correctly; the address-class heuristic adds nothing and broke a real class of servers.

**Changes:**
- Remove `isPrivateAddress()` function and its call site from `connectMCPWithFallbackImpl` — SSE is now tried for every address type after StreamableHTTP fails (non-401), consistent with ≤6.6.0 behavior
- Delete `test/unit/mcp-private-address-sse-gate.test.js` — tested a now-deleted function
- Add two regression tests to `test/unit/mcp-discovery-sse-fallback.test.js` documenting that SSE must be attempted for loopback addresses
- Add patch changeset

**What tested:**
- `node --test test/unit/mcp-discovery-sse-fallback.test.js` — 9/9 pass (includes 2 new loopback regression tests)
- `node --test test/unit/mcp-sse-auth-fallback.test.js test/unit/mcp-url-normalization.test.js` — 14/14 pass
- `npm run typecheck` — pre-existing TS2688 / TS5107 errors unrelated to this diff (present on `main` before this change)
- `npm run build:lib` — blocked by missing `tsx` in CI environment; no source changes affect compilation paths

**Pre-PR review:**
- code-reviewer: approved — no blockers; notes regression tests exercise the `testEndpoint` wrapper rather than `connectMCPWithFallbackImpl` directly (acceptable since the removed code is deleted with no surviving branch)
- dx-expert: approved — no blockers; nits: changeset could name the observable symptom, migration doc could gain a one-liner for deferred upgraders (both optional)

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01Qq62ZkmrmbMWr4tNiLZE4d

---
_Generated by [Claude Code](https://claude.ai/code/session_01Qq62ZkmrmbMWr4tNiLZE4d)_


bokelley:
## Root cause re-analysis — recommend closing this PR

The `isPrivateAddress` gate this PR removes is **intentional behavior** introduced to fix issue #1054 (a separate, confirmed transport-selection bug). The gate's rationale: every SDK-managed local server (`createAdcpServer`, `LocalAgentRunner`, mock servers, stubs) uses StreamableHTTP exclusively and returns `405` on SSE GET probes, which was masking the real failure in the original #1054 scenario. The comment at `src/lib/protocols/mcp.ts:74-78` documents this explicitly, and the gate has its own 30-case unit-test suite (`test/unit/mcp-private-address-sse-gate.test.js`). Removing it would re-introduce the #1054 regression.

The actual root cause of #1438 — now confirmed and documented by @bokelley in the issue — was that `create-adcp-server.ts`'s collision check throws lazily inside the request handler rather than at `createAdcpServer()` boot time. When adcp@main's training agent passed a stale `customTools["update_rights"]` registration after SDK 6.7.0 promoted that tool to framework-registered, every MCP POST returned a 500 HTML body. The SDK's discovery code correctly classified that as "no valid MCP protocol response." The `isPrivateAddress` gate was never in the call path.

This PR's changes (removing the gate + its 30 tests + adding new loopback-SSE regression tests) are based on a misdiagnosis. There's no evidence of any real SSE-only server legitimately binding to a private address in this codebase — all private-address servers are `createAdcpServer`-based and StreamableHTTP-only.

**Recommendation: close this PR.** The two actionable follow-ups from the root-cause analysis are:
1. **Migration doc gap** (docs-only, in progress): `docs/migration-6.6-to-6.7.md` needs a note that `customTools["update_rights"]` must be removed on upgrade to 6.7.0, with the lazy-throw/discovery-failure timing called out.
2. **Boot-time collision check** (library improvement, separate issue): moving the `create-adcp-server.ts:3787` collision check from lazy-first-request to `createAdcpServer()` boot time so misconfigurations surface as startup errors rather than discovery failures.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01V1MgRaahiwFFgpMc66eD2k

---

bokelley:
## Triage follow-up — please close this PR

This PR was opened by the triage bot on a misdiagnosis. The original analysis identified `isPrivateAddress()` in `connectMCPWithFallbackImpl` as the root cause of the `discovery_failed` regression; that was wrong.

**Actual root cause (per @bokelley's diagnosis on #1438):** SDK 6.7.0 promoted `update_rights` to a framework-registered first-class tool (PR #1349 / commit `522015d9`). The `adcp@main` training agent still registered it via `customTools`, causing the `createAdcpServer` collision check to throw inside the lazy tenant-build path. Every MCP POST returned a 500 HTML body, which the client correctly read as "no MCP response" and reported as `discovery_failed`. The discovery code and the SSE fallback were working correctly.

**What actually shipped:** PR #1448 (merged 2026-05-03, commit `f4bb1cd9`), which improves the collision error message to name the platform-handler superseding the `customTools` entry, adds migration recipe #16, and adds boot-vs-lazy guidance.

**This PR should be closed.** The change it makes (removing `isPrivateAddress()`) removes a guard that doesn't exist in the current `main` codebase (the branch base is `4f59334`, already past the 6.7.0 refactor) and the regression test it adds tests a deleted code path. Merging would be a no-op at best and a misleading diff at worst.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01486kP3T4yse3kepoGPkfDd

---

PR #1357, opened 2026-05-02
feat(server): add InMemoryImplicitAccountStore reference adapter

bokelley:
Refs #1340

Adds a reference `AccountStore` implementation for `resolution: 'implicit'` platforms (the sync_accounts-first pattern). Previously each adopter independently re-derived key extraction, TTL semantics, and error handling — and commonly got them wrong (keying on `authInfo.sub` instead of `credential.client_id`, returning `AUTH_REQUIRED` instead of `ACCOUNT_NOT_FOUND` when sync hasn't happened).

## What ships

- **`InMemoryImplicitAccountStore<TCtxMeta>`** in `src/lib/adapters/implicit-account-store.ts` — implements `AccountStore` with `resolve()` + `upsert()`. Configurable `keyFn` (defaults to `credential.client_id` / `credential.key_id` / `credential.agent_url`), configurable `ttlMs` (default 24h), and a `buildAccount` factory. Test helpers: `clear()`, `authKey()`, `size`. Exported from both `@adcp/sdk/server` and `@adcp/sdk`.
- **`docs/guides/account-resolution.md`** — one-pager covering all three resolution modes with a deep dive on `'implicit'`: key derivation rationale, the `ACCOUNT_NOT_FOUND` vs `AUTH_REQUIRED` error contract, TTL/staleness guidance, and a Postgres cross-link.
- **`examples/decisioning-platform-implicit-accounts.ts`** — runnable wiring example.
- **JSDoc fix** — both `AccountStore.resolve()` and `refAccountId()` example snippets updated from deprecated `authInfo?.clientId` to `credential.client_id` / `credential.key_id`.

## Deferred from issue scope

- **Postgres-backed adapter** — deferred; `docs/guides/POSTGRES.md` already covers the DDL pattern and the doc cross-links it.
- **Storyboard end-to-end test** — a storyboard exercising `sync_accounts → create_media_buy` without `account_ref` requires new phases in the `media_buy_*_seller` storyboard YAML, which is a spec-side change (`adcontextprotocol/adcp`). Deferred to a follow-up PR once upstream storyboard support lands.

## What was tested

- `npm run format:check` — clean
- `npx tsc --project tsconfig.lib.json --noEmit` — only pre-existing TS2688/TS5107 env errors (confirmed on `main` before branching; not introduced by this diff)
- `tsx` not available in this environment; unit tests require built `dist/` — pre-existing environment constraint

## Pre-PR review

[...]

PR #1467, opened 2026-05-03
feat(adapters): createDerivedAccountStore — Shape D AccountStore factory (#1462)

### LINKED ISSUE #1462: Shape D: createDerivedAccountStore — stateless single-tenant AccountStore factory
## Context

Working through the doha-v20 6.7 migration. The published 6.7 has clean factories for three of four canonical AccountStore shapes:

- **Shape A** `InMemoryImplicitAccountStore` — buyer self-onboards via `sync_accounts`. (#1340)
- **Shape B** `createOAuthPassthroughResolver` — upstream OAuth API owns the roster. (#1369)
- **Shape C** `createRosterAccountStore` — publisher ops curates the roster. (#1409)
- **Shape D** — *(missing)* — stateless single-tenant where every request authenticates with the same per-request credential and there's no account concept.

The wire spec already has `accountResolution: 'derived'` for Shape D ("single-tenant agents where there is no account_id on the wire at all and the auth principal alone identifies the tenant"). What's missing is the factory.

## The pattern Shape D adopters write today

From [`packages/adapter-audiostack/src/platform.ts`](https://github.com/scope3data/agentic-adapters/blob/main/packages/adapter-audiostack/src/platform.ts) — a creative agent that produces an audio ad from script + voice + music. No buyer accounts. No account roster. Every API call uses the buyer's API key directly:

```ts
// ~30 LOC of identical boilerplate every Shape D adopter writes:
export class AudioStackAccountStore implements AccountStore<AudioStackAccountMeta> {
  readonly resolution = 'explicit' as const

  async resolve(ref: AccountReference | undefined, ctx?: ResolveContext): Promise<AudioStackAccount | null> {
    const accessToken = ctx?.authInfo?.token
    if (!accessToken) {
      throw new AdcpError('AUTH_REQUIRED', { /* ... */ })
    }
    const accountId = (ref as { account_id?: string } | undefined)?.account_id
    return {
      id: accountId ?? '__audiostack_publisher_wide__',
      name: accountId ? `AudioStack (${accountId})` : 'AudioStack (publisher-wide)',
      status: 'active',
      ctx_metadata: { accessToken },
      authInfo: { kind: 'api_key', token: accessToken },
    }
  }
}
```

Same shape repeats in `adapter-flashtalking` (creative-only), `adapter-criteo` and `adapter-citrusad` (single-namespace retail-media). Five+ adopters in our monorepo wrote roughly the same 25–30 LOC.

## Note: today many adopters use `'explicit'` for this — should be `'derived'`

Audiostack et al. all declare `resolution: 'explicit'` even though there's no buyer-supplied `account_id` for them to resolve against — the wire field is ignored. The correct mode per the SDK JSDoc is `'derived'`. A factory would standardize the right setting and prevent this drift.

## Proposal

```ts
import { createDerivedAccountStore } from '@adcp/sdk/server';

const accounts = createDerivedAccountStore<AudioStackAccountMeta>({
  toAccount: (ctx) => ({
    id: 'audiostack',                          // stable, same id every time
    name: 'AudioStack',
    status: 'active',
    ctx_metadata: { accessToken: ctx?.authInfo?.token ?? '' },
  }),
  authKind: 'api_key',                         // one of 'oauth' | 'api_key' | 'signature' | 'public'
});
```

The factory:

1. Sets `resolution: 'derived'`.
2. Throws `AUTH_REQUIRED` when `ctx?.authInfo?.token` is missing (configurable via an option).
3. Calls `toAccount(ctx)` and attaches `authInfo` from `ctx.authInfo` to the result.
4. Ignores buyer-supplied `AccountReference` — single-tenant by definition. (Could optionally throw `INVALID_REQUEST` if a buyer DOES pass `account_id` for a `'derived'`-mode adapter, matching the enforcement #1365 added for `'implicit'`.)
5. Omits `list` / `upsert` — single-tenant adapters have nothing to enumerate or write. Adopters who want either compose via spread.

## Why a factory over inline

Same rationale as Shape A/B/C — the boilerplate is repetitive across adopters and easy to misclassify (`'explicit'` vs `'derived'`). One factory bug-fixed centrally beats five adopter implementations drifting.

The savings per adopter are smaller than `createRosterAccountStore` (~25 LOC vs ~150) but the surface is also strictly simpler — probably a 50-line implementation in the SDK + tests. Cheaper to ship than Shape C was.

## Adopter usage in our monorepo (Shape D candidates)

| Adapter | Current | Could become |
|---|---|---|
| `adapter-audiostack` | 30-LOC AccountStore class, `'explicit'` (wrong) | 5 lines, `'derived'` |
| `adapter-flashtalking` | similar | similar |
| `adapter-criteo` (single-tenant variant) | similar | similar |
| `adapter-citrusad` | similar | similar |

Net: ~100 LOC removed across our monorepo, plus correct `'derived'` declaration on four adapters that today claim `'explicit'`.

## Source

Surfaced while assessing whether the audiostack adapter is "crazy easy" yet under 6.7 ([scope3data/agentic-adapters PR #244 thread](https://github.com/scope3data/agentic-adapters/pull/244)). Answer was "no, audiostack still has 30 LOC of AccountStore boilerplate that 6.7 helpers don't address." Filing this would close that gap and complete the four-shape factory family.

Happy to send a PR if the team agrees on the shape — the implementation is straightforward following the Shape A/B/C precedents.

bokelley:
## Summary

Closes #1462. Completes the four-shape `AccountStore` factory family by adding `createDerivedAccountStore` for `resolution: 'derived'` single-tenant agents (audiostack, flashtalking, single-namespace retail-media — every API call authenticates with the buyer's per-request bearer; there is no `account_id` on the wire).

| Shape | Resolution | Helper | This PR |
|---|---|---|---|
| A | `'implicit'` | `InMemoryImplicitAccountStore` | (#1357) |
| B | `'explicit'` | `createOAuthPassthroughResolver` | (#1369) |
| C | `'explicit'` | `createRosterAccountStore` | (#1409) |
| **D** | `'derived'` | **`createDerivedAccountStore`** | **this PR** |

The factory:
- Sets `resolution: 'derived'`.
- Throws `AdcpError('AUTH_REQUIRED', { recovery: 'correctable' })` when `ctx.authInfo` carries no credential signal. Accepts the discriminated `credential` shape (preferred, post-#1269) AND the deprecated `token` / `clientId` fields populated by pre-Stage-3 authenticators during the N+1 deprecation window — fail-closed only when none are present.
- Calls `toAccount(ctx)` and ignores any buyer-supplied `AccountReference` (single-tenant by definition).
- Omits `list` / `upsert` — adopters compose via spread when they need them.
- `skipAuthCheck: true` escape hatch for genuinely public agents (rare; doc-discouraged).

Re-exported from `@adcp/sdk`, `@adcp/sdk/server`, `@adcp/sdk/adapters`. Replaces ~25–30 LOC of bearer-extract + `AUTH_REQUIRED` + return-singleton boilerplate Shape D adopters write today and standardizes the correct `'derived'` declaration (many such adapters today claim `'explicit'`).

## Expert review applied

Ran the change through `code-reviewer`, `ad-tech-protocol-expert`, `security-reviewer`, and `dx-expert` in parallel. Convergent feedback addressed in this PR:

- **Auth predicate** (code-reviewer + security): Original predicate `ctx?.authInfo?.credential === undefined` would refuse legitimately-authenticated requests on the legacy `token`-only `ResolvedAuthInfo` shape (still legal through N+1 per `account.ts:259`). Now accepts any of `credential | token | clientId`; tests added for both legacy paths.
- **"When NOT to use" callout** (DX): Wrong-shape misuse (picking Shape D when you need B/C) silently routes every buyer to the same singleton `id` — added an explicit JSDoc preamble pointing at the right alternatives.
- **`skipAuthCheck` test-harness hint** (DX): Added prose discouraging "tests don't have authInfo" as a reason to flip the flag (use `dispatchTestRequest` / a fixture instead).
- **`docs/llms.txt`** (DX): Updated the three-shape paragraph to four shapes so an LLM agent bumping SDKs finds it without reading migration docs.

## Follow-up filed

The protocol expert flagged that the framework already refuses inline `account_id` for `'implicit'` mode (`refuseImplicitAccountId` at `from-platform.ts:172-185`) but does not for `'derived'`. The Shape D factory ignores inline ids rather than refusing them; matching `'implicit'`'s refuse-semantic in the framework would catch the wire-contract drift for hand-rolled `'derived'` stores too. Worth a follow-up issue to extend the framework-side refusal — out of scope here.

## Test plan

- [x] Unit tests: 16 in `test/derived-account-store.test.js`, all pass with `NODE_ENV=test`.
- [x] Regression: ran the full four-shape factory test suite (74 → 76 tests, all pass).
- [x] Build: `tsc --noEmit` clean.
- [x] Format: `npm run format:check` clean.
- [x] Changeset: `.changeset/derived-account-store.md` (minor bump).


bokelley:
Downstream adopter perspective from `scope3data/agentic-adapters#245` (audiostack rebuild — Shape D in the wild). Three notes, in priority order:

**1. Confirms audiostack's "crazy easy" floor.** Today's hand-rolled inline store there is ~17 LOC; with this helper + the CTX-METADATA-SAFETY pattern (credentials live on `ctx.authInfo`, `ctx_metadata: {}`, specialism methods read `ctx.account.authInfo?.token`), audiostack drops to ~7 LOC of substance:

```ts
const accounts = createDerivedAccountStore<Record<string, never>>({
  toAccount: () => ({
    id: 'audiostack',
    name: 'AudioStack (publisher-wide)',
    status: 'active',
    ctx_metadata: {},
  }),
})
```

Will swap as a follow-up PR once 6.8 ships. The example in the JSDoc (`ctx_metadata: {}` with the "tokens stay on ctx.authInfo, not here" comment) is exactly the shape we want adopters to copy — the credential-discipline win is as valuable as the LOC win.

**2. Please file the follow-up the PR description flags — framework-side refusal of inline `account_id` for `'derived'`.** Without it, a buyer that sends `{ account_id: "foo" }` to audiostack gets silent drop with the singleton response and no signal that the field was ignored. Matching `'implicit'`'s `refuseImplicitAccountId` semantic (from-platform.ts:172-185) for `'derived'` makes the wire-contract violation surface cleanly to the buyer. Useful for hand-rolled `'derived'` stores too, not just helper users. Happy to file if you haven't.

**3. Migration-guide nudge.** The recipe-17 row says "mechanical" but the real value is the credential-discipline shift, not the LOC drop — adopters who hand-roll `'derived'` stores today often stash bearers in `ctx_metadata` (audiostack #245 included). Worth a sentence in the recipe-17 cell pointing at CTX-METADATA-SAFETY for "while you're in there, drop the bearer out of ctx_metadata too" — turns a mechanical swap into a security-posture upgrade.

No blockers. This is the right shape; ship it.

bokelley:
Thanks for the detailed notes, @bokelley — all three addressed:

**Note 1** — good to have real-world confirmation of the LOC floor. The `ctx_metadata: {}` + comment in the JSDoc example is deliberately copy-paste-shaped for exactly that pattern.

**Note 2** — filed as #1469 ("framework: refuse inline `account_id` for `resolution: 'derived'` stores (match `'implicit'` semantic)"). Covers both helper users and hand-rolled Shape D stores; acceptance criteria mirror what `refuseImplicitAccountId` already does for `'implicit'`.

**Note 3** — pushed a follow-up commit to this branch (`08fc763`):
- Recipe-17 "Do this in 6.7" cell now ends with: _"While here, move any bearer tokens out of `ctx_metadata` — see [CTX-METADATA-SAFETY](./guides/CTX-METADATA-SAFETY.md)."_
- Complexity tag upgraded from `mechanical` → `mechanical (security-relevant)` to match the recipe-4 / recipe-13 pattern.
- Fixed an off-by-one in the tl;dr header while I was there (`sixteen` → `seventeen`).

---
Session: https://claude.ai/code/session_012165RTSCncWf1oKri7LAH4

---

bokelley:
Addressed all three. Quick recap:

**1. Audiostack confirmation — noted.** ~17 LOC → ~7 LOC of substance + correct `'derived'` declaration is the win. The JSDoc example deliberately models `ctx_metadata: {}` for that reason.

**2. Follow-up filed → #1468.** Framework-side refusal of inline `account_id` for `'derived'` mode, mirroring `'implicit'`'s `refuseImplicitAccountId` precedent at `from-platform.ts:172-185`. Companion section in migration-6.6-to-6.7.md now references it so adopters who hit the silent-ignore behavior have a tracking issue (`2c741159`).

**3. Recipe-17 nudge — landed.** Your commit `08fc7635` covers the table row + tl;dr count fix. I added a longer "Security-posture upgrade" callout to the companion section explaining the rationale (`2c741159`) — wire-strip ≠ log/error safety, re-derive from `ctx.account.authInfo` per request. Same guidance, two surfaces: quick-fix in the table, with-rationale in the deep-dive.

Force-pushed (your commit was already on the branch when I tried to push mine; rebased on top to keep your row edit intact and avoid a duplicate).

bokelley:
Acknowledged — all three noted. Issue #1469 filed for the framework-side `'derived'` refusal, and the Recipe-17 / complexity-tag / header fixes land in `08fc763`. PR looks clean from here; ready for maintainer review.

---

bokelley:
Thanks for the summary, @bokelley. All three threads accounted for — #1468 filed for the framework-side refusal, and the two-surface Recipe-17 landing (table row + rationale callout) looks solid. Nothing outstanding on my end.

---

bokelley:
Noted — acknowledged the three feedback items, #1469 filed for the framework-side `'derived'` refusal, and fixes in `08fc763`. PR is in the maintainer review queue.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_014ZYN1a5i7pA7Dyg4XoNyGR_

---

Issue #1364, opened 2026-05-02
fix(server): account.ts 'implicit' resolution — docstring claims framework refuses inline account_id but enforcement isn't wired


bokelley:
## Context

Surfaced from the #1340 triage thread. Doha-v20 reports:

> **`account.ts:367-368`** "Framework refuses inline `account_id`" docstring vs enforcement. Aspirational from where we sit — we never observed the framework rejecting an inline `account_id` against an `'implicit'`-resolution platform.

## Problem

`src/lib/server/decisioning/account.ts:367-368` (the `AccountStore.resolution` JSDoc) says:

> `'implicit'` — buyer must `sync_accounts` first; subsequent requests are resolved from the auth principal's pre-synced linkage (LinkedIn, some retail-media operators). **Framework refuses inline `account_id` references for these platforms.**

But no enforcement was found in `src/lib/server/runtime/from-platform.ts` or the dispatcher. Adopters configuring `resolution: 'implicit'` and observing inline-`account_id` requests being accepted have no way to tell whether:

- (a) The enforcement is intended but not yet implemented
- (b) The docstring is aspirational and the enforcement will never land
- (c) The enforcement exists but the test path didn't exercise it

## Ask

Pick one and act:

**Option A** — implement the enforcement. Reject `AccountReference` of shape `{ account_id: ... }` against an `'implicit'`-resolution platform with a typed error before reaching `accounts.resolve`. Probably maps to `INVALID_REQUEST` with `field: 'account.account_id'` and a `suggestion: 'Call sync_accounts first; this platform resolves accounts from your authenticated principal'`.

**Option B** — fix the docstring. Drop the "Framework refuses inline `account_id`" claim, replace with "Adopters MAY refuse inline `account_id` in their `resolve()` impl by returning null." Adopters then own the rejection.

I lean **A** — the docstring exists because protocol-level enforcement is the right answer (otherwise every Shape B adopter writes the same `if (ref?.account_id) return null;` branch), and the implicit-resolution flow is a security boundary that benefits from framework-level guarantees.

## Acceptance

If A: enforcement at the dispatcher layer with a test asserting the typed error envelope. Storyboard test exercising the `'implicit'` + inline-`account_id` combination.

If B: docstring edited to match reality. Migration note for the existing 6.6 → 6.7 work calling out the change in expected behavior.

## Source

#1340 triage thread, adopter-context comment from doha-v20 monorepo.

bokelley:
## Triage

[...]

PR #1479, opened 2026-05-03
fix(server): enforce inline account_id refusal for resolution: 'derived' platforms

### LINKED ISSUE #1468: Framework-side refusal of inline `account_id` for `resolution: 'derived'` (mirror #1365)
## Context

#1462 added Shape D (`createDerivedAccountStore`) for `resolution: 'derived'` single-tenant agents. The factory ignores buyer-supplied `account_id` (single-tenant by definition), but the framework does not refuse it — a buyer that sends `{ account_id: "foo" }` to a `'derived'` agent gets a silent drop with the singleton response and no signal that the field was meaningless.

Asymmetry with `'implicit'`: per #1365, the framework refuses inline `{account_id}` for `'implicit'`-resolution platforms with `AdcpError('INVALID_REQUEST', { field: 'account.account_id' })` *before* `accounts.resolve` runs. See [`refuseImplicitAccountId`](https://github.com/adcontextprotocol/adcp-client/blob/main/src/lib/server/decisioning/runtime/from-platform.ts#L172-L185). The `'derived'` mode has the same wire-contract claim (\"there is no `account_id` on the wire\") but no enforcement.

## Proposal

Extend the existing pre-resolve refusal to also fire when `resolution === 'derived'`. Either:

1. Rename `refuseImplicitAccountId` → `refuseInlineAccountIdWhenForbidden` and have it fire on `resolution === 'implicit' || resolution === 'derived'`. One predicate, both modes.
2. Add a sibling `refuseDerivedAccountId` and call both at the same site.

Option (1) is cleaner — the two modes share the same wire contract for inline ids ('not allowed'), so collapsing them into one check keeps the refusal logic in one place.

## Why it matters

- **Catches buyer-side cargo-culting.** A buyer that copies `account_id` from a Shape B/C call into a Shape D call gets a clean `INVALID_REQUEST` rather than confusing 'singleton ignored my id' behavior.
- **Helps hand-rolled `'derived'` stores too.** Adopters who don't use the factory still get the wire-contract enforcement for free.
- **Matches `'implicit'` precedent.** No reason to enforce the contract in one mode and silently ignore it in the other.

## Out of scope

The `createDerivedAccountStore` factory's behavior (silently ignore inline ids) stays correct for the `skipAuthCheck`/public-catalog edge cases — the framework refusal would short-circuit before factory `resolve` runs, so the factory doesn't need a per-call ignore-vs-throw flag.

## Source

Surfaced in #1462 review by `ad-tech-protocol-expert` and confirmed by adopter feedback in the audiostack rebuild thread (scope3data/agentic-adapters#245).

### LINKED ISSUE #1469: framework: refuse inline account_id for resolution: 'derived' stores (match 'implicit' semantic)
## Summary

The framework already refuses buyer-supplied inline `account_id` for `resolution: 'implicit'` accounts via `refuseImplicitAccountId` at `from-platform.ts:172-185`. No equivalent guard exists for `resolution: 'derived'`.

Currently a buyer that sends `{ account_id: "foo" }` to a `'derived'` agent gets the singleton response silently — the field is ignored with no signal to the buyer that the contract was violated. This is the same wire-contract drift that #1274 closed for `'implicit'`.

## Expected behaviour

When the framework receives a request carrying `account_id` on a `'derived'`-resolution agent, it should return an appropriate error (e.g. `VALIDATION_ERROR` or `INVALID_REQUEST`) rather than silently discarding the field. Matching `'implicit'`'s `refuseImplicitAccountId` semantic is the right analogue.

## Why this matters

- Applies to **all** `'derived'` stores, not just `createDerivedAccountStore` helper users — hand-rolled Shape D adapters benefit equally.
- Without it, buyers that accidentally send `account_id` get an opaque mismatch instead of a correctable error, making debugging difficult and violating the principle of fail-loudly on wire-contract errors.
- `createDerivedAccountStore` (added in #1467) already ignores the field at the factory level; the framework guard closes the gap for hand-rolled stores.

## Acceptance criteria

- [ ] Framework-side guard added alongside or as a complement to `refuseImplicitAccountId` — refuses requests carrying `account_id` when `resolution === 'derived'`.
- [ ] Guard applies to all `'derived'` stores (helper and hand-rolled).
- [ ] Error response is buyer-actionable (code + human-readable reason).
- [ ] Unit test: `{ account_id: "foo" }` on a `'derived'` agent returns the expected error.
- [ ] Existing `createDerivedAccountStore` tests still pass (no regression on the factory's own ignore-logic).

## References

- `from-platform.ts:172-185` — `refuseImplicitAccountId` for `'implicit'`  
- #1467 — `createDerivedAccountStore` (the PR that surfaced this gap; see PR description "Follow-up filed" section)
- `scope3data/agentic-adapters#245` — audiostack rebuild; real-world case where the silent-drop was observed

bokelley:
Closes #1469, Closes #1468

Extends the framework wire-contract enforcement to `'derived'`-resolution (single-tenant) platforms, mirroring the `'implicit'` refusal added in #1364 / recipe #10. Before this change, a buyer sending `{ account_id: "foo" }` to a derived-mode agent received the singleton response silently — the field was discarded with no signal. The framework now rejects it with `AdcpError('INVALID_REQUEST', { field: 'account.account_id' })` before `accounts.resolve` is called. The `{ brand, operator }` arm remains permitted in both modes.

**Implementation:** Renamed private function `refuseImplicitAccountId` → `refuseInlineAccountIdWhenForbidden`; extended the mode guard from `=== 'implicit'` to cover `'derived'` as well; branched error message and suggestion per mode so `'derived'` never emits `sync_accounts` guidance (unsupported in that mode). All 3 existing call sites updated. `account.ts` docstring and inline comments updated to match.

**Migration:** New recipe #10b added to `docs/migration-6.6-to-6.7.md` (callout box, audit list, and full recipe section). Three-row action table covers no-op, behavior-change, and the `resolve()` reads-ref regression case.

## What tested

- `npm run format:check` — clean ✅
- `npm run build:lib` — clean ✅
- Pre-push hook (format + typecheck + build) — passed ✅
- `test/server-decisioning-implicit-account-id.test.js` — 13/13 pass (7 existing `'implicit'` + 6 new `'derived'`; includes `tasks_get`, `get_account_financials`, brand+operator permit, omitted-account path, and message-distinctness assertions) ✅
- Full suite regression: 272 failures after vs 275 before — 3 fewer, no regressions introduced ✅

**Nits noted, not fixed (per reviewer):**
- `buildDerivedPlatform` omits `upsert`/`list` stubs (tests pass without them; parity with `buildImplicitPlatform` optional)
- Guard clause ordering in `refuseInlineAccountIdWhenForbidden` could use a `switch` (current logic is unambiguously correct)

## Pre-PR review

- **code-reviewer**: approved — blockers addressed: resolution-specific error message (no `sync_accounts` in `'derived'` suggestion), migration doc seam between recipe #10b and `createRosterAccountStore` companion note fixed
- **dx-expert**: approved — blockers addressed: suggestion text narrowed to `account.account_id` (brand+operator arm still permitted), migration doc header updated to "Four exceptions" with #10b bullet in callout box and audit list

---

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01Mg8BhdBSsyHD2Y1FppTYVB

---
_Generated by [Claude Code](https://claude.ai/code/session_01Mg8BhdBSsyHD2Y1FppTYVB)_

### DIFF
--- a/.changeset/derived-resolution-account-id-refusal.md
+++ b/.changeset/derived-resolution-account-id-refusal.md
@@ -1,0 +1,13 @@
+---
+"@adcp/sdk": minor
+---
+
+**breaking** — `accounts.resolution: 'derived'` now enforces inline-`account_id` refusal (closes #1469, mirrors recipe #10 for `'implicit'`).
+
+Pre-6.7, a buyer sending `{ account_id: "foo" }` to a `'derived'`-resolution (single-tenant) agent received the singleton response silently — the field was discarded with no signal. The framework now refuses it with `AdcpError('INVALID_REQUEST', { field: 'account.account_id' })` before `accounts.resolve` is called, consistent with the `'implicit'` enforcement added in #1364.
+
+The error message for `'derived'` is distinct from `'implicit'`: it says "single-tenant agent — account is derived from your auth credential" and does not mention `sync_accounts` (which is irrelevant and unsupported for derived-mode agents).
+
+**Migration:** See recipe #10b in `docs/migration-6.6-to-6.7.md`. Adopters whose `resolve()` reads `ref.account_id` under `'derived'` mode must remove that branch — the refusal fires before `resolve()` is called.
+
+Internal rename: `refuseImplicitAccountId` → `refuseInlineAccountIdWhenForbidden` (private function, not exported).
--- a/docs/migration-6.6-to-6.7.md
+++ b/docs/migration-6.6-to-6.7.md
@@ -2,13 +2,17 @@
 
 > **Status: GA in 6.7.** Most changes are additive — adopters running on
 > 6.6 today see no behavior change on `npm update @adcp/sdk` unless they
-> opt in. **Three exceptions** require attention before bumping:
+> opt in. **Four exceptions** require attention before bumping:
 >
 > - **`accounts.resolution: 'implicit'` adopters**: the framework now
 >   actually refuses inline `{account_id}` references (the docstring
 >   was aspirational pre-6.7). If your platform declared `'implicit'`
 >   but accepted inline ids, those calls now reject with
 >   `INVALID_REQUEST`. See recipe **#10** below.
+> - **`accounts.resolution: 'derived'` adopters**: the same inline
+>   `{account_id}` refusal now applies to single-tenant derived-resolution
+>   platforms. Previously the field was silently ignored; it now rejects
+>   with `INVALID_REQUEST`. See recipe **#10b** below.
 > - **Adopters with `: SalesPlatform<Meta>` field annotations claiming
 >   `sales-guaranteed` / `sales-non-guaranteed` / `sales-broadcast-tv` /
 >   `sales-catalog-driven`**: `SalesPlatform` is now structurally
@@ -22,9 +26,9 @@
 >   client probe — including discovery — masquerading as a transport bug.
 >   See recipe **#16**.
 
-## Audit first — the three breaking recipes
+## Audit first — the four breaking recipes
 
-Before bumping, read recipes **#10**, **#11**, and **#16**. Everything
+Before bumping, read recipes **#10**, **#10b**, **#11**, and **#16**. Everything
 else is additive and can be applied incrementally.
 
 - **#10 — `accounts.resolution: 'implicit'` enforces inline-`account_id`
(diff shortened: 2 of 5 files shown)


bokelley:
Runtime change already on main via PR #1475 (commit 6c47db5 — `feat(server): refuse inline account_id for resolution: 'derived' (#1468) (#1475)`). `refuseInlineAccountIdWhenForbidden` is wired in 4 call sites on main. Closing this PR as superseded.

**Migration doc gap to follow up:** main's `docs/migration-6.6-to-6.7.md` still says the derived refusal is "tracked at adcp-client#1468" (line ~666) — that text is stale now that the refusal shipped. Filing a docs-only follow-up to add recipe #10b proper.

Issue #1469 should also close — same scope as #1468 which #1475 already closed.

bokelley:
Acknowledged — PR closed as superseded by #1475. Migration-docs stale reference and #1469 closure noted for your follow-up. Nothing further from me here.

---

bokelley:
Acknowledged. Migration-docs stale reference and #1469 open-status flagged for the next triage sweep.

---
_Generated by [Claude Code](https://claude.ai/code/session_01J8Vmx2aMMHiSwMVnXwe2Wg)_

---

PR #1489, opened 2026-05-03
fix(conformance): get_media_buys enricher resolves account via resolveAccount(options)

### LINKED ISSUE #1487: storyboard runner: get_media_buys enricher should resolve account via resolveAccount(options) like create_media_buy
## Background

Surfaced while landing adcontextprotocol/adcp#3989 (the `inventory_list_targeting` storyboard fixture fix). The fixture-side patch unblocks the worked example, but the underlying defect lives in the runner's enricher asymmetry.

## The asymmetry

In `FIXTURE_AWARE_ENRICHERS`:

- **`create_media_buy`** enricher builds the request from scratch via `resolveAccount(options)`, which inherits `sandbox: true` from the test-kit defaults. The adapter routes to its sandbox account namespace.
- **`get_media_buys`** enricher returns only `{ media_buy_ids: [...] }`. The request-builder then merges in the storyboard fixture's raw `account` block, which has no `sandbox` flag unless the fixture author remembered to add one. The adapter routes to its production namespace.

Same fixture, two different `accountId`s across the create / get round-trip → framework-side `mediaBuyStore` cannot backfill `targeting_overlay` onto `get_media_buys` responses → `property_list` / `collection_list` assertions in `verify_create_persisted` and `verify_update_persisted` fail with `Expected ..., got undefined`.

## Why fixture-side `sandbox: true` is not the right durable fix

adcp#3989 lands `sandbox: true` on every account block in the fixture. That works, but it makes \"remember to set sandbox: true on every account fixture\" implicit storyboard-authoring lore. The next storyboard author who omits it will trip the same misrouted-namespace failure with no obvious diagnostic — the round-trip just silently misses the store and an assertion fails far from the cause.

## Proposed fix

Make the `get_media_buys` enricher (and any other lifecycle enricher that takes an account) inherit account resolution from the same `resolveAccount(options)` path `create_media_buy` uses, so that test-kit defaults (notably `sandbox`) flow through consistently regardless of what the fixture author wrote inline.

Alternatively, if the design intent is that the fixture's account is authoritative, document it and have the runner emit a warning when create-vs-get account references diverge in the same scenario.

## Refs

- adcontextprotocol/adcp#3989 — fixture-side fix that unblocks the worked example
- adcontextprotocol/adcp-client commit `5fd83f17` (#1453) — `Account.mode` trust boundary that introduced the sandbox-vs-production routing split
- adcontextprotocol/adcp-client#1415 — original SDK-side gap that #3989 closes from the fixture side

bokelley:
Closes #1487

## Summary

`get_media_buys` and `get_media_buy_delivery` enrichers did not include an `account` field, so the generic fixture-authoritative merge (`{ ...enriched, ...fixture }`) let the fixture's raw `account` block win. When a fixture's `account` lacks `sandbox: true`, the adapter routes to the production namespace — different from the sandbox namespace used by `create_media_buy`. This breaks the create→get round-trip: the framework-side `mediaBuyStore` cannot backfill `targeting_overlay` because the two calls resolve to different account IDs.

Fix: add both enrichers to `FIXTURE_AWARE_ENRICHERS` and resolve account via `context.account ?? resolveAccount(options)`, matching the pattern already used by `create_media_buy` and `update_media_buy`. Fixture fields other than `account` are preserved via explicit spread so storyboards that author filters, status, or pagination fields continue to work.

## What was tested

- `npm run typecheck` — pre-existing `TS2688`/`TS5107` environment errors only (present on `main` before this change)
- `npm run build:lib` — succeeded
- `node --test test/lib/request-builder.test.js test/lib/request-builder-schema-roundtrip.test.js` — **116 tests, 0 failures** (4 new regression tests added)

## Pre-PR review

- **code-reviewer:** approved — blocker (duplicate test title, missing delivery context.account test) addressed before this PR; nit (FIXTURE_AWARE_ENRICHERS comment parity) fixed
- **ad-tech-protocol-expert:** approved — non-breaking per spec; `account` is an optional field on both `GetMediaBuysRequest` and `GetMediaBuyDeliveryRequest` schemas; fixture-field preservation via spread addresses the filter/pagination drop blocker

## Nits (not fixed — noted for reviewers)

- `get_media_buy_delivery` uses `media_buy_ids` (plural array) consistent with the existing schema shape; a schema cross-check for singular vs plural is noted but out of scope for this fix.
- `ENVELOPE_FIELDS` omitting `correlation_id` is a pre-existing gap unrelated to this PR.

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01HZwDMPZBZmcMGQncxKgPwG

---
_Generated by [Claude Code](https://claude.ai/code/session_01HZwDMPZBZmcMGQncxKgPwG)_


bokelley:
**Fold candidate — Issue #1505.** The fix for #1505 (add a symmetric `record` hook on `update_media_buy` in `createMediaBuyStore`, then drop the `get_after_update` / `adcp-client#1505` entries from both `EXPECTED_FAILURES` blocks) overlaps directly with this PR's diff — those very entries were added here. The change is modest (~20–30 lines in `from-platform.ts`), so folding before merge avoids a sequential PR. Skip if you prefer to keep this PR's scope narrow; #1505 is tracked and will resurface on merge.

---

bokelley:
Folded — pushed as a follow-up commit (`e32105f`).

**Root cause (confirmed by static analysis):** the `create_media_buy` context extractor never captured `packages[0].package_id` from the create response, so `$context.package_id` in the storyboard's `update_media_buy` fixture was left as the literal string `"$context.package_id"` by `injectContext`. `mergeFromUpdate` then stored the overlay under that bogus key; `backfill` on `get_after_update` looked up the real seller-assigned id and found nothing.

The fix is two lines in `src/lib/testing/storyboard/context.ts` — not `from-platform.ts`. `persistTargetingOverlayFromUpdate` / `mergeFromUpdate` / `backfill` were already correct and wired up (from PR #1424); the pipeline only needed `context.package_id` populated upstream of the enricher.

Both `get_after_update` / `adcp-client#1505` entries are dropped from `EXPECTED_FAILURES`, and the changeset is updated to reference both #1487 and #1505.

---

bokelley:
Thanks for the root-cause write-up — the `context.package_id` gap in the context extractor is a good catch. Noted that `e32105f` lands the two-line fix in `context.ts`, drops both `get_after_update` / #1505 from `EXPECTED_FAILURES`, and updates the changeset to reference both issues. No further action needed from triage.

---
_Generated by [Claude Code](https://claude.ai/code/session_016e1AEuFymcUUpboB36q3L1)_

---

bokelley:
Acknowledged — glad the root-cause write-up was useful. Noted that `e32105f` closes the loop on both issues. No further triage action from me.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nbs89dvzRuG7Qmw3qUZ6Na)_

---

PR #1528, opened 2026-05-04
feat(cli): --soft-fail, --summary-file, and storyboard show --specialism (#1527)

bokelley:
Closes #1527

Adds three additive CLI features to `adcp storyboard` to address the silent-CI-failure pattern surfaced by `examples/multi_platform_seller/` having 14 masked storyboard failures across several PRs. All changes are in `bin/adcp.js` only; no library surface touched.

## Changes

**A. `--summary-file [PATH]`** — write a Markdown run summary to disk after the run. Defaults to `storyboard-result-summary.md` when `PATH` is omitted. Auto-activates from `$GITHUB_STEP_SUMMARY` for GitHub Actions job summary integration. Implemented for capability-driven (`comply()`), single-storyboard (`--file`), and `--local-agent` run paths. Multi-instance and agents-map paths emit a warning when the flag is passed explicitly (env-var activation is silenced to avoid noise in every CI run).

**B. `adcp storyboard show --specialism <slug>`** — resolves which storyboards are graded when an agent claims a specialism. Looks up the parent protocol from the compliance index (required by `resolveStoryboardsForCapabilities` to avoid `specialism_parent_protocol_missing` throw), then renders the full list with `[track]` labels and capability-gating annotations. `--json` output is structured. The positional-filter uses an index guard (`specialismSlug === null || i !== specialismIdx + 1`) to avoid dropping the storyboard ID on plain `show <id>` calls.

**C. `--soft-fail`** — exit 0 even when storyboards fail (suppresses exit 3 only; exit 1 and 2 are preserved). Prints `STORYBOARD FAILURES (N): scenario1, …` to stderr; silent in `--json` mode (failures already present in the JSON payload). Threaded through all 8 exit-code-3 sites: capability-driven, single `--file`, `--local-agent` (×3), multi-instance, agents-map (×2).

## What was tested

- `npm run format:check` — passed
- `node --check bin/adcp.js` — passed (syntax valid)
- `npm run typecheck` — pre-existing environment errors (missing `@types/node`, deprecated `moduleResolution=node10`); not introduced by this PR (confirmed by checking the base branch)
- `npm run build:lib` / `npm test` — cannot run in this environment (`tsx` not available); no `.ts` files were modified

[...]

PR #1533, opened 2026-05-04
feat(server): createDynamicRegistry — atomic hot-reloadable multi-registry bundles (#1531)

### LINKED ISSUE #1531: createDynamicRegistry helper for hot-reloadable tenant routing (atomic build-then-swap)
## Problem

Any adopter that hot-reloads tenants from a database hits the same recurring multi-registry-atomicity problem:

1. Multiple parallel registries — `adapters` (PlatformIdentity), `v6Platforms` (DecisioningPlatform), and increasingly `operationalPlatforms` (OperationalPlatform when shipped, see adcontextprotocol/adcp-client#TODO-issue-2).
2. A `refreshAdapters()` cycle reads from a config store, builds new versions of each registry from scratch, and atomically swaps them in.
3. Concurrent readers (per-request lookups) MUST never see a half-rebuilt mix where, say, `adapters` has a new storefront but `operationalPlatforms` doesn't.

We've now built this primitive **three times** in our agentic-adapters codebase:
- v5 `PlatformAdapter` registry refresh
- v6 `DecisioningPlatform` registry refresh
- v6 `OperationalPlatform` registry refresh (in scope3data/agentic-adapters#248)

Each iteration has its own hard-won lessons:

- `clear() + for-set` looks atomic but isn't across `await`. Caught in PR #248 round-2 review.
- Three separate pointer reassignments are atomic against synchronous readers but not across `await`. Caught in PR #248 round-3 review by Emma.
- Concurrent `refreshAdapters()` calls without an in-flight guard race — last-writer-wins regardless of which `store.list()` snapshot is fresher. Caught in PR #248 round-2 review.

Every adopter rebuilding this primitive will independently hit each of these. The pattern belongs in the SDK.

## Proposal

Ship `createDynamicRegistry()` that bakes the atomic-bundle-swap idiom + in-flight refresh guard:

```ts
import { createDynamicRegistry } from '@adcp/sdk/server'

const registry = createDynamicRegistry({
  registries: ['adapters', 'v6', 'operational'] as const,
  staticIds: () => Array.from(staticPlatformIds),
  refresh: async (pending) => {
    // adopter populates the pending Maps from their config store
    // returns a count or arbitrary metadata
    const configs = await store.list()
    for (const config of configs) {
      pending.adapters.set(config.platformId, buildAdapter(config))
      pending.v6.set(config.platformId, buildPlatform(config))
      pending.operational.set(
        config.platformId,
        deriveOperational(buildAdapter(config)),
      )
    }
    return configs.length
  },
})

// Per-request lookups
registry.get('adapters', platformId)
registry.get('operational', platformId)

// Hot-reload
await registry.refresh()  // overlapping calls coalesce automatically
```

Internals to bake in (each is a lesson from this PR cycle):

1. **Single-pointer atomic swap** — internal `let bundle: Bundle` reassigned in one statement, all reads go through `bundle.<name>.get(id)`. Atomic across `await`.
2. **In-flight refresh guard** — concurrent `refresh()` calls return the same Promise; sequential calls get fresh ones.
3. **Static-carry-forward** — adopter-supplied `staticIds()` are preserved across refreshes without rebuild.
4. **Unregister cleans all bundles** — `unregister(id)` deletes from every named registry in lock-step.
5. **Per-registry typed `get`** — returns `T | undefined` typed per registry name.

## Estimated savings per adopter

Our `router.ts` is ~430 LOC, of which ~150 LOC is the multi-registry plumbing. Each adopter doing the same pattern lands in the same range with the same hard-won bug fixes. Centralizing saves ~150 LOC × N adopters AND prevents N independent rediscoveries of the atomicity lessons.

## Concrete prior art

- [`packages/shared/src/server/router.ts`](https://github.com/scope3data/agentic-adapters/blob/bokelley/operational-platform-shim/packages/shared/src/server/router.ts) — current implementation. Note especially the `Registries` interface + single-pointer reassignment in `refreshAdapters()` (round-3 fix), and the `inflightRefresh` guard at module top (round-2 fix).
- PR scope3data/agentic-adapters#248 review thread — Emma's findings on atomicity claims and the iterations that followed.

## Related

- adcontextprotocol/adcp-client#1467 — `createDerivedAccountStore` (the precedent for "we kept rolling this; ship it")

bokelley:
Closes #1531

Adds `createDynamicRegistry<TRegistries>()` — a factory for adopters that maintain multiple correlated Maps of platform objects (adapters, v6 platforms, operational platforms, etc.) refreshed atomically from a config store. Centralises three atomicity lessons learned through repeated hand-rolled implementations in scope3data/agentic-adapters#248.

## What this does

**Single-pointer bundle swap.** All reads go through one `let bundle` pointer. The swap `bundle = pending` executes in one synchronous microtask-boundary step — readers never observe a half-rebuilt state where, e.g., `adapters` is at version N+1 while `v6` is at version N.

**In-flight refresh coalescing with `finally`-guarded release.** Concurrent `refresh()` calls return the same in-flight Promise. The guard is cleared in `finally`, so a thrown refresh never permanently freezes the registry. The previous bundle survives unchanged on any throw; the error is re-thrown to all coalesced callers.

**Static-id carry-forward with `unregister()` denylist.** `staticIds()` is polled on each `refresh()` (so the static set can change at runtime). For each id returned, if the `refresh` callback didn't populate it, the previous bundle's entry is preserved. `unregister(id)` removes from all Maps atomically and adds to a permanent denylist that suppresses `staticIds` carry-forward on all future refreshes. A future `refresh()` that explicitly populates the id in `pending` will make it available again.

## Type-safe API

```ts
import { createDynamicRegistry } from '@adcp/sdk/server';

const registry = createDynamicRegistry<{
  adapters: PlatformIdentity;
  v6: DecisioningPlatform;
  operational: OperationalPlatform;
}>({
  registries: ['adapters', 'v6', 'operational'],
  staticIds: () => Array.from(hardcodedPlatformIds),   // optional
  refresh: async (pending) => {
    const configs = await store.list();
    for (const config of configs) {
      pending.adapters.set(config.platformId, buildAdapter(config));
      pending.v6.set(config.platformId, buildPlatform(config));
      pending.operational.set(config.platformId, deriveOperational(config));
    }
  },
});

// Per-request lookups — typed per registry name:
const adapter = registry.get('adapters', platformId); // PlatformIdentity | undefined
const v6      = registry.get('v6', platformId);       // DecisioningPlatform | undefined

// Hot-reload — overlapping calls coalesce:
await registry.refresh();
```

Type parameter `TRegistries` is the value-type map per registry name. The `pending` callback parameter is typed `{ [K in keyof TRegistries]: Map<string, TRegistries[K]> }` — no `as const` footgun, no casting at call sites.

## What was tested

- `npx tsc --project tsconfig.lib.json` — pre-existing `@types/node` / `moduleResolution=node10` errors only (same as PR #1528; not introduced by this PR)
- `npm run build:lib` — `tsx` unavailable in this environment; ran `tsc` directly, output verified in `dist/`
- `node --test test/lib/dynamic-registry.test.js` — **12/12 tests pass**

Test coverage:
- Concurrent `refresh()` → same Promise identity (atomicity of in-flight guard)
- Sequential `refresh()` → distinct Promises
- Thrown `refresh()` → inflight guard cleared via `finally`, previous bundle survives, error rethrown to all callers
- Concurrent callers both receive the thrown error (callback's explicit `throw` after resolved deferred)
- Bundle swap atomicity — both registries reflect the same version synchronously
- `unregister()` removes from all Maps, returns `true/false`
- `staticIds` carry-forward on second refresh
- `staticIds` polled per refresh (set can expand at runtime)
- Unregistered static id suppressed from carry-forward
- Initial state: `get()` returns `undefined` before first `refresh()`
- `unregister()` before `refresh()` returns `false` without throwing

**Known nits (not fixed — out of scope):**
- Test for "staticIds polled per refresh" creates `registry2` mid-test; could be split into two focused tests for clarity.
- `registries: []` (empty array) is not guarded — `get()` on any name returns `undefined` safely (optional chaining), but `refresh()` produces an empty bundle silently. Document-only fix if desired.
- `createDynamicRegistry` not yet in `docs/llms.txt` or `docs/TYPE-SUMMARY.md` — post-merge housekeeping.

## Pre-PR review

- **code-reviewer:** approved — no blockers; flagged one test dead-code defect (concurrent-error test was testing rejected-deferred propagation rather than the callback's explicit `throw`) and a denylist memory note. Both fixed before PR open.
- **dx-expert:** approved with nits — confirmed all three original blockers resolved (type-map API shape, `staticIds` semantics, error recovery); two nits applied as fixup commit: doc correction on `unregister` re-registration semantics and `get()` optional-chaining safe-miss guard.

---

> issue). To unblock:
>
> - **Push fixup commits directly:** `gh pr checkout 1533` →
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01TX1rajPsAv3z64R4Y9DBPk

### DIFF
--- a/.changeset/create-dynamic-registry.md
+++ b/.changeset/create-dynamic-registry.md
@@ -1,0 +1,13 @@
+---
+"@adcp/sdk": minor
+---
+
+feat(server): add `createDynamicRegistry` for atomic hot-reloadable multi-registry bundles
+
+Adds `createDynamicRegistry<TRegistries>()` — a factory for adopters that maintain multiple correlated Maps (adapters, v6 platforms, operational platforms, etc.) refreshed together from a config store. Centralises three atomicity lessons learned through repeated hand-rolled implementations:
+
+1. Single-pointer bundle swap — `bundle = pending` in one synchronous statement; readers never see a half-rebuilt mix across multiple Maps.
+2. In-flight refresh guard — concurrent `refresh()` calls coalesce onto the same Promise; the guard is cleared in `finally` so a thrown refresh never permanently freezes the registry.
+3. Static-id carry-forward with `unregister()` denylist — `staticIds()` is polled on each refresh for ids to preserve; `unregister(id)` removes from the live bundle AND suppresses from future carry-forward.
+
+Exported from `@adcp/sdk/server`. Replaces ~150 LOC of multi-registry plumbing per adopter deployment.
--- a/src/lib/server/dynamic-registry.ts
+++ b/src/lib/server/dynamic-registry.ts
@@ -1,0 +1,259 @@
+/**
+ * `createDynamicRegistry` — atomic hot-reload helper for adopters that manage
+ * multiple correlated Maps of platform objects (adapters, v6 platforms,
+ * operational platforms, etc.) refreshed together from a config store.
+ *
+ * ## The problem this solves
+ *
+ * Any adopter that hot-reloads tenants from a database maintains multiple
+ * parallel Maps that must be swapped together — a reader must never see
+ * `adapters` at version N+1 while `v6Platforms` is still at N. Three
+ * atomicity pitfalls appear in every hand-rolled implementation:
+ *
+ *   1. `clear() + for-set` looks atomic but is not across `await`.
+ *   2. Three separate pointer reassignments are not atomic across `await`.
+ *   3. Concurrent `refresh()` calls without an in-flight guard race
+ *      (last-writer-wins regardless of which snapshot is fresher).
+ *
+ * This factory bakes in the correct idioms: single-pointer bundle swap,
+ * in-flight coalescing guard (with `finally`-based release so a thrown
+ * refresh never freezes the registry), and static-id carry-forward with a
+ * denylist that respects explicit `unregister()` calls.
+ *
+ * ## Usage
+ *
+ * ```ts
+ * import { createDynamicRegistry } from '@adcp/sdk/server';
+ * import type { PlatformIdentity, DecisioningPlatform } from '@adcp/sdk/server';
+ *
+ * const registry = createDynamicRegistry<{
+ *   adapters: PlatformIdentity;
+ *   v6: DecisioningPlatform;
+ * }>({
+ *   registries: ['adapters', 'v6'],
+ *   staticIds: () => Array.from(hardcodedPlatformIds),   // optional
+ *   refresh: async (pending) => {
+ *     const configs = await store.list();
+ *     for (const config of configs) {
+ *       pending.adapters.set(config.platformId, buildAdapter(config));
+ *       pending.v6.set(config.platformId, buildPlatform(config));
+ *     }
+ *   },
+ * });
+ *
+ * // Per-request lookups — typed per registry name:
+ * const adapter = registry.get('adapters', platformId); // PlatformIdentity | undefined
+ * const v6      = registry.get('v6', platformId);       // DecisioningPlatform | undefined
+ *
+ * // Hot-reload — overlapping calls coalesce onto the same Promise:
+ * await registry.refresh();
+ * ```
+ *
+ * ## Before the first `refresh()`
+ *
+ * All Maps are empty; `get()` returns `undefined` for every id. Callers
+ * that dispatch requests before the first refresh completes must handle
+ * `undefined` (or await the first `refresh()` during startup).
+ *
+ * ## Error behavior
+ *
+ * When `refresh()` throws, the previous bundle is preserved — the registry
+ * continues serving the last known-good state. The error is re-thrown to all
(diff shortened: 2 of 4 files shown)


bokelley:
Superseded by #1537 () — hand-shipped implementation of #1531 with the same atomicity model plus pinned-flag refinement, post-merge stale-pin fix (#1542), and 27 tests covering atomic-swap-across-await, in-flight refresh coalescing, lock-step unregister, and per-registry typed get. Closing this branch.

bokelley:
Acknowledged — closing this branch in favour of #1537. Thanks for the follow-through on the atomicity model and the expanded test coverage.

---

bokelley:
Understood — nothing further to do here. Good luck with #1537.

---

Issue #1530, opened 2026-05-04
First-class OperationalPlatform contract for in-process consumers (poller / fan-out / scheduled jobs)


bokelley:
## Problem

v6 `DecisioningPlatform` is shaped for buyer-facing MCP request dispatch. Every method takes a `RequestContext` carrying:

- `account` (resolved by `AccountStore.resolve` from `authInfo`)
- `state` (sync state reads — workflow steps, governance context)
- `resolve` (async framework-mediated fetches — property lists, formats)
- `ctxMetadata` (opaque-blob round-trip cache)
- `handoffToTask` (background-task hand-off)

These all assume there's an in-flight buyer MCP request to derive context from. Adopters with **in-process operational consumers** — background pollers, audience-sync watchers, scheduled jobs, storefront fan-out tasks — don't have an MCP request. They have a stored task with an access token (or no token at all, e.g. server-side internal scans). They can't honestly satisfy `RequestContext`. Every adopter doing operational work will reinvent this seam.

We did. PR scope3data/agentic-adapters#248 introduces a parallel `OperationalPlatform` interface in our shared package because the v6 surface didn't fit. The same shape will recur across every adopter that does anything beyond pure buyer-facing dispatch (which is approximately everyone running production traffic).

## Proposal

Ship `OperationalPlatform` as a first-class SDK concept alongside `DecisioningPlatform`. Five methods covering the real operational surface:

```ts
export interface OperationalPlatform<
  TCtx extends OperationalContext = OperationalContext,
> {
  readonly platformId: string

  /**
   * Synthesize a per-call context from a stored token. The SINGLE
   * documented credential-synthesis path outside `AccountStore.resolve`.
   */
  extractOperationalContext(
    args?: Record<string, unknown>,
    sessionToken?: string,
    requireAuth?: boolean,
  ): Promise<TCtx>

  /** Update an upstream media buy from a poller / fan-out caller. */
  updateMediaBuy?(
    ctx: TCtx,
    request: UpdateMediaBuyRequest,
  ): Promise<Result<UpdateMediaBuyResponse, PlatformError>>

  /** Fetch delivery metrics from a poller / dashboard caller. */
  getMediaBuyDelivery?(
    ctx: TCtx,
    mediaBuyId: string,
    startTime: string,
    endTime: string,
  ): Promise<Result<GetMediaBuyDeliveryResponse, PlatformError>>

  /** Poll upstream for current audience-sync status. */
  pollAudienceStatus?(
[...]

PR #1535, opened 2026-05-04
feat(server): credentialPolicy args-bag scan (#1529 L1)

### LINKED ISSUE #1529: Lock down credential-bearing surfaces in adapter args (security hardening)
## Problem

The framework currently allows adapters to read credentials from buyer-supplied request args. Concretely, every existing v6 adapter we ship reads patterns like:

```ts
extractContext(args, sessionToken) {
  const token =
    args.<platform>_access_token ??         // top-level
    args.context?.<platform>_access_token ??  // nested in context
    args.ext?.<platform>_access_token ??      // nested in ext
    sessionToken                              // authInfo bearer
  return { accessToken: token, ... }
}
```

This pattern is the natural shape of `extractContext` since v5 — the SDK never told adopters that buyer args are an unsafe credential channel — so it's spread across every adapter we have. Empirically across **three rounds of expert review** on PR scope3data/agentic-adapters#248 we kept finding new credential-smuggling vectors in storefront fan-out paths:

- **Round 1:** buyer's `request.context.<platform>_access_token` survives into downstream `extractContext`
- **Round 2:** allowlist drift between fan-out scrubbers
- **Round 3:** top-level `request.<platform>_access_token` AND `request.account: { brand: ... }` pivot — neither caught by the round-1 fix

The pattern recurs because the SDK boundary doesn't enforce what's safe to read from buyer args. Storefronts (any platform that fans out to upstream platforms acting as the buyer) are the highest-risk shape: the storefront's TLS, IP reputation, and request envelope all vouch for whatever credential the buyer-controlled args carry through. Confused-deputy by default.

## Proposal

Make the SDK boundary the single perimeter for credential-bearing data. Two levels, in priority order:

### Level 1 (cheap, high-value): Add a runtime credential discipline check

`createAdcpServer` (or `defineSalesPlatform`, `defineAudiencePlatform`, etc.) gains an optional `credentialPolicy: 'authInfo-only' | 'lax'` mode (default `'lax'` for back-compat, opt-in `'authInfo-only'` for hardened deployments). When set to `'authInfo-only'`, the framework's pre-dispatch hook scans the args bag for keys matching `/_access_token$|_secret$|_password$/i` and rejects with `INVALID_REQUEST` if any are present at any depth.

Adopters who legitimately accept buyer-supplied credentials (rare — typically just storefronts running in proxy mode) opt out per-tool with `credentialPolicy: 'lax'`.

### Level 2 (harder, real fix): Deprecate `args.*` credential reads in favor of the `authInfo` channel exclusively

`authInfo.token` is already the canonical bearer in v6. The SDK could:

1. Make the args type passed to `extractContext` schema-strict — only the wire-spec fields, no `passthrough()`.
2. Document that adapters MUST read credentials from `authInfo.token` and `account.ctx_metadata`, never from args.
3. Provide a migration path: ship a codemod or lint rule that flags `args.<platform>_access_token` reads in adapter code.

The existing `AccountStore.resolve` already gets `authInfo` and is the One Blessed Path for buyer-facing dispatch. The framework could close the args channel entirely.

## Concrete benefit

Three rounds of review on the same conceptual hole on a single PR, plus a class of bugs every storefront-shaped adopter will hit independently. Closing this at the SDK boundary kills the entire bug category instead of relying on every adopter to scrub correctly. Per-adopter scrubbing (what we shipped in scope3data/agentic-adapters#248) is recoverable but it's the wrong layer.

## Concrete prior art in this codebase

- [`packages/shared/src/platforms/managed/platform.ts:854` — `synthesizeArgsForFanout`](https://github.com/scope3data/agentic-adapters/blob/main/packages/shared/src/platforms/managed/platform.ts) is the per-storefront allowlist we hand-rolled. Belongs in the SDK.
- [`packages/shared/src/platforms/managed/tasks/update-media-buy.ts:scrubRequestForFanout`](https://github.com/scope3data/agentic-adapters/blob/bokelley/operational-platform-shim/packages/shared/src/platforms/managed/tasks/update-media-buy.ts) is the second hand-roll — same shape, different call site.
- [Round-3 review on PR #248](https://github.com/scope3data/agentic-adapters/pull/248#issuecomment-4367546641) lays out the three vectors in detail.

## DX note

`{ ...request, paused: true }` is a one-line natural pattern. `pickWireSpecFields(request)` is 15 lines and unnatural. The TS shape encourages the unsafe path. Whatever the SDK ships, the safe pattern needs to be the easier pattern.

## Related

- adcontextprotocol/adcp-client#1146 — earlier credential-discipline thread (CTX-METADATA-SAFETY)

bokelley:
## Summary

Closes #1529 L1. Opt-in `credentialPolicy: 'authInfo-only'` scans every incoming request's args bag for credential-shaped keys at any depth and rejects with `INVALID_REQUEST`. Closes the buyer-args credential-smuggling vector class observed across three rounds of review on PR scope3data/agentic-adapters#248 — top-level `<platform>_access_token`, nested `context.<platform>_access_token`, nested `ext.<platform>_access_token` — at the framework boundary, regardless of how each adapter's `extractContext` reads from args.

Default mode is `'lax'` (back-compat). Opt in to enforce.

## Design

`credentialPolicy` accepts a string shorthand or a config object:

```ts
// simple
createAdcpServer({ credentialPolicy: 'authInfo-only' })

// extensible patterns + per-tool overrides
createAdcpServer({
  credentialPolicy: {
    policy: 'authInfo-only',
    patterns: { extend: [/^bearer$/i, /credentials/i] },
    tools: { activate_signal: 'lax' },
  },
})
```

Default patterns: `_access_token$`, `_secret$`, `_password$`, `accessToken`, `refreshToken`. Per the [follow-up comment on #1529](https://github.com/adcontextprotocol/adcp-client/issues/1529#issuecomment-4367740748), the SDK ships configurable matchers from day one — adopters know their platform vocabulary best, and round-4 vectors land in adopter config rather than SDK releases.

The rejection envelope reports paths only (no values) and **bypasses `finalize`** so `injectContextIntoResponse` can't echo the offending `params.context` value back through the response. Tested explicitly.

Hook point: in `toolHandler`, before request schema validation. Uniform behavior across `validation: 'off' | 'warn' | 'strict'`.

## Files

- `src/lib/server/credential-policy.ts` (new) — `scanArgsForCredentials`, `resolveCredentialPolicyForTool`, `getCredentialPatterns`, `DEFAULT_CREDENTIAL_PATTERNS`, types
- `src/lib/server/create-adcp-server.ts` — `credentialPolicy?` field on `AdcpServerConfig`, scan wired into `toolHandler`
- `test/server-credential-policy.test.js` (new) — 25 tests covering scanner, policy resolution, all three round-N vectors, no-echo invariant, per-tool override, extensible patterns
- `docs/guides/CTX-METADATA-SAFETY.md` — defense-in-depth section
- `.changeset/credential-policy-l1.md` — minor bump

## Test plan

- [x] `npm run build` clean
- [x] `npm run lint` baseline (no new errors)
- [x] 25/25 new credential-policy tests pass
- [x] 1210/1210 server-side test suite pass (`test/server-*.test.js`) — no regression
- [x] Rejection envelope does NOT echo offending value back (explicit test)
- [x] All three round-N vectors caught in a single call (explicit test)

## What this is NOT

- L2 of #1529 (typed wire-spec args + `WireSafe<T>` brand at the operational boundary) ships separately after #1530's `OperationalPlatform` interface lands.
- A change to default behavior. `'lax'` is the default; `'authInfo-only'` is opt-in.


bokelley:
Author here, but reviewing this with fresh eyes — feedback from the consumer side now that I've read the diff cold.

## What I really like

- **`finalize` bypass for the rejection envelope.** This is exactly the right call — `injectContextIntoResponse` echoing `params.context` would have round-tripped the credential the scan just caught. Easy detail to miss, comment captures the why well.
- **Hook point BEFORE request schema validation.** Uniform across `'off' | 'warn' | 'strict'`. Adopters in `'off'` mode (e.g. running the comply test controller) get the same protection.
- **Cycle-safe via `WeakSet`.** A buyer who sends a self-referential JSON object can't infinite-loop the scanner. Defensive and correct.
- **"The next platform-specific vector lives in adopter config, not in the SDK"** — great framing in the changeset. Keeps the SDK honest, gives adopters real control.
- **`'lax'` default for back-compat.** Right call for a minor bump.

## Concerns worth addressing in this PR

### 1. Default patterns leave easy escapes

The defaults catch `_access_token$`, `_secret$`, `_password$`, `accessToken`, `refreshToken`. They miss several common shapes:

- **`api_key` / `apiKey`** — extremely common, esp. in retail-media adapters. `criteo_api_key`, `audiostack_api_key`, etc.
- **`_token$` (without `_access` prefix)** — `bearer_token`, `id_token`, `session_token`. Trivial to write, escapes the regex.
- **PascalCase variants of camelCase patterns** — `^accessToken$` is anchored exact. `AccessToken: 'x'` (PascalCase) slips past. The `_access_token$` regex uses `/i` but the camelCase ones don't. Asymmetry.

Suggested defaults extension:
```ts
/_token$/i,        // catches bearer_token, id_token, session_token
/api[_-]?key/i,    // catches api_key, apiKey, api-key
/^bearer$/i,
```

…and make the camelCase patterns case-insensitive (`/^accessToken$/i`).

### 2. The `account` ref pivot vector isn't covered — and shouldn't be conflated with credentials

The PR description says "closes the buyer-args credential-smuggling vector class" but Emma's round-3 review on scope3data/agentic-adapters#248 actually identified TWO categories:

- **Credentials** — `<platform>_access_token`, secrets, etc. → `credentialPolicy` handles ✓
- **Identity** — `request.account: { brand: 'attacker.com' }` to pivot onto a different ad account → NOT handled here

The mitigation for the second is `AccountStore.resolve` org-gating, not args-bag scanning. Worth either:

- Documenting explicitly: "this closes credentials but not identity smuggling — see `AccountStore.resolve` for that"
- Adding a separate `identityPolicy` flag (probably future work)

Recommend the doc note in this PR. Otherwise adopters who read this will think they're fully protected.

### 3. `tools: { activate_signal: 'lax' }` per-tool overrides are typo-fragile

If an adopter writes `tools: { activte_signal: 'lax' }` (typo), the override silently no-ops and `'authInfo-only'` applies — which would actually fail-closed correctly **for that case** but the broader concern is the other direction: adopter writes `tools: { my_legitimate_buyer_creds_tool: 'lax' }`, typos the tool name, and prod traffic to that tool starts rejecting. The framework knows the registered tool names (it registers them). Could validate at server construction that `Object.keys(policy.tools ?? {})` are all registered tools, throw at startup if not.

Could be follow-up — flag it now.

### 4. `field: hits[0]` under-discloses

`adcpError('INVALID_REQUEST', { field: hits[0], details: { credential_paths: hits } })` — `field` is an AdCP-spec single-path indicator. Reporting only the first hit is slightly arbitrary when 3+ vectors are present together. Adopters debugging client integrations might miss the full picture.

Two options:
- Set `field` to a non-path value like `'__credential_policy__'` (or omit) and put the full list in `details.credential_paths` only
- If the spec allows array-of-fields, surface all paths

Minor, but the current shape implies "this one path" when the truth is "any of these".

## DX observations

- Consider shipping a lint rule (or `npx adcp-doctor` mode) that scans an adopter's `extractContext` implementations for `args.<word>_access_token` reads. The pattern is so widespread in v5-shaped adapters that build-time detection would be a much bigger DX win than runtime rejection.
- For v7, recommend flipping default to `'authInfo-only'` and requiring `'lax'` to be explicit. Secure-by-default beats secure-by-opt-in for new adopters.
- Documentation in CTX-METADATA-SAFETY.md should call out concretely the three round-N vectors with example payloads. Adopters who don't know what's at stake won't opt in.

## Nits

- `getCredentialPatterns` is a one-liner used once. Could be inlined.
- `extend` + `matcher` precedence (matcher wins) is documented but could throw at config time when both are set, since the silent override is footgun-y.

bokelley:
Addressed in `e7dffa52`. Walking the comment top-to-bottom:

## In this PR

**1. Default patterns broadened** ✅
New set: `/_token$/i`, `/_secret$/i`, `/_password$/i`, `/api[_-]?key/i`, `/^bearer$/i`, `/^accessToken$/i`, `/^refreshToken$/i`. Dropped `_access_token$` (subsumed by `_token$`). camelCase patterns now case-insensitive — catches `AccessToken`, `ACCESSTOKEN`, etc. Test coverage for each new vector.

Punted on broader `*Token$` PascalCase suffix — `paymentToken` could be a legitimate wire field. Documented in the new pattern-set comment block; adopters extend.

**2. Identity-pivot disclaimer** ✅
New section in `CTX-METADATA-SAFETY.md`: "What `credentialPolicy` does NOT cover" with the concrete `account: { brand: 'attacker.com' }` example and the redirect to `AccountStore.resolve` org-gating. Explicit: "If your resolver doesn't gate on the principal's authorized accounts, `credentialPolicy: 'authInfo-only'` will not save you."

**3. Typo'd `tools[]` keys throw at construction** ✅
New `validateCredentialPolicy(policy, knownToolNames)` runs after the tool-registration loop. `tools: { activte_signal: 'lax' }` throws with the registered tool list in the error message before first traffic. Test covers the typo path.

**4. `field` dropped from rejection envelope** ✅
Was `field: hits[0]` (arbitrary single-path), now omitted. Full list still in `details.credential_paths`. Test asserts `field === undefined`.

**5. `extend` + `matcher` both set throws** ✅
Previous behavior silently chose `matcher`. Now throws both at the runtime helper and at construction-time validation. Tests cover both paths.

**6. Round-N example payloads in docs** ✅
Three concrete jsonc examples (top-level, nested context, nested ext) in `CTX-METADATA-SAFETY.md` showing exactly what `'authInfo-only'` rejects.

**Nit: `getCredentialPatterns`** ✅
Inlined into `create-adcp-server.ts`. One ternary.

## Deferred (with reasons)

**v7 default flip to `'authInfo-only'`** — agreed in principle, but a major-bump default change. Filing separately.

**Lint rule for v5 `extractContext` patterns** — bigger DX project. The `args.<word>_access_token` pattern is so widespread that build-time detection would beat runtime rejection. Worth its own issue (`@adcp/sdk/lint` package?). Will file.

## Test results

- 36/36 credential-policy tests pass (+11 from previous 25)
- 1199/1199 server suite pass (no regressions)
- format / typecheck / lint clean

Ready for re-review.

bokelley:
Thanks for the thorough walkthrough — the six addressed items look well-scoped and the deferred items (v7 default flip, lint rule) are sensible punts with clear rationale. 36/36 + 1199/1199 with no regressions is a clean baseline for re-review.

---

bokelley:
Reviewed commit `e7dffa5` against all four concerns. All addressed — verified locally (36/36 tests pass after rebuild):

1. **Default patterns broadened** ✓ — `_token$/i`, `api[_-]?key/i`, `/^bearer$/i` added; `accessToken`/`refreshToken` now case-insensitive; `_access_token$` dropped (subsumed by `_token$`).
2. **Identity-pivot vector documented** ✓ — CTX-METADATA-SAFETY.md now has a clear "what `credentialPolicy` does NOT cover" section pointing to `AccountStore.resolve` org-gating.
3. **Typo'd `tools[]` keys throw at construction** ✓ — `validateCredentialPolicy(policy, knownToolNames)` runs post-registration loop; `activte_signal` gets caught before first traffic.
4. **`field` under-disclosure fixed** ✓ — `field` dropped entirely from the rejection envelope; full path list lives in `details.credential_paths` only.

Also caught that the upstream commit had a 5th fix (concurrent `extend`+`matcher` throws rather than silently dropping `extend`), which is a good catch.

One note for reviewers: the dist in this repo isn't committed, so the test failures visible in the pre-push validation were a stale-dist artifact — source is correct and all 36 tests pass against a fresh build.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01MmepaE5djoQso4ordzDyfP

---

bokelley:
Glad the walkthrough landed cleanly. Marking ready for re-review — let me know if anything surfaces on your pass.

---

bokelley:
Acknowledged — nothing to flag from my side on this pass. The design looks clean and the test coverage is solid. Ready for CODEOWNERS review.

---

bokelley:
Thanks — noted. No further action from this side; standing by for CODEOWNERS review.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_016EyH4GRNRLKRnMttiVYYaH_

---

bokelley:
Ran the PR by three experts (security-reviewer, code-reviewer, ad-tech-protocol-expert) in parallel. Eight findings, all addressed in `7412f0ab`.

## Protocol-correctness — error code change

**`INVALID_REQUEST` → `PERMISSION_DENIED`** (per ad-tech-protocol-expert review)

The reviewer's spec citation was decisive: AdCP request schemas set `additionalProperties: true`, so a credential-bearing payload is structurally well-formed. `INVALID_REQUEST` is for schema violations; this is a seller POLICY violation. Spec text on `PERMISSION_DENIED` matches exactly: *"the authenticated caller is not authorized for the requested action under the seller's own policies"* (`enums/error-code.json:99`).

New envelope shape:
```json
{
  "code": "PERMISSION_DENIED",
  "message": "Request args carry credential-shaped keys. Credentials must arrive on authInfo, not in the request body.",
  "recovery": "correctable",
  "details": {
    "scope": "credentials",
    "credential_paths": ["snap_access_token", "context.linkedin_access_token"]
  }
}
```

Mirrors the existing buyer-agent-status rejection (`details.scope: 'agent'`).

## Security findings

- **`validateCredentialPolicy` ordering bug**: was running before `get_adcp_capabilities` was registered, so `tools: { get_adcp_capabilities: 'lax' }` would spuriously throw. Moved to after the capabilities-tool registration.
- **Getter defense**: walker now uses `Object.getOwnPropertyDescriptors` and only recurses into data descriptors. Credential-named getters are flagged by name fail-closed, but the getter is never invoked. Defends hand-built / proxy-trapped inputs against throw-and-log-`params` DoS.
- **Default patterns broadened**: added `^authorization$/i`, `^cookie$/i`, `private[_-]?key/i` (HTTP-header smuggling, JWT/HTTPSig signing keys).

## Code quality findings

- **`/g` and `/y` flag stripping** in `extend` regexes. `/credentials/gi` is a natural footgun — `RegExp.prototype.test` advances `lastIndex` and would skip alternating inputs across calls. `/i` preserved.
- **Public surface trimmed**: `validateCredentialPolicy` and `resolveCredentialPolicyForTool` marked `@internal` and dropped from `index.ts`. Adopters can't construct `knownToolNames` outside the framework anyway.
- **`ResolvedMatcher` interface inlined** as a function type — single-method interfaces add no type safety.

## New test coverage

- Getter defense: credential-named getter is flagged by name without invocation
- `/g`/`/y` stripping (state-leak between calls)
- `/i` preservation when stripping `/g`/`/y`
- `matcher.path` parameter exercise (was untested)
- **Idempotency-ordering pin**: a rejected credential-bearing payload does NOT poison the replay cache, so a retry with the same `idempotency_key` and a CLEAN payload succeeds. Locks in the dispatcher ordering.
- New defaults: `authorization` / `cookie` / `private_key` variants

## Deferred

Three items kept out of L1 scope, will file as follow-ups:

1. **Spec PR on `adcontextprotocol/adcp`** clarifying credentials MUST arrive on `authInfo` (RFC 6750 §2 alignment).
2. **Scanning `ctx.authInfo.extra`** as part of L2's broader perimeter.
3. **Per-tool `{ allow: [...] }` granular allowlist** (L2 — narrower than the current `'lax'` opt-out).

## Test results post-fix

- 43/43 credential-policy tests pass (+7 from prior 36)
- 1238/1238 server suite pass
- Format / typecheck / lint clean

Ready for admin merge once CI greens.

bokelley:
Acknowledged. The `INVALID_REQUEST` → `PERMISSION_DENIED` correction is the right call per spec — credential-bearing payloads are structurally valid, so this is purely a policy violation on the seller's side. The getter defense and `/g`/`/y` stripping are solid hardening. Will stand by for CI.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_012Rg1koR28wmheYGdw8kbNT

---

PR #1536, opened 2026-05-04
feat(server): OperationalPlatform interface (#1530)

bokelley:
Closes #1530.

## Summary

Adds a first-class SDK contract for in-process operational consumers — price-optimization pollers, audience-sync task pollers, scheduled jobs, storefront fan-out paths — that don't carry an MCP request and therefore can't honestly satisfy `RequestContext`.

Distinct from `DecisioningPlatform` (buyer-facing MCP dispatch). Five-method interface:

| Method | Required | Purpose |
|---|---|---|
| `extractContext(args, sessionToken?, requireAuth?)` | ✅ | Synthesize per-call context. Single credential-synthesis path outside `AccountStore.resolve`. |
| `updateMediaBuy(ctx, request)` | ✅ | Storefront fan-out + price-step pollers. |
| `getMediaBuyDelivery(ctx, id, start, end, args?)` | ✅ | Pollers + dashboards. |
| `pollAudienceStatus?(platformData, accessToken)` | optional | Audience-sync pollers only. |
| `getProducts?(ctx, brief, contextId?, brand?, sourceChain?)` | optional | Storefront bundle composition only. |

## Design calls

- **Throw `AdcpError`, not `Result<T, E>`** (per the post-merge sequencing decision on PR #1535). Matches `DecisioningPlatform`'s convention. Avoids forcing `neverthrow` as a peer dep on every adopter — the shim used `Result` for storefront-internal ergonomics, but for an SDK contract that an adopter learns once, throw-based aligns with everything else they touch.

- **Sibling of `DecisioningPlatform`, not a field on it.** Operational fan-out is a wiring pattern inside an adopter's sales handler (or a poller's main loop), not a separate AdCP surface. Keeping `OperationalPlatform` as a top-level interface avoids baking it into the `DecisioningPlatform` surface until a second pattern wants it.

- **Combined `extractContext(args, sessionToken?, requireAuth?)` signature.** Tests pin the three call patterns (poller / storefront fan-out / server-side scan). Per the [follow-up comment on #1530](https://github.com/adcontextprotocol/adcp-client/issues/1530#issuecomment-4367741058), the post-migration split into `synthesizeFromToken` / `synthesizeFromArgs` ships as a follow-up once v5 shells are retired. Today's combined shape lets v5 adapters duck-type-satisfy without a wrapper.

[...]

PR #1537, opened 2026-05-04
feat(server): createDynamicRegistry — multi-registry hot-reload (#1531)

### LINKED ISSUE #1531: createDynamicRegistry helper for hot-reloadable tenant routing (atomic build-then-swap)

bokelley:
Closes #1531.

## Summary

Packages the multi-registry-atomicity idiom every adopter that hot-reloads tenants from a database independently rebuilds. The shim in scope3data/agentic-adapters built this three times before the pattern crystallized.

```ts
import { createDynamicRegistry } from '@adcp/sdk/server';

interface MyRegistries {
  adapters: PlatformIdentity;
  v6: AnyDecisioningPlatform;
  operational: OperationalPlatform;
}

const registry = createDynamicRegistry<MyRegistries>({
  registries: ['adapters', 'v6', 'operational'] as const,
  refresh: async pending => {
    for (const tenant of await store.list()) {
      pending.adapters.set(tenant.id, build(tenant));
      pending.v6.set(tenant.id, buildPlatform(tenant));
      pending.operational.set(tenant.id, deriveOperational(tenant));
    }
  },
});

registry.register('adapters', 'snap', snapAdapter, { pinned: true });
registry.register('v6', 'snap', snapPlatform, { pinned: true });

await registry.refresh();  // concurrent calls coalesce
registry.get('adapters', tenantId);
```

## Five lessons baked in

| # | Lesson | Test |
|---|---|---|
| 1 | Single-pointer atomic swap (concurrent reader sees consistent snapshot across `await`) | "concurrent reader during refresh sees old bundle until swap" |
| 2 | In-flight refresh guard (concurrent calls coalesce onto reference-equal Promise) | "concurrent refresh() calls coalesce" |
| 3 | Pinned-carry-forward (pin always wins; callback can't override) | "pin always wins: callback writing a pinned id is silently overridden" |
| 4 | Lock-step unregister (clears every registry in one op) | "removes from every registry in one operation" |
| 5 | Per-registry typed `get` (TS-side; via `TRegistries` generic) | n/a — type-only |

## Design refinements over the original issue

Two changes from the issue's pinned-flag comment thread:

1. **`pinned: true` flag at register time** replaces the parallel `staticIds()` callback / Set the shim used. Single source of truth, no drift hazard.
2. **Duplicate registration throws by default.** Silent tenant clobbering doesn't ship to production. `{ overwrite: true }` opt-in for the rare legitimate replace.

## Subtle implementation details

- **`refresh()` is non-async deliberately.** `async refresh()` would wrap the inflight Promise per-caller and break the reference-equal coalesce contract. The non-async signature returns the shared inflight Promise directly. Pinned by a test (line 224).
- **Inflight cleared in `finally`.** Even when the callback throws, the next `refresh()` call gets a fresh invocation. Pinned by a test ("inflight is cleared even when callback throws").
- **Pin snapshot captured at refresh-start.** Adopters who call `register({ pinned: true })` between refresh start and the callback's first `await` still get carry-forward.
- **Pin always wins at swap time.** The new bundle merges pending values FIRST, then pinned values from the live bundle on top — so a callback that writes a pinned id into `pending` is silently overridden.

## Files

- `src/lib/server/dynamic-registry.ts` (new) — module
- `test/server-dynamic-registry.test.js` (new) — 22 tests
- `.changeset/dynamic-registry.md` (new) — minor bump

## Test plan

- [x] 22/22 new dynamic-registry tests pass
- [x] 1260/1260 server suite pass — no regression
- [x] format / typecheck / lint clean
- [x] Concurrency tests use `block` Promises to deterministically pause refresh mid-callback and verify atomic-swap behavior

## Independent of #1530 / #1535

This PR doesn't depend on `OperationalPlatform` (#1536) or `credentialPolicy` (#1535). The `TRegistries` generic is shape-agnostic — adopters use it for any heterogeneous Map-of-Maps need. Common shape is the v5/v6/operational triple shown above; nothing in the SDK requires that shape.

### DIFF
--- a/.changeset/dynamic-registry.md
+++ b/.changeset/dynamic-registry.md
@@ -1,0 +1,9 @@
+---
+'@adcp/sdk': minor
+---
+
+Add `createDynamicRegistry<TRegistries>` — multi-registry plumbing with atomic-bundle-swap, in-flight refresh coalescing, and pinned-carry-forward semantics. Packages the multi-registry-atomicity idiom every adopter that hot-reloads tenants from a database independently rebuilds (the shim in scope3data/agentic-adapters built this three times before the pattern crystallized).
+
+Five lessons baked in: single-pointer atomic swap (concurrent readers see consistent snapshots across `await`), in-flight refresh coalescing (parallel `refresh()` calls share one Promise), pinned-carry-forward (entries with `{ pinned: true }` survive every refresh; pin always wins over `pending` writes), lock-step unregister (clears across all registries), per-registry typed `get`.
+
+Two design refinements over the original issue: `pinned: true` flag at register time replaces the parallel `staticIds()` Set (single source of truth, no drift hazard); duplicate registration throws by default (`{ overwrite: true }` opt-in) so silent tenant clobbering doesn't ship to production. See #1531.
--- a/src/lib/server/dynamic-registry.ts
+++ b/src/lib/server/dynamic-registry.ts
@@ -102,6 +102,13 @@ export interface DynamicRegistryRegisterOptions {
    * Use for built-in tenants registered at startup. Dynamic tenants
    * loaded from config should NOT be pinned — they live in `pending`
    * and rebuild on every refresh.
+   *
+   * **Pinned values are immutable until `unregister`.** Refresh-source
+   * updates do not propagate to pinned entries. Adopters who rotate
+   * credentials for a pinned tenant must explicitly call
+   * `unregister(id)` and `register()` again with the new value — a
+   * `refresh()` cycle alone will not pick up the rotation. If you want
+   * refresh-driven updates, do not pin.
    */
   pinned?: boolean;
 
@@ -177,7 +184,9 @@ export interface DynamicRegistry<TRegistries extends RegistryShape> {
    * the bundle from scratch; pinned entries carry forward. Concurrent
    * calls coalesce onto the same Promise.
    *
-   * Throws if no `refresh` callback was supplied at construction.
+   * Returns a rejected Promise if no `refresh` callback was supplied
+   * at construction (does not throw synchronously — callers can `await`
+   * uniformly).
    */
   refresh(): Promise<void>;
 }
@@ -222,6 +231,21 @@ export function createDynamicRegistry<TRegistries extends RegistryShape>(
 ): DynamicRegistry<TRegistries> {
   const names = [...config.registries];
 
+  // Validate construction: empty registries are useless; duplicate names
+  // would shadow each other in the bundle Maps and produce confusing
+  // "unknown registry" errors at lookup time. Surface both at
+  // construction with specific messages.
+  if (names.length === 0) {
+    throw new Error('createDynamicRegistry: `registries` must contain at least one name.');
+  }
+  const dedupedNames = new Set(names);
+  if (dedupedNames.size !== names.length) {
+    const dupes = names.filter((n, i) => names.indexOf(n) !== i);
+    throw new Error(
+      `createDynamicRegistry: duplicate registry name(s): ${dupes.map(n => JSON.stringify(n)).join(', ')}.`
+    );
+  }
+
   function emptyBundle(): InternalBundle<TRegistries> {
     const maps = {} as InternalBundle<TRegistries>['maps'];
     const pinned = {} as InternalBundle<TRegistries>['pinned'];
@@ -243,7 +267,7 @@ export function createDynamicRegistry<TRegistries extends RegistryShape>(
   let inflight: Promise<void> | null = null;
 
   function assertKnownRegistry(name: string): void {
-    if (!names.includes(name as keyof TRegistries & string)) {
+    if (!dedupedNames.has(name)) {
       throw new Error(
         `createDynamicRegistry: unknown registry name ${JSON.stringify(name)}. ` +
           `Known names: ${names.map(n => JSON.stringify(n)).join(', ')}.`
@@ -264,6 +288,14 @@ export function createDynamicRegistry<TRegistries extends RegistryShape>(
             `Pass { overwrite: true } if the duplicate is intentional.`
         );
       }
+      // Non-pinned register during in-flight refresh writes to the
+      // LIVE bundle. The live bundle becomes `liveBundle` in the
+      // refresh closure; at swap, the new bundle is built from
+      // `pending` ∪ `liveBundle.pinned` — so a non-pinned mid-refresh
+      // register is silently dropped at swap time. This is documented
+      // semantics; callers who race `register` against `refresh`
+      // either pin the entry (so it survives the swap) or serialize
+      // the operations.
       map.set(id, value);
       if (options?.pinned === true) {
         bundle.pinned[name].add(id);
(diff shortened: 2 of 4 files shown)


bokelley:
Two expert reviews back (security, code-reviewer). Eight findings, all addressed in `c2cad8c0`.

## Race semantics — newly documented + tested
Both reviewers flagged the lost-write footgun: a non-pinned `register()` during in-flight refresh writes to the live bundle, but the swap rebuilds from `pending` ∪ `liveBundle.pinned` only, so the new entry vanishes at swap. Three new tests pin the behavior:
- non-pinned register-during-refresh dropped at swap
- pinned register-during-refresh survives
- unregister-during-refresh removes the entry post-swap

JSDoc on `register` documents the contract: adopters racing operations either pin or serialize.

## Pinned-rotation surprise — louder doc
Per security M2: adopters rotating credentials for a pinned tenant via the refresh source will be surprised pinned values are immutable. Added explicit JSDoc warning on `pinned`: "Pinned values are immutable until `unregister`. Adopters who want refresh-driven updates don't pin."

## Test determinism
Replaced `await new Promise(setImmediate)` with sentinel `started` promises the refresh callback resolves as its first synchronous statement. Deterministic across Node versions / future microtask-scheduling refactors.

## Construction-time validation
Added throws for empty `registries: []` and duplicate registry names. Surfaces clearer errors than the silent shadowing or "unknown registry" misdirect they'd otherwise produce.

## Hot-path improvement
`assertKnownRegistry` swapped from `Array.includes` (O(n)) to `Set.has` (O(1)). Marginal but free.

## Future-proofing
`// CRITICAL: do not mark async` comment promoted from inside the function body to the function declaration line. Future refactors scanning signatures catch it.

## JSDoc tightening
`refresh()` returns a rejected Promise (not "throws synchronously") when no callback is configured. Adopters can `await` uniformly without try/catch on the construction-config check.

## Tests
- 27/27 pass (was 22, +5 new). Format / typecheck clean.

## Deferred
- Soft-disable for pinned entries (security L1) — incident-response ergonomics. Filing as separate issue.
- Memory-growth note for static-only adopters (security L3) — doc note for follow-up.

Ready for re-review.

bokelley:
## Re-review

All 8 prior findings are clean. Two experts reviewed the updated diff. Both sign off **approved with nits** — one semantic issue to address before merge, three nits.

---

### Semantic issue — stale pin on `overwrite: true` (code-reviewer)

**`register()` does not clear a stale pin when overwriting without `{ pinned: true }`.**

```ts
// Current behavior
r.register('adapters', 'snap', valueA, { pinned: true });
r.register('adapters', 'snap', valueB, { overwrite: true }); // no pinned: true
// bundle.pinned['adapters'] still contains 'snap'
// → 'snap' carries forward on next refresh, contrary to caller intent
```

The pin Set is appended on `register` but never cleared on overwrite. Fix:

```ts
map.set(id, value);
if (options?.pinned === true) {
  bundle.pinned[name].add(id);
} else {
  bundle.pinned[name].delete(id); // ← add this
}
```

Add a test: register pinned, overwrite without `pinned`, call `refresh()`, assert the entry is gone.

---

### Nits

**n1 (code-reviewer):** Test file header comment and PR body checklist both say "22 tests"; actual count after the fix commit is 27. Update the comment/checklist so the number stays accurate.

**n2 (dx-expert):** `registries` JSDoc doesn't mention the `as const` requirement. Without it TypeScript widens to `string[]` and `get`/`register` silently lose per-registry types. One line: _"Pass `as const` to preserve per-registry value types in `get` and `register`."_

**n3 (dx-expert):** No `has(name, id)` method — adopters will fall back to `get(...) !== undefined` for presence checks. Low cost to add; optional if you'd rather keep the surface minimal.

---

**Pre-PR review:**
- code-reviewer: approved with nits — stale-pin issue + test-count nit
- dx-expert: approved with nits — `as const` JSDoc gap + missing `has()`

Session: https://claude.ai/code/session_016fgGnkRfXtM3mq6PxP4LSQ

---

bokelley:
Applied the semantic fix and nits n2/n3 in draft PR #1542. The four nits from the review are surfaced in the PR body for the reviewer's discretion — not pre-fixed, per convention.

Nit n1 (test-count in the PR body checklist saying "22/22") refers to this closed PR's body, which can't be updated — noted for awareness only; the actual test count is 29 after #1542.

---

bokelley:
Acknowledged — noted that the semantic fix and nits n2/n3 landed in #1542, and that n1 (test-count in this closed PR's body) is an immutable artifact. Nothing further to act on here.

---

Issue #1541, opened 2026-05-04
feat(eslint-plugin): phase 1 — `no-credential-read-from-args` rule in new `@adcp/eslint-plugin` workspace


bokelley:
## Problem

Surfaced during expert DX review of #1535.

The `args.<vendor>_access_token` pattern in adopter `extractContext` implementations is widespread — every v5 adapter we ship reads it, and the pattern recurs in adapter-side fan-out paths because `extractContext`'s `args: Record<string, unknown>` makes the unsafe read trivial.

Runtime rejection via `credentialPolicy: 'authInfo-only'` (#1535) catches these at dispatch time, but build-time detection beats runtime detection — adopters see the issue when they write the code, not when a buyer's request hits production.

## Proposal

Ship `@adcp/sdk/lint` (or `npx adcp-doctor` mode) — a static-analysis tool that scans adopter codebases for the `extractContext`-credential-read antipattern.

Concrete starting rule set:

1. **`no-credential-read-from-args`** — flag `args.<word>_access_token` / `args.<word>_secret` / `args.<word>_password` / `args.context.<word>_access_token` reads in any function whose name matches `/^extract(Context|Operational)Context$/` or whose declaring object has `OperationalPlatform` as a type.
2. **`require-account-store-resolve-for-credentials`** — flag any `args.access_token` read outside a `resolveAccount` callback or `extractContext`.
3. **`prefer-authinfo-credential-channel`** — suggestion-level: code that reads from `args.<credential>` should consider `ctx.authInfo.token` / `ctx.authInfo.credential` instead.

Distribution:
- (a) ESLint plugin (`@adcp/sdk/eslint-plugin`) — fits adopter CI, low integration cost
- (b) `npx adcp-doctor` mode — one-shot scan, useful for adopters not on ESLint

Recommend (a) primary, (b) as wrapper.

## Related

- #1535 (L1 credentialPolicy) — runtime sibling
- DX review on #1535: dx-expert recommended this as the bigger DX win than runtime rejection
- v7 default flip (separate tracking) — lint ships as the migration path

bokelley:
## Triage

**Classification:** Feature request
**Bucket(s):** library, dx
**Status:** ready-for-human

**What the experts said:**
[...]

Issue #1540, opened 2026-05-04
OperationalPlatform: split extractContext into synthesizeFromToken / synthesizeFromArgs


bokelley:
## Problem

Follow-up from #1530 / #1536.

The shipped `OperationalPlatform.extractContext(args, sessionToken?, requireAuth?)` overloads three distinct call patterns into one method:

| Caller | \`args\` | \`sessionToken\` | Goal |
|---|---|---|---|
| Poller (price opt, audience) | \`{}\` | stored token | synthesize from token |
| Storefront fan-out | scrubbed buyer args | sometimes | derive from request, fallback to token |
| Scheduled job / server-side scan | sometimes | sometimes | flexible |

The combined signature is a v5-compat artifact. v5 \`PlatformAdapter.extractContext\` has the same shape, so v5 adapters duck-type-satisfy the v6 \`OperationalPlatform\` during migration without a wrapper. Once v5 shells retire, the SDK can do better.

## Proposal

Post-v5-retirement, split into two methods:

\`\`\`ts
interface OperationalPlatform<TCtx> {
  /** Poller / scheduled-job path. Token required. */
  synthesizeFromToken(sessionToken: string): Promise<TCtx>;

  /** Fan-out path. Args are scrubbed buyer input; token is optional fallback. */
  synthesizeFromArgs(
    args: WireSafe<RequestType>,  // assumes #1529 L2
    fallbackToken?: string,
  ): Promise<TCtx>;
}
\`\`\`

Two methods, two contracts, no overloading. Caller picks obviously. Type system enforces which fields each caller MUST have.

## Migration path

1. Add \`synthesizeFromToken\` / \`synthesizeFromArgs\` as optional methods alongside \`extractContext\`.
2. Dispatcher prefers the new methods when present; falls back to \`extractContext\` when not.
3. Mark \`extractContext\` \`@deprecated\` with a one-line migration note.
4. Drop in a major version.

## Status

Tracked. Implementation depends on:
- v5 shells retiring across reference adapters
- #1529 L2 (\`WireSafe<T>\`) landing for the typed \`synthesizeFromArgs.args\` parameter

## Related

- #1530 / #1536 — base
- #1529 L2 — typed args dependency

bokelley:
## Triage

**Classification:** Feature request
**Bucket(s):** library
**Status:** deferred

[...]

Issue #1539, opened 2026-05-04
credentialPolicy: scan ctx.authInfo.extra in L2's broader perimeter


bokelley:
## Problem

Surfaced during expert security review of #1535.

L1 (#1535) scans the buyer args bag at any depth. Custom authenticators that stamp credential-shaped values into `authInfo.extra` (e.g. token-introspection responses, signature claim sets, OAuth scope blobs) bypass the L1 scan entirely — `extra` is on `ctx.authInfo`, not `params`.

Concrete vector: an adopter with a custom JWT authenticator stores a refresh token claim into `authInfo.extra.refresh_token` for downstream use. A buggy `BuyerAgentRegistry.resolve` factory that throws an error embedding `extra` lands the refresh token in `logger.error('reason', { extra })`. The redactor at `src/lib/server/redact.ts` only catches labeled credentials and ≥32-char base64 blobs; shorter or unlabeled tokens slip through.

## Proposal

Extend the credentialPolicy scan to cover `ctx.authInfo.extra` when `policy: 'authInfo-only'`:

```ts
// pseudocode for the dispatcher
if (effectivePolicy === 'authInfo-only') {
  const argHits = scanArgsForCredentials(params, patterns);
  const extraHits = ctx.authInfo?.extra
    ? scanArgsForCredentials(ctx.authInfo.extra, patterns).map(p => `authInfo.extra.${p}`)
    : [];
  const hits = [...argHits, ...extraHits];
  if (hits.length > 0) reject(...);
}
```

Open questions for adopter feedback before implementation:
1. Should the scan run at every dispatch (perf cost) or once per resolved authInfo (cached)?
2. Should it apply when `authInfo.extra` is opaque-typed (`Record<string, unknown>`) vs. when adopter narrows the type?
3. Should this be its own opt-in flag (`credentialPolicy.scanAuthInfo: true`) so adopters who legitimately stash structured data in `extra` aren't broken?

Recommend: ship with #1529 L2 since it's the natural perimeter expansion; gate on `credentialPolicy.scanAuthInfo: true` (default `false`) to avoid breaking adopters with structured `extra`.

## Related

- #1535 (L1 credentialPolicy) — base
- #1529 (L2 wire-safe args) — fold in

bokelley:
## Triage

**Classification:** Feature request — security hardening
**Bucket(s):** library
**Status:** ready-for-human

**What the experts said:**
[...]

PR #1544, opened 2026-05-04
feat(server): WireSafe<T> brand + pickWireSpecFields (#1529 L2)

### LINKED ISSUE #1529: Lock down credential-bearing surfaces in adapter args (security hardening)

bokelley:
Closes #1529.

## Summary

Ships the L2 half of the credential-discipline plan designed in `.context/operational-fanout-design.md`.

L1 (#1535 credentialPolicy) catches credential-shaped keys at the buyer-facing dispatch boundary. **L2 catches structural leakage at the operational fan-out boundary** — storefront fan-out code that picks per-target args from a buyer request and forwards them upstream, where the buyer-facing scan has already passed.

## Three pieces

1. **`WireSafe<T>` brand** via `unique symbol`. `{...buyerReq}` is `T`, not `WireSafe<T>` — so passing it where `WireSafe<T>` is required is a compile error. The brand is the load-bearing element: code that spreads a buyer request directly cannot satisfy it.

2. **`pickWireSpecFields(req, schemaName)`** — canonical constructor of `WireSafe<T>`. Strips a buyer request to the AdCP wire-spec field allowlist for `schemaName`, returns a branded value. Drops top-level credential-shaped keys, unknown attacker payload, and anything not in the schema's `properties` list.

3. **`scrubExtensions(safe, options)`** — per-target ext/context scrub. Filters `ext`/`context` (which `pickWireSpecFields` keeps whole because they're wire-spec fields) to a caller-specified key allowlist, and merges in caller-injected values. Returns `WireSafe<T>` so the brand survives the chain.

## Adopter usage

```ts
import { pickWireSpecFields, scrubExtensions } from '@adcp/sdk/server';

async updateMediaBuy(buyerReq, ctx) {
  // L2 step 1: strip to wire-spec fields. Drops buyer creds.
  const safe = pickWireSpecFields(buyerReq, 'UpdateMediaBuyRequest');

  // L2 step 2: per-target ext/context scrub + storefront cred inject
  const targets = await resolveTargets(buyerReq, ctx);
  for (const target of targets) {
    const perTarget = scrubExtensions(safe, {
      allowedExtKeys: new Set(['scope3_api_key', 'partner_request_id']),
      inject: {
        context: {
          managed_access_token: target.token,
          managed_advertiser_id: target.advertiserId,
        },
      },
    });
    await operational.updateMediaBuy(ctxFor(target), perTarget);
  }
}
```

The 'storefront fan-out flow' end-to-end test demonstrates all three round-N vectors (top-level, nested context, nested ext) being eliminated through this pipeline while storefront credentials get injected.

## Codegen

`scripts/generate-wire-spec-fields.ts` walks `schemas/cache/{version}/**/*-request.json` and emits `src/lib/server/wire-spec-fields.generated.ts` with field arrays per request type. **Drift between the helper and the schema is structurally impossible** — both are emitted from the same codegen pass.

Wired into `npm run build:lib` so it runs alongside the existing version-sync. Restricted to fan-out-relevant request types (every mutating tool in `MUTATING_TASKS` + `get_media_buy_delivery`); 29 types total. Read-only tools like `list_creative_formats` are excluded — they aren't fan-out targets, and some have cross-protocol shape divergence in the schema cache that would block codegen.

## Source-compatible with #1530

`OperationalPlatform` method signatures keep their existing plain-typed parameters (`request: UpdateMediaBuyRequest`). The brand is opt-in at the type level — adopters who use `pickWireSpecFields` get the safety benefit at the picking site (compile error if they accidentally spread buyer args into the variable typed `WireSafe<T>`).

Adopters not using the helper see zero behavior change. If a future major bump wants to make the brand mandatory at the operational boundary, that's a v7 change — discussed in #1540 (extractContext split, also depends on the brand).

## Files

- `src/lib/server/wire-safe.ts` (new) — module
- `src/lib/server/wire-spec-fields.generated.ts` (new, codegen output) — 29 schemas
- `scripts/generate-wire-spec-fields.ts` (new) — codegen
- `package.json` — `build:lib` calls codegen
- `test/server-wire-safe.test.js` (new) — 17 tests
- `docs/guides/CTX-METADATA-SAFETY.md` — adopter guide for the L2 flow
- `.changeset/wire-safe-args.md` — minor bump

## Test plan

- [x] 17/17 new wire-safe tests pass
- [x] 1291/1291 server suite pass — no regression
- [x] Format / typecheck clean
- [x] Codegen idempotent (re-runs are no-ops when schemas unchanged)
- [x] End-to-end test covers all three round-N vectors plus storefront-cred injection

## Tracking next

Per the broader plan (`/.context/operational-fanout-design.md`):
- #1539 (scan `ctx.authInfo.extra`) is the natural follow-up; folds the L1 scanner into the L2 perimeter.
- #1540 (`extractContext` split) and #1541 (`@adcp/sdk/lint`) remain queued as separate work.


bokelley:
Two expert reviews (security, code) flagged convergent and independent findings. All addressed in `95307a79`.

## Blockers fixed

- **Recursive credential scan in `scrubExtensions`** (security M1). The original top-level-only filter left a real escape hatch — buyer hides credentials at depth, survives the gate. Now `scrubExtensions` walks `ext`/`context` values using the L1 pattern set, dropping credential-shaped keys at any depth. Default-on; new `recursiveCredentialScan: false` option for adopters who validate shapes themselves. Two new tests pin the behavior.

- **`pickWireSpecFields` return type narrowing** (code-reviewer #1). Codegen now emits `{ fields, __type }` per entry; return type is `WireSafe<(typeof WIRE_SPEC_FIELDS)[K]['__type']>` so adopters get strong per-schema narrowing.

## Defense-in-depth fixes

- **Runtime-frozen arrays + outer object** (security L1). `Object.freeze` in codegen output. Test asserts behavioral effect (length unchanged) rather than mode-dependent throw.
- **`inject` JSDoc warning** (security L2). "Values here MUST be storefront-derived, never from the buyer request."
- **Migration footgun doc** (security #8). `pickWireSpecFields` ALONE reopens round-2/3 — must chain `scrubExtensions`. Atomic-migration recipe in `CTX-METADATA-SAFETY.md`.

## Smaller fixes

- `scrubExtensions` generic constraint widened from `T extends Record<string, unknown>` to `T extends object` — `WireSafe<T>` brand assignability.
- SI naming in codegen — `si-*` files emit `SI*` PascalCase (matches `json-schema-to-typescript`'s acronym handling).
- Recursive scan strategy: when `allowedExtKeys` is set, scan starts at depth 1 (top level already gated); when not set, scan starts at depth 0.

## Subtle scan semantics

When an adopter writes `allowedExtKeys: new Set(['scope3_api_key'])`, that's an explicit affirmation that `scope3_api_key` is legitimate even though the L1 default pattern would flag it. So the recursive scan respects that — it doesn't second-guess top-level keys when there's an allowlist. The scan still walks deeper to drop nested credentials inside allowlisted values (`partner: { partner_access_token: '...' }` becomes `partner: {}`).

## Tests

- 20/20 wire-safe tests pass (was 17, +3 new — recursive scan with allowlist, recursive scan without, opt-out)
- 1287/1287 server suite pass — no regression
- Format / typecheck clean

## Deferred

- Code-reviewer's idempotency-helper-consistency note (use shared `writeFileIfChanged`) — small refactor, file follow-up.
- `ci:schema-check` should fail on stale generated file — same.

Both are separate cleanup PRs, neither blocks landing this.

Ready for re-review.

bokelley:
## Re-review — `95307a79`

Verified the fixes against the original security and code-reviewer findings.

**Security logic correct:**
- `fromDepth=1` semantics: step 1 has already filtered ext to allowlisted keys before `stripNestedCredentials` runs, so the scan correctly ignores depth-0 names (adopter affirmed them) and drops credentials at depth 1+. The `partner: { partner_access_token: '...' }` round-4 vector is closed.
- Inject-after-scan ordering is correct — storefront tokens like `managed_access_token` are merged after the recursive scan, so they don't get dropped by the same pattern set.
- `Object.freeze` on generated arrays + outer object: meaningful defense against shared-state mutation of the allowlist.

**Type narrowing correct:**
- `__type: null as unknown as RequestType` phantom + `(typeof WIRE_SPEC_FIELDS)[K]['__type']` indexed lookup resolves correctly with `as const`. Adopters get `WireSafe<UpdateMediaBuyRequest>`, not `WireSafe<Record<string, unknown>>`.
- `T extends object` widening: correct and necessary — generated request types don't have index signatures, so the prior `Record<string, unknown>` constraint fought brand assignability.

**One item worth adding before merge:**

The docs say bare `token` (no `_` separator) is *intentionally* not in the default pattern set (false-positive risk on `paymentToken`-style fields). That claim isn't pinned as a behavioral contract in the test suite — a regression in `DEFAULT_CREDENTIAL_PATTERNS` would silently break adopters who inject `token` inside an allowlisted key. A test asserting `{ partner: { token: 'kept' } }` survives with `recursiveCredentialScan: true` would close that gap. Not a blocker for landing, but worth tracking.

**Nit:** `Object.keys(properties).sort()` in `generate-wire-spec-fields.ts` uses locale collation. An explicit comparator (`localeCompare` with `'en'` + `'variant'` sensitivity, or a byte-order sort) makes the generated output reproducible across locales. Low risk today (all schema keys are ASCII), but cheap to harden.

Deferred items (idempotency-helper consistency, `ci:schema-check` stale-file gate) are reasonable scope splits — neither affects the security properties here.

**Ready to merge** from a security + correctness standpoint.

---
Reviewed by Claude Code. Session: https://claude.ai/code/session_01SHgCadn9gHDZXet5cjMAxD

---

bokelley:
## Follow-up — one blocker missed in the first re-review pass

The security reviewer completed after my earlier comment. One finding upgrades to **blocker**.

### Blocker — entry objects in `WIRE_SPEC_FIELDS` are not frozen

`Object.freeze(WIRE_SPEC_FIELDS)` makes the outer map's slots non-writable, so you can't replace `WIRE_SPEC_FIELDS.UpdateMediaBuyRequest`. The inner `fields` arrays are frozen. But the **entry objects themselves** are not frozen, so their `fields` property is writable:

```js
WIRE_SPEC_FIELDS.UpdateMediaBuyRequest.fields = ['media_buy_id', 'snap_access_token'];
```

This silently succeeds. After it, every call to `pickWireSpecFields(req, 'UpdateMediaBuyRequest')` copies `snap_access_token` from the buyer request, defeating L2 entirely for that schema. The JSDoc says the freeze prevents "shared-state mutation of the allowlist" — that claim is false today.

**Fix in `scripts/generate-wire-spec-fields.ts`:** wrap each entry in `Object.freeze`:

```diff
- UpdateMediaBuyRequest: {
+ UpdateMediaBuyRequest: Object.freeze({
    fields: Object.freeze([...]) as readonly string[],
    __type: null as unknown as UpdateMediaBuyRequest,
- },
+ }),
```

Also add a test assertion: `assert.ok(Object.isFrozen(WIRE_SPEC_FIELDS.UpdateMediaBuyRequest))` — the existing test only checks the outer map and the arrays, not the entry objects.

---

### Nit — contradictory JSDoc on `inject`

`ScrubExtensionsOptions.inject` JSDoc says two contradictory things:

> "Adopter-controlled values to merge into `ext` and/or `context` AFTER the allowlist filter and recursive scan run."  
> … "The inject precedes the allowlist filter, so a buyer value injected here would survive every check."

The second sentence is wrong about ordering — inject is Step 3, after both filter and scan. The warning intent (inject bypasses the scan) is correct but the mechanism description isn't. Replace the second sentence with something like: "Values injected here are not passed through the recursive credential scan — they are merged in after the scan completes."

---

Sorry for the split — the entry-freeze gap wasn't in my initial analysis. The first comment's "ready to merge" call is retracted until this is addressed.

---
Reviewed by Claude Code. Session: https://claude.ai/code/session_01SHgCadn9gHDZXet5cjMAxD

---

bokelley:
## Follow-up — `06a6c13`

Addressed both items from the re-review:

**Test pinned** (`test/server-wire-safe.test.js`): Added `'bare token (no _ separator) passes through recursive scan by default'` in the `scrubExtensions` suite. Asserts `{ partner: { token: 'kept' } }` survives `scrubExtensions` with `recursiveCredentialScan: true` — the behavioral contract for `DEFAULT_CREDENTIAL_PATTERNS` omitting bare `token` is now a regression gate, not just prose.

**Sort comparator hardened** (`scripts/generate-wire-spec-fields.ts`): Replaced `Object.keys(properties).sort()` with `sort((a, b) => (a < b ? -1 : a > b ? 1 : 0))`. Byte-order sort; reproducible across locales. All schema keys are ASCII so the output is unchanged, but the comparator is now explicit.

Neither change touches generated files or the public API surface.

---

bokelley:
Acknowledged — both items look clean. The bare-token regression gate and the explicit byte-order comparator are exactly the right hardening for a generated file. Nothing further needed from this side.

---

bokelley:
Noted — thanks for the sign-off, @bokelley. Nothing further queued on this end.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_019mbM4ZrQPnyrN9wTxwESav_

---

bokelley:
Both findings addressed in draft PR #1547:

- **Blocker (entry objects not frozen)** — every entry in `WIRE_SPEC_FIELDS` is now wrapped in `Object.freeze({...})` in the codegen script and generated file. `WIRE_SPEC_FIELDS.UpdateMediaBuyRequest.fields = [...]` now throws in strict mode / silently no-ops in sloppy mode, same as reassigning a frozen outer-map slot.
- **Medium (iterator poisoning)** — `pickWireSpecFields` now uses an indexed `for` loop instead of `for-of`, bypassing `Array.prototype[Symbol.iterator]` dispatch entirely.
- **Nit (inject JSDoc)** — corrected to say inject bypasses the scan and runs after it completes.

Tests: 23/23 pass (2 new: entry-freeze + iterator-poisoning).

---

bokelley:
Thanks for the follow-through — noted that all three findings (entry-freeze blocker, iterator-poisoning, JSDoc nit) are addressed in #1547. No further action needed on this PR.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_01BY8P3FJdG7GfsR4SZDWc2M_

---

PR #1555, opened 2026-05-04
feat(server): ctx.handoffToTask accepts optional task_id override

### LINKED ISSUE #1554: ctx.handoffToTask should accept a caller-supplied task_id (force_create_media_buy_arm contract)
## Summary

The framework's submitted-arm projection (`from-platform.js:1438-1448`) rejects hand-rolled `{ status: 'submitted', task_id }` envelopes with the message `Specialism handler returned a hand-rolled \`{status: 'submitted', task_id}\` envelope. The framework owns the submitted envelope — return \`ctx.handoffToTask(...)\``. The only way into the submitted arm is via `ctx.handoffToTask(fn)`, which assigns a framework-issued `taskId` exposed read-only on `TaskHandoffContext.id`.

This conflicts with the spec contract for `comply_test_controller` scenario `force_create_media_buy_arm` (adcp#3816 + adcp-client#1477):

> The next create_media_buy call from this sandbox account will return **this** task_id.

`task_id` is a **caller-supplied** parameter on the directive. The runner captures it via `forced.task_id` and asserts the next `create_media_buy.task_id` matches byte-for-byte (`media_buy/scenarios/create_media_buy_async.yaml`). When the framework auto-issues a different task_id, the storyboard fails:

```
× create_media_buy_submitted: SERVICE_UNAVAILABLE: Tool create_media_buy handler threw:
  Specialism handler returned a hand-rolled `{status: 'submitted', task_id}` envelope.
  The framework owns the submitted envelope — return `ctx.handoffToTask(...)` ...
```

## Suggested fix

Allow callers to pass an explicit task_id through `handoffToTask`:

```ts
ctx.handoffToTask<TResult>(
  fn: (taskCtx: TaskHandoffContext) => Promise<TResult>,
  options?: { task_id?: string }
): TaskHandoff<TResult>;
```

When `options.task_id` is set, the framework reuses the supplied id (after validating shape — non-empty, ≤128 chars per spec) instead of minting a fresh one. `taskCtx.id` reflects the supplied value.

This is narrowly justified — the only callers should be sellers that have a deterministic external-system task identifier they want to project (the test-controller directive is one; adopters that mint task_ids upstream of the framework are another). Production HITL flows continue to use the auto-issued id.

## Why not just use ctx.handoffToTask without the directive's task_id

The training-agent currently returns the directive's task_id from a hand-rolled v5 envelope. With this SDK fix, the v6 platform can do:

```ts
createMediaBuy: async (req, ctx) => {
  const directive = consumeForcedArmDirective(...);
  if (directive?.arm === 'submitted') {
    return ctx.handoffToTask(async () => directive.completionResult, { task_id: directive.taskId });
  }
  // ... sync arm
}
```

Without it, the test-controller's deterministic-task_id contract isn't reachable through the framework projection.

## Impact

Blocks the `media_buy_seller/create_media_buy_async` conformance storyboard on framework-routed sellers. Sellers on the v5 monolith path (e.g. legacy `/mcp` mount) don't hit the projection so they pass.

## Related

- adcp#3965 (training-agent regression cluster — last storyboard regression after the cluster work)
- adcp-client#1477 (added typed force.create_media_buy_arm slot — the directive-storage half is now first-class; this issue closes the response-emission half)
- spec: `static/compliance/source/protocols/media-buy/scenarios/create_media_buy_async.yaml`

bokelley:
Closes #1554

## Summary

The framework's `dispatchHitl` always minted a fresh `task_${randomUUID()}` id, making the `force_create_media_buy_arm` storyboard contract unreachable: the spec requires the seller to echo the directive-supplied `task_id` verbatim on the submitted arm.

This PR adds `options?: { task_id?: string }` to `ctx.handoffToTask(fn, options?)`. When set, the framework uses that exact string as the submitted task_id instead of minting a fresh one. `taskCtx.id` reflects the supplied value. Callers without options are unaffected.

**Changed files:**
- `async-outcome.ts` — `TaskHandoffOptions` interface; single `taskHandoffEntries` WeakMap stores `{ fn, options }` atomically; `_extractTaskFn` renamed to `_extractHandoffEntry`
- `context.ts` — `handoffToTask` signature + JSDoc
- `decisioning/index.ts` — exports `TaskHandoffOptions` by name
- `to-context.ts` — validation (non-empty, ≤ 128 chars, throws `Error`); forwards options to `_createTaskHandoff`
- `task-registry.ts` — `overrideTaskId?` on `TaskRegistry.create`; in-memory collision guard
- `postgres-task-registry.ts` — `overrideTaskId?`; `ON CONFLICT DO NOTHING` with `(rowCount ?? 0) === 0` guard
- `from-platform.ts` — `routeIfHandoff` extracts `options` from entry, threads `options.task_id` to `dispatchHitl`
- `test/server-decisioning-handoff-task-id.test.js` — 7 tests: custom id emitted verbatim, no-options regression, empty/long id rejection, registry collision unit tests

## What tested

- `prettier --check` — passed
- `tsc --project tsconfig.lib.json` — pre-existing `@types/node` + `moduleResolution=node10` environment diagnostics only (present on main before this change, exit 2); no new type errors introduced
- `npm test` — requires `node_modules`; CI will validate

**Nits surfaced by pre-PR review (not fixed — deferred):**
- Postgres collision under concurrent multi-instance retry (same `overrideTaskId` race) surfaces as 500 rather than idempotent response; pre-existing for minted IDs too — follow-up issue
- `to-context.ts` `typeof !== 'string'` guard unreachable from TS callers (reachable from plain-JS shims); test coverage gap noted
- `decisioning.type-checks.ts` missing `@ts-expect-error` negative case for `task_id: 42`

**Pre-PR review:**
- code-reviewer: approved after one blocker fixed (`rowCount === 0` → `(rowCount ?? 0) === 0` for pg null-rowCount on conflict)
- ad-tech-protocol-expert: approved — spec contract confirmed normative ("seller MUST emit verbatim"), `taskCtx.id` reflects override correctly, 128-char cap is SDK-defined/consistent with read-side cap, cross-specialism exposure correct

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01L7vuBEvg7NKMCPMW7Pk13q

---
_Generated by [Claude Code](https://claude.ai/code/session_01L7vuBEvg7NKMCPMW7Pk13q)_

### DIFF
--- a/.changeset/handoff-task-id-option.md
+++ b/.changeset/handoff-task-id-option.md
@@ -1,0 +1,14 @@
+---
+"@adcp/sdk": minor
+---
+
+feat(server): ctx.handoffToTask accepts optional task_id override
+
+Adds `options?: { task_id?: string }` to `ctx.handoffToTask(fn, options?)`.
+When `options.task_id` is set, the framework uses that exact string as the
+submitted task_id instead of minting a fresh one. Validates non-empty, ≤ 128
+characters. Closes #1554.
+
+Required for the `force_create_media_buy_arm` comply_test_controller scenario,
+which asserts the seller echoes the directive-supplied task_id verbatim on
+the create_media_buy submitted arm.
--- a/src/lib/server/decisioning/async-outcome.ts
+++ b/src/lib/server/decisioning/async-outcome.ts
@@ -185,15 +185,37 @@ export class AdcpError extends Error {
 const TASK_HANDOFF_BRAND: unique symbol = Symbol.for('@adcp/decisioning/task-handoff');
 
 /**
- * The handoff function lives in a WeakMap keyed by the `TaskHandoff`
- * marker object — NOT on the marker itself. This means the framework
- * can extract the function (it has the WeakMap reference) but adopters
- * holding only a `TaskHandoff<T>` value cannot invoke or inspect it.
- * Closes round-6 CR-3 / Protocol-L2: `_taskFn` was previously a
- * type-visible field that adopters could forge with their own
+ * Optional overrides for `ctx.handoffToTask`.
+ *
+ * @public
+ */
+export interface TaskHandoffOptions {
+  /**
+   * Use this exact string as the `task_id` instead of letting the framework
+   * mint a fresh one. Required when an upstream system (or a test-controller
+   * directive such as `force_create_media_buy_arm`) has already issued the ID
+   * and the spec contract demands the response echoes it verbatim.
+   *
+   * Constraints: non-empty, ≤ 128 characters. Throws if violated.
+   * The caller is responsible for uniqueness within the account.
+   */
+  task_id?: string;
+}
+
+type TaskHandoffEntry = {
+  fn: (taskCtx: TaskHandoffContext) => Promise<unknown>;
+  options?: TaskHandoffOptions;
+};
+
+/**
+ * The handoff fn and options live in a WeakMap keyed by the `TaskHandoff`
+ * marker object — NOT on the marker itself. Both are stored atomically in a
+ * single entry so `isTaskHandoff` and `_extractHandoffEntry` always see a
+ * consistent pair. Closes round-6 CR-3 / Protocol-L2: `_taskFn` was
+ * previously a type-visible field that adopters could forge with their own
  * `Symbol.for(...)` call.
  */
-const taskHandoffFns = new WeakMap<object, (taskCtx: TaskHandoffContext) => Promise<unknown>>();
+const taskHandoffEntries = new WeakMap<object, TaskHandoffEntry>();
 
 /**
  * Marker the framework recognizes as "promote this call to a task."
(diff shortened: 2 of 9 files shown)


bokelley:
**Assessment: not ready to merge**

Triage approvals stand, but the PR's own 4 new tests fail in CI ([run 25316500441](https://github.com/adcontextprotocol/adcp-client/actions/runs/25316500441/job/74215117488)):

```
not ok 1 - emits the caller-supplied task_id on the wire verbatim
  + actual: undefined / - expected: 'submitted'
not ok 2 - without options, framework mints a fresh task_ prefixed id
not ok 3 - rejects empty string task_id at call time
  + actual: undefined / - expected: true
not ok 4 - rejects task_id longer than 128 characters at call time
```

All four assertions are reading `undefined` where they expect either the projected `submitted` envelope (#1, #2) or a thrown validation error captured via `try/catch` (#3, #4). That's consistent with the test harness's `createMediaBuy` handler not actually being routed through the framework's submitted-arm projection — the `ctx.handoffToTask(fn, { task_id })` call returns, but the projection that produces `{ status: 'submitted', task_id }` isn't invoked.

Two likely causes to check at `test/server-decisioning-handoff-task-id.test.js:63`:
1. The harness builds the platform via the wrong factory (e.g., raw `createAdcpServer` vs `createAdcpServerFromPlatform`), so the projection at `from-platform.js:1438-1448` never fires.
2. The `handoffToTask` overload is registered but the WeakMap lookup happens after the projection — the supplied `task_id` is read but the envelope is still constructed with the auto-issued id, leaving `result.status` unset on this code path.

The validation tests (#3, #4) failing the same way (expecting a thrown error, getting undefined) suggest the call chain swallows or skips validation rather than the validation logic itself being wrong.

Leaving as draft pending a fix pass.

PR #1576, opened 2026-05-07
fix(a2a): reroute SDK polling-path terminal-state error through unwrap pipeline

### LINKED ISSUE #1575: storyboard runner: failed-Task DataPart adcp_error not unwrapped on polling/blocking path (sibling of #1571)
## Symptom

Storyboard runner reports `Task <uuid> is in terminal state: 3` instead of decoding the `adcp_error` DataPart on `TASK_STATE_FAILED` A2A responses.

```
media_buy_seller/inventory_list_targeting/get_products_brief [media_buy]:
  A2A agent returned error: Task 9a364eb3-17c5-467a-8dcb-4d08d41474b3 is in terminal state: 3

media_buy_seller/creative_fate_after_cancellation/get_products_brief [media_buy]:
  A2A agent returned error: Task 2263bfc5-4868-42e3-8bb3-c31e05eabebb is in terminal state: 3
```

This is the same asymmetry class as #1571 (fixed in 6.13.1) — the runner doesn't unwrap the AdCP-mandated A2A error envelope on failed Tasks. **6.13.1 fixed the JSON-RPC `result.error` path; this issue is the parallel `Task.state=failed + adcp_error DataPart` path used by the polling/blocking client.**

## Repro

Against any spec-compliant AdCP seller. Standalone single-step calls **pass** because they don't enable polling; only batch runs trip this:

```bash
npx @adcp/sdk@6.13.1 storyboard run https://<agent-url>/ media_buy_seller \
  --auth <token> --protocol a2a --format json --timeout 180
```

In our case (Wonderstruck staging), 2 of the failing steps reproduce — both are `get_products_brief` early-cascade steps in scenarios where a later step is supposed to fail; the polling client trips before the storyboard ever reaches the supposed-to-fail step.

Standalone counter-example (passes because no polling):
```bash
npx @adcp/sdk@6.13.1 storyboard step <agent-url>/ \
  media_buy_seller/inventory_list_targeting get_products_brief \
  --auth <token> --protocol a2a --json
```

## Server response (spec-compliant per transport-errors.mdx §A2A Binding)

The seller emits exactly what the spec mandates — `_send_adcp_error` in the framework produces:

```json
{
  "result": {
    "kind": "task",
    "status": {"state": "failed"},
    "artifacts": [{
      "parts": [
        {
          "kind": "data",
          "data": {
            "adcp_error": {
              "code": "TERMS_REJECTED",
              "message": "...",
              "recovery": "correctable",
              "field": "..."
            },
            "context": {"correlation_id": "..."}
          }
        },
        {"kind": "text", "text": "..."}
      ]
    }]
  }
}
```

`pb.TaskState.TASK_STATE_FAILED == 3`. The literal "is in terminal state: 3" message doesn't appear in `@adcp/sdk@6.13.1` source — it's constructed in the polling client when a polled task settles to `state=failed`. The runner then surfaces the raw transport message without trying to extract the `adcp_error` DataPart.

## Fix region

* **`@adcp/sdk` `dist/lib/protocols/a2a.js:241-253`** — currently handles only JSON-RPC-level `error` responses. Needs a parallel branch for `result.kind === "task" && result.status.state === "failed"` that pulls `adcp_error` out of `result.artifacts[*].parts[*].data` and surfaces it through the same path the success unwrapper uses.

* **`@adcp/sdk` `dist/lib/core/TaskExecutor.js:137-174` `unwrapTasksGetEnvelope`** already handles the AdCP A2A unwrap pattern correctly for `result.kind === "task"` artifacts. The runner's A2A protocol shim (`a2a.js`) just isn't routing failed tasks through it.

Likely a tiny patch — same shape as the 6.13.1 fix, but on the polling code path.

## Versions

- `@adcp/sdk@6.13.1`
- `@a2a-js/sdk` (whatever 6.13.1 transitively depends on)

## Related

- adcontextprotocol/adcp-client#1571 — the JSON-RPC `error` path fix that 6.13.1 shipped
- This issue is the polling/Task path of the same class

Happy to PR the fix if you can confirm the right code location.

bokelley:
## Summary

Closes #1575

When `@a2a-js/sdk` polls a task to completion on the blocking/SSE path, it annotates `result.error` with `"Task X is in terminal state: 3"` for failed tasks — but `result.artifacts` still carries the spec-canonical `adcp_error` DataPart. `callA2AToolImpl` was throwing immediately on `result.error`, discarding the artifacts and surfacing the opaque SDK sentinel string rather than the structured `adcp_error`.

- **`src/lib/protocols/a2a.ts`**: Before throwing on `messageResponse?.result?.error`, check whether the response is a terminal-failed A2A task (`kind: 'task'`, non-empty `artifacts`, `status.state` in `failed/rejected/canceled`). If so, return `messageResponse` instead of throwing, routing it through the existing `handleAsyncResponse → extractResponseData → unwrapA2AResponse` pipeline that already handles this case (landed in #1572).
- **`src/lib/utils/response-unwrapper.ts`**: Export `TERMINAL_FAILURE_A2A_STATES` (the failure-only subset of terminal states) so the state membership check is maintained in one place rather than inlined separately in `a2a.ts`.
- **`test/lib/protocol-compliance.test.js`**: Add regression test covering the exact response shape `@a2a-js/sdk` produces on the polling path: `{ result: { kind: 'task', status: { state: 'failed' }, error: {...}, artifacts: [...] } }`. Asserts that `callA2ATool` returns (does not throw) and preserves `result.artifacts` for downstream `adcp_error` extraction.

## What was tested

- Unit test added in `test/lib/protocol-compliance.test.js` ("Error Response Handling" describe block) exercises the exact SDK response shape from the polling path.
- Confirmed existing tests in the describe block still cover: bare JSON-RPC `error`, nested `result.error` without artifacts (both must still throw), and the new bypass-and-return path.
- Build environment (`tsx` / `@types/node`) is pre-broken on this host — pre-existing, unrelated to this change. CI will verify.

## Pre-PR review

- **Protocol expert (ad-tech-protocol-expert)**: Confirmed the A2A binding shape (`result.artifacts[].parts[kind=data].data`) is correct per the current spec. The `docs/proposals/transport-error-mapping.md` describing the older `status.message.parts` path is a stale pre-GA proposal, not the current wire contract.
- **Code reviewer (code-reviewer)**: Confirmed fix approach is correct. Actionable findings applied: (1) exported `TERMINAL_FAILURE_A2A_STATES` to avoid state-set drift between `a2a.ts` and `response-unwrapper.ts`; (2) added `debugLogs` entry when taking the artifact-bypass path so operators can observe the reroute; (3) reviewer's catch-block property-cascade concern does not apply to this implementation (the fix is in the `if (messageResponse?.result?.error)` branch, not in `catch`).

---

https://claude.ai/code/session_01CZWJDUsvc5mSHaGTgnHXsT

---
_Generated by [Claude Code](https://claude.ai/code/session_01CZWJDUsvc5mSHaGTgnHXsT)_

### DIFF
--- a/.changeset/fix-a2a-polling-terminal-state.md
+++ b/.changeset/fix-a2a-polling-terminal-state.md
@@ -1,0 +1,17 @@
+---
+"@adcp/sdk": patch
+---
+
+fix(a2a): reroute SDK polling-path terminal-state errors through unwrap pipeline
+
+When `@a2a-js/sdk` polls a task to completion on the blocking/SSE path, it
+annotates `result.error` with "Task X is in terminal state: 3" for failed
+tasks — but the `result.artifacts` DataPart still carries the spec-canonical
+`adcp_error`. `callA2AToolImpl` was throwing immediately on `result.error`,
+discarding the artifacts. Now it detects terminal-failed tasks with artifacts
+(`kind: 'task'`, non-empty `artifacts`, state in `failed/rejected/canceled`)
+and returns the full response so `handleAsyncResponse` can extract `adcp_error`
+via `unwrapA2AResponse`. Fixes #1575.
+
+Also exports `TERMINAL_FAILURE_A2A_STATES` from `response-unwrapper` so the
+state membership check is maintained in one place.
--- a/src/lib/protocols/a2a.ts
+++ b/src/lib/protocols/a2a.ts
@@ -12,6 +12,7 @@ import { withSpan, injectTraceHeaders } from '../observability/tracing';
 import { isAgentCardPath, buildCardUrls } from '../utils/a2a-discovery';
 import { buildAgentSigningFetch, signingContextStorage, type AgentSigningContext } from '../signing/client';
 import { redactIdempotencyKeyInArgs } from '../utils/idempotency';
+import { TERMINAL_FAILURE_A2A_STATES } from '../utils/response-unwrapper';
 import { wrapFetchWithCapture } from './rawResponseCapture';
 import { wrapFetchWithSizeLimit } from './responseSizeLimit';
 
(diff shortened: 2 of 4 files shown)


bokelley:
Superseded by #1577 (merged) which shipped the same fix — protocol-layer guard in src/lib/protocols/a2a.ts now defers to the canonical DataPart artifact when a seller emits both a transport-level error hint and a structured AdCP envelope. Sibling unwrap-layer fix landed in #1579. Closing as duplicate. Issue #1575 closed by #1577.

bokelley:
Acknowledged — closing this branch out. #1577 + #1579 cover the full fix; #1575 is closed. No further action from triage.

---

PR #1592, opened 2026-05-08
fix(a2a): handle stale pendingTaskId causing "Task not found" in batch storyboard runs

### LINKED ISSUE #1585: storyboard runner: A2A "Task X not found" on completed-task lookups during batch runs
## Summary

Sibling of #1571 and #1575. After the unwrap fixes in 6.13.1 (#1571) and 6.13.2 (#1575), the A2A storyboard runner still fails on success-path `get_products` calls during batch runs because it appears to re-look-up tasks the seller has already completed and discarded.

## Version

`@adcp/sdk@6.13.2`

## Symptom

Two storyboard steps fail under A2A only (MCP path is unaffected):

```
media_buy_seller/inventory_list_targeting/get_products_brief [media_buy]:
  A2A agent returned error: Task <uuid> not found

media_buy_seller/creative_fate_after_cancellation/get_products_brief [media_buy]:
  A2A agent returned error: Task <uuid> not found
```

These are SUCCESS-path `get_products` calls — not error scenarios. The seller is a spec-compliant `adcp` Python framework deployment (Wonderstruck staging, `https://wonderstruck.sales-agent.scope3.com/`).

## Reproduction

- Run the `media_buy_seller` storyboard batch against an A2A endpoint with `--blocking false` polling behavior.
- The `inventory_list_targeting` and `creative_fate_after_cancellation` scenarios both fail at their initial `get_products_brief` step with "Task not found".
- Direct standalone calls to the same endpoint succeed and return products immediately:
  ```
  npx adcp https://wonderstruck.sales-agent.scope3.com/ get_products '{...}' --auth <token>
  ```
  Single-step / direct invocations do NOT hit this. Only batch storyboard runs do.

## Suspected cause

When `--blocking false`, the runner polls until the task settles. After the task completes, the runner appears to issue another `tasks/get` lookup against the now-completed task ID. The seller's task store does not retain completed tasks (or there is a race on cleanup), so the lookup returns "Task not found".

Either:
- The runner should not re-look-up tasks that have already settled to a terminal state, OR
- The seller is expected to retain completed tasks for some TTL (this would need to be specified)

## Cascading impact

This is not isolated to the two failing steps. `get_products_brief` is the gating step for both scenarios — every dependent storyboard step in `inventory_list_targeting` and `creative_fate_after_cancellation` is blocked by this failure.

## Related

- #1571 (initial unwrap fix, 6.13.1)
- #1575 (follow-up unwrap fix, 6.13.2)

This issue is the next layer down — it survives both prior fixes.

bokelley:
Closes #1585

In batch storyboard runs, all scenarios share a single `AgentClient` instance (via `getOrCreateClient` caching). When a prior scenario ends with an A2A working-state response, `AgentClient.retainSession` sets `pendingTaskId`. The storyboard poll path (`waitForCompletion` in `task-map.ts`) bypasses `retainSession`, so that task ID is never cleared. The next scenario inherits it; A2A 0.3.x defines no minimum retention TTL for completed tasks, so a seller that has evicted the completed task responds with "Task \<uuid\> not found" and the scenario fails.

**Three fixes (root cause + two defense-in-depth):**

1. **Root cause** — `AgentClient.clearPendingTask()` (`@internal`) + `task-map.ts` `finally` block. After `waitForCompletion` resolves (or throws / times out), `clearPendingTask()` is called so the stale `pendingTaskId` is gone before the next scenario starts.

2. **Defense-in-depth** — `callA2AToolImpl` (`a2a.ts`): when `session.taskId` is set and the seller returns "Task not found", retry once without the task ID. The guard is self-limiting: the recursive call passes `taskId: undefined`, so the guard cannot fire again.

3. **Defense-in-depth** — `pollTaskCompletion` (`TaskExecutor.ts`): catch "Task not found" from `getTaskStatus` and return a descriptive `failed` `TaskResult` with an actionable error message rather than an opaque uncaught exception.

## What-tested

- `npx tsc --noEmit`: 2 pre-existing errors only (`@types/node` not installed, `moduleResolution=node10` deprecation warning). No new errors introduced. Verified by running before and after changes.
- `npm run ci:quick`: pre-existing environment failure (`tsx: not found`). Confirmed pre-existing by stashing changes and re-running.
- Nits from second expert pass addressed: changeset body updated to describe all three fixes; one-line self-limiting comment added on the retry guard.

## Pre-PR review

- **code-reviewer**: approved — no blockers after second pass; nits addressed (changeset body updated, self-limiting comment added)
- **ad-tech-protocol-expert**: approved — root-cause fix satisfies the spec-level concern; defense-in-depth retry correctly framed (dropping a stale prior-scenario `taskId`, not the current call's own ID); `pollStartTime` fix is sound

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01NXzCFi3jJDNqXaDgQrPzKK

---
_Generated by [Claude Code](https://claude.ai/code/session_01NXzCFi3jJDNqXaDgQrPzKK)_

### DIFF
--- a/.changeset/fix-a2a-task-not-found-batch-race.md
+++ b/.changeset/fix-a2a-task-not-found-batch-race.md
@@ -1,0 +1,29 @@
+---
+"@adcp/sdk": patch
+---
+
+fix(a2a): handle stale pendingTaskId causing "Task not found" in batch storyboard runs
+
+When multiple storyboard scenarios share a single `AgentClient` instance, the client's
+`pendingTaskId` can be left set from a prior scenario's async working-state response if
+the poll path (via `waitForCompletion`) bypasses `absorbServerSideMetadata`. The next
+scenario then includes this stale task ID in its `message/send` call; if the seller has
+already completed and evicted that task (A2A 0.3.x defines no minimum retention TTL),
+the agent returns "Task <uuid> not found" and the scenario fails.
+
+Three fixes:
+
+1. **Root cause** — `AgentClient.clearPendingTask()` (`src/lib/core/AgentClient.ts`) +
+   `task-map.ts` `finally` block: after `waitForCompletion` resolves, `clearPendingTask()`
+   is called so the stale `pendingTaskId` is cleared before the next scenario starts.
+   This is the primary fix; the two below are defense-in-depth.
+
+2. `callA2AToolImpl` (`src/lib/protocols/a2a.ts`): when a "Task not found" error is
+   returned and `session.taskId` is set, retry once without the stale task ID so the
+   call starts a fresh server-side task. (Guard is self-limiting: the retry passes
+   `taskId: undefined`, so the guard cannot fire a second time.)
+
+3. `pollTaskCompletion` (`src/lib/core/TaskExecutor.ts`): when explicit polling
+   (`tasks/get`) receives a "Task not found" error, return a descriptive `failed`
+   `TaskResult` with an actionable error message rather than letting the error
+   propagate as an uncaught exception.
--- a/src/lib/core/AgentClient.ts
+++ b/src/lib/core/AgentClient.ts
@@ -216,6 +216,18 @@ export class AgentClient {
   }
 
   /**
+   * Clear any retained `pendingTaskId` without requiring a new round-trip.
+   * Called by the storyboard poll path (`task-map.ts`) after `waitForCompletion`
+   * resolves, because that path bypasses `retainSession` and would otherwise
+   * leave a stale task ID on the session (adcp-client#1585).
+   *
+   * @internal
+   */
+  clearPendingTask(): void {
+    this.pendingTaskId = undefined;
+  }
+
+  /**
    * Returns the AdCP protocol version this client speaks. Mirrors
    * `SingleAgentClient.getAdcpVersion()`. See {@link SingleAgentClientConfig.adcpVersion}.
    */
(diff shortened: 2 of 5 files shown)


bokelley:
Closing in favor of:

- **#1588 (merged)** — `resetContext()` per storyboard at the runner level. Addresses the cross-storyboard root cause for `comply()`-style harnesses.
- **#1593 (merged)** — narrows `AgentClient` `pendingTaskId` auto-threading to same-`(contextId, taskName)`. Addresses the SDK-level root cause for any adopter reusing one `AgentClient` across logically distinct conversations.
- **#1596 (open)** — extracts the surviving useful piece from this PR: `TaskExecutor.pollTaskCompletion` returning a descriptive `failed` `TaskResult` on "Task not found" instead of an uncaught exception. Includes a regression test.

The two pieces from this PR that were _not_ extracted:

1. `clearPendingTask()` + `task-map.ts` `finally` block — superseded by #1593's narrowing. Clearing on every `waitForCompletion` would also drop intentional within-storyboard task resumption, which #1593 preserves correctly via `(contextId, taskName)` matching.
2. `callA2AToolImpl` retry on "Task not found" — became risky after #1593 narrowed auto-threading. Post-#1593 the only ways a stale `taskId` reaches `callA2AToolImpl` are caller-supplied (e.g., HITL resume); silently retrying would mask a real eviction failure the buyer needs to surface.

bokelley:
Acknowledged — thanks for the clear decomposition rationale. The three-way split (#1588 + #1593 + #1596) is a cleaner outcome: runner-level reset, SDK-level narrowing, and the surviving `pollTaskCompletion` improvement each land in the right layer without the coupling issues this PR carried. No further action from triage.

---

Issue #1612, opened 2026-05-08
storyboard run: A2A pre-flight comply() ~3x slower than MCP, hits 180s timeout for healthy agents


bokelley:
## Symptom

Running \`@adcp/sdk@6.15.0 storyboard run media_buy_seller\` against the same agent on both transports, the A2A pre-flight \`comply()\` hits the default 180s timeout while MCP completes the entire storyboard (pre-flight + 19 attempted steps) in 147s.

```
── MCP ──
→ https://wonderstruck.sales-agent.scope3.com/mcp
  Steps:    14 passed, 5 failed, 40 skipped
  Duration: 147.1s

── A2A ──
→ https://wonderstruck.sales-agent.scope3.com/
  Steps:    0 passed, 1 failed, 0 skipped
  Duration: 307.1s
  Failing steps:
    - pre-flight/comply [core]: comply() timed out after 180000ms
```

A2A never gets past pre-flight, so we can't tell whether the storyboard would otherwise pass.

## What's surprising

Per-call latency on A2A is **faster** than MCP. Probed each transport directly with \`curl\` against the same deployed agent (Wonderstruck on Fly.io IAD):

| Probe | Latency |
|---|---|
| TLS handshake (control) | 54ms |
| MCP \`tools/list\` | 380ms |
| MCP \`get_media_buys\` (6 buys returned) | ~700ms |
| MCP \`list_creative_formats\` (151KB body) | 1.8s |
| MCP \`get_products\` (brief processing) | 2.2s |
| A2A \`agent-card.json\` | 53ms |
| A2A \`get_adcp_capabilities\` | 117ms |
| A2A \`get_media_buys\` | 235ms |
| A2A sandboxed \`get_products\` | 120ms |

Every A2A call returned \`status.state: \"completed\"\` synchronously — the seller is not parking work in long-running tasks.

So whatever \`comply()\` is doing per-method, it's >>> 235ms per A2A call. With 10 advertised skills + signing/auth probes, even 30 sequential calls @ 235ms = 7s. To fill 180s the SDK would need to be making thousands of calls, or waiting on something it never sees from the seller.

## Hypothesis

\`comply()\` (still invoked as a pre-flight by \`storyboard run\` even though the CLI subcommand is marked \`DEPRECATED — will be removed in v5\`) appears to add per-call polling overhead on the A2A path that doesn't exist on MCP. Plausible:

[...]

PR #1613, opened 2026-05-08
fix(a2a/comply): three-layer A2A timeout fix — pollTaskCompletion + comply signal + discovery (#1612)

bokelley:
Refs #1612 — fixes the A2A transport path only; the MCP sequential-call timeout is a separate root cause tracked in #1612.

`storyboard run media_buy_seller --transport a2a` was timing out at 180s because `pollTaskCompletion` had no exit branch for `status: 'unknown'` — when a seller's `tasks/get` response didn't match any known envelope shape (flat AdCP, MCP `structuredContent`, or A2A DataPart artifact), the while-true loop fell through to `sleep(pollInterval)` and retried indefinitely. Compounded by orphaned poll promises that kept firing A2A calls after the outer 30s storyboard race timer resolved, accumulating across multiple storyboard steps until comply() burned its full 180s budget.

Four-part fix:

1. **`status: 'unknown'` exit branch** (`TaskExecutor.ts`) — `pollTaskCompletion` now returns a `failed` `TaskResult` immediately when `mapTasksGetResponseToTaskInfo` returns `status: 'unknown'`, with a message naming the likely seller-side cause.

2. **AbortSignal threading** (`TaskExecutor.ts`, `ConversationTypes.ts`, `runner.ts`) — `pollTaskCompletion` gains an optional `signal?: AbortSignal` 5th param; `SubmittedContinuation.waitForCompletion` gains a matching `signal?: AbortSignal` 2nd param. The storyboard runner passes `AbortSignal.timeout(timeoutMs)` so the inner loop exits when the outer race timer fires — no more orphaned A2A calls. Public buyers can now also pass a signal to `waitForCompletion()`.

3. **A2A transport-status fallback** (`TaskExecutor.ts` `unwrapTasksGetEnvelope`) — when an A2A Task response has no DataPart artifacts, surface `result.status.state` + `result.id` as a last-resort hint so terminal transport states (`completed`, `failed`, etc.) propagate instead of collapsing to `'unknown'`.

4. **Post-sleep abort re-check** — a second `if (signal?.aborted) continue` after `sleep(pollInterval)` limits abort latency to one sleep interval rather than one full poll cycle.

Known follow-ups (not in this PR):
- **MCP sequential-call timeout** — a second user confirmed that bare `storyboard run` on MCP (default 120s) also times out at comply() pre-flight. Root cause: `complyImpl`'s `for (const sb of applicableStoryboards)` loop does not forward the AbortSignal into `runStoryboard()`, so the combined signal only fires *between* storyboards, not *inside* them. Fix: thread `signal` through `StoryboardRunOptions` into per-step MCP calls. Tracked in #1612.
- **`tasks/get` via `message/send`** — `tasks/get` is dispatched via `message/send` skill invocation for A2A; the native A2A 0.3.0 `tasks/get` JSON-RPC method would be more correct for sellers that implement that path directly. Tracked separately.

## What tested

- `npm run typecheck` — no new errors (pre-existing `TS2688`/`TS5107` environment-level errors unrelated to these changes)
- Build toolchain (`tsx`) unavailable in CI environment; TypeScript compilation via `tsc --noEmit` confirmed clean

## Pre-PR review

- **code-reviewer**: approved — no blockers; issues noted (waitForCompletion signal threading, transport-status task_id fallback) were fixed in follow-up commit; remaining nits surfaced in PR body above
- **ad-tech-protocol-expert**: approved — "Sound. All four changes are correct, consistent with AdCP's submitted-arm polling contract, and introduce no new protocol concerns." Confirmed blocker (waitForCompletion signal gap) resolved; optional-param addition is non-breaking per AdCP versioning policy

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01QQeuhhr7XAMjKg2qZGVaza

---
_Generated by [Claude Code](https://claude.ai/code/session_01QQeuhhr7XAMjKg2qZGVaza)_

### DIFF
--- a/.changeset/fix-a2a-discovery-1612.md
+++ b/.changeset/fix-a2a-discovery-1612.md
@@ -1,0 +1,44 @@
+---
+"@adcp/sdk": patch
+---
+
+fix(cli/discovery): three compounding A2A timeout bugs (#1612)
+
+Investigation of the original #1612 symptom against
+`wonderstruck.sales-agent.scope3.com` (`storyboard run … --transport a2a`
+hanging at 307s+) found three independent bugs whose interaction produced
+the observed wall-clock:
+
+1. **`--transport` flag was silently dropped.** CLI accepted only
+   `--protocol`. The original repro and many users use `--transport`
+   (A2A SDK convention). Now accepted as an alias; explicit `--protocol`
+   wins on conflict.
+
+2. **`detectProtocol` returned `'mcp'` on any non-200 response from the
+   well-known A2A card path** — including 5xx, 401, 403, 429, and network
+   timeouts. A host that returns 503 on `/.well-known/agent.json` is far
+   more likely to be A2A-with-an-unhealthy-card than MCP-at-root. New
+   classification: 5xx/401/403/429/network-error → A2A suspect, only 4xx
+   (other than auth/rate) and clean miss → MCP fallback.
+
+3. **`discoverAgentProfile` had no AbortSignal awareness.** When the SDK
+   was misled into trying MCP discovery against a non-MCP root, the
+   internal `getAgentInfo()` call could spin past comply()'s timeout
+   (we observed 425s of orphaned MCP retries). `comply()` now passes its
+   combined timeout/external signal to `discoverAgentProfile`, which
+   races the underlying transport calls so the comply pipeline unblocks
+   on abort. The orphaned in-flight transport request still resolves in
+   the background; the fix bounds caller-visible latency, not the
+   transport's own retry budget.
+
+**Empirical verification against Wonderstruck**:
+
+| Invocation | Before | After |
+|---|---|---|
+| `--transport a2a` (typo for `--protocol`) | 425s | 31s |
+| bare URL auto-detect | 425s | 41s |
+| `--protocol mcp` against `/mcp` | 60.8s | 31s |
+
+These three fixes are independent — each addresses a distinct failure mode
+exposed by the same symptom. The `pollTaskCompletion` hardening from the
+prior commits remains relevant for the per-step polling case.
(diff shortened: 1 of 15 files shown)


bokelley:
**Pre-merge expert reviews — all three converged on approve.**

- **code-reviewer:** "Admin-merge — the third commit is correctly scoped, well-tested, and the residual issues are pre-existing or stylistic." Two nits filed as follow-ups (pre-existing CLI-parse parity bug at site 3; coverage gap on second `getAdcpCapabilities` call abort).
- **ad-tech-protocol-expert:** "Sound-with-caveats. Protocol semantics do NOT block merge." Notes the 5xx→A2A reclassification could misclassify an MCP-at-root host behind a generic-503 WAF, but the cost is bounded (one extra A2A handshake) vs the 425s symptom this fixes. Open question (non-blocker): wire `tasks/cancel` so orphan A2A polls don't keep billing the seller.
- **security-reviewer:** "Do not block admin-merge. All five concerns examined; no exploitable findings." Two low-severity defense-in-depth notes: (1) add a "do not log response bodies" comment in `raceWithSignal` — orphans carry auth-bearing responses; (2) consider an internal-IP denylist in `detectA2AOrMcp` if comply ever runs server-side against tenant-supplied URLs.

CI red-X is `npm audit --audit-level=high` flagging `fast-uri` (GHSA-q3j6-qgpj-74h6) — same advisory failing on `main`, not introduced by this PR. Admin-merging.

bokelley:
Acknowledged — noted the three converged approvals and the admin-merge decision. The follow-up items (logging comment in `raceWithSignal`, internal-IP denylist consideration in `detectA2AOrMcp`, `tasks/cancel` wiring, coverage gap on second `getAdcpCapabilities` abort) are non-blocking and can be tracked as separate issues when the PR lands.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_014GfvnvnfhuQW8sWpVYjKwa_

---

PR #1620, opened 2026-05-09
feat(comply): fire-and-forget A2A tasks/cancel when pollTaskCompletion aborts (#1617)

bokelley:
Refs #1617

## Summary

- Adds `cancelA2ATask(agentUrl, taskId, authToken)` to `src/lib/protocols/a2a.ts` — a raw JSON-RPC 2.0 POST (`method: "tasks/cancel"`, `params: { id }`, `id: null`) as required by A2A 0.3.0 §7.4.
- Wires it into `pollTaskCompletion` (`TaskExecutor.ts`) at the top-of-loop abort check: when `signal.aborted` and the agent is A2A and `taskId` is non-empty, fires the cancel as fire-and-forget; network errors, `TaskNotCancelableError`, and terminal-state races are swallowed.
- The caller's `TaskResult` is unchanged (`{ success: false, status: 'failed', error: '…cancelled' }`).
- Updates `SubmittedContinuation.waitForCompletion` JSDoc to document the new behavior.
- Adds changeset `a2a-cancel-on-abort.md` (minor).

## Signing note

`signingContextStorage` AsyncLocalStorage is **not active** during `pollTaskCompletion` (different call stack from `callA2ATool`). The signing branch in `cancelA2ATask` is a forward-compatibility hook. In practice Phase 1 cancel calls are unsigned; a `signed-requests` seller will 401 non-fatally. Phase 2 (tracked on the issue) can address this.

## What was tested

- `npm run typecheck` — passes
- `npm run format:check` — passes
- `npm run build:lib` — passes (with local stub schemas; CI will use real schemas)
- 4 unit tests in `test/lib/poll-task-completion-a2a-cancel.test.js` covering:
  - cancel fired with correct JSON-RPC body (`method`, `params.id`, `id: null`)
  - cancel NOT fired for MCP protocol agents
  - cancel failure is non-fatal (TaskResult unaffected)
  - cancel skipped when `taskId` is empty string

## Pre-PR review

Three expert subagents (protocol, product, security) consulted before implementation.

**Protocol expert:** Confirmed `tasks/cancel` is the correct A2A 0.3.0 §7.4 verb; `id: null` is correct JSON-RPC 2.0 notification form; raw fetch needed since `@a2a-js/sdk` buyer client has no `cancelTask()` method.

**Product expert:** Approved fire-and-forget framing; confirmed cancel prevents seller billing for abandoned tasks; Phase 1 scope (A2A only) is appropriate.

[...]

PR #1625, opened 2026-05-09
feat(security): SSRF policy gate on buyer-side discovery (#1618)

### LINKED ISSUE #1618: security(discovery): SSRF denylist on detectProtocol / discoverAgentProfile (#1612 follow-up)
## Summary

`detectProtocol` in `src/lib/utils/protocol-detection.ts` issues `fetch()` calls to `<agentUrl>/.well-known/agent.json` and `<agentUrl>/.well-known/agent-card.json` with no allowlist or internal-IP filtering. Today this is operator-controlled (CLI arg, saved alias, brand-resolver output) so the SSRF surface is the operator's own host. **If `comply()` is ever exposed as a server-side service** taking buyer-supplied agent URLs (multi-tenant SaaS comply, hosted CI integration, agent registry probe-on-demand, …), this becomes an unbounded SSRF vector.

This was raised as a defense-in-depth note in the security review of #1613:

> "Low (defense in depth): if `agentUrl` is ever sourced from a less-trusted channel (e.g., a buyer-supplied registry record consumed by a server-side comply runner), `fetch()` against arbitrary hosts with no allowlist is the ambient risk; consider an internal-IP block at the `discoverAgentProfile` boundary if comply ever runs server-side against tenant-supplied URLs. Not a blocker." — `security-reviewer`

## Current state

Not exploitable in the shipped use case (CLI on a trusted operator host). All probes resolve via `new URL(path, url)` so they cannot escape the host of `agentUrl`. There is no internal-IP / loopback / link-local filter.

## Concrete risks (post-shipping if a server-side comply lands)

A tenant-supplied `agentUrl` of any of these would hit infrastructure the operator did not intend to probe:

- `http://169.254.169.254/.well-known/agent.json` — AWS / GCP / Azure metadata service
- `http://localhost:6379/.well-known/agent.json` — Redis / cache backends behind the comply runner
- `http://127.0.0.1:8200/.well-known/agent.json` — Vault, internal admin UIs
- `http://10.0.0.0/8` / `http://172.16.0.0/12` / `http://192.168.0.0/16` — internal subnets
- `http://[::1]/.well-known/agent.json` — IPv6 loopback
- `http://[fe80::]/.well-known/agent.json` — IPv6 link-local

The 5xx-as-A2A-suspect classification from #1613 means even denied/blocked responses still influence the SDK's downstream protocol selection — though the actual data exfiltration concern is separate from the protocol classification.

## Proposed approach

Bolt SSRF-safe URL resolution into the discovery boundary. Probably:

1. **A new helper** `assertPublicHttpUrl(url, options)` in `src/lib/utils/`.
   - Resolves the host via DNS.
   - Rejects literals that resolve to RFC 1918, RFC 4193, RFC 3927, loopback, link-local, multicast, or reserved ranges.
   - Optional allowlist (operator config) for testing/dev hosts the operator does want to reach (e.g. `localhost` during local dev — gated by `NODE_ENV === 'test' | 'development'` per memory `feedback_node_env_allowlist.md`).
   - Optional explicit-deny override (`ADCP_ALLOW_INTERNAL_PROBES=1`) for the rare server-side operator who genuinely needs to probe their own infrastructure — same allowlist gate.

2. **Apply at `detectA2AOrMcp` and `discoverAgentProfile` entry points.** Both are buyer-side and the only places `fetch(...)` is issued against an unvalidated agent URL.

3. **TOCTOU mitigation:** resolve the URL once, validate the IP, and `fetch` against the resolved IP with a `Host:` header — otherwise an attacker can DNS-rebind between validation and fetch. This is the expensive part; punt to a follow-up if not already needed.

## Out of scope

- Filtering data sent in the request (we only ever GET; no body leakage).
- Retroactive validation of saved aliases (operator-trust path).
- Anything in the buyer-side AgentClient transport layer beyond discovery — those calls happen after the buyer has chosen to engage with the agent.

## When this becomes blocking

The moment a server-side comply runner lands (hosted CI, registry probe-on-demand, agent marketplace listing-validation), this needs to be done. Today the CLI is operator-trusted, so it's defense-in-depth, not a vulnerability.

## Tracking

Filed as the security-reviewer-flagged follow-up to #1612 / #1613. Specifically called out as "not a blocker for this PR" — file before the next server-side comply integration.

bokelley:
## Summary

Closes #1618. Adds a hostname-literal SSRF refusal at the two buyer-side discovery integration sites identified in triage. Implementation matches the [greenlit design comment](https://github.com/adcontextprotocol/adcp-client/issues/1618#issuecomment-4412924925).

## Two integration sites

1. **`src/lib/utils/protocol-detection.ts:detectA2AOrMcp`** — gate fires BEFORE the existing try/catch loop. Critical: placing it inside would let `catch { suspect = true }` silently convert a denied URL into `'a2a'` (the catch-swallow class flagged by code-reviewer in triage).
2. **`src/lib/testing/client.ts:createTestClient`** — gate fires at URL construction. Every downstream transport call inherits the agent URI's policy verdict — guarding once here covers `getAgentInfo` / `getAdcpCapabilities` / etc. without instrumenting them.

## Default policy

| Range | Default | Why |
|---|---|---|
| Loopback (`127.0.0.0/8`, `::1`, `localhost`) | allow | Local dev loops, mock-server tests, `npm run dev` |
| Cloud metadata (`169.254.169.254`, `fe80::/10`) | always-deny | IMDS exfiltration is never legitimate |
| RFC-1918 / link-local / IPv6 ULA / CGNAT | deny by default | Internal subnets behind a server-side comply runner |
| Public IPv4/IPv6 | allow | The whole point |

## Single opt-out

`ADCP_ALLOW_INTERNAL_PROBES=1` widens the default to allow RFC-1918/link-local/ULA. **Read once at module load** — no `NODE_ENV` gate (per security review note: unsafe in multi-tenant staging where `NODE_ENV=test` images would inherit looser SSRF posture than production). IMDS stays refused regardless.

## Reuses existing infrastructure

- `src/lib/net/address-guards.ts` (`isAlwaysBlocked`, `isPrivateIp`) — Node `BlockList`-backed, canonicalizes IPv4-mapped IPv6 natively (closes the `::ffff:169.254.169.254` bypass discussed in triage).
- `src/lib/net/ssrf-fetch.ts` (`SsrfRefusedError`) — existing typed error with `code` field. No new error class.
- New `src/lib/utils/probe-policy.ts` is the policy layer above those primitives — synchronous, no DNS, returns a typed verdict.

## Known gap (deliberate)

Hostname-literal classification only. A DNS rebind that resolves `evil.example.com` → `169.254.169.254` between the gate and `fetch` would slip past. The DNS-pinned `ssrfSafeFetch` primitive (already used in JWKS / revocation / agent-resolver paths) covers that vector — wiring `detectProtocol` to route through it is the natural follow-up but adds undici-fetch as a hot-path dependency. Filed as TOCTOU follow-up; not blocking this PR.

## Hostname-only error messages

Error text names only the hostname, never the resolved IP. Per the security-reviewer's earlier note: a counterparty-supplied hostname that resolves into the buyer's internal address space would otherwise leak network topology into compliance reports and log aggregators. The resolved address is on `.address` for programmatic debugging.

## Test coverage

- **`test/lib/probe-policy.test.js`** (16 tests) — every range in the table above, IPv4-mapped IPv6 normalization, hostname-only error text, `ADCP_ALLOW_INTERNAL_PROBES=1` opt-in via spawned subprocess (env is read once at module load, can't flip mid-test).
- **`test/lib/ssrf-discovery-integration.test.js`** (11 tests) — explicit catch-swallow regression guard for `detectA2AOrMcp`, `createTestClient` accepts loopback / refuses IMDS / refuses RFC-1918, IPv6 link-local rejection.

Full suite: 8387 / 8390 pass (3 pre-existing skips, 0 fail).

## Bundling with #1620

The Release PR for 6.16.0 (#1622) is currently open with #1620's changeset. Merging this PR adds `fix-1618-ssrf-policy.md` to the same Release PR — the auto-aggregator will roll both into a single 6.16.0 minor.

## Test plan

- [x] New tests: 27/27 pass
- [x] Full suite: 8387/8390 pass (only pre-existing flakes remain)
- [ ] Expert review (code-reviewer + security-reviewer) — to be triggered after this PR opens

### DIFF
--- a/.changeset/fix-1618-ssrf-policy.md
+++ b/.changeset/fix-1618-ssrf-policy.md
@@ -1,0 +1,46 @@
+---
+"@adcp/sdk": minor
+---
+
+feat(security): SSRF policy gate on buyer-side discovery (adcp-client#1618)
+
+When the SDK auto-detects an agent's protocol (`detectProtocol`) or builds
+a `TestClient` for a comply / storyboard run (`createTestClient`), a new
+URL-policy gate refuses obvious SSRF targets before any network probe.
+
+**Default policy** (matches the design proposed and greenlit in #1618):
+
+| Range | Default | Why |
+|---|---|---|
+| Loopback (`127.0.0.0/8`, `::1`, `localhost`) | allow | Local dev loops, mock-server tests, `npm run dev` |
+| Cloud metadata (`169.254.169.254`, `fe80::/10`) | always-deny | IMDS exfiltration is never legitimate |
+| RFC-1918 / link-local / IPv6 ULA / CGNAT | deny | Internal subnets behind a server-side comply runner |
+| Public IPv4/IPv6 | allow | The whole point |
+
+**Single opt-out:** `ADCP_ALLOW_INTERNAL_PROBES=1` widens the default to
+allow RFC-1918/link-local/ULA destinations. Read **once at module load** —
+no `NODE_ENV` gate (unsafe in multi-tenant staging where `NODE_ENV=test`
+images run in production posture). Cloud-metadata addresses stay refused
+even with the opt-in.
+
+**IPv4-mapped IPv6 normalization:** `::ffff:169.254.169.254` and
+`::ffff:127.0.0.1` are canonicalized via Node's `BlockList` so attackers
+can't bypass the policy by choosing a non-standard textual form (URL
+parsers canonicalize to binary form `::ffff:a9fe:a9fe`).
+
+**Refusal type:** existing `SsrfRefusedError` (from `src/lib/net/`) with
+`code: 'always_blocked_address' | 'private_address'`. User-visible error
+text names only the hostname — no resolved IP echo, so refusal logs in
+compliance reports don't leak internal network topology.
+
+**Known gap (TOCTOU):** `classifyProbeUrl` is hostname-literal only; a
+DNS rebind that resolves `evil.example.com` → `169.254.169.254` between
+this gate and `fetch` would slip past. The DNS-pinned `ssrfSafeFetch`
+primitive (already used elsewhere in the SDK) covers that vector.
+Wiring `detectProtocol` to route through `ssrfSafeFetch` for full
+TOCTOU defense is tracked as a follow-up.
+
+**Breaking-ish:** patches existing CLI usage that targets RFC-1918
+addresses without setting `ADCP_ALLOW_INTERNAL_PROBES=1`. The dominant
+case (CLI against `localhost` or public agents) is unchanged. Minor
+bump per semver.
--- a/src/lib/testing/client.ts
+++ b/src/lib/testing/client.ts
@@ -107,16 +107,15 @@ export function createTestClient(agentUrl: string, protocol: 'mcp' | 'a2a' = 'mc
   // in the user-visible text; see `probe-policy.ts` rationale).
   const policy = classifyProbeUrl(agentUrl);
   if (!policy.allowed) {
-    let parsed: URL | undefined;
-    try {
-      parsed = new URL(agentUrl);
-    } catch {
-      /* leave parsed undefined; SsrfRefusedError carries the raw url field */
-    }
+    // `classifyProbeUrl` already returned `{ allowed: true }` for any URL
+    // that fails `new URL(...)`, so reaching the !allowed branch implies the
+    // URL parses cleanly. Reparse only to extract the bare hostname for the
+    // SsrfRefusedError meta.
+    const hostname = new URL(agentUrl).hostname.replace(/^\[|\]$/g, '');
     throw new SsrfRefusedError(
       policy.code === 'always_blocked' ? 'always_blocked_address' : 'private_address',
       policy.reason,
-      { url: agentUrl, hostname: parsed?.hostname.replace(/^\[|\]$/g, '') ?? agentUrl }
+      { url: agentUrl, hostname }
     );
   }
 
--- a/src/lib/testing/client.ts
+++ b/src/lib/testing/client.ts
@@ -18,6 +18,8 @@ import type {
 import type { TestOptions, TestStepResult, AgentProfile, TaskResult, Logger } from './types';
 import { TOOL_RESPONSE_SCHEMAS } from '../utils/response-schemas';
 import { parseCapabilitiesResponse } from '../utils/capabilities';
+import { classifyProbeUrl } from '../utils/probe-policy';
+import { SsrfRefusedError } from '../net/ssrf-fetch';
 
 /**
  * Extract a principal identifier from TestOptions auth.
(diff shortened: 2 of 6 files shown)


bokelley:
**Pre-merge expert reviews — all three converged on approve.**

- **security-reviewer:** "Verdict: ship." All four checks verified (catch-swallow placement, `localhost` exact-match, env-flag handling, hostname-only error contract). Two known follow-up gaps listed (out of scope — JWKS/signing already use `ssrfSafeFetch`).
- **code-reviewer:** "Ready to push." All six checks confirmed. Confirmed existing tests don't hit RFC-1918 (no breakage). Subprocess-based env tests are CI-portable.
- **ad-tech-protocol-expert:** "Sound-with-caveats." Spec-aligned with OWASP ASVS V12.6 + Octokit/prebid/IAB precedent. Three caveats are documentation/follow-up suggestions, not blockers.

Two small fold-ins applied as commit `f1fa5748`:
- Dropped dead defensive try/catch around `new URL(url)` at both integration sites — `classifyProbeUrl` returns `{ allowed: true }` for unparseable URLs, so the `!allowed` branch implies clean parsing.
- Linked TOCTOU follow-up #1627 from `probe-policy.ts` header so next reader knows the literal gate isn't sufficient on its own.

Larger follow-ups tracked separately:
- **#1627** — wire `detectProtocol` through `ssrfSafeFetch` for DNS-rebind defense.
- Documentation note for intranet sellers (protocol-expert caveat) — to be filed against `adcontextprotocol/adcp` separately.

Bundles into 6.16.0 alongside #1620's `tasks/cancel` work via auto-aggregation in Release PR #1622.

bokelley:
Acknowledged. Three-way expert convergence noted, fold-ins landed in `f1fa5748`, follow-ups tracked in #1627 and the cross-repo doc note. No further action from triage — this is in human hands for merge.

---

bokelley:
**Triage note — fold candidate:** Issue #1627 ("wire detectProtocol through ssrfSafeFetch for DNS-rebind defense") touches the same file (`src/lib/utils/protocol-detection.ts`) and was filed by the same author as a direct follow-on to this PR's known gap. The scope is small — swapping the `fetch()` call in `probeWellKnown` for `ssrfSafeFetch` with `allowPrivateIp: isInternalProbesAllowed()`, plus a `maxBodyBytes` cap.

Consider folding #1627 into this diff before merge rather than landing a separate follow-up PR. The "Why deferred" note in #1627 cites undici hot-path concern + perf check — if those have been addressed or are acceptable given the PR's existing undici footprint, now is the cheapest merge window.

If you'd prefer to keep the scope as-is and ship separately, no action needed — #1627 resurfaces automatically when this PR closes.

---

bokelley:
Holding scope as-is. The undici hot-path + connection-pooling concern in #1627 isn't resolved, and all three expert sign-offs cover the current diff — expanding scope now would invalidate those reviews. The deliberate-gap framing in the PR body and changeset was the right call. #1627 will resurface automatically when this closes.

---

PR #1632, opened 2026-05-09
fix(security): detectProtocol through ssrfSafeFetch for DNS-pin / TOCTOU defense (#1627)

### LINKED ISSUE #1627: security(discovery): wire detectProtocol through ssrfSafeFetch for DNS-rebind defense (#1618 TOCTOU follow-up)
## Summary

`classifyProbeUrl` (added in #1625, shipping in 6.16.0) is hostname-literal only — it inspects `URL.hostname` without resolving DNS. A hostname like `evil.example.com` that resolves to `169.254.169.254` between the policy gate and the actual `fetch` call slips past the literal check.

## Current mitigation

The DNS-pinned `ssrfSafeFetch` primitive in `src/lib/net/ssrf-fetch.ts` already exists for this vector — it resolves once, validates the address set, and pins the connect to the first validated address via an undici `Agent` whose `connect.lookup` returns the pinned tuple. Already used in:
- `src/lib/signing/jwks-https.ts`
- `src/lib/signing/revocation-https.ts`
- `src/lib/signing/agent-resolver/fetch-helpers.ts`

But **NOT used by `detectProtocol`** — that path still uses native `fetch`. Closing the gap means routing the well-known card probe through `ssrfSafeFetch` (with `allowPrivateIp: !STRICT_MODE`).

## Why deferred

The shape change adds a hot-path dependency on undici fetch and shifts behavior under load (different connection pooling). Worth landing alongside a baseline-perf check.

## Fix sketch

```ts
// src/lib/utils/protocol-detection.ts:detectA2AOrMcp
async function probeWellKnown(url: string, timeoutMs: number) {
  // classifyProbeUrl already ran upstream
  const result = await ssrfSafeFetch(url, {
    timeoutMs,
    allowPrivateIp: isInternalProbesAllowed(),
    headers: { Accept: 'application/json, */*' },
    maxBodyBytes: 4 * 1024, // card is small; cap tightly
  });
  return result.status;
}
```

Follow-on: same treatment in `discoverAgentProfile` or any other place that fetches an agent-supplied URL.

## Tracking

Filed from the protocol-expert review of #1625 and the security-reviewer's TOCTOU-acknowledged caveat. Not blocking the 6.16.0 release.

bokelley:
## Summary

Closes #1627. Closes the TOCTOU rebind window left open in #1618's hostname-literal gate.

Before this change, `detectProtocol` used native `fetch`, which performs its own DNS lookup at connect time — a hostname like `evil.example.com` that resolves to `169.254.169.254` would slip past the literal-IP gate and reach IMDS regardless. Native fetch also auto-followed `Location:` headers, so a 302 to an internal URL would bounce through the SSRF defense.

## Change

`detectA2AOrMcp` now routes the well-known card probe through `ssrfSafeFetch`, which:

- resolves DNS once,
- validates the full address set against `address-guards` (handles IPv4-mapped IPv6 canonicalization natively via Node's `BlockList`),
- pins the connect to the first validated address via undici's `Agent.connect.lookup` (defeats rebind between validation and connect),
- sets `redirect: 'manual'` so a 302 to an internal URL is not auto-followed,
- caps the response body at 4 KiB (the agent card is small; tightens malicious-slow-body window).

The hostname-literal `classifyProbeUrl` gate from #1618 stays in place as cheap synchronous defense in depth.

## SSRF-error classification preserved

The catch handler distinguishes policy refusals (must propagate — caller is reaching for SSRF targets) from runtime/network conditions (treat as suspect — host unreachable but might be a slow A2A seller):

| Code | Disposition |
|---|---|
| `always_blocked_address`, `private_address`, `body_exceeds_limit`, `scheme_not_allowed`, `non_https_without_opt_in`, `invalid_url` | **propagate** — silently downgrading to `suspect` reintroduces the catch-swallow class flagged in #1618 review |
| `dns_lookup_failed`, `dns_empty` | **suspect** — preserves pre-#1627 behavior so CLI tests using `*.example.invalid` keep their existing exit codes |

## Test migration

Classification tests moved from `globalThis.fetch` mocks to real loopback HTTP servers (matches the `net-ssrf-fetch.test.js` pattern). New `protocol-detection-toctou.test.js` adds explicit DNS-pin defense assertions:

- IMDS literal refusal (sync, hostname-literal gate)
- IPv4-mapped IPv6 IMDS (canonicalization via `BlockList`)
- IPv6 link-local literal
- **Catch-swallow regression guard** — explicit assertion that `SsrfRefusedError` propagates and is NOT converted to `suspect = true`
- 302 `Location:` not auto-followed (probes the existence of a follow-on path against the loopback server, asserts it is never visited)

Some redundant RFC-1918 tests removed from same-process integration files — exhaustively covered in `probe-policy.test.js` (subprocess-based env-flag tests).

## No public API change

- Default behavior on public hosts: unchanged
- Default behavior on `http://` URLs: now refused (matches `ssrfSafeFetch`'s scheme guard) — operator must set `ADCP_ALLOW_INTERNAL_PROBES=1` to permit. Production AdCP agents must terminate TLS per spec.
- Refusal types: existing `SsrfRefusedError` (no new error class)

## Test plan

- [x] All 5 detection-related test files (60+ tests): pass
- [x] Full suite: 8329/8332 pass (3 pre-existing skips, 0 fail)
- [ ] Expert review (security-reviewer + ad-tech-protocol-expert) — to be triggered after this PR opens

### DIFF
--- a/.changeset/fix-1627-detect-ssrf-safe.md
+++ b/.changeset/fix-1627-detect-ssrf-safe.md
@@ -1,0 +1,49 @@
+---
+"@adcp/sdk": patch
+---
+
+fix(security): route detectProtocol through ssrfSafeFetch for DNS-pin / TOCTOU defense (adcp-client#1627)
+
+Closes the TOCTOU rebind window left open in #1618's hostname-literal gate.
+Before this change, `detectProtocol` used native `fetch`, which performs
+its own DNS lookup at connect time — a hostname like `evil.example.com`
+that resolves to `169.254.169.254` would slip past the literal-IP gate
+and reach IMDS regardless. Native fetch also auto-followed `Location:`
+headers, so a 302 to an internal URL would bounce through the SSRF
+defense.
+
+`detectA2AOrMcp` now routes the well-known card probe through
+`ssrfSafeFetch`, which:
+
+- resolves DNS once,
+- validates the full address set against `address-guards`,
+- pins the connect to the first validated address via undici's
+  `Agent.connect.lookup` (defeats rebind between validation and connect),
+- sets `redirect: 'manual'` so a 302 to an internal URL is not followed,
+- caps the response body at 4 KiB (the agent card is small; tightens the
+  malicious-slow-body window).
+
+The hostname-literal `classifyProbeUrl` gate from #1618 stays in place
+as cheap synchronous defense in depth.
+
+**SSRF-error classification preserved:** policy refusals
+(`always_blocked_address`, `private_address`, `body_exceeds_limit`,
+`scheme_not_allowed`, `non_https_without_opt_in`, `invalid_url`)
+propagate to the caller — silently converting them to `suspect = true`
+would reintroduce the catch-swallow class. DNS conditions
+(`dns_lookup_failed`, `dns_empty`) fall through to suspect, matching the
+pre-#1627 behavior so CLI tests that use `*.example.invalid` keep
+exiting with the existing CLI exit codes.
+
+**Test migration:** the existing classification tests moved from
+`globalThis.fetch` mocks to real loopback HTTP servers (matches the
+`net-ssrf-fetch.test.js` pattern). New `protocol-detection-toctou.test.js`
+adds explicit DNS-pin defense assertions: IMDS literal refusal,
+IPv4-mapped IPv6 IMDS, IPv6 link-local, catch-swallow regression guard,
+and confirmation that 302 `Location:` headers are not auto-followed.
+
+**No API change.** Default behavior on public hosts is unchanged.
+Behavior on `http://` URLs is now refused without
+`ADCP_ALLOW_INTERNAL_PROBES=1` (matches `ssrfSafeFetch`'s scheme guard).
+Compliance runs against bare-`http://` agents need the env opt-in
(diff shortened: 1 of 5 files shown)


bokelley:
**Pre-merge expert reviews — both converged on approve.**

- **security-reviewer:** "No High or Medium findings. Ship it." All four invariants verified (catch-handler classification, body-cap timing, `redirect: 'manual'` invariant, cross-cutting native-fetch sites listed).
- **code-reviewer:** Two real issues to fix (302-no-bounce test passed for wrong reason; `body_exceeds_limit` shouldn't propagate). Both folded in as `3d040f3b`.

Cross-cutting follow-up filed as #1633 — `network-consistency-checker.ts` + `property-crawler.ts` still use native `fetch` against agent-supplied URLs and have the same TOCTOU window. Out of scope for this PR.

`SSRF_TRANSIENT_CODES` centralization (security-reviewer suggestion) deferred to that follow-up since it touches multiple call sites.

bokelley:
Acknowledged. Both expert reviews converged on approve, the two code-reviewer findings are folded into `3d040f3b`, and #1633 tracks the cross-cutting `network-consistency-checker.ts` / `property-crawler.ts` follow-up. No further triage action needed on this PR.

---

Issue #1617, opened 2026-05-09
feat(comply): wire A2A tasks/cancel when comply abort fires (#1612 follow-up)


bokelley:
## Summary

When `comply()` aborts (timeout fires or external signal cancels) mid-flight, the SDK currently unblocks the caller but does **not** notify the seller. The orphan transport call keeps running on the seller side until it self-terminates — billing the seller for compute/network/inventory holds the buyer no longer cares about.

This was raised as a non-blocker follow-up in the protocol-correctness review of #1613:

> "should A2A-suspect path eventually wire `tasks/cancel` so the orphan A2A poll doesn't continue billing the seller after comply gives up? File-follow-up, not merge-blocker." — `ad-tech-protocol-expert`

## Current behavior

After #1613 (shipped in `@adcp/sdk@6.15.1`):
- `comply()` AbortSignal fires.
- `discoverAgentProfile`'s `raceWithSignal` wrapper rejects.
- `pollTaskCompletion` exits its loop on the next `signal.aborted` check.
- The underlying transport requests (MCP `tools/call`, A2A `message/send`, A2A `tasks/get` poll cycle) keep running until they naturally settle.

For long-running tool calls (`create_media_buy` with seller-side approval workflow, `sync_creatives` with provider validation, etc.), the seller continues to do work for a buyer who has already given up.

## Spec landscape

- **A2A 0.3.0 §7.4** — defines `tasks/cancel` JSON-RPC method. The seller responds with the canceled task or `TaskNotCancelableError`.
- **AdCP 3.0** — has no buyer→seller cancellation verb. Task lifecycle defines `canceled` as a terminal state but no operation that drives it.
- **MCP** — the protocol itself has no equivalent (tool calls are request/response). Cancellation is whatever the underlying transport (`fetch` AbortController) supports, and most MCP servers don't propagate transport-level cancel into upstream work.

## Proposed approach

**Phase 1 (A2A only, additive):**
- When `pollTaskCompletion`'s AbortSignal fires AND the transport is A2A AND we have a `taskId`, fire-and-forget an A2A `tasks/cancel` before exiting.
- Failure to cancel (network error, `TaskNotCancelableError`, terminal-state race) is non-fatal; the buyer was already abandoning the task.

**Phase 2 (cross-protocol, spec dependency):**
[...]

PR #1640, opened 2026-05-10
fix(a2a): sign tasks/cancel POST against signed-requests sellers (#1617 Phase 2)

bokelley:
## Summary

Closes the Phase 2 follow-up from #1617. Phase 1 (6.16.0) sent the A2A `tasks/cancel` POST unsigned — documented limitation. A `signed-requests` seller that enforces signing on the cancel path would 401 it, defeating the orphan-prevention goal of cancel-on-abort.

## Change

`cancelA2ATask(agentUrl, taskId, authToken)` → `cancelA2ATask(agent, taskId)`. When `agent.request_signing` is configured, routes the cancel POST through `createSigningFetch` (inline keys) or `createSigningFetchAsync` (provider keys). When unset, falls through to the unsigned Phase 1 path.

## Why rebuild signer per cancel (not replay ALS)

`signingContextStorage` is set around `callA2AToolImpl` and doesn't extend into `pollTaskCompletion`'s sibling promise tree, so capture-and-replay isn't viable from the cancel call site. Rebuilding from `agent.request_signing` is equivalent — that's the same source of truth `buildAgentSigningContext` reads at submission time, and the agent's signer key is the right primitive to grab here.

## Why bypass `buildAgentSigningFetch`'s capability gate

That path consults `request_signing.supported_for` to decide whether to sign, but `tasks/cancel` is an A2A 0.3.0 §7.4 protocol method, not an AdCP tool. Sellers typically list AdCP tool names in `supported_for` (test-agent's 28 entries are all AdCP tool names — none of them `tasks/cancel`). Real-world sellers either (a) check signing on every mutating POST uniformly at the HTTP layer, or (b) only check on specific AdCP tools. Behavior the buyer wants in both cases: if the agent claims signing at all, sign the cancel.

## Test fixture

In-process loopback HTTP server runs the SDK's own `verifyRequestSignature` against incoming `tasks/cancel` POSTs. Mirrors what a real signed-requests seller does at the verifier seam. Three tests:

1. **Signed and accepted** — the Phase 2 fix. Verifier inside the test server returns 200 → cancel succeeds.
2. **Unsigned regression guard** — agent without `request_signing` still goes through the Phase 1 unsigned path; no Signature headers attached.
3. **Bearer + signature coexistence** — confirms `Authorization: Bearer` and `x-adcp-auth` stay on the signed path (signing wraps the auth-injecting fetch, doesn't replace it).

## Upstream coordination

Filed [`adcontextprotocol/adcp#4314`](https://github.com/adcontextprotocol/adcp/issues/4314) asking test-agent to add a per-session strict-mode header (`X-Test-Require-Signing`). Once that lands, an integration test can exercise this fix against the production fixture too. Until then, this in-process fixture covers the regression surface.

## API change

`cancelA2ATask` is an internal helper. The single call site (`TaskExecutor.pollTaskCompletion`) is updated; no public API impact.

## Test plan

- [x] New `cancel-a2a-task-signing.test.js`: 3/3 pass
- [x] Full suite: 8401/8449 pass (1 known order-dependent flake in `server-a2a-adapter`, passes in isolation)
- [ ] Expert review (security-reviewer + ad-tech-protocol-expert) — to be triggered after this PR opens

### DIFF
--- a/.changeset/fix-1617-phase2-cancel-signing.md
+++ b/.changeset/fix-1617-phase2-cancel-signing.md
@@ -1,0 +1,42 @@
+---
+"@adcp/sdk": patch
+---
+
+fix(a2a): sign tasks/cancel POST against signed-requests sellers (#1617 Phase 2)
+
+Phase 1 (`@adcp/sdk@6.16.0`) sent the A2A `tasks/cancel` POST unsigned —
+documented limitation. A `signed-requests` seller that enforces signing on
+the cancel path would 401 it, defeating the orphan-prevention goal of
+cancel-on-abort.
+
+`cancelA2ATask` now takes the full `AgentConfig` and, when
+`agent.request_signing` is configured, routes the cancel POST through
+`createSigningFetch` (inline keys) or `createSigningFetchAsync` (provider
+keys). The signing wrapper is rebuilt per cancel rather than replayed from
+`signingContextStorage` — that ALS scope is set around `callA2AToolImpl`
+and doesn't extend into `pollTaskCompletion`'s sibling promise tree, so
+capture-and-replay isn't viable. Rebuilding from `agent.request_signing`
+is equivalent: that's the same source of truth `buildAgentSigningContext`
+reads from at submission time.
+
+**Bypass `buildAgentSigningFetch`'s capability gate**: that path consults
+the seller's `request_signing.supported_for` to decide whether to sign,
+but `tasks/cancel` is an A2A protocol method, not an AdCP tool. Sellers
+typically list AdCP tool names there, not protocol-level methods. The
+shape that matches actual seller behavior: if the agent claims signing
+at all, sign every mutating POST we send. Sellers with uniform "must be
+signed" policies accept this; sellers that only check signing on specific
+AdCP tools simply ignore the extra signature.
+
+**Test fixture**: in-process loopback HTTP server runs the SDK's own
+`verifyRequestSignature` against incoming `tasks/cancel` POSTs (mirrors
+what a real signed-requests seller does at the verifier seam). Three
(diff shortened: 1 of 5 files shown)


bokelley:
**Pre-merge expert reviews — both approve.**

- **security-reviewer:** "Verdict: Ship. No High or Medium findings." All five concerns walked: capability-gate bypass right-shape, no reserved-header collisions, provider rotation is intentional rotation semantics, AbortSignal survives the wrap, body re-stringification clean. One Low (typed FetchLike export) folded in.
- **ad-tech-protocol-expert:** "Sound-with-caveats." Confirmed `supported_for` keys on AdCP tool names per the 3.0.8 schema, so capability-gate bypass is correct. Caught a real defense-in-depth gap: `cancelA2ATask` should assert `protocol === 'a2a'` since POSTing tasks/cancel JSON-RPC at an MCP endpoint would 404. Folded in.

Spec gap (Q2/Q4) filed upstream as **adcp#4318** — declare verifier coverage of A2A protocol methods (tasks/cancel, tasks/get) in request_signing capability schema. Once the spec defines protocol-method coverage, the buyer's "always sign mutating POST when key present" default narrows to spec-driven.

Both fold-ins applied as `4f6a1d9e`. Ready for admin-merge once CI clears.

bokelley:
Acknowledged — both expert sign-offs logged, fold-ins confirmed at `4f6a1d9e`, spec gap tracked upstream as adcp#4318. Standing by; no action needed from triage until CI result or a follow-up request lands.

---
Triaged by Claude Code. Session: https://claude.ai/code/${CLAUDE_CODE_REMOTE_SESSION_ID}

---

PR #1651, opened 2026-05-10
fix(conformance): write compliance/CACHE_VERSION marker to detect stale storyboard cache

### LINKED ISSUE #1648: @adcp/sdk 6.17.0 ships compliance/cache/3.0.6 — 2 spec versions stale
## Summary

The published `@adcp/sdk@6.17.0` bundles `compliance/cache/3.0.6/...` storyboard YAMLs, but the AdCP spec is at **3.0.8**. So even when fixes land in `adcontextprotocol/adcp` (e.g. #4218, #4230, the remaining static-key sweep filed today), they don't reach SDK adopters until the bundle is regenerated and a new SDK version published.

## Repro

```bash
$ npx -y @adcp/sdk@6.17.0 --version
@adcp/sdk@6.17.0

$ ls node_modules/@adcp/sdk/compliance/cache/
3.0.6/    # <-- two minor versions behind spec
```

Inspecting the cached YAMLs shows behavior the spec source has presumably moved past — e.g. the surviving 24 static `idempotency_key` strings we're filing as a follow-up to adcp#4230. Even after that follow-up lands in spec source, SDK users will still hit the static keys until the cache is regenerated.

## Why this matters

Storyboard validation is the buyer-side proof of seller compliance. If the cached scenarios in the SDK lag the spec, the validation grades sellers against an old contract — false positives (failing on behavior the seller actually fixed) and false negatives (passing on behavior the spec has tightened) both happen.

We're hitting this concretely: shape changes in our seller surface as storyboard failures because the SDK bundle's static-key scenarios collide with cached idempotency responses.

## Proposed fix

Either:

1. **Regenerate the bundled cache on every SDK release** — CI step that pulls `compliance/source/...` from `adcontextprotocol/adcp@<spec_version>` matching the SDK's pinned spec version, runs whatever build step produces the cache. Bumps cache version with each spec bump.

2. **Fetch latest cache at install/invocation time** — `npx -y @adcp/sdk` could fetch the current spec's cache from a CDN or from `adcontextprotocol/adcp` directly. Avoids the rebuild-cycle entirely but introduces a network dependency at storyboard-runtime.

(1) is more robust and is what most SDK-spec relationships look like. (2) is more responsive to spec changes but riskier for offline users.

Either way, the SDK should *advertise* which spec version its bundled cache corresponds to (already does via the path), and its release notes / `--version` should mention any spec-version drift the user might care about. Currently SDK 6.17.0 says "AdCP 3.0.8" in its compliance report header but ships 3.0.6 cache — silent inconsistency.

## Related

- adcp#XXXX (follow-up to #4230 — surviving static keys we just filed)
- adcp-client#1612, #1614 (other SDK-vs-spec drift filed earlier)

## Environment

- @adcp/sdk: 6.17.0
- Spec advertised in CLI output: 3.0.8
- Bundled cache path: `compliance/cache/3.0.6/`

bokelley:
Closes #1648

## Summary

`compliance/cache/` is gitignored (populated at sync/publish time), so `schema-sync.yml`'s diff guard was blind to compliance-only spec bumps — new storyboard YAMLs, fixed `idempotency_key` strings — any spec release that didn't also change TypeScript schemas would leave stale storyboards in the published npm artifact. This caused the silent mismatch the issue describes: `--version` reports `AdCP 3.0.8` while `compliance/cache/` still contains `3.0.6` storyboards, producing false-positive and false-negative results in buyer-side storyboard validation.

Root cause is two-part:
1. The diff guard in `schema-sync.yml` watched only `src/lib/types/`, `src/lib/agents/`, and `package.json` — invisible to compliance-only spec changes.
2. No validation caught the version mismatch before publishing.

The fix avoids removing `compliance/cache/` from `.gitignore` (which would add large YAML bulk to git history). Instead, `sync-schemas.ts` writes a tiny tracked marker file `compliance/CACHE_VERSION` (one level above the gitignored `compliance/cache/`) that CI can diff against.

**Changes:**
- `scripts/sync-schemas.ts`: writes `compliance/CACHE_VERSION` after every tarball sync, using `adcp_version` from the tarball's `index.json`. The per-file fallback path (tarball endpoint 404) also writes the marker using the `ADCP_VERSION` pin — compliance storyboards aren't actually synced on that path, but the marker prevents a hard CI failure.
- `.github/workflows/schema-sync.yml`: diff guard now includes `compliance/CACHE_VERSION` via `git status --short` (detects both new/untracked and modified marker files). Version capture moved before the diff check so PR titles are correct for compliance-only bumps too.
- `.github/workflows/ci.yml`: post-sync validation step asserts `compliance/CACHE_VERSION` matches `ADCP_VERSION`, catching version drift before any PR can merge.

**Known limitation:** The per-file fallback path writes the marker with the `ADCP_VERSION` pin rather than the tarball's embedded `adcp_version`, since no tarball is available. This reflects intent (the pin) rather than actual cached content — the pre-existing `console.warn` documents that compliance storyboards are not synced on this path. The asymmetry is safe for semver pins but would produce a mismatch if `ADCP_VERSION` were ever set to `"latest"` (pre-existing concern, not introduced here).

## What tested

- `npm run format:check`: ✅ passes
- Pre-existing typecheck errors (`Cannot find type definition file for 'node'`, deprecated `moduleResolution`) verified to exist on `main` before this branch — not introduced here
- `git diff HEAD~2 HEAD`: reviewed manually; all imports already present in `sync-schemas.ts`; `path.dirname(COMPLIANCE_CACHE_DIR)` resolves to `compliance/` (confirmed by reviewers); `compliance/` guaranteed by `replaceTree`'s `mkdirSync` before the marker write

## Nits from pre-PR review (not fixed)

- `ci.yml` lines 39/45: `cat FILE | tr -d` is useless-use-of-cat; `tr -d '[:space:]' < FILE` is idiomatic — no correctness impact
- `schema-sync.yml` line 58: `2>/dev/null` suppresses stderr on `git status`; if git exits non-zero for any reason the compliance bump is silently missed — ci.yml's hard check is the safety net, so no release-safety gap, but the failure mode is invisible
- `sync-schemas.ts` tarball path (~line 388): a `// compliance/ guaranteed by replaceTree above` comment would clarify why `mkdirSync` is absent there but present in the per-file fallback path

## Pre-PR review

- **code-reviewer** (round 1): approved after fixing dead `|| cat ADCP_VERSION` fallback in `schema-sync.yml` (pipe through `tr` exits 0 on empty input, silently producing empty string). Fixed in second commit.
- **ad-tech-protocol-expert**: approved after adding per-file fallback marker write in `sync()` — the blocker was that `syncSchemasPerFile` path never wrote the marker, which would have caused `ci.yml` to hard-fail with "not found". Fixed in second commit.
- **code-reviewer** (round 2): approved — no blockers; three nits surfaced above.

> issue). To unblock:
>
> - **Push fixup commits directly:** `gh pr checkout 1651` →
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01JACGQ4QjAH319KJNKjCwQy

### DIFF
--- a/.changeset/fix-compliance-cache-version-marker.md
+++ b/.changeset/fix-compliance-cache-version-marker.md
@@ -1,0 +1,25 @@
+---
+"@adcp/sdk": patch
+---
+
+Fix compliance cache staleness: write tracked `compliance/CACHE_VERSION` marker
+
+`compliance/cache/` is gitignored (populated at sync/publish time) so
+compliance-only spec bumps — new storyboard YAMLs, fixed `idempotency_key`
+strings — never triggered a PR from `schema-sync.yml`. The diff guard only
+watched `src/lib/types/`, `src/lib/agents/`, and `package.json`, making it
+blind to spec changes that don't touch TypeScript schemas.
+
+Fixes the silent mismatch where `--version` advertised one AdCP version while
+`compliance/cache/` contained storyboards from an older release — causing false
+positives and false negatives in buyer-side storyboard validation.
+
+Changes:
+- `scripts/sync-schemas.ts` now writes `compliance/CACHE_VERSION` (tracked in
+  git, one level above the gitignored `compliance/cache/`) after every tarball
+  sync; the file contains the `adcp_version` from the tarball's `index.json`.
+- `schema-sync.yml` diff guard now includes `compliance/CACHE_VERSION` so
+  compliance-only spec bumps trigger an automated PR.
+- `ci.yml` gains a post-sync validation step that asserts
+  `compliance/CACHE_VERSION` matches `ADCP_VERSION`, catching version drift
+  before it reaches npm.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,23 @@ jobs:
       - name: Sync schemas from AdCP
         run: npm run sync-schemas:all
 
+      - name: Validate compliance cache version matches spec
+        run: |
+          EXPECTED=$(cat ADCP_VERSION | tr -d '[:space:]')
+          if [ ! -f compliance/CACHE_VERSION ]; then
+            echo "❌ compliance/CACHE_VERSION not found after sync — sync-schemas may have failed silently."
+            echo "   Run: npm run sync-schemas:all"
+            exit 1
+          fi
+          CACHED=$(cat compliance/CACHE_VERSION | tr -d '[:space:]')
+          if [ "$CACHED" != "$EXPECTED" ]; then
+            echo "❌ Compliance cache version mismatch: bundled '$CACHED', expected '$EXPECTED'"
+            echo "   Stale storyboards cause false pass/fail in buyer-side validation."
+            echo "   Run: npm run sync-schemas:all"
+            exit 1
+          fi
+          echo "✅ compliance/cache version matches spec: $CACHED"
+
       - name: Validate generated files are in sync
         run: |
           npm run generate-types
--- a/.github/workflows/schema-sync.yml
+++ b/.github/workflows/schema-sync.yml
@@ -48,7 +48,8 @@ jobs:
           npm run generate-wellknown-schemas
 
           # Always capture version so PR title is correct for compliance-only bumps too.
-          ADCP_VERSION=$(cat compliance/CACHE_VERSION 2>/dev/null | tr -d '[:space:]' || cat ADCP_VERSION)
+          # compliance/CACHE_VERSION is always written by sync-schemas (tarball or per-file path).
+          ADCP_VERSION=$(cat compliance/CACHE_VERSION | tr -d '[:space:]')
           echo "adcp_version=$ADCP_VERSION" >> $GITHUB_OUTPUT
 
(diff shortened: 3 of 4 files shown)


bokelley:
Closing — source issue #1648 was a misdiagnose (npx cache pointed at a stale install at a different path), so the hazard this guards against hasn't materialized. The existing `ci:schema-check` already catches compliance-only spec drift via metadata-comment changes in generated files (validated on today's 3.0.10 bump in #1655 — the `entity-hydration` / `manifest` / `core.generated` headers all flipped and would have failed CI if the sync had been skipped). A tracked marker file would be defense-in-depth on top of that, but it's surface area for the same outcome.

bokelley:
Acknowledged — makes sense. The `ci:schema-check` coverage you validated on the 3.0.10 bump in #1655 addresses the underlying concern, so this branch isn't needed.

---

bokelley:
Understood — closing the loop here. #1655's `ci:schema-check` coverage makes this branch redundant; no further action needed.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_01CR6y6uzgoLTEt4L7RF1cit_

---

PR #1671, opened 2026-05-11
fix(cli): propagate --allow-http to MCPOAuthProvider.validateResourceURL

bokelley:
Closes #1664

## Summary

`MCPOAuthProvider.validateResourceURL` threw unconditionally on any non-HTTPS `resource` URL, blocking OAuth flows against `http://localhost` dev servers even when `--allow-http` was passed. The flag was parsed by `comply`/`storyboard`/`grade` but never threaded into the `MCPOAuthProvider` constructor.

**Fix (option 3 from the issue):**
- Loopback hosts (`localhost`, `127.0.0.1`, `[::1]`) are always allowed in resource URL validation — matching the `ClientCredentialsFlow.ts` precedent and RFC 6749 §3.1.2.1 loopback carve-out
- Non-loopback HTTP resource URLs are gated on `allowHttp` (CLI: `--allow-http`)
- `--allow-http` is now threaded from all five CLI OAuth call sites through `createCLIOAuthProvider` to the provider
- `--allow-http` added to `OAuthProviderConfig`, `createCLIOAuthProvider` options, and `createNonInteractiveOAuthProvider` options
- `--allow-http` added to top-level `--help` OPTIONS output

## What was tested

- `npm run format:check` ✅
- `npx tsc --project tsconfig.lib.json` — no errors in changed files (2 pre-existing env errors: missing `@types/node` and deprecated tsconfig `node10` option — both pre-date this branch)
- 9 unit tests in `test/mcp-oauth-provider-validate-resource-url.test.js` covering: HTTPS pass-through, loopback carve-out for all three loopback forms (`localhost`, `127.0.0.1`, `[::1]`), non-loopback HTTP rejection with actionable error message, `--allow-http` acceptance, `localhost.attacker.com` non-bypass
- Tests fail in this dev environment due to missing `@modelcontextprotocol/sdk` in `node_modules` (same failure as pre-existing `test/oauth-storyboard-plumbing.test.js`); will pass in CI with full dep install

## Pre-PR review

- **code-reviewer:** approved — no blockers; all 5 CLI OAuth call sites correctly covered; `[::1]` check is correct (bare `::1` unreachable after WHATWG URL normalization — clarified in follow-up commit); `--save-auth` call site re-parses `args` because it's a different block scope without a hoisted `allowHttp` (intentional); changeset present and correctly typed `patch`
[...]

PR #1682, opened 2026-05-11
fix(conformance): grade storyboards not_applicable when required_tools are missing

### LINKED ISSUE #1680: comply() grades storyboards 'partial' when required_tools are missing — should grade not_applicable
## Summary

When a storyboard declares `required_tools` and the discovered toolset on the agent is missing one of them, the runner currently:

1. Attempts the storyboard anyway.
2. Marks the missing-tool step as `Skipped: prior stateful step "<X>" skipped (missing_tool); state never materialized.`
3. Cascades that skip across subsequent steps in the scenario.
4. Reports the scenario as non-passing inside its track, pulling the track to `partial`.

The right behavior is to grade the **storyboard** `not_applicable` at the storyboard level and exclude its scenarios from the track's pass/fail rollup.

## Repro

Wonderstruck (`sales-agent.scope3.com`) declares `specialisms: ["sales-non-guaranteed"]`, not governance. The runner attempts the governance storyboard suite:

- `media_buy_seller/governance_conditions/buy_with_conditions`
- `media_buy_seller/governance_denied/buy_denied`
- `media_buy_seller/governance_denied_recovery/buy_denied` + retried scenario

Each fails with `Skipped: prior stateful step "sync_governance" skipped (missing_tool); state never materialized.`. `sync_governance` is in the storyboard's `required_tools` and is not in the agent's discovered toolset (12 tools, none of them governance).

The `media_buy` track then reports `59/66 partial` instead of the clean `59/59 + 7 not-applicable` that reflects reality.

Same pattern: `pagination_integrity_creative_formats/pagination_walk` requires `seed_creative_format` via `comply_test_controller` (not in Wonderstruck's public toolset).

## Why it matters for AAO grading

AAO badge issuance (`processAgentBadges` in adcontextprotocol/adcp) derives per-role verification from per-storyboard pass/fail. A storyboard that grades `partial` because it requires tools the agent doesn't expose is a false signal — the agent isn't claiming the specialism that storyboard tests. Net effect: heartbeats look messier than they are, and operators have to read storyboard internals to understand why their track is partial.

## Fix

Before running each storyboard, compute the applicability check:

```ts
const required = new Set(storyboard.required_tools ?? []);
const missing = [...required].filter(t => !discoveredTools.has(t));
if (missing.length > 0) {
  return {
    storyboard_id: storyboard.id,
    status: 'not_applicable',
    reason: `missing required_tools: ${missing.join(', ')}`,
    scenarios: [],
  };
}
```

Exclude `not_applicable` storyboards from track-level pass/fail aggregation.

For the AAO grader contract: `not_applicable` is already in the grading vocabulary (see #4226 / #4228 in adcontextprotocol/adcp), so this aligns with the existing post-#4379 (Sandbox) framing.

## Cross-link

Same Wonderstruck probe as #1676–#1679. Findings doc in adcontextprotocol/adcp `.context/wonderstruck-findings.md`.

bokelley:
Closes #1680

## Summary

`comply()` was running every capability-resolved storyboard even when the agent's discovered toolset was missing one or more tools declared in the storyboard's `required_tools` field. The cascading step-level skips caused the storyboard to grade `partial`, which propagated to the track as a false failure — visible in the Wonderstruck probe (`59/66 partial` instead of `59/59 + 7 not-applicable`).

**Changes:**
- In `complyImpl`, after `expandScenarios`, build `runnableStoryboards` by filtering out storyboards/scenarios whose `required_tools` aren't in the agent's discovered toolset. Filtered entries are pushed to `notApplicable` with `reason: "missing required_tools: <list>"` and get a synthetic `not_applicable` result (consistent with the existing version-gating path).
- Explicit storyboard IDs (`options.storyboards`) bypass the filter — they are an operator override and run regardless of `required_tools`.
- Corrects `groupByTrack`, `poolTrackSet`, `extractFailures`, and `storyboards_executed` to reference `runnableStoryboards` rather than the full expanded set.
- Adds `storyboards_not_applicable?: string[]` to `ComplianceResult` so AAO consumers can distinguish intentional exclusions from missing coverage without parsing per-track results.

## What was tested

- `npm run format:check` ✅
- `npx tsc` — only pre-existing env errors (missing `@types/node`, deprecated `moduleResolution=node10`); no errors in changed files. node_modules not installed in this environment — full test suite runs in CI (same constraint as #1671).
- Manual type verification: `Storyboard.title: string` (non-nullable, `storyboard/types.ts:18`); `required_tools?: string[]` (`types.ts:33`); `NotApplicableStoryboard.storyboard_title: string` (`storyboard/compliance.ts:118`) — all types align.

## Pre-PR review

- **code-reviewer:** approved — explicit-storyboard bypass correctly guarded; `storyboards_not_applicable` field added; `sb.title` is non-nullable so no fallback needed; all five `applicableStoryboards` references correctly updated to `runnableStoryboards`
- **ad-tech-protocol-expert:** approved — `not_applicable` with `overall_passed: true` is spec-correct per adcp#4226/#4228; AAO track rollup is clean; explicit-storyboard bypass semantics documented in comment; `storyboards_not_applicable` satisfies AAO grader contract. Note: `runWithDegradedProfile` doesn't emit `storyboards_not_applicable` (vacuously harmless — degraded profile produces no capability-driven exclusions) — follow-up before grader contract is finalized.

## Nits noted (not fixed in this PR)

- No unit test added for the filter path — `node_modules` not available in this environment; test coverage for `comply.ts` is a known gap
- `runWithDegradedProfile` doesn't emit `storyboards_not_applicable` — currently harmless (degraded `notApplicable` is always empty) but worth adding before the AAO grader reads this field

> issue). To unblock:
>
> - **Push fixup commits directly:** `gh pr checkout 1681` →
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_019y9VpBe6T9oSU69GzRicDX

---
_Generated by [Claude Code](https://claude.ai/code/session_019y9VpBe6T9oSU69GzRicDX)_


bokelley:
Tracking under #1685 — hat 5 of 5 (`partial` grade inflation). **Do not merge before #1684** (the visibility keystone). #1690 builds on this PR's `not_applicable` plumbing for the `webhook_receiver` requirement; sequence is #1684 → #1683/#1681 → this → #1690.

bokelley:
Noted — merge sequence recorded: **#1684 → #1683/#1681 → this PR → #1690**. This PR stays draft until #1684 lands. Tracking parent: #1685 (hat 5/5).

---

PR #1681, opened 2026-05-11
fix(normalizer): fail closed on pre-3.0 PackageRequest shapes

bokelley:
Closes #1677

## Summary

`normalizePackageParams` was silently passing through pre-3.0 `PackageRequest` shapes — `packages[].product_ids: string[]` and `packages[].budget: { total, currency }` — which caused 10 conformance scenarios to fail with `INVALID_REQUEST` from 3.0-strict sellers. Both shapes cannot be translated to AdCP 3.0 equivalents without data loss (which id wins from `product_ids[]`? which currency from `budget.currency`?), so the normalizer now fails closed with an early `ValidationError` at the client boundary rather than letting a malformed payload reach the wire.

Also fixes a TypeScript-only cast in `request-builder.ts` (`baseSample.budget as number | undefined`) that had no runtime effect — replaced with a `typeof` check so storyboard fixtures with object budgets correctly fall through to discovery-derived values instead of flowing into `normalizePackageParams`.

v2 sunset policy: v2 unsupported as of 3.0 GA (April 2026); security-only until Aug 1 2026. No translation obligation.

## What was tested

- `npx prettier --check` ✅
- `npx tsc --project tsconfig.lib.json` — no source-file errors in changed files (2 pre-existing env errors: missing `@types/node`, deprecated `moduleResolution=node10` — both predate this branch)
- New test file `test/request-normalizer-package-params.test.js` — 7 cases covering: `product_ids[]` throw, object-budget throw (`{total,currency}` and `{amount,currency}` variants), empty `product_ids[]` passthrough (no ambiguity), valid 3.0-shape passthrough, null/non-object passthrough, and both throw paths via `normalizeRequestParams('create_media_buy', ...)` integration
- Tests require `npm install` (no `node_modules` in triage env — same constraint as #1671/#1670); will run on CI

## Pre-PR review

- **code-reviewer:** approved — no blockers; noted empty `product_ids[]` edge case (fixed: added `.length > 0` guard), simplified budget null check (`!== null && typeof === 'object'`), TypeScript-only cast fix in request-builder is correct
- **ad-tech-protocol-expert:** approved — no spec-defined backward-compat obligation for these shapes at the client normalizer layer; fail-closed is correct; added comment clarifying `ValidationError` is a client-boundary throw (not wire-forwarded) and documenting the intentional asymmetry vs `get_products.product_ids` (warn+delete) pattern

## Nits noted (not fixed)

[...]

PR #1683, opened 2026-05-11
fix(client): remove account_from_brand fabrication shim in normalizeRequestParams

### LINKED ISSUE #1676: request-normalizer fabricates account from brand on create_media_buy — silently ships invented data as 3.0
## Summary

`src/lib/utils/request-normalizer.ts` (compiled at `dist/lib/utils/request-normalizer.js:99-110`) silently fabricates a 3.0-shaped `account` object using `brand.domain` as a stand-in for `operator` when the caller omits `account`. The wire then looks like a valid AdCP 3.0 `create_media_buy` request — but the `account.operator` value is data the caller never supplied.

```js
// dist/lib/utils/request-normalizer.js:99-110
if (taskType === 'create_media_buy' && !normalized.account && normalized.brand?.domain) {
    warnOnce('account_from_brand',
        'create_media_buy: account is required. Inferring from brand for backward compatibility.');
    normalized.account = {
        brand: normalized.brand,
        operator: normalized.brand.domain,    // fabricated
    };
}
```

## Why this is wrong (not just stale)

1. **No account ≠ a back-compat case.** Missing `account` is the caller's bug. The normalizer should surface it (hard error), not invent a value to make the request shape-valid.
2. **`operator = brand.domain` is fiction.** Per the AdCP 3.0 spec, `account.operator` identifies the seller/operator of the billing account, not the brand's own domain. Conflating them isn't a rename — it's a hallucination.
3. **v2 sunset is past.** v2 unsupported as of 3.0 GA (April 2026), security-only until Aug 1 2026. The original "backward compatibility" framing on this shim no longer applies — there are no pre-3.0 callers we owe support to.
4. **Conformance results aren't what they claim.** When the compliance harness goes through this path, the badge issued certifies "agent handles a 3.0 envelope wrapping a fabricated account" — not "agent handles 3.0 `create_media_buy`." Different claim; the badge promises the second.

## Repro

Comply run against a 3.0-conformant seller (e.g. `https://wonderstruck.sales-agent.scope3.com/mcp`):

```
[@adcp/sdk] DEPRECATED: create_media_buy: account is required.
            Inferring from brand for backward compatibility.
```

Wonderstruck accepts the fabricated account (their schema is permissive on `AccountReference2`), so the test passes — but no real validation of caller-supplied account semantics happened.

## Fix

Replace lines 99-110 with:

```js
if (taskType === 'create_media_buy' && !normalized.account) {
    throw new ADCPValidationError(
        'create_media_buy: account is required. ' +
        'Pass account as { account_id } or { brand, operator, sandbox? }.'
    );
}
```

Then run the comply harness and surface the new failures — the harness is the dominant caller hitting this shim, and the right fix downstream is to give the runner an actual account to thread (via `list_accounts`, seller `seed_account`, or a session-config field, similar to how `governance_agent_url` is threaded today).

## Cross-link

This was discovered while probing Wonderstruck (Scope3 Storylab) for AAO compliance. Full findings + code coordinates in adcontextprotocol/adcp's `.context/wonderstruck-findings.md` on branch `bokelley/sdk-bump-and-wonderstruck-findings`.

bokelley:
Closes #1676

## Summary

`normalizeRequestParams` contained a v2 backward-compat shim that silently set `account.operator = brand.domain` when a `create_media_buy` call omitted `account`. This PR removes it and throws `ValidationError` instead.

**Why the shim was wrong:**
- `account.operator` identifies the billing account's operator (e.g. `pinnacle-media.com`), not the brand's own domain — the fabricated value was semantically correct only for direct-buy topologies and wrong for any agency/DSP-intermediated call.
- Missing `account` is a caller bug, not a back-compat case. The normalizer was masking it.
- The v2 sunset (AdCP 3.0 GA, April 2026) removes the compat rationale entirely.
- Compliance badges issued through this path certified "handles 3.0 envelope wrapping fabricated account" rather than actual `create_media_buy` account semantics.

**Behavior change (two cases):**
1. `create_media_buy` with `brand.domain` but no `account` — previously fabricated `{ brand, operator: brand.domain }`; now throws `ValidationError(field='account')`.
2. `create_media_buy` with neither `account` nor `brand` — previously fell through to seller-side schema rejection; now throws `ValidationError` client-side.

**Note on `executeAndHandle` vs `executeTask`:** `executeTask` (generic method) wraps all errors in a `TaskResult.failed` return value. `executeAndHandle` (used by typed methods like `createMediaBuy`) does not — `ValidationError` propagates as a thrown exception from typed methods, consistent with how `this.validateRequest` already throws from that path. This pre-existing asymmetry is not addressed in this PR; it's documented here as a known limitation. Callers of typed methods should `try/catch` for pre-flight validation errors.

**Known follow-up:** The `schema_validation` storyboard step that deliberately sends `create_media_buy` without `account` (to test seller error handling) needs an `omit_account` escape hatch analogous to `omit_idempotency_key` — otherwise it tests client-side error behavior rather than seller-side. That storyboard path was already broken by the old shim (shim was fabricating account, seller never saw the missing-account case). Follow-up needed in storyboard runner.

## What was tested

- `npm run typecheck` — pre-existing errors unrelated to this diff (`TS2688: @types/node`, `TS5107: moduleResolution=node10 deprecated`); both present on `main` before this change
- Build: `dist/` pre-seeded with compiled changes; full build requires `npm install` (CI will verify)
- Tests: `test/lib/request-normalizer.test.js` added with 7 cases covering throw + no-throw paths; full test suite requires `npm install` (CI will verify)

**Pre-PR review:**
- code-reviewer: approved with nits — flagged `executeAndHandle` no-catch (pre-existing, documented above) and missing tests (tests are at `test/lib/request-normalizer.test.js`)
- ad-tech-protocol-expert: approved — "removal is correct and overdue"; `operator = brand.domain` is precisely wrong per `AccountReference` discriminated union; condition widening is correct (explicitly documented in changeset); qualified `sync_accounts` guidance to implicit-account sellers only in error message

**Nits (not fixed):**
- `ValidationError` `constraint` parameter receives a recovery sentence rather than a pure constraint description — pre-existing pattern issue in the error class, out of scope for this PR.

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01WYQpHoctUsoEBa2wuSpnY4

---
_Generated by [Claude Code](https://claude.ai/code/session_01WYQpHoctUsoEBa2wuSpnY4)_

### DIFF
--- a/.changeset/remove-account-from-brand-shim-1676.md
+++ b/.changeset/remove-account-from-brand-shim-1676.md
@@ -1,0 +1,9 @@
+---
+"@adcp/sdk": minor
+---
+
+Remove deprecated `account_from_brand` shim in `normalizeRequestParams` (#1676).
+
+The shim silently fabricated `account.operator = brand.domain` when a `create_media_buy` call omitted `account`. This was semantically wrong for any caller with a buying-side intermediary, and caused compliance badges to certify requests with invented data rather than caller-supplied account semantics. With AdCP 3.0 GA (April 2026) sunsetting v2 support, the back-compat rationale no longer applies.
+
+`create_media_buy` calls that omit `account` now throw `ValidationError` (exported as `ADCPValidationError`) with field `account` and a message describing the two valid forms: `{ account_id }` or `{ brand, operator, sandbox? }`. Use `list_accounts` to discover an existing `account_id`, or `sync_accounts` to register a natural-key account first.
--- a/.changeset/remove-account-from-brand-shim-1676.md
+++ b/.changeset/remove-account-from-brand-shim-1676.md
@@ -6,4 +6,8 @@ Remove deprecated `account_from_brand` shim in `normalizeRequestParams` (#1676).
 
 The shim silently fabricated `account.operator = brand.domain` when a `create_media_buy` call omitted `account`. This was semantically wrong for any caller with a buying-side intermediary, and caused compliance badges to certify requests with invented data rather than caller-supplied account semantics. With AdCP 3.0 GA (April 2026) sunsetting v2 support, the back-compat rationale no longer applies.
 
-`create_media_buy` calls that omit `account` now throw `ValidationError` (exported as `ADCPValidationError`) with field `account` and a message describing the two valid forms: `{ account_id }` or `{ brand, operator, sandbox? }`. Use `list_accounts` to discover an existing `account_id`, or `sync_accounts` to register a natural-key account first.
+**Behavior change (two cases):**
+1. `create_media_buy` with `brand.domain` but no `account` — previously fabricated `{ brand, operator: brand.domain }` (wrong for non-direct topologies); now throws `ValidationError`.
+2. `create_media_buy` with neither `account` nor `brand` — previously fell through to seller-side schema rejection; now throws `ValidationError` client-side.
+
+`create_media_buy` calls that omit `account` now throw `ValidationError` (exported as `ADCPValidationError`) with field `account`. Use `list_accounts` to discover an existing `account_id`, or `sync_accounts` to register a natural-key account (implicit-account sellers only).
--- a/src/lib/utils/request-normalizer.ts
+++ b/src/lib/utils/request-normalizer.ts
@@ -7,6 +7,7 @@
  */
 
 import { brandManifestToBrandReference, promotedProductsToCatalog } from '../types/compat';
+import { ValidationError } from '../errors/index';
 import { warnOnce } from './deprecation';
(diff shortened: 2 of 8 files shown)


bokelley:
Tracking under #1685 — hat 1 of 5 (account fabrication). **Do not merge before #1684** (the visibility keystone). Pairs with #1681 (the fabrication twin); after #1684 lands, this and #1681 can ship in either order.

bokelley:
Acknowledged — merge dependency on #1684 (visibility keystone) noted. This PR will stay draft until #1684 lands; #1681 and this can ship in either order after that.

---

Issue #1707, opened 2026-05-12
Schema URLs and Zod validators are baked at SDK build time — drift from agent's advertised `adcp_version`


bokelley:
## Summary

The runner cites schema URLs and runs Zod validation against schemas pinned to the **SDK's build-time `ADCP_VERSION`** rather than the agent's advertised version. When the two are skewed (which happens any time an adopter installs an `@adcp/sdk` version older than the agent they're probing), the runner reports schema URLs that don't reflect what's actually on the wire and validates responses against a schema set the agent has already moved past.

> ## 📌 Update 2026-05-12: scope correction after BidMachine investigation
>
> Three findings shift the urgency and shape of this issue:
>
> 1. **The BidMachine symptom (adcp#4419) has a different root cause.** Verified the published @adcp/sdk tarballs 5.25.1, 6.12.0, and current all use `.passthrough()` on `SyncAccountsResponseSchema.accounts[]`. Zod silently accepts `authorization`. The actual rejection is `context.no_secret_echo`'s name-based dragnet flagging the spec-legitimate `authorization` field. Fix landed via adcp-client#1713 / PR #1714.
>
> 2. **Deliverables 1 (`.strict()` → `.passthrough()`) and 2 (codegen regen) are not needed.** The published codegen already uses passthrough on every `additionalProperties: true` schema. fgranata's "Zod `.strict()` codegen lag" diagnosis (per adcp-client#1711) doesn't match what's actually in the tarballs. Removed from scope.
>
> 3. **The agent doesn't advertise a full semver.** `get_adcp_capabilities` exposes only `adcp.major_versions: integer[]` (e.g. `[3]`) — no `adcp_version` patch-level string. So the "cite the agent's version in `schemas_used`" framing (deliverable 7 as originally written) has no agent-side signal to consult at full-semver granularity. The honest cite today is the SDK's build-time pin (current behavior).
>
> What remains real: **adopters who install a stale @adcp/sdk validate against stale Zod schemas regardless of what the agent emits**. That's a real correctness drift, but its blast radius is limited to: (a) typed-shape rejection of newer fields the codegen doesn't know about, and (b) report-time URL drift. Both are bounded by `additionalProperties: true` permissiveness on most response schemas. There's no urgent unblocker; this is architectural cleanup, not an emergency.
>
> **Revised recommendation**: hold #1707 for adopter demand to surface. The two empirical symptoms it would address are:
[...]

PR #1705, opened 2026-05-11
feat(runner): structured RunnerNotice advisory surface on StoryboardResult / ComplianceResult

bokelley:
Closes #1704

## Summary

- Adds `notices: RunnerNotice[]` to `StoryboardResult` (always-present, default `[]`) and `notices?: RunnerNotice[]` to `ComplianceResult` (optional, deduped by `code` across all storyboard runs)
- New exported types: `RunnerNotice`, `NoticeCode` (stable union), `NoticeSeverity` (`'info' | 'deprecation' | 'future_required'`)
- Two spec-grounded day-one notices from `collectCapabilityNotices()`:
  - `request_signing_required_in_4_0` (`future_required`) — on `signed_requests` storyboards (or storyboards with a `request_signing_probe` step) when `request_signing.supported` is absent/false
  - `legacy_hmac_fallback_removed_in_4_0` (`deprecation`) — on webhook-related storyboards (id matches `/webhook/i` or has webhook assertion steps) when `webhook_signing.legacy_hmac_fallback: true`
- All early-return paths (`buildCapabilityUnsupportedResult`, `buildRequirementUnmetResult`, `buildRequiredToolsMissingResult`, `buildDiscoveryFailedResult`, `buildNotApplicableStoryboardResult`, `checkAccountDiscoveryGate`) carry `notices: []`
- CI gates and JUnit consumers can key on stable `code` values instead of parsing prose `skip.detail` strings

Deferred: `signed_requests_specialism_deprecated` — pending upstream deprecation of the `'signed-requests'` specialism value in adcontextprotocol/adcp#4418; the value is still active in the spec.

## Test plan

- [x] 13 new tests in `test/lib/storyboard-notices.test.js` (5 describe blocks, `node:test` runner, `_profile` injection — no live agent or schema cache needed)
- [x] `notices` always-present on StoryboardResult including early-return paths
- [x] `request_signing_required_in_4_0` fires on signed_requests storyboard and request_signing_probe storyboards; does not fire when signing declared or on unrelated storyboards
- [x] `legacy_hmac_fallback_removed_in_4_0` fires on webhook-scoped storyboards only; does not fire on non-webhook storyboards even when flag is set
- [x] Notice codes are unique per storyboard result
- [x] No notices when `raw_capabilities` absent (standalone runner without pre-fetched profile)
- [x] TypeScript: `tsc --project tsconfig.lib.json` clean (no errors)
- [x] Prettier: format check passes

## Pre-PR review

[...]

Issue #1702, opened 2026-05-11
comply(): autodetect `request_signer` and `oauth_metadata` requirements (mirror webhook_receiver autodetect from #1678)


bokelley:
## Summary

#1678 / `17d1c9a` (shipped in 7.0) taught `comply()` to autodetect the `webhook_receiver` requirement from `{{runner.webhook_url:…}}` token presence and grade webhook storyboards `not_applicable` when no receiver is configured. The same pattern is needed for two more requirement types that currently produce ~44 false-negative step failures per run on agents that don't satisfy them.

This is a direct extension of the #1685 stance ("the SDK is a witness, not a translator") to two more requirement surfaces.

## Evidence

A 7.0.0 `comply()` run against `https://wonderstruck.sales-agent.scope3.com/mcp` — a bearer-only sales agent with no runner-side JWKS registration — reports:

```
agent_url: https://wonderstruck.sales-agent.scope3.com/mcp
overall_status: partial
steps_passed:  57
steps_failed:  49   <-- 44 of these belong to the two storyboards below
steps_skipped: 64
storyboards_executed: 24
storyboards_not_applicable: []
storyboards_missing_tools: [11]   <-- #1695 / e93306c split works correctly
```

| Storyboard | Step failures | Root cause |
|---|---|---|
| `signed_requests` | 39 | Runner has no RFC 9421 signer keyed to the agent's JWKS; every positive/negative vector returns `401 request_signature_required`. |
| `security_baseline` | 5 | Probes OAuth 2.1 Protected-Resource Metadata discovery (`/.well-known/oauth-protected-resource`, `/.well-known/oauth-authorization-server`); the agent uses static bearer and has no PRM endpoint. |

Both surfaces are operating-as-designed for this agent class. The runner currently reports them as `fail` instead of `not_applicable`, dragging the track verdict from `pass` to `partial`. This is the same false-negative pattern that #1678 addressed for webhooks.

## Proposed change

Extend `detectImplicitRequires` (or its successor) to recognize two additional requirement names and corresponding implicit declarations.

### `request_signer`

- **Declaration signal:** any storyboard step whose `sample_request` or step config references a request-signing component (e.g., storyboard YAML carrying `signing_vector:`, `sig_input:`, or a `$test_kit.request_signer.*` task placeholder).
[...]

Issue #1685, opened 2026-05-11
Tracking: ship 1676–1680 as one stance — the SDK is a witness, not a translator


bokelley:
Five issues were filed in a row from the same probe (a 3.0-strict seller — Wonderstruck). On their own they read as five spot fixes. They aren't. They're one bug wearing five hats: **at every seam between adopter input and the seller wire, the SDK is being polite instead of honest.**

That politeness was invisible against permissive sellers. Against a 3.0-strict seller it surfaces all at once.

## The five hats

| # | Politeness move | What it hides | PR |
|---|---|---|---|
| #1676 | Invents `account` from `brand.domain` | Adopter never wired account resolution | #1683 |
| #1677 | Accepts pre-3.0 `product_ids[]` / object `budget` and rewrites in place | Adopter never migrated to 3.0 shape | #1681 |
| #1678 | Ships literal `{{runner.webhook_url:…}}` when no receiver is configured | Test config gap that should be `not_applicable` | #1690 |
| #1679 | Flattens wire-level `INVALID_REQUEST` detail to `{}` | The other four become silent — grader sees no diagnostic | #1684 |
| #1680 | Grades unrunnable scenarios as `partial` | Inflates cert scores; "couldn't run" reads as "got close" | #1682 |

## The rule that emerges

> The SDK is a witness, not a translator. Don't fabricate, don't normalize stale shapes, don't placeholder when prereqs are unmet, don't flatten the wire diagnostic, don't upgrade a skip to a partial pass.

Fail closed at every seam. The same principle already governs capability declarations (capabilities must match actual behavior — see the precedent set in #931 where a test-mode flag that changed runtime behavior was required to flip the capability block). These five extend that principle one layer up: requests and grades, not just the capabilities body.

## Merge order matters

**#1684 (for #1679) ships first.** It's the keystone — purely visibility. While `step.error` arrives as `{}`, the other four can regress silently because every failure looks like `INVALID_REQUEST: ¯\\_(ツ)_/¯`. Land #1684 and every future politeness regression becomes a tripwire. Land the others first and the next one sneaks back in unannounced.

Suggested order:

1. **#1684** (closes #1679) — visibility keystone. Unblocks everything.
2. **#1683** (closes #1676) and **#1681** (closes #1677) — fabrication twins. Can ship in either order after #1684.
3. **#1682** (closes #1680) — `not_applicable` routing.
[...]

PR #1714, opened 2026-05-12
fix(invariant): no_secret_echo only fails on string-valued suspect-named fields (#1713)

### LINKED ISSUE #1713: no_secret_echo invariant flags spec-legitimate field names on structured-value fields
## Summary

The default `context.no_secret_echo` invariant flags any response field whose **name** is in a hardcoded `SUSPECT_PROPERTY_NAMES` set (`'authorization'`, `'api_key'`, `'apikey'`, `'bearer'`, `'x-api-key'`) — regardless of the field's value. This catches structured object/array fields whose names happen to overlap with credential-shaped names but whose values aren't credentials.

Empirical impact: every BidMachine `sync_accounts` failure in [adcp#4419](https://github.com/adcontextprotocol/adcp/issues/4419) / [adcp-client#1711](https://github.com/adcontextprotocol/adcp-client/issues/1711) surfaces as `context.no_secret_echo`. The diagnostic framing in those reports points at Zod codegen lag — but I verified the published @adcp/sdk 5.x / 6.x tarballs use `.passthrough()` (not `.strict()`) on `SyncAccountsResponseSchema.accounts[]`, so Zod silently accepts the `authorization` field. The actual rejection is the invariant's NAME dragnet at `src/lib/testing/storyboard/default-invariants.ts:184-190`:

```ts
if (v !== null && typeof v === 'object') {
  for (const [key, inner] of Object.entries(v as Record<string, unknown>)) {
    if (SUSPECT_PROPERTY_NAMES.has(key.toLowerCase())) {
      return `contains suspect property name "${key}"`;
    }
    stack.push(inner);
  }
}
```

The dragnet fires on the NAME alone. If a seller returns `accounts: [{ authorization: { type: 'oauth', token_endpoint: '...' } }]` — a perfectly reasonable structured authorization config — the invariant fails with "contains suspect property name 'authorization'".

## Why `authorization` specifically matters now

The spec uses `authorization` as a real field name in legitimate contexts:

- `compliance/cache/3.0.11/property/validation-result.json` — `authorization` is the structured object containing authorization-validation results (not a credential).
- Sellers extending `sync_accounts` per `additionalProperties: true` may carry their own structured `authorization` field describing per-account auth posture.

Both are spec-compliant and should not fail compliance grading.

## Why the existing value-based detection is sufficient for the actual leak surface

The invariant already does value-based detection in the same function:

- Lines 173-178: scans every string value for the `BEARER_TOKEN_PATTERN` regex (`\bbearer\s+[A-Za-z0-9._~+/=-]{10,}`) and for verbatim caller-supplied secret literals.
- Bearer-prefixed tokens, API keys, and caller secrets all trip via the VALUE check.

The NAME dragnet adds coverage only for:

- Bare tokens (`"eyJabc..."`) on suspect-named fields where the value isn't bearer-prefixed and isn't a known caller secret.

That's a real signal — but it should only fire when the value is a STRING. Object/array values bearing the same name are structured config, not credential echoes.

## Proposed fix

Tighten `findSecretEcho` so the `SUSPECT_PROPERTY_NAMES` check only triggers when the field value is a string:

```ts
if (v !== null && typeof v === 'object') {
  for (const [key, inner] of Object.entries(v as Record<string, unknown>)) {
    if (SUSPECT_PROPERTY_NAMES.has(key.toLowerCase()) && typeof inner === 'string' && inner.length > 0) {
      return `contains suspect property name "${key}" with string value`;
    }
    stack.push(inner);
  }
}
```

Same behavior for the actual leak case (bare token on a credential-named string field) and the existing bearer-pattern / caller-secret checks. Permissive for spec-legitimate structured fields bearing a credential-shaped name.

## Acceptance criteria

- BidMachine's `sync_accounts` response with `authorization` carrying a structured object value passes `context.no_secret_echo`.
- A response carrying `authorization: "eyJabc..."` (string value, looks like a token) still fails.
- A response carrying `authorization: "Bearer abc..."` (bearer-prefixed string anywhere) still fails via the existing pattern.
- A response carrying a caller-supplied API key as a verbatim string value still fails via the existing secret-set match.

## Tests

- New `test/lib/no-secret-echo-name-dragnet.test.js` covering all four acceptance criteria above.
- Existing `test/lib/default-invariants-no-secret-echo.test.js` continues to pass (regression guard for the value-based detection).

## Related work

- adcp-client#1709 (merged via #1712) — Zod-reject error attribution. That fix addresses a different misattribution path (when Zod actually rejects). This issue addresses the case where Zod ACCEPTS but the invariant over-rejects on field name.
- adcp-client#1707 — dynamic schema fetch + strict→passthrough + codegen regen. Real architectural cleanup but **does not** unblock BidMachine — verified the published SDK already uses `.passthrough()`.
- adcp-client#1711 (fgranata) — BidMachine's bug report. Their diagnosis of "Zod `.strict()` codegen lag" is incorrect for the published SDK; the actual culprit is this invariant. Will close once this lands and fgranata retests.
- adcp-client#1685 — "the SDK is a witness, not a translator" stance. The dragnet fabricates a credential-leak signal against a structured non-credential field — same anti-pattern.

## Refs

- `src/lib/testing/storyboard/default-invariants.ts:93-194` — the invariant implementation
- `src/lib/testing/storyboard/default-invariants.ts:102` — `SUSPECT_PROPERTY_NAMES`
- `src/lib/testing/storyboard/default-invariants.ts:184-190` — the name dragnet (the over-rejection)
- `src/lib/types/schemas.generated.ts:5407` — `SyncAccountsSuccessSchema.accounts[]` (uses `.passthrough()`; confirms Zod isn't the rejection point)

bokelley:
Closes #1713.

## The actual BidMachine root cause

Investigating #1707/#1711 surfaced something unexpected: BidMachine's `sync_accounts` `context.no_secret_echo` failures **aren't** caused by Zod `.strict()` codegen lag. I verified published @adcp/sdk tarballs at 5.25.1, 6.12.0, and current main — every one uses `.passthrough()` on `SyncAccountsResponseSchema.accounts[]`. Zod silently accepts the `authorization` field.

The rejection happens inside `context.no_secret_echo`'s name dragnet at `src/lib/testing/storyboard/default-invariants.ts:184-190`:

```ts
if (v !== null && typeof v === 'object') {
  for (const [key, inner] of Object.entries(v as Record<string, unknown>)) {
    if (SUSPECT_PROPERTY_NAMES.has(key.toLowerCase())) {
      return `contains suspect property name "${key}"`;
    }
    stack.push(inner);
  }
}
```

The dragnet trips on field NAME alone (`authorization`, `api_key`, `apikey`, `bearer`, `x-api-key`) regardless of value. BidMachine's response carries `authorization` with a structured object value (per the spec — `compliance/cache/3.0.11/property/validation-result.json` declares `authorization` as a structured authorization-validation payload, and the `additionalProperties: true` on `sync_accounts_response.accounts[]` permits seller extensions). The structured value isn't a credential leak; the dragnet false-positives.

## The fix

`findSecretEcho` now requires both a suspect name AND a non-empty string value:

```ts
if (
  SUSPECT_PROPERTY_NAMES.has(key.toLowerCase()) &&
  typeof inner === 'string' &&
  inner.length > 0
) {
  return `contains suspect property name "${key}" with string value`;
}
stack.push(inner);
```

Object/array values on suspect-named fields pass through to the recursive walk, which still scans nested strings via `BEARER_TOKEN_PATTERN` and caller-supplied secret literals. The actual leak shapes (bearer tokens at any depth, caller secret echoes, bare credential strings on suspect-named fields) all stay caught.

## Tests

Five new cases in `describe('default-invariants: context.no_secret_echo (widened whole-body scan)')`:

| # | Case | Result |
|---|---|---|
| 1 | Suspect name (`authorization`) carrying an OBJECT value — the BidMachine case | passes ✓ |
| 2 | Suspect name (`api_key`) carrying an ARRAY value | passes ✓ |
| 3 | Suspect name carrying an empty string | passes ✓ |
| 4 | Suspect name carrying a non-empty string — existing dragnet preserved | fails ✓ |
| 5 | Bearer literal nested INSIDE a structured suspect-named object | fails (regression guard) ✓ |

113/113 invariant tests pass.

## Why this matters for the coordinated stance

- adcp-client#1709 (merged via #1712) addresses Zod-reject misattribution (different path — when Zod DOES reject, attribute to `response_schema`).
- adcp-client#1707 — dynamic schema fetch + strict→passthrough + codegen regen. Real architectural cleanup but **does not unblock BidMachine** (verified the published SDK already uses passthrough).
- adcp-client#1711 (fgranata) — BidMachine's report. Their "Zod `.strict()` codegen lag" diagnosis was incorrect for the published SDK; **this PR is the actual unblocker** for the 27-point delta (63/128 comply vs 45/59 CLI). Will close once fgranata retests.

So the order BidMachine cares about:

1. **#1714 (this PR)** — closes BidMachine's immediate symptom.
2. **#1707** — independent architectural cleanup; lower urgency now.

## Test plan

- [x] `npm run build` clean
- [x] `node --test test/lib/storyboard-default-invariants.test.js` — 113/113 pass (5 new + all existing)
- [ ] After merge, BidMachine retest should show `sync_accounts` failures resolved.

Coordinated stance: #1685 (the SDK is a witness, not a translator). Same anti-pattern as #1702 / #1705 / #1712: the SDK fabricated a credential-leak signal against a structured non-credential field the spec legally permits.

### DIFF
--- a/.changeset/no-secret-echo-structured-value-relaxation-1713.md
+++ b/.changeset/no-secret-echo-structured-value-relaxation-1713.md
@@ -1,0 +1,56 @@
+---
+'@adcp/sdk': patch
+---
+
+fix(invariant): no_secret_echo only fails on string-valued suspect-named fields (#1713)
+
+The default `context.no_secret_echo` invariant flagged any response field
+whose name was in `SUSPECT_PROPERTY_NAMES` (`'authorization'`, `'api_key'`,
+`'apikey'`, `'bearer'`, `'x-api-key'`) regardless of the field's value.
+This over-rejected spec-legitimate structured fields — notably the
+`authorization` object on `validation-result.json` (a structured
+authorization-validation payload, not a credential) and any seller-side
+extension fields named `authorization` under
+`sync_accounts_response.accounts[]` (`additionalProperties: true`).
+
+Symptom: BidMachine's `sync_accounts` failures in
+adcontextprotocol/adcp#4419 all surfaced as `context.no_secret_echo`.
+Diagnosis in #4419 pointed at Zod `.strict()` codegen lag, but
+verification of the published SDK tarballs (5.25.1 / 6.12.0 / current)
+shows `SyncAccountsResponseSchema.accounts[]` already uses
+`.passthrough()`. Zod silently accepts the `authorization` field; the
+NAME dragnet in this invariant is the actual rejection.
+
+Fix: narrow `findSecretEcho` so the suspect-name check only fires when
+the field VALUE is a non-empty string. Structured object/array values
+on suspect-named fields pass through to the recursive walk, which still
+scans nested strings against `BEARER_TOKEN_PATTERN` and caller-supplied
+secret literals. The actual leak shapes the invariant was designed to
+catch (bearer tokens, API keys, caller secret echoes) remain caught.
+
+Adds five new test cases covering:
+
+- Suspect name with object value passes (the BidMachine case)
+- Suspect name with array value passes
+- Suspect name with empty string passes (no leak)
+- Suspect name with non-empty string value still fails (existing dragnet)
+- Bearer literal nested inside a structured suspect-named object still
+  fails via the value-scan regression guard
+
+Sequencing relative to other in-flight work:
+
+- adcp-client#1709 (PR #1712, merged) — Zod-reject error attribution.
+  Addresses a different misattribution path (when Zod DOES reject).
+  This fix addresses the case where Zod ACCEPTS but the invariant
+  over-rejects on field name.
+- adcp-client#1707 — dynamic schema fetch + strict→passthrough +
+  codegen regen. Real architectural cleanup; does NOT unblock
+  BidMachine (the published SDK already uses passthrough).
+- adcp-client#1711 (fgranata) — BidMachine's report. The "Zod
+  `.strict()` codegen lag" diagnosis was incorrect for the published
+  SDK; this fix is the actual unblocker. Closes once fgranata retests.
+
+Coordinated stance: adcp-client#1685 (the SDK is a witness, not a
+translator). Same anti-pattern: the SDK fabricated a credential-leak
+signal against a structured non-credential field that the spec legally
+permits.
--- a/src/lib/testing/storyboard/default-invariants.ts
+++ b/src/lib/testing/storyboard/default-invariants.ts
@@ -183,8 +183,27 @@ function findSecretEcho(value: unknown, secrets: Set<string>): string | null {
     }
     if (v !== null && typeof v === 'object') {
       for (const [key, inner] of Object.entries(v as Record<string, unknown>)) {
-        if (SUSPECT_PROPERTY_NAMES.has(key.toLowerCase())) {
-          return `contains suspect property name "${key}"`;
+        // Name-based dragnet. The SUSPECT_PROPERTY_NAMES set catches the
+        // "bare token on a credential-named field" leak shape that the
+        // BEARER_TOKEN_PATTERN value-scan misses (a JWT without `Bearer `
+        // prefix won't match the regex). But the dragnet over-rejects
+        // when the field's VALUE is a structured object or array — those
+        // are spec-legitimate config payloads, not credential echoes.
+        // `compliance/cache/{version}/property/validation-result.json`
+        // declares an `authorization` object field carrying authorization-
+        // validation metadata, not a token; sellers extending
+        // `sync_accounts` via the schema's `additionalProperties: true`
+        // may carry their own structured `authorization` posture. Both
+        // are spec-conformant.
+        //
+        // Gate the name-based fail on `typeof inner === 'string'` and
+        // non-empty so structured values pass through to the recursive
+        // walk below. The recursion still scans nested string values, so
+        // a `Bearer xyz...` literal embedded in a structured object is
+        // still caught by the regex at the top of this loop.
+        // Spec: adcp-client#1713 / adcp#4419.
+        if (SUSPECT_PROPERTY_NAMES.has(key.toLowerCase()) && typeof inner === 'string' && inner.length > 0) {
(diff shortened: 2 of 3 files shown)


fgranata:
Following up on your guinea-pig invitation. Ran our prod against `@adcp/sdk@7.1.0` (the version with #1714's `findSecretEcho` fix) using `npx -y -p '@adcp/sdk@7.1.0' adcp storyboard run bm-prod ...` since `@adcp/client@latest` is still on the 5.25.1 compat shim.

**Empirically confirms your analysis.** Two probes against `https://adcp.bidmachine.io/adcp/mcp`:

| Storyboard | Result |
|---|---|
| `media_buy_seller` (+ 10 sub-storyboards) | 47 pass / 3 fail / 9 skip |
| `sales_non_guaranteed` | 26 pass / 3 fail / 8 skip |

`account_setup` is **cleared** — not in the failure list with `@adcp/sdk@7.1.0`. Score should jump once your grading infra picks up 7.1.0. Will you re-run automatically, or should we wait?

**Three remaining failures we captured locally** (each independent of #1714):

1. `media_buy_seller/delivery_reporting/simulate_delivery` — our `comply_test_controller` returns `UNKNOWN_SCENARIO` (we hadn't wired this scenario). Shipping the dispatch now.
2. `media_buy_seller/inventory_list_targeting/get_products_brief` — our sandbox product emits `channels: ["mobile_app"]` which isn't in the canonical `MediaChannel` enum. One-line fix to `["display"]`.
3. `sales_non_guaranteed/create_media_buy` — our handler reads `bid_price` only from `packages[0]`; multi-package case silently drops `packages[1]`. Refactoring handler + mapper.

Shipping all three this afternoon. When grading retests against 7.1.0, would love a per-storyboard failure-list dump so we can diff against this local capture. Thanks again — the #1714 diagnostic trail made triangulation easy.

PR #1748, opened 2026-05-14
feat: guard example adapter domains

### LINKED ISSUE #1747: feat(examples): fail-fast startup guard against `.example` TLD constants in adapter examples
## Problem

Every `hello_*_adapter_*.ts` worked example has module-level constants like `KNOWN_PUBLISHERS = ['acmeoutdoor.example', 'pinnacle-agency.example', ...]` that adopters must flip when forking. The FORK CHECKLIST header is the only reminder. Adopters who fork and ship without flipping put `.example` TLD domains into production tenant lookups.

## Proposal

Each `hello_*_adapter_*.ts` calls a small helper at module-load (or in the bootstrap that starts the server):

```ts
import { assertNoExampleTlds } from '@adcp/sdk/server';

assertNoExampleTlds({ KNOWN_PUBLISHERS }, {
  allowIn: ['test', 'development'],
  // NODE_ENV outside allowIn → throw with a pointer to the FORK CHECKLIST line
});
```

Behavior:
- In `NODE_ENV=test` / `development`: no-op (example fixtures use `.example` deliberately)
- Otherwise: scan the provided record's string/string-array values for `.example` TLDs and throw `Error('Adapter forked without flipping <key>; see FORK CHECKLIST in examples/hello_seller_adapter_guaranteed.ts')` if found
- Fail-closed at startup, not at first request

## Why this and not typed config

Discussed in #1382. The typed `defineAdapterConfig` shape would catch the same class via tsc but requires per-specialism config types + rewriting all six examples to read from `config.*` instead of module-level constants. The runtime guard catches the production-leak case (forked-and-shipped-without-changes) at startup with ~5 LOC per example and zero new vocabulary.

## Acceptance

- `assertNoExampleTlds(constants, opts)` exported from `@adcp/sdk/server`
- Called in each of the six `hello_*_adapter_*.ts` files near the FORK CHECKLIST
- Gated on NODE_ENV allowlist per [feedback_node_env_allowlist] (use `{test, development}`, not `=== 'production'`)
- Unit test that throws when `.example` present and passes when adopter has flipped the constant
- Changeset (patch)

## Refs

- #1382 — original typed-config proposal (closed in favor of this)
- `examples/hello_seller_adapter_guaranteed.ts:103` — canonical `KNOWN_PUBLISHERS` site

fizyxbt:
## Summary
- add `assertNoExampleTlds` to `@adcp/sdk/server` for startup checks outside test/development
- wire the guard into the six checklist-backed adapter examples
- add regression coverage and a patch changeset

Fixes #1747.

## Checks
- `npm run build:lib`
- `node --test test/lib/example-tld-guard.test.js`
- `npm run typecheck:examples`
- `NODE_ENV=test node --test-timeout=180000 --test-force-exit --test-concurrency=1 --test test/examples/hello-seller-adapter-guaranteed.test.js test/examples/hello-seller-adapter-non-guaranteed.test.js test/examples/hello-seller-adapter-social.test.js test/examples/hello-creative-adapter-template.test.js test/examples/hello-creative-adapter-ad-server.test.js test/examples/hello-si-adapter-brand.test.js`
- `npm run test:lib`

Pre-push also ran `npm run format:check`, `npm run typecheck` and `npm run build:lib`.

### DIFF
--- a/.changeset/example-tld-guard.md
+++ b/.changeset/example-tld-guard.md
@@ -1,0 +1,5 @@
+---
+'@adcp/sdk': patch
+---
+
+Add a server-side guard that fails fast when forked adapter examples still contain `.example` tenant domains outside test or development.
--- a/examples/hello_creative_adapter_ad_server.ts
+++ b/examples/hello_creative_adapter_ad_server.ts
@@ -60,6 +60,7 @@ import {
   memoryBackend,
   serve,
   verifyApiKey,
+  assertNoExampleTlds,
   type AccountStore,
   type Account,
   type CreativeAdServerPlatform,
@@ -86,6 +87,10 @@ const ADCP_AUTH_TOKEN = process.env['ADCP_AUTH_TOKEN'] ?? 'sk_harness_do_not_use
 const PUBLIC_AGENT_URL = process.env['PUBLIC_AGENT_URL'] ?? `http://127.0.0.1:${PORT}`;
 
 const KNOWN_PUBLISHERS = ['creative-network.example', 'acmeoutdoor.example', 'pinnacle-agency.example'];
+assertNoExampleTlds(
+  { KNOWN_PUBLISHERS },
+  { allowIn: ['test', 'development'], checklistPath: 'examples/hello_creative_adapter_ad_server.ts' }
+);
 const SANDBOX_ID_PREFIX = 'sandbox_';
 
 // ---------------------------------------------------------------------------
--- a/examples/hello_creative_adapter_template.ts
+++ b/examples/hello_creative_adapter_template.ts
@@ -38,6 +38,7 @@ import {
   memoryBackend,
   AdcpError,
   defineCreativeBuilderPlatform,
+  assertNoExampleTlds,
   type DecisioningPlatform,
   type CreativeBuilderPlatform,
   type BuildCreativeReturn,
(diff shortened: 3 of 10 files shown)


fizyxbt:
I have read the IPR Policy

bokelley (approved):
Approving. Clean, minimal implementation that satisfies #1747 — guard ships in `@adcp/sdk/server`, fires at module-load (fail-fast, not first-request), and the regex correctly distinguishes `.example` placeholders from legit `.example.com` / `.example.org` domains via the trailing terminator lookahead.

The `env !== undefined && allowIn.includes(env)` shape gives the right fail-closed default for unset `NODE_ENV` (raw Lambda / custom containers / some K8s setups) — same allowlist pattern we used for `idempotency: 'disabled'` in #931, which is what this guard is meant to mirror.

A couple of small observations, none blocking:

1. **Scope of constants per example is uneven.** `hello_creative_adapter_template.ts` and `hello_si_adapter_brand.ts` pass `DEFAULT_LISTING_WORKSPACE` / `DEFAULT_LISTING_BRAND` (defaults like `'ws_acme_studio'`, no `.example` substring) — harmless but won't ever fire. Meanwhile `hello_seller_adapter_guaranteed.ts` only passes `KNOWN_PUBLISHERS` and skips `ADCP_AUTH_TOKEN` / `PUBLIC_AGENT_URL`, which can also be overridden to `.example` placeholders by an adopter. Worth a follow-up to standardize on "pass every module-level constant that an adopter is expected to flip" — but I'd rather not block this PR on it.

2. **Running examples bare locally now throws.** `node examples/hello_seller_adapter_guaranteed.ts` without `NODE_ENV` set will hit the throw. That's the intended fail-closed behavior; a one-liner in the FORK CHECKLIST header (`# run with NODE_ENV=development to skip the .example guard`) would prevent confusion for first-time readers. Optional follow-up.

3. **`PUBLIC_AGENT_URL` in the template example** — nice catch including it, since an adopter swapping `127.0.0.1` for `https://creative.example` would currently get caught here even though the issue only called out `KNOWN_PUBLISHERS`. Worth applying the same treatment in the seller adapters in a future PR.

LGTM as-is.

PR #1753, opened 2026-05-14
feat(testing): per-tool seeded callbacks on `TestControllerBridge` for platform-proxy sellers (closes #1002)

### LINKED ISSUE #1002: comply_test_controller: read-path interception for platform-proxy sellers
## Context

We're a seller running 13 platform adapters (Snap, Meta, TikTok, Pinterest, LinkedIn, Google, Amazon, Reddit, Spotify, Criteo, CitrusAd, Flashtalking, UniversalAds) against the framework via [`createAdcpServer`](https://github.com/adcontextprotocol/adcp-client/blob/main/src/lib/server/create-adcp-server.ts) — every adapter is a thin wrapper that proxies AdCP tool calls to the upstream platform's API (`snapClient.getCreatives(...)`, `metaClient.getMediaBuy(...)`, etc.). State of record lives upstream; the seller doesn't persist domain objects in its own data layer.

We tried to wire AdCP storyboards in-process via `runAgainstLocalAgent` (PR [scope3data/agentic-adapters#123](https://github.com/scope3data/agentic-adapters/pull/123)). That surfaced a structural mismatch in the `comply_test_controller` design.

## Problem

The controller pattern, as documented in [`comply-controller.d.ts`](https://github.com/adcontextprotocol/adcp-client/blob/main/src/lib/testing/comply-controller.d.ts), assumes the seller's read path consults the same data layer the controller writes to:

> "Seed adapters persist the fixture to the seller's data layer. … the helper builds the `previous_state` / `current_state` envelope from its own idempotency cache."

For **stateful-native sellers** (the seller stores media buys / creatives / accounts in its own DB and serves them on read), this composes cleanly:
1. Storyboard fires `seed_creative` → controller writes to `stateStore`
2. Storyboard fires `sync_creatives` → handler reads/writes `stateStore`, returns canned data

For **platform-proxy sellers** (which is most production ad-tech: DSPs, retail-media, walled gardens, programmatic guaranteed marketplaces), the read path doesn't touch `stateStore`:
1. Storyboard fires `seed_creative` → controller writes to `stateStore`
2. Storyboard fires `list_creatives` → handler calls `snapClient.getCreatives(...)` → real Snap API → returns real (or 401) data
3. Seeded `stateStore` is never read

Net: seeded fixtures are dead writes, and storyboards either need real upstream OAuth or every adapter needs per-platform upstream-client stubs (~150-300 LOC per platform × 13 platforms = unsustainable).

## Proposed surface

Two paths, in order of preference:

### Option A — framework-level read-path interception (preferred)

A new opt-in flag on `AdcpServerConfig` that, when a controller is registered, makes every framework-registered read handler consult `stateStore` first and fall back to the registered domain handler:

```ts
createAdcpServer({
  ...,
  stateStore: sharedStateStore,
  testController: createComplyController({ ... }),
  // NEW:
  testControllerReadInterception: true,  // default: false, only meaningful when testController is registered
});
```

Semantics:
- `list_creatives`, `get_creative`, `get_media_buy`, `list_accounts`, `get_account_financials`, etc. — read `stateStore` first; if the lookup hits, return the seeded fixture; if it misses, call through to the registered domain handler.
- Mutating handlers (`create_media_buy`, `sync_creatives`, etc.) stay unchanged — they call the registered handler, and the controller's `force_*` calls update `stateStore` after the fact for the next read.
- Gated behind a top-level config flag so production behavior is unchanged.

This unlocks the controller pattern for every proxy-shaped seller without per-adapter rework.

### Option B — explicit docs + a recommended adapter-stubbing pattern

If interception is too much surface area: add a section to `comply-controller.d.ts` (or a top-level guide) for platform-proxy sellers explaining that controller adoption requires per-adapter upstream-client stubs (the equivalent of `vi.mock('snapClient')` in unit tests, but inside the seller's runtime when sandbox mode is active). Could ship a `createSandboxClientStub<T>(real: T, fixtures: Fixtures): T` helper that automates the "intercept calls, serve seeded data, fall back to real" pattern at the *adapter's upstream client* boundary.

Option B is smaller but moves the cost onto every seller. Option A pays the SDK once, every consumer benefits.

## Why this matters beyond us

Walled-garden ad tech is structurally proxy-shaped — sellers don't get to be authoritative for the platform's domain objects. Snap, Meta, TikTok, Pinterest, LinkedIn, Reddit, Spotify, Amazon, Criteo, CitrusAd, FlashTalking, UniversalAds, and most retail-media networks all fit this pattern. Without this gap closed, storyboard-driven compliance verification only works for the (rarer) stateful-native sellers — primarily DSP-side and pure marketplace sellers.

## What we did instead

For now: the in-process storyboard test in our codebase is gated as best-effort across all 3 platforms we tried (snap, meta, tiktok), and the load-bearing artifact for compliance verification is a sandbox-runner CLI that points at deployed sandbox URLs with real OAuth (slow, credential-dependent, not CI-friendly). Tracked internally as [scope3data/agentic-adapters#122](https://github.com/scope3data/agentic-adapters/issues/122).

## Acceptance criteria (Option A)

- [ ] `AdcpServerConfig` accepts a `testControllerReadInterception` flag (or equivalent)
- [ ] When set, framework-registered read handlers consult `stateStore` before delegating
- [ ] Default behavior (flag unset) is unchanged for every existing consumer
- [ ] `runAgainstLocalAgent` works end-to-end against a platform-proxy adapter when the seller wires `comply_test_controller` + the interception flag
- [ ] Test in `test/server/test-controller-bridge.test.js` (or similar) covering the proxy-seller path

cc Anthropic SDK team — happy to pair on the design if that's faster.

bokelley:
Closes #1002.

## Context

Platform-proxy sellers (DSPs, walled gardens, retail-media networks) run adapters that proxy AdCP tool calls to upstream platforms — state of record lives upstream, not in the seller's data layer. Today `TestControllerBridge` only exposes `getSeededProducts`, so storyboard seeds against every other read tool (`seed_creative`, `seed_media_buy`, etc.) are dead writes; without per-platform upstream-client stubs (~150–300 LOC × 13 platforms in Scope3's case), conformance runs either need real upstream OAuth or skip those steps.

This PR extends the same bridge pattern across the read-side tool surface so a platform-proxy seller wires one `TestControllerBridge` (typically via `bridgeFromSessionStore`) and the framework feeds seeded fixtures into the read path on sandbox requests — without bypassing adapter code (that was the rejected Option A from the triage on #1002; this is Option B).

cc Scope3 (per #1002).

## Summary

Five new opt-in callbacks on `TestControllerBridge<TAccount>`, all mirroring `getSeededProducts` exactly (post-handler merge, opt-in by presence, gated on sandbox + resolved-account + controller-present, warn-and-drop validation):

| Callback                          | Tool                      | Semantics                                                                     | Dedup key                                              |
| --------------------------------- | ------------------------- | ----------------------------------------------------------------------------- | ------------------------------------------------------ |
| `getSeededCreatives`              | `list_creatives`          | append; seeded wins on collision                                              | `creative_id`                                          |
| `getSeededMediaBuys`              | `get_media_buys`          | append; seeded wins on collision                                              | `media_buy_id`                                         |
| `getSeededAccounts`               | `list_accounts`           | append; seeded wins on collision                                              | `account_id`                                           |
| `getSeededAccountFinancials`      | `get_account_financials`  | **singleton replace** — pick fixture matching request's `account.account_id`  | `account.account_id`                                   |
| `getSeededCreativeFormats`        | `list_creative_formats`   | append; seeded wins on collision                                              | canonical `format_id.agent_url\|format_id.id`          |

`bridgeFromSessionStore` gets matching `selectSeeded*` selectors so adopters wiring storyboards via the session-store pattern get all bridges in one helper. Per-tool callbacks are omitted from the returned bridge when no selector is provided.

Production traffic is unchanged: bridges run only on sandbox-flagged requests (`account.sandbox === true` or `context.sandbox === true`) against a registered controller, with the additional resolved-account `sandbox: true` cross-check when `resolveAccount` is wired.

### Notes on shape

- `get_account_financials` is the singleton exception — it returns one account's envelope, so "merge" reduces to "replace if the request's `account.account_id` matches a seeded fixture". The bridge interface and validation pattern still mirror the others.
- `list_creative_formats` composes with `SalesPlatform.listCreativeFormats` / `CreativeBuilderPlatform.listCreativeFormats` — the adapter handler runs first, then this bridge supplements. Not a replacement for those handler-side hooks; storyboards use the bridge to inject test-only formats.
- Type derivation: seeded entry types are derived via lookup (`ListCreativesResponse['creatives'][number]`, etc.) so they stay in lockstep with the generated wire schema. The named entity types `Account` and `Format` are imported directly. `SeededCreative`, `SeededMediaBuy`, `SeededAccountFinancials` are exported from `@adcp/sdk/server` and `@adcp/sdk/testing`.

## Test plan

- [x] `test/lib/seed-per-tool-wiring.test.js` — 27 new tests covering, per callback: seeded-only path, handler-only path, merge path (handler entries first, seeded append, seeded wins on collision), sandbox-gating refusal, validation drop.
- [x] `test/lib/test-controller-bridge.test.js` and `test/lib/seed-get-products-wiring.test.js` still pass (no regression in `getSeededProducts` precedent).
- [x] `bridgeFromSessionStore` per-tool selector wiring — 6 tests covering each new selector and the "no selector → no callback" omission.
- [x] Full test suite: 8638 tests pass / 0 fail / 3 skipped.
- [x] `npm run build` clean.

## Migration

None — additive. Existing `testController: { getSeededProducts }` adopters are unaffected.

### DIFF
--- a/.changeset/test-controller-bridge-per-tool.md
+++ b/.changeset/test-controller-bridge-per-tool.md
@@ -1,0 +1,19 @@
+---
+'@adcp/sdk': minor
+---
+
+feat(testing): per-tool seeded callbacks on `TestControllerBridge` for platform-proxy sellers (closes #1002)
+
+`TestControllerBridge` previously exposed only `getSeededProducts`, leaving platform-proxy sellers (DSPs, walled gardens, retail-media networks) without a way to feed seeded fixtures into the read path of every other read tool — storyboard seeds against `seed_creative` / `seed_media_buy` were dead writes when the adapter proxied to upstream APIs.
+
+Extends `TestControllerBridge<TAccount>` with five opt-in callbacks mirroring `getSeededProducts` (post-handler merge, sandbox + resolved-account + controller-present gating, warn-and-drop validation, never throws):
+
+- `getSeededCreatives(ctx)` → merged into `list_creatives` (dedup by `creative_id`)
+- `getSeededMediaBuys(ctx)` → merged into `get_media_buys` (dedup by `media_buy_id`)
+- `getSeededAccounts(ctx)` → merged into `list_accounts` (dedup by `account_id`)
+- `getSeededAccountFinancials(ctx)` → replaces `get_account_financials` envelope when seeded `account.account_id` matches the request (singleton response, replace semantics)
+- `getSeededCreativeFormats(ctx)` → merged into `list_creative_formats` (dedup by canonical `format_id.agent_url|format_id.id`)
+
+`BridgeFromSessionStoreOptions` gains matching `selectSeeded*` selectors so adopters wiring storyboards via the session-store pattern get all bridges in one helper. Per-tool callbacks are omitted from the returned bridge when no selector is provided.
+
+Production traffic is unchanged: bridges run only on sandbox-flagged requests against a registered controller.
--- a/docs/guides/VALIDATE-YOUR-AGENT.md
+++ b/docs/guides/VALIDATE-YOUR-AGENT.md
@@ -235,6 +235,41 @@ Omit adapters you don't support — they auto-return `UNKNOWN_SCENARIO`. Throw `
 
 For domain state that carries internal structure (packages, revision, history) read by production tools, use `registerTestController(server, store)` — flat store surface, session-scoped factory. See `examples/seller-test-controller.ts`. Pick by state shape, not by helper tier — both sit on the same primitives and both auto-emit the capability block.
 
+### Platform-proxy sellers (state-of-record lives upstream)
+
+Sellers whose read path proxies to upstream platforms (`snapClient.getCreatives(...)`, `metaClient.getMediaBuy(...)`) don't read seeded fixtures from their own data layer — controller-seeded state is a dead write. Wire `TestControllerBridge` to feed those fixtures back into the read path on sandbox requests, without stubbing 13 upstream clients per adapter.
+
+```typescript
+import { createAdcpServer, bridgeFromSessionStore } from '@adcp/sdk/server';
+
+const server = createAdcpServer({
+  // ... usual config + per-adapter handlers ...
+  testController: bridgeFromSessionStore({
+    loadSession: input => sessionStore.load(sessionKey(input)),
+
+    // Each selector is opt-in by presence — wire the ones whose storyboards
+    // your specialism gates on. Returned entries are validated (warn-and-drop
+    // on shape errors, never throw) and merged into the matching tool's
+    // response on sandbox requests only.
+    selectSeededProducts:          s => s.seededProducts,          // get_products
+    selectSeededCreatives:         s => s.seededCreatives,         // list_creatives
+    selectSeededMediaBuys:         s => s.seededMediaBuys,         // get_media_buys
+    selectSeededAccounts:          s => s.seededAccounts,          // list_accounts
+    selectSeededAccountFinancials: s => s.seededFinancials,        // get_account_financials
+    selectSeededCreativeFormats:   s => s.seededFormats,           // list_creative_formats
+  }),
+});
+```
+
+Bridge contract:
+
+- **Triply gated.** Bridge runs only when the bridge is registered, the request carries a sandbox marker (`account.sandbox === true` or `context.sandbox === true`), and — if `resolveAccount` produced a record — that record is `sandbox: true` too. Production traffic untouched.
+- **Post-handler merge.** The adapter's real handler runs first (so a broken `snapClient.getCreatives()` still fails the conformance gate — the bridge supplements, it does not replace adapter behavior). Seeded entries append; on id collision the seeded fixture wins.
+- **Singleton exception.** `get_account_financials` returns one account's envelope, so the bridge picks the seeded fixture whose `account.account_id` matches the request's `account.account_id` and REPLACES the handler envelope for that account. Other accounts pass through unchanged.
+- **Validation drops invalid fixtures.** Entries missing the dedup id (`creative_id`, `media_buy_id`, `account_id`, `account.account_id`, `format_id.{agent_url,id}`) are warn-and-dropped, not thrown — a broken test fixture shouldn't tank the request under test.
+
+Plain hand-rolled `TestControllerBridge` (no session store) is also supported — wire `getSeededProducts` / `getSeededCreatives` / etc. directly. See `src/lib/server/test-controller-bridge.ts` for the interface.
+
 ---
 
 ## Schema-driven validation (catch field drift at dev time)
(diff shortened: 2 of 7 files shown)


bokelley:
Applied four review-driven fixes in `f540d7c1` (commit on the same branch — no new PR):

### Fix 1 — brand+operator `AccountReference` resolution for `get_account_financials`
`pickSeededAccountFinancialsForRequest` previously read `account.account_id` off the raw request, which is `undefined` for the brand+operator variants of `AccountReference`. Storyboards seeding financials against a brand+operator-referenced sandbox silently no-op'd. The helper now accepts an optional `resolvedAccountId` (preferred when set, falls back to the request field for single-tenant adopters who don't wire `resolveAccount`). Dispatcher passes `ctx.account.account_id` through.

### Fix 2 — `query_summary` / `pagination.total_count` drift on array merges
After verifying the bundled schemas in `schemas/cache/3.0.11/`: only `list_creatives` carries a `query_summary` block (with `total_matching` + `returned`); `get_media_buys` and `list_accounts` use `pagination.total_count` (optional). Now:
- `returned` (creatives) = final array length
- `total_matching` (creatives) and `pagination.total_count` (all three) += count of NEW seeded entries (entries that did NOT collide on id)
- Full id-collision yields **zero** drift (handler's count passes through unchanged)
- Handlers that omit `pagination.total_count` still pass through untouched

### Fix 3 — preserve handler `context` / `ext` on singleton replace
`replaceAccountFinancialsIfSeeded` now spreads the seeded fixture and re-stamps `handler.context` / `handler.ext` on top. The seeded fixture is authoritative on the financials body (spend / period / account / currency); the framework owns the envelope echo (`adcp_version` / `request_id`). JSDoc on `getSeededAccountFinancials` and the helper updated.

### Fix 4 — warn-and-drop duplicate seeded `account.account_id`
`filterValidSeededAccountFinancials` now logs `logger.warn(...)` and drops second/Nth occurrences of the same `account.account_id`. First occurrence wins (consistent with array helpers' dedupe pattern). Includes the offending `account_id` in the warning meta.

### Test plan delta
6 new test suites in `test/lib/seed-per-tool-wiring.test.js` (11 new test cases total):
- `get_account_financials — brand+operator account reference resolution` (2): seeded fixture matches via resolved id; no-match passes handler through
- `get_account_financials — handler context/ext preserved on singleton replace` (1): asserts `context` / `ext` survive the replace
- `get_account_financials — duplicate seeded account_ids` (1): asserts first wins, second logs warning
- `list_creatives — query_summary count updates on merge` (2): new-entry path (50→53), full-collision path (10→10 / no drift)
- `get_media_buys — pagination.total_count updates on merge` (2): new-entry path (50→53), full-collision path (10→10)

Full suite: 8670/8670 passing (was 8638 before; +32 from new + the cross-suite test-count drift the runner already exhibits between rebuilds). `npm run format:check` and `tsc --noEmit` clean.

### Note on surprises
The user's fix-2 description mentioned `query_summary` on `get_media_buys` / `list_accounts` — the AdCP 3.0.11 schemas only have `query_summary` on `list_creatives`. The other two surface counts via `pagination.total_count` (optional). I treated both shapes consistently so drift can't surface on either, regardless of which field the handler populates.

Issue #1755, opened 2026-05-14
feat(testing): next-phase `TestControllerBridge` seeded callbacks for proxy sellers


bokelley:
## Context

Follow-up to #1002 / PR #1753 (which shipped 5 per-tool `getSeeded*` callbacks on `TestControllerBridge`).

During expert review of #1753, both ad-tech-protocol-expert and adtech-product-expert independently called out the same priority list of next-phase bridges that platform-proxy sellers (Scope3-style, 13 platform adapters) will need to fully unblock storyboard-driven conformance.

Filing as one umbrella so the next batch is scoped and prioritized together, mirroring the per-tool bridge pattern already established in PR #1753.

## Phase 1 (highest priority — file as #1755-or-similar)

- [ ] **`getSeededMediaBuyDelivery`** — for `get_media_buy_delivery`. Called out as the highest-value proxy-seller read after `get_media_buys` themselves. Storyboards that gate on delivery pacing can't seed without this. **Both reviewers independently flagged this as #1 priority.**

## Phase 2 (governance / retail-media adopters)

- [ ] **`getSeededAuthorizedProperties`** — for `list_authorized_properties`. Proxy sellers often advertise upstream's property catalog; seeding lets storyboards assert publisher inventory composition.
- [ ] **`getSeededPropertyLists`** — for `list_property_lists` / `get_property_list` (governance-aware-seller specialism).
- [ ] **`getSeededContentStandards`** — for `list_content_standards` / `get_content_standards` (governance specialisms).
- [ ] **`getSeededCollectionLists`** — for `list_collection_lists` / `get_collection_list` (program-level brand-safety via IMDb/Gracenote/EIDR IDs).

## Phase 3 (signals + walled-garden completeness)

- [ ] **`getSeededSignals`** — for `get_signals` (signals-marketplace and signals-owned specialisms).
- [ ] **`getSeededAudiences`** — for `list_audiences`. Upstream-proxied on most walled gardens.
- [ ] **`getSeededTargetingCategories`** — for `list_targeting_categories`. Same walled-garden proxy pattern.
- [ ] **`getSeededCreativeDelivery`** / **`getSeededCreativeFeatures`** — for `get_creative_delivery` / `get_creative_features` (creative-template / generative specialisms).

## Phase 4 (brand-rights + SI — different seam)

- [ ] **`getSeededBrandIdentity`** / **`getSeededRights`** — for the brand-rights specialism.
[...]

PR #1759, opened 2026-05-14
feat(testing): getSeededMediaBuyDelivery bridge callback + nit tests (#1755 phase 1)

bokelley:
## Summary

Adds the sixth opt-in callback on `TestControllerBridge<TAccount>` —
`getSeededMediaBuyDelivery(ctx)` — so platform-proxy sellers can seed
delivery-snapshot fixtures into conformance storyboards without driving real
measurement through the upstream adapter. Also adds the two nit tests called
out on PR #1753 review.

Closes phase 1 of #1755. Nit tests cover the precedence + count-drift seams
on the bridges that landed in #1753.

### `getSeededMediaBuyDelivery` callback (new)

- `TestControllerBridge<TAccount>.getSeededMediaBuyDelivery?(ctx)` callback,
  mirroring the validate-and-drop + post-handler-merge pattern of the other
  five bridges.
- `filterValidSeededMediaBuyDeliveries` — drops entries missing
  `media_buy_id` (warn, never throw).
- `mergeSeededMediaBuyDeliveryIntoResponse(handlerResponse, seededEntries)`
  — appends non-colliding seeded entries (**handler wins on collision** —
  measurement stays authoritative on the handler; deviation from the
  seeded-wins precedent on the other array bridges), then calls
  `recomputeAggregatedTotals`.
- `recomputeAggregatedTotals(deliveries, handlerAggregated)` — pure helper
  that recomputes the response's required `aggregated_totals` block from
  the merged per-delivery `totals`. Policy:

| Field group | Behavior |
|---|---|
| `impressions`, `spend`, `media_buy_count` (required) | Always recomputed |
| `clicks`, `completed_views`, `views`, `conversions`, `conversion_value` (optional sums) | Recomputed only when every merged delivery populates the field; partial population falls back to handler's value |
| `roas`, `completion_rate`, `cost_per_acquisition` (derived ratios) | Recomputed only when both input fields recomputed AND divisor is non-zero; otherwise fall back to handler's value or omit |
| `reach`, `reach_unit`, `frequency`, `new_to_brand_rate` (pass-through) | Handler's value preserved verbatim — not derivable from per-delivery `totals` |

- `BridgeFromSessionStoreOptions.selectSeededMediaBuyDelivery` selector
  wired through `bridgeFromSessionStore`.
- Triply-gated dispatch in `create-adcp-server.ts` (sandbox + resolved-
  account + controller-present), parallel to the existing five dispatches.

### Nit tests on #1753 bridges

- `getSeededAccountFinancials` — explicit precedence test: when both
  `ctx.account.account_id` and the request's `account.account_id` are
  present but DIFFERENT, the bridge keys on the resolved id. Documents the
  contract that makes seeded fixtures interchangeable across
  `AccountReference` variants.
- `list_creatives` — explicit mixed-collision test: when handler returns
  `[shared, h-only]` and bridge seeds `[shared, s-new]`, the merged
  response has `creatives.length === 3`, `query_summary.returned === 3`,
  and `query_summary.total_matching === 51` (handler's 50 + newCount=1,
  NOT + seededCount=2).

### Docs

- `docs/guides/VALIDATE-YOUR-AGENT.md` — adds `selectSeededMediaBuyDelivery`
  to the session-store table with the specialism each callback unblocks,
  documents the `aggregated_totals` re-aggregation policy, and notes the
  resolved-id-wins precedence on `get_account_financials`.

## Test plan

- [x] `NODE_ENV=test node --test test/lib/seed-per-tool-wiring.test.js` — 51/51 pass (+16 new)
- [x] `npm test` — 8713 pass, 0 fail
- [x] `npm run typecheck` — no new errors (pre-existing dep-missing errors only)
- [x] Pre-push hook (typecheck + build) — passes


bokelley:
## Correction: collision semantic flipped to seeded-wins

The initial commit shipped `getSeededMediaBuyDelivery` with handler-wins on `media_buy_id` collision, reasoning that delivery data is measurement-authoritative on the handler side. On review, the existing bridges (`mergeSeededCreatives` / `mergeSeededMediaBuys` / `mergeSeededAccounts`) all use seeded-wins — adopters learn one mental model: storyboards seed deliberately, so a seeded fixture for an existing ID is an explicit author override.

Mixing semantics across bridges would be poor DX. Flipped to match the precedent. Tests updated; `recomputeAggregatedTotals` still produces wire-correct totals across the final merged set.

PR #1764, opened 2026-05-14
fix(client): route synthetic-v2 sellers through v2 adapter instead of refusing

nastassiafulconis:
## Summary

- `requireSupportedMajor` previously threw `VersionUnsupportedError(reason: 'synthetic')` when capabilities were synthesized from `tools/list` and the inferred version was `'v2'`. That blocked buyers from calling sellers that simply hadn't implemented `get_adcp_capabilities`. The SDK now routes those requests through the v2 wire-shape adapter and emits a one-time per-client warning via `maybeWarnSyntheticV2` so adopters can audit the routing decision; idempotency-TTL guarantees remain unknown for these sellers and BYOK retry callers should treat them as such. `SingleAgentClient.isSyntheticV2()` exposes the routing decision as a predicate so retry frameworks can branch before dispatch.
- Synthetic-v3 behavior, real declared-v2 sellers, and real-v3-missing-idempotency-TTL sellers are unchanged. The `'synthetic'` member of `VersionUnsupportedReason` is retained for back-compat with consumers that pattern-match the union (now tagged `@deprecated`), but no SDK call site throws it anymore.
- Tests in `synthetic-v3-fallback.test.js` and `v3-guard.test.js` flipped from rejecting synthetic-v2 to accepting it (plus a new one-shot warning assertion). Patch changeset included. `version.ts` and `package-lock.json` are catch-up sync to `package.json` 7.3.0 via `sync-version`.
- Buyer guidance: `skills/call-adcp-agent/SKILL.md` symptom table now points readers at `client.isSyntheticV2()` for retry tightening when they see the routing warning.

Spec footing: filed adcontextprotocol/adcp#4578 to confirm the "absence of `get_adcp_capabilities` => infer v2" inference at the spec level.

Replaces #1763 (auto-closed by branch rename).

## Test plan

- [ ] `npm test` covers the updated synthetic-v2 / synthetic-v3 / v3-guard suites


bokelley:
Direction is defensible; a few things I'd push back on before merge. The "compliant v3 seller would declare itself via `get_adcp_capabilities`, so absence = v2" inference is reasonable, and replacing a `throw` with a typed warning + `isSyntheticV2()` for retry frameworks is the right shape — the previous fail-closed was actively hurting the most common case (buyers calling sellers that haven't shipped v3 discovery yet).

**Concerns**

1. **`requireV3()` semantics regress, not just `requireSupportedMajor`.** The test diff at `test/lib/v3-guard.test.js` flips `requireV3('sync_creatives')` from rejecting synthetic-v2 to accepting it. That's a bigger change than the PR summary advertises. A caller of `requireV3()` is explicitly saying "I need v3" — letting a synthetic-v2 seller through with a warning erodes the function's contract. I'd keep `requireV3()` strict (still throw `reason: 'synthetic'`) and only loosen `requireSupportedMajor` / `requireV3ForMutations`. Two functions, two contracts.

2. **Dead `'synthetic'` reason is a trap.** The type union keeps it, the error message was genericized, and no SDK call site emits it. Downstream code that switches on `reason === 'synthetic'` will look correct and never fire. Either remove it (major bump, cleaner), or — if you keep it for the reason in the changeset — pair it with a `@deprecated` JSDoc tag so TS surfaces it.

3. **`package-lock.json` has an inverted bump.** In the `eslint-plugin` workspace: `\"@adcp/sdk\": \"^7.3.0\"` → `\"^7.1.0\"`. Looks like `sync-version` regenerated it the wrong way; should be the opposite direction.

4. **Spec footing.** Changing fail-closed to fail-open on a discovery-version inference is the kind of SDK opinion that wants a spec issue first confirming "absence of `get_adcp_capabilities` = v2." If `adcontextprotocol/adcp` is silent on this, the SDK is baking opinion into ecosystem behavior. Link a spec issue in the PR description.

5. **Buyer-facing docs not updated.** `skills/call-adcp-agent/SKILL.md` and the retry guidance should mention `isSyntheticV2()` as the gate for "TTL unknown, tighten retries." The PR references the pattern but ships no doc to point buyers at it.

6. **Naming nit.** `isSyntheticV2()` couples two things (synthesized + routed v2). For a retry-policy consumer the more honest predicate is "idempotency-TTL guarantee is unknown" — which is true for both synthetic-v2 *and* synthetic-v3. Today synthetic-v3 gets a free pass on retry-safety because the SDK assumes "v3 spec mandates TTL." Worth at least documenting that synthetic-v3 also has unknown TTL despite returning `false` here.

**Verdict:** approve once #1, #3, and #4 are addressed. #2 and #5 are nice-to-haves but not blocking. #6 is a future-naming concern worth a comment.

bokelley:
Looked again. Two of my concerns above are wrong — retracting them.

**Concern #1 (`requireV3()` semantics regress) — retracted.** `requireV3()` on this branch is a deprecated alias that just calls `requireSupportedMajor()` (`SingleAgentClient.ts:3223`). There's no separate contract to preserve; the `v3-guard.test.js` flip is testing the same underlying function via the legacy entry point. The test rename in that file is fine, though the file/describe titles could probably move off "requireV3" to reduce confusion for readers.

**Concern #3 (inverted package-lock bump) — retracted.** Checked main: `packages/eslint-plugin/package.json` declares `"@adcp/sdk": "^7.1.0"`, but the lockfile entry was stale at `^7.3.0`. This PR's `sync-version` regen brings the lockfile back into agreement with the package.json. It's a corrective sync, not an inversion. (Separate question worth asking: should the eslint-plugin's declared range be bumped to `^7.3.0` now? But that's a follow-up, not this PR's job.)

**Still standing:**
- #2 (dead `'synthetic'` reason — `@deprecated` tag would help)
- #4 (spec issue link for the "absence = v2" inference)
- #5 (buyer-facing docs / `call-adcp-agent` SKILL mention of `isSyntheticV2()`)
- #6 (synthetic-v3 also has unknown TTL but returns `false` — worth documenting)

Sorry for the noise.

package-lock.json
```diff
@@ -6329,7 +6329,7 @@
       "version": "0.0.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@adcp/sdk": "^7.3.0",
+        "@adcp/sdk": "^7.1.0",
```
EmmaLouise2018, line 6332: 🔴 **Looks like an accidental downgrade.** The CLI workspace above correctly bumps `@adcp/sdk` to `^7.3.0`, but here `packages/eslint-plugin` goes the other way — `^7.3.0` → `^7.1.0`. Probably a `sync-version` artifact. Re-run `npm install` (or hand-fix) so both workspaces land on `^7.3.0`; otherwise the eslint-plugin workspace will resolve against an older SDK than the CLI.

src/lib/core/SingleAgentClient.ts
```diff
@@ -246,15 +246,22 @@ export interface SingleAgentClientConfig extends ConversationConfig {
...
-   * Throws `VersionUnsupportedError` before the request is sent. Bypass
-   * with `allowV2` or — process-wide as a fallback — `ADCP_ALLOW_V2=1`.
+   * Sellers whose capabilities are synthesized from `tools/list` (no
+   * authoritative `get_adcp_capabilities` response) route through the
+   * v2 adapter with a one-time warning — a compliant v3 seller would
+   * declare itself, so absence of a declaration is read as v2. Adopters
+   * who need a hard "definitely-v3" gate should validate
+   * `(await client.getCapabilities())._synthetic === false` directly.
+   *
+   * Throws `VersionUnsupportedError` before the request is sent when
```
EmmaLouise2018, line 262: 🟡 **Contract narrowing — worth flagging in the changeset.** Pre-PR, `requireV3ForMutations: true` was a *hard* fail-loud gate: synthetic capabilities threw `VersionUnsupportedError`. Post-PR, the same flag silently dispatches through v2 with a one-shot `console.warn`. Adopters who set this flag specifically to refuse unverifiable sellers now have to call `isSyntheticV2()` manually before every mutation to recover the old behavior — and a `console.warn` fired once per client lifetime is easy to miss in production logs.

Two questions:
1. Is `minor` the right semver bump here? The flag's safety contract narrowed.
2. Worth offering a `'strict'` mode (`requireV3ForMutations: 'strict'`) that preserves throw-on-synthetic for callers who want the old behavior without having to bolt their own predicate gate in front of every mutating call?

src/lib/core/SingleAgentClient.ts
```diff
@@ -3008,6 +3038,29 @@ export class SingleAgentClient {
...
+   *
+   * Caveat for synthetic v3: the predicate returns `false`, but TTL is
+   * still unknown for those sellers — `getIdempotencyReplayTtlSeconds()`
+   * throws until the agent's capabilities endpoint is fixed (issue
+   * #1217). Retry-policy consumers that need a complete "TTL unknown"
+   * gate should additionally check `getCapabilities()._synthetic`.
+   */
+  async isSyntheticV2(): Promise<boolean> {
+    const capabilities = await this.getCapabilities();
+    return capabilities._synthetic === true && capabilities.version !== 'v3';
```
EmmaLouise2018: Minor: `version !== 'v3'` is permissive — if `version` is `undefined` or some unknown string for any synthetic path, this still returns `true`. Probably fine in practice since the synthetic builder is the only source, but `version === 'v2'` would be more explicit and would prevent future drift if someone adds a `'v4'` or unknown-version synthetic path later.

src/lib/errors/index.ts
```diff
@@ -497,7 +504,7 @@ export class VersionUnsupportedError extends ADCPError {
       case 'idempotency':
         return `seller reports v3 but omits adcp.idempotency.replay_ttl_seconds (required by spec).`;
       case 'synthetic':
-        return `capabilities were synthesized from a tool list — v3 claim is unverifiable.`;
+        return `seller's capabilities could not be verified against the required major version.`;
```
EmmaLouise2018: Nit: since the SDK no longer emits `'synthetic'`, this `case` branch only runs if a downstream consumer constructs the error themselves. The original message (`capabilities were synthesized from a tool list — v3 claim is unverifiable.`) was strictly more informative than the new generic one. Worth keeping the original wording for those consumers — they're the only ones who'll see it.

EmmaLouise2018 (commented):
Overall: solid fix with good spec footing (#4578) and the one-shot warning cadence matches existing `maybeWarnV2Sunset` / `maybeWarnSyntheticV3`. Two things worth attention — one is a real bug in the lockfile, one is a semver/contract question worth a thread. Inline comments below.

**🟢 Strengths**
- Spec footing filed (adcontextprotocol/adcp#4578) before baking SDK opinion in — exactly the triage discipline `CLAUDE.md` calls for.
- `'synthetic'` retained in the union as `@deprecated` rather than removed — preserves pattern-match compat for downstream consumers.
- `isSyntheticV2()` test matrix covers all four quadrants (declared v2/v3, synth v2/v3) cleanly.
- SKILL.md symptom-table entry is a nice DX touch — pulls buyers straight to the predicate.

package-lock.json
nastassiafulconis: brian raised this too and then retracted after checking main: packages/eslint-plugin/package.json declares "@adcp/sdk": "^7.1.0" but the lockfile entry was stale at "^7.3.0". sync-version brings the lockfile back into agreement with the package.json, so it's a corrective sync rather than a downgrade. that said, the eslint-plugin's declared range probably should bump to ^7.3.0, i'll open a follow-up so it doesn't get lost.

src/lib/core/SingleAgentClient.ts
nastassiafulconis: on the semver bump: changeset is already `minor` (set when the isSyntheticV2() method landed in 53663644). agree patch would've been too quiet, adopters who pattern-matched `reason: 'synthetic'` in a catch block lose a typed signal, even if undocumented as load-bearing. minor reflects the additive surface (new method, new warning) plus the contract narrowing.

on `'strict'` mode: i'd push back. adding a config knob to preserve behavior we deliberately moved away from feels like over-engineering, the recovery path is two lines:

```ts
if (await client.isSyntheticV2()) {
  throw new Error('refusing dispatch to unverified seller');
}
```

callers who want hard refusal write that gate explicitly; callers who want the new permissive routing get it by default. one knob, two contracts feels worse than one method, two consumer-side patterns. happy to revisit if you've seen adopters in the wild stitching this on every mutation.

EmmaLouise2018 (approved):

PR #1769, opened 2026-05-15
feat(registry): add scope=public option to lookupOperator / lookupPublisher

EmmaLouise2018:
## Summary

Adds an `opts?: { scope?: 'public' | 'all' }` parameter to `RegistryClient.lookupOperator` and `RegistryClient.lookupPublisher`. When `scope: 'public'` is passed, the client appends `?scope=public` to the request so the AAO server forces the anonymous-equivalent view regardless of the caller's API tier.

## Why

The current AAO operator endpoint includes `members_only` agents whenever the caller's key resolves to an org with API-tier access. That's correct for buyer/seller flows authenticating with their own AAO membership — but breaks for platform integrations holding an admin-scoped API key.

## What changed

- \`lookupOperator(domain, opts?)\` — appends \`&scope=public\` when \`opts.scope === 'public'\`.
- \`lookupPublisher(domain, opts?)\` — same shape for symmetry. Today the publisher endpoint doesn't vary by visibility tier, so the option is a no-op there but reserved for forward compatibility.
- Both default to \`scope=all\` (omitted on the wire) when no opts are passed → fully backwards compatible.

## Server side

Companion PR on adcontextprotocol/adcp adds the server-side enforcement of \`?scope=public\` on \`/registry/operator\`. This SDK PR is safe to merge before the server PR — older AAO servers will ignore the unknown query param and continue to return the current behavior.

## Test plan

- [x] \`npm run build\` — clean, dist regenerated
- [ ] New unit test asserting \`?scope=public\` is appended only when opts.scope === 'public' (follow-up — keeping this PR scoped to the API addition)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/.changeset/registry-scope-public.md
+++ b/.changeset/registry-scope-public.md
@@ -1,0 +1,9 @@
+---
+'@adcp/sdk': patch
+---
+
+feat(registry): add `scope=public` option to `lookupOperator` / `lookupPublisher` (#1769)
+
+`RegistryClient.lookupOperator(domain, opts?)` and `RegistryClient.lookupPublisher(domain, opts?)` now accept an optional `{ scope?: 'public' | 'all' }` argument. When `scope: 'public'` is passed, the client appends `?scope=public` so the AAO server returns the anonymous-equivalent view regardless of the caller's API tier — useful for platform integrations using an admin-scoped key that want to render a public picker.
+
+Default behavior (no opts, or `scope: 'all'`) is unchanged: the server honors the caller's tier and includes `members_only` agents when authorized. The publisher endpoint does not vary by tier today; the option is forwarded for symmetry and forward compatibility.
--- a/src/lib/registry/index.ts
+++ b/src/lib/registry/index.ts
@@ -389,16 +389,43 @@ export class RegistryClient {
 
   // ====== Operator & Publisher Lookups ======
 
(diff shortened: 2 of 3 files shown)


bokelley:
Force-pushed cleanup on your branch — happy to merge once green:

**Fixed:**
1. Rebased on `main` (which is already at 7.3.0, so the manual `version.ts`/`package.json` edits are now no-ops and dropped).
2. Reset `package-lock.json` to match `main` (the workspace-dep regression artifact is gone).
3. Added `.changeset/registry-scope-public.md` (patch — additive API).
4. Added 5 unit tests in `test/lib/registry.test.js`:
   - `lookupOperator` appends `scope=public` only when `opts.scope === 'public'`
   - `lookupOperator` omits the param when `scope: 'all'` or no opts
   - `lookupPublisher` same coverage
5. Prettier'd the new code (one long-signature reflow).

**Untouched:** your `RegistryClient` change itself. The `URLSearchParams` migration + scope-public semantics + doc comments are exactly what you wrote.

Two pre-existing lint errors in `task-map.test.ts` and `compliance/runtime.ts` are on `main` already — not from this PR.

bokelley (approved):
LGTM. Cleanup pushed in 0a99528... and ede65538 — rebase + lockfile reset + changeset + 5 unit tests. Test & Build green, changeset check green. Emma's RegistryClient logic is untouched and exactly right.

Issue #1756, opened 2026-05-14
Codegen handling of allOf+$ref produces broken union types (blocks adcp#4510 schema dedup)


bokelley:
## Background

`adcp-client`'s type generation uses `json-schema-to-typescript` (per `scripts/generate-types.ts`). Spec repo adcp#4510 proposes a hygiene refactor that extracts duplicated fields from the 6 `authorized_agents[*]` variants in `adagents.json` into a shared base schema, then `allOf`'s the base into each variant.

Python codegen (`datamodel-codegen`, used by `adcp-client-python`) handles this beautifully — produces a real `AuthorizedAgents1(AuthorizedAgentBase)` inheritance hierarchy with field-level duplication eliminated.

`json-schema-to-typescript` does not. The spike (adcp#4510 comment) found that the same refactored schema produces an **incorrect union** in TS:

```typescript
authorized_agents: [
  ( AuthorizedAgentBaseFields | { url; authorized_for; encryption_keys?; last_updated? } ),
  ...
]
```

That's "either just the base fields, or a duplicate of base fields plus the variant-specific ones" — semantically wrong (should be an intersection / interface-extends) and worse for consumers than the current shape.

This blocks adcp#4510 from landing today. The spec refactor would degrade TS SDK types until this is resolved.

## Why this is the `allOf + $ref + siblings` problem

JSON Schema draft-07 historically had ambiguous semantics for `$ref` adjacent to other keywords. Most validators ignored the siblings; some merged them; codegen tools made their own calls. `json-schema-to-typescript` builds variant types by traversing variants and synthesizing per-variant shapes — when it sees `allOf: [{$ref}, {required: [...], properties: {...}}]` at the variant root, it treats them as alternative shapes rather than the intersection JSON Schema actually requires.

This is the same family of failure described in the `product.publisher_properties` codegen-gap note in adcp#4537: `allOf: [{$ref: selector}, {not: {required: [publisher_domains]}}]` is correct JSON Schema, but several tools drop the `not.required` constraint silently. Different symptom, same root cause.

## What success looks like

Either of these unblocks adcp#4510:

1. **Codegen produces a real intersection / extends.** E.g., for the refactored variant:
   ```typescript
   export interface AuthorizedAgentVariantPropertyIds extends AuthorizedAgentBaseFields {
     authorization_type: "property_ids";
     property_ids: PropertyID[];
     // ...variant-specific only
   }
   ```
[...]

PR #1777, opened 2026-05-16
fix(codegen): bridge JSON Schema constraints across TS->Zod hop

### LINKED ISSUE #1745: Zod codegen drops JSON Schema numeric constraints (minimum/maximum/pattern)
## Summary

Surfaced during #1736 fix: the Zod schemas in \`src/lib/types/schemas.generated.ts\` are generated from JSON Schema but **do not carry numeric/string constraints** like \`minimum\`, \`maximum\`, \`pattern\`, \`minLength\`, \`maxLength\`. They're dropped during codegen.

## Why this matters

The SDK has two validation paths:

1. **Zod path** — \`validateResponseSchema\` in \`src/lib/testing/client.ts\` (uses generated Zod schemas)
2. **AJV path** — \`validateResponse\` via the storyboard runner (parses raw JSON Schema)

Only the AJV path actually enforces \`minimum: 1\` on \`revision\`. Consumers using the Zod path (typed SDK usage, IDE inference, runtime validation outside the storyboard runner) silently accept non-conformant responses.

## Concrete repro

A response with \`revision: 0\`:
- AJV path → emits a constraint violation (keyword: \`minimum\`)
- Zod path → no error; \`MediaBuy['revision']\` parses as a generic \`number\`

Same for any field with \`minimum\`, \`maximum\`, \`pattern\`, \`minLength\`, \`maxLength\`, \`format\`. The Zod schemas use the loose \`z.number()\` / \`z.string()\` shape instead of the constrained \`z.number().min(1)\` / \`z.string().regex(...)\` shape.

## Acceptance

- Codegen pipeline (likely \`scripts/generate-schemas.ts\` or wherever \`schemas.generated.ts\` is produced) translates JSON Schema \`minimum\`/\`maximum\`/\`pattern\`/\`minLength\`/\`maxLength\` into Zod \`.min()\`/\`.max()\`/\`.regex()\` chains
- Regression test pins at least one constrained field (e.g. \`MediaBuy.revision\` rejects \`0\`)
- AJV and Zod paths produce equivalent violation reports for the same input

## Refs

- Surfaced in #1736 / PR #1744 (storyboard evaluator constraint vs missing fix)
- Codegen lives in: TBD (search \`schemas.generated.ts\` build target)

bokelley:
## Summary

- Generated Zod schemas now enforce `minimum`, `maximum`, `minLength`, `maxLength`, `pattern`, and `format` from the upstream JSON Schemas. Before, those keywords were silently lost during the `JSON Schema → TypeScript → Zod` codegen because TypeScript cannot carry a numeric range or regex on a `number`/`string` field. `MediaBuySchema.revision` was emitted as `z.number().optional()` instead of `z.number().min(1).optional()`, so typed validators accepted inputs that AJV correctly rejected.
- New helper `injectJsdocConstraints` (in `scripts/schema-utils.ts`) walks each JSON Schema and encodes the constraints into the `description` field as `@minimum` / `@maximum` / `@minLength` / `@maxLength` / `@pattern` / `@format` JSDoc tags before `json-schema-to-typescript` compiles. `ts-to-zod` natively reads those tags, so the constraints round-trip into `.min()` / `.max()` / `.regex()` / `.iso.datetime()` / `.email()` chains. ~900 new validator chains land across `schemas.generated.ts`.
- `exclusiveMinimum` / `exclusiveMaximum`, patterns containing newlines, and unsupported `format` values (e.g. `iri-reference`) continue to be enforced at runtime by AJV against the unstripped schema.

Closes #1745. This is the bigger of two paired codegen fixes; the sibling PR for #1756 (allOf+$ref merger) also touches `scripts/generate-types.ts` but is independent — merge order doesn't matter.

## Test plan

- [x] New unit test (`test/jsdoc-constraints-codegen.test.js`) exercises `injectJsdocConstraints` on a synthetic schema covering all six tag kinds plus a nested object, then runs the real `json-schema-to-typescript` + `ts-to-zod` pipeline on the result and asserts the emitted TS and Zod.
- [x] New pinning test confirms `MediaBuySchema.revision` rejects `0`, `MediaBuySchema.total_budget` rejects `-1`, `BrandReferenceSchema.domain` rejects an invalid domain, and `BusinessEntitySchema.address.country` rejects lowercase.
- [x] `npm run generate-types && npm run generate-zod-schemas` regenerates cleanly; large additive diff in `core.generated.ts` / `tools.generated.ts` / `schemas.generated.ts` (~1900 lines of new JSDoc + Zod chains).
- [x] `npm run build` passes.
- [x] `npm test` — full suite passes (the previously-flagged flaky perf test `InMemoryReplayStore: has() stays sub-linear` is unrelated and passes on a clean re-run).
- [x] `npm run format:check` passes.

### DIFF
--- a/.changeset/zod-codegen-numeric-string-constraints.md
+++ b/.changeset/zod-codegen-numeric-string-constraints.md
@@ -1,0 +1,15 @@
+---
+'@adcp/sdk': minor
+---
+
+fix(codegen): bridge JSON Schema validation constraints across the lossy TS → Zod hop
+
+Generated Zod schemas now enforce the `minimum`, `maximum`, `minLength`, `maxLength`, `pattern`, and `format` keywords from the upstream JSON Schemas. Before this change, those keywords were silently lost during the `JSON Schema → TypeScript (json-schema-to-typescript) → Zod (ts-to-zod)` codegen because TypeScript can't carry a numeric range or a regex on a `number` / `string` field. The emitted Zod for `MediaBuy.revision` was `z.number().optional()` instead of `z.number().min(1).optional()`, so typed validators accepted constraint-violating values that Ajv (which loads the raw JSON Schema) correctly rejected.
+
+The codegen now pre-processes each schema before it reaches `json-schema-to-typescript`, encoding each property's constraints into the JSDoc-bound `description` field as `@minimum` / `@maximum` / `@minLength` / `@maxLength` / `@pattern` / `@format` tags. `ts-to-zod` natively reads those JSDoc tags, so the constraints round-trip into `.min()` / `.max()` / `.regex()` / `.iso.datetime()` / `.email()` / etc. on the emitted Zod schemas. About 900 new validator chains land across `schemas.generated.ts` covering currency codes, ISO country codes, ISBN/IBAN/BIC patterns, RFC-conformant URLs, ISO-8601 timestamps, and numeric ranges (`revision >= 1`, `priority >= 1`, all monetary `total_budget`/`rate` >= 0, etc.).
+
+Behavioral impact for adopters: typed SDK validators (`MediaBuySchema.parse(...)`, etc.) now reject inputs they previously silently accepted. This matches what real Ajv-driven server-side validation already did — the SDK no longer hands a typed buyer payload the wire would refuse. Adopters who relied on the looser typed validators to construct fixtures or intermediate stages will need to populate the constraint-valid values.
+
+`exclusiveMinimum` / `exclusiveMaximum` are not injected (ts-to-zod has no exclusive variant); they continue to be enforced at runtime by Ajv against the unstripped schema. Pattern values containing newlines or unsupported `format` keywords (e.g. `iri-reference`) are likewise skipped — same Ajv fallback.
+
+Fixes #1745.
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,9 @@ schemas/cache/
 compliance/cache/
 .adcp-sync-*/
 
+# Test harness scratch dirs (test/jsdoc-constraints-codegen.test.js)
+.tmp-jsdoc-constraints-*/
+
 # Environment variables
 .env
 .env.local
--- a/scripts/generate-types.ts
+++ b/scripts/generate-types.ts
@@ -3,7 +3,7 @@
 import { writeFileSync, mkdirSync, readFileSync, existsSync, readdirSync, statSync } from 'fs';
 import { compile } from 'json-schema-to-typescript';
 import path from 'path';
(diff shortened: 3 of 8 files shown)


bokelley:
Triage review pass — one fix pushed (`3c22f74d`).

## Review synthesis

**Code review (helper + wiring):** clean. The recursion mirrors `enforceStrictSchema`'s traversal; the 7 wiring sites in `generate-types.ts` are consistent (`enforceStrictSchema(removeArrayLengthConstraints(injectJsdocConstraints(schema)))`); idempotency via exact-line match is sound; the helper preserves the `minimum` / `maximum` / etc. keywords on the schema (does not delete them), so Ajv still enforces them against the unstripped JSON Schema in parallel with the Zod chain.

**Protocol alignment:** all six injected constraint kinds match AdCP 3.0.11 spec semantics. `minimum: 0` lives on budgets / counts / deadlines — none of which legitimately go negative. `format: hostname` is **not** in `TS_TO_ZOD_SUPPORTED_FORMATS`, so `Brand.domain` is enforced via `pattern` (not `format`) — the FQDN pattern matches `localhost`, IPv4 segments fail (no all-numeric labels), single-label hosts match, empty strings fail. That's correct per spec. Lowercase ISO country codes (`us`) now fail vs uppercase (`US`) — also correct (`^[A-Z]{2}$`) and explicitly pinned by the test.

**Test coverage:** synthetic end-to-end already covers all six tag kinds, nested recursion, unsupported-format skip, idempotency, description preservation, and slash escaping. Pinning slice exercises real generated Zod against the dist build. Gap noted (not blocker): `oneOf` / `anyOf` / `allOf` branch constraints don't round-trip through ts-to-zod's JSDoc reader, but Ajv enforces them at runtime — acceptable.

## Blocker fixed in 3c22f74d

CodeQL flagged `scripts/schema-utils.ts:130` as "incomplete string escaping" (1 high). The `.replace(/\//g, '\\/')` is correct for the AdCP 3.0.11 schema corpus but has two real edge cases:

1. A pattern already containing `\/` would become `\\/`, breaking the `/PATTERN/` wrapper ts-to-zod emits.
2. A pattern with an unpaired trailing `\` would let the closing `/` of the regex literal get consumed as an escape.

Replaced with a single-pass walker that treats `\X` as one regex token (so `\d`, `\.` keep regex-source semantics — escaping `\` would turn `\d` into the literal sequence `\d`, not the digit class), escapes bare `/` → `\/`, and skips patterns with an unpaired trailing `\` (Ajv still enforces them). No diff in the generated files — none of the current AdCP schemas hit either edge case. Two new test cases pin the behavior.

`npm test` (8741 tests, 0 fail), `npm run format:check`, `npm run build`, `npm run typecheck` all clean locally.

PR #1783, opened 2026-05-16
fix(codegen): pre-merge allOf[$ref] siblings to unbreak jsts union emission

### LINKED ISSUE #1756: Codegen handling of allOf+$ref produces broken union types (blocks adcp#4510 schema dedup)

bokelley:
Closes #1756. Unblocks adcp#4510 (the spec-side schema dedup spike that reverted on this codegen bug).

## Summary

JSON Schema objects that mix sibling `properties`/`required` with `allOf: [{ $ref }]` make `json-schema-to-typescript` emit a broken union `( Base | { variant + duplicated base fields } )` where the intent is `Base & { variant }`. This is most visible inside `oneOf` discriminator variants — `GetContentStandardsResponse` collapsed its success arm to bare `ContentStandards`, silently dropping the variant's own `context` and `ext` fields.

`enforceStrictSchema` now pre-merges any `allOf` member that is a single `$ref` into the parent shape when the parent already declares its own `properties`/`required` — variant-level fields win on collision; the base's `additionalProperties` is inherited only when the variant didn't override it. Local `#/$defs/...` fragments are left in place (existing Individual*Asset post-processor handles them). `vendor-pricing-option`-style allOf-only roots stay untouched (jsts handles those correctly today).

This is paired with the sibling PR for #1745 (JSDoc constraint injection) — independent edits to different parts of `scripts/generate-types.ts`, mergeable in either order.

## Knock-on type changes (intentional)

- `BriefAsset` / `CatalogAsset` flatten from `Base & { ... }` to merged interfaces. Field set is identical; the named base type is no longer referenced via intersection. **This matches option 2 in the issue's acceptance criteria** ("flattens the allOf into a single merged shape — less ideal but acceptable"). Recovering the `Base & { variant }` intersection form would be a follow-up polish pass and is out of scope here.
- `GetContentStandardsResponse` success arm regains its `context?` / `ext?` fields — the bug this PR fixes.
- `CreativeBrief` no longer transitively reaches `tools.generated.ts`; `src/lib/index.ts` re-exports it from `core.generated` instead. Existing import sites are unaffected.

## Test plan

- [x] Unit tests for `enforceStrictSchema` cover three scenarios — `vendor-pricing-option`-style pass-through, single-variant merge, and a two-variant `oneOf` ending in a clean discriminated union (`test/type-generator-allof-ref-merge.test.js`).
- [x] `npm run generate-types` regenerates cleanly; diff confined to the four type changes listed above plus their downstream Zod schemas.
- [x] `npm test` passes (1 known-flaky perf-test unrelated to this change: `request-signing-replay-perf.test.js`'s `has() sub-linear` assertion — passes in isolation).
- [x] `npm run format:check` passes.
- [x] Pre-push validation (`tsc --noEmit` + `build:lib`) passes.

### DIFF
--- a/.changeset/codegen-allof-ref-merge.md
+++ b/.changeset/codegen-allof-ref-merge.md
@@ -1,0 +1,17 @@
+---
+'@adcp/sdk': patch
+---
+
+fix(codegen): pre-merge `allOf: [{ $ref }]` siblings into parent shape (#1756)
+
+`json-schema-to-typescript` mishandles JSON Schema objects that combine `properties` / `required` at their own level with an `allOf: [{ $ref }]` sibling — instead of emitting `BaseFields & { variant-specific }` it emits a broken union `( BaseFields | { variant-specific + duplicated base fields } )`. This is most visible inside `oneOf` discriminator variants, where the success arm of `get_content_standards` collapsed to bare `ContentStandards`, silently dropping the variant's own `context` and `ext` fields.
+
+`enforceStrictSchema` now pre-merges any `allOf` member that is a single `$ref` into the parent schema when the parent already declares its own `properties` / `required` — variant-level fields win on collision, and the base schema's `additionalProperties` is inherited only when the variant didn't override it. Refs we can't resolve through the cache (local `#/$defs/...` fragments, missing schemas) are left in place for jsts to handle as before. The `vendor-pricing-option`-style schemas — allOf-only at root, no sibling properties — are intentionally untouched.
+
+Side effects on existing emitted types:
+
+- `BriefAsset` and `CatalogAsset` change from `CreativeBrief & { asset_type: 'brief' }` / `Catalog & { asset_type: 'catalog' }` aliases to flat interfaces with the merged shape. Field set is identical; the named base reference is gone. Tracked as option 2 in the adcp#4510 acceptance criteria.
+- `GetContentStandardsResponse` success variant now correctly includes its own `context?` and `ext?` fields alongside the merged `ContentStandards` shape (the bug this PR fixes).
+- `CreativeBrief` is no longer transitively pulled into `tools.generated.ts` (it's still emitted in `core.generated.ts`); `src/lib/index.ts` re-exports it from `core.generated` instead.
+
+Unblocks adcp#4510 (schema dedup spike), which the spec team reverted on the bug surfaced by this codegen path.
--- a/scripts/generate-types.ts
+++ b/scripts/generate-types.ts
@@ -192,6 +192,37 @@ function tightenMutualExclusionOneOf(schema: any): any {
 }
 
 /**
+ * Resolve an external `$ref` for the purpose of pre-merging an
+ * `allOf: [{ $ref }]` member into its parent. Returns `null` for
+ * unresolvable refs (including local `#/$defs/...` refs, which require
+ * document context this helper doesn't have). Suppresses the warning that
+ * `loadCachedSchema` would emit for unresolvable paths — a miss here just
+ * means we leave the `allOf` member in place for jsts to handle.
+ */
+function resolveAllOfRefForMerge(ref: string): any | null {
+  if (!ref || typeof ref !== 'string') return null;
+  // Only external schema refs are resolvable through the cache. Local
+  // `#/$defs/...` and other fragment-only refs are left to jsts.
+  if (!ref.startsWith('/schemas/')) return null;
+  // Suppress the warn from loadCachedSchema for legitimate misses (e.g. a
+  // schema path we don't have cached yet). The original `allOf` member stays
+  // in place if resolution fails.
+  const originalWarn = console.warn;
+  console.warn = () => {};
+  let raw: any;
+  try {
+    raw = loadCachedSchema(ref);
+  } finally {
+    console.warn = originalWarn;
+  }
+  if (!raw) return null;
+  // Apply the same preprocessing the ref resolver uses for normal $ref reads —
+  // otherwise minItems/maxItems constraints from the base schema would leak
+  // through the merge and resurrect as `@minItems`/tuple types in jsts output.
+  return removeArrayLengthConstraints(raw);
+}
+
+/**
  * Recursively remove additionalProperties: true from schema to enforce strict typing
  * This prevents [k: string]: unknown in generated TypeScript types
  *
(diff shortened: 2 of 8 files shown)


bokelley:
Coordinator review (3-lens sweep: code / protocol / testing). One real concern worth surfacing; the rest are nits or follow-ups. Not blocking — flagging for your call before merge.

## Real concern: `additionalProperties: true` leaks through the merge

`resolveAllOfRefForMerge` returns the raw resolved base (only `removeArrayLengthConstraints` applied — not `enforceStrictSchema`). The merge then propagates `resolved.additionalProperties` directly:

```ts
if (strictSchema.additionalProperties === undefined && resolved.additionalProperties !== undefined) {
  strictSchema.additionalProperties = resolved.additionalProperties;
}
```

`creative-brief.json` and `catalog.json` both declare top-level `additionalProperties: true`. Pre-PR, `enforceStrictSchema` stripped that at line 253 when emitting the standalone `CreativeBrief` interface, so `CreativeBrief & { asset_type: 'brief' }` had no index signature. Post-PR, the merged `BriefAsset` gains `[k: string]: unknown | undefined` (core.generated.ts:2158, also at line 2233 for `CatalogAsset`).

Net: the new merged shape is **wider** than both the old intersection form and the current standalone `CreativeBrief` interface. Adopters who relied on the closed shape lose that guarantee.

Two ways to tighten if you want strictness parity:
- Run `enforceStrictSchema(resolved)` inside `resolveAllOfRefForMerge` before returning — the top-level `additionalProperties: true` strip then fires before the merge sees it.
- After the merge, re-check: `if (strictSchema.additionalProperties === true && !mustPreserveProperties) delete strictSchema.additionalProperties;`

Both fix it. Calling `enforceStrictSchema` on the resolved base also handles nested base allOf chains for free (currently silently dropped — see follow-up #2 below).

If you prefer to ship as-is given the "merged shape is acceptable / option 2" framing in the issue, fine — but worth a code comment noting the index-signature widening is a deliberate side effect of the merge so the next codegen maintainer doesn't get surprised.

## Follow-ups (file separately, not blocking)

1. **Test coverage gap**: no test pins the `additionalProperties` propagation behavior either way. A subtest "base with `additionalProperties: true` does/doesn't leak `[k: string]` into merged shape" would lock the contract regardless of which direction you go above.
2. **Resolved base's own `allOf` is silently dropped**: merge only inherits `properties` / `required` / `additionalProperties` from `resolved`. If a future base schema chains its own `allOf: [{ $ref }, ...]`, those intersections vanish. No live targets today (creative-brief and catalog have no root-level allOf). Cheap fix is the same as above — recurse through `enforceStrictSchema` on `resolved` before merging.
3. **Collision precedence is documented but not tested**: variant-level field winning on collision is asserted in the code comment but not pinned by a test. A one-line subtest (base has `name: string`, variant has `name: { const: 'brief' }`, assert variant wins) would prevent regression.
4. **`runHarness` temp-dir leak on `npx tsx` crash**: `.allof-ref-merge-harness-*` directories under REPO_ROOT are cleaned in `finally`, but `spawnSync` failures throw before — actually wait, no, `finally` runs after throw. Disregard.

## Confidence

CI is effectively green (Test & Build at "Run unit tests ✓", finishing post-test steps). Two-lens checks I ran:

- `BriefAsset` / `CatalogAsset` field set is identical to pre-PR. Verified by diffing the property keys against the old `CreativeBrief & { asset_type }` form.
- `GetContentStandardsResponse` success arm correctly surfaces `context?` and `ext?` at tools.generated.ts:10970-10971. That's the bug fix.
- `CreativeBrief` is still emitted in `core.generated.ts:8487` and re-exported through `src/lib/index.ts:546`. No consumption-path breakage for adopters who import `CreativeBrief` from `@adcp/sdk`.
- `tightenMutualExclusionOneOf` runs before the merge (line 241), then merge runs before the `not` / `if-then-else` filter (line 398). Sequencing is correct — non-ref allOf members fall through to the existing filter.

Mergeable. The strictness regression on `BriefAsset` / `CatalogAsset` is the only judgement call.

bokelley:
Addressed the reviewer's `additionalProperties` widening concern in eab40000.

**What changed**

`resolveAllOfRefForMerge` now applies `enforceStrictSchema` to the resolved base before returning it. Previously it returned the raw schema with only `removeArrayLengthConstraints` applied, which let the top-level `additionalProperties: true` flag on `creative-brief.json` and `catalog.json` propagate into the merged shape and emit a `[k: string]: unknown | undefined` index signature on `BriefAsset` / `CatalogAsset`. The fix runs the resolved base through the same strict-schema pipeline the parent uses, so the strip happens before the merge.

**Recursion verified safe**: `loadCachedSchema` reads a fresh JSON document per call (no shared mutable cache), and AdCP schemas aren't cyclic at the `allOf:[{ $ref }]` sibling level. Both schemas in question have no top-level `allOf`, so the new pre-merge pass on the resolved base doesn't even re-enter the same code path here.

**Diff scope** (tight, as expected):
- `scripts/generate-types.ts`: +11 lines (the normalization + comment).
- `src/lib/types/core.generated.ts`: removed `[k: string]: unknown | undefined` from `BriefAsset` and `CatalogAsset`.
- `src/lib/types/tools.generated.ts`: same two index signatures removed.
- `test/type-generator-allof-ref-merge.test.js`: new test pinning that resolved-base `additionalProperties: true` does NOT propagate. Verified the test fails on the prior `resolveAllOfRefForMerge` and passes after the fix.

Full test suite passes, format/build clean. The two pre-existing lint errors (`prefer-const` in storyboard runner, `task-map.test.ts` parsing) are unrelated and present on the parent commit.

The three "nice-to-have" items you flagged (collision-precedence pin, nested-base allOf test, etc.) are out of scope for this fix — happy to file follow-ups if useful.

PR #1786, opened 2026-05-16
feat(testing): bridge participation marker (#1775)

bokelley:
## Summary

Closes #1775 (SDK half — consumer half tracked in #1782 and `adcontextprotocol/adcp`).

When the `TestControllerBridge` merges seeded fixtures into a handler response, stamp a non-normative `_bridge: { callback, tool, merged_count }` marker on `structuredContent` (mirrored to `content[0].text` when JSON). This is the runner-visible signal that distinguishes _"this pass exercised the adopter's adapter against upstream"_ from _"this pass exercised wire conformance against fixture data merged by the SDK"_. Storyboard runners and compliance leaderboards read it to attribute bridge participation in run records.

Why now: the bridge surface has grown from 5 tools (PR #1754) to **13** through phases 2–4 (#1759, #1761, #1772, #1773). With 13 fixture-augmented tools, every storyboard pass through fixture-merge reads identically to one that ran through a real adapter — exactly where the gap matters most (walled-garden proxies). JSDoc disclaimers don't survive contact with a leaderboard score.

## What ships

- **`stampBridge` helper** (parallel to `stampReplayed`) sets `_bridge` on `structuredContent` + opportunistic JSON-text mirror.
- **19 dispatch sites across 16 bridge callbacks** — every bridge merge point in `create-adcp-server.ts` stamps per-tool with the originating callback name, so runners can attribute participation at the `getSeededCreatives` / `getSeededMediaBuys` / etc. granularity.
- **Append-merge tools** stamp when the callback returned ≥ 1 valid entry. **Singleton-replace tools** (`get_account_financials`, `get_brand_identity`, `si_get_offering`, `get_property_list`, `get_collection_list`, `get_content_standards`) stamp only when a seeded fixture actually matched the request id and replaced the handler payload.
- **Marker absent** on non-sandbox traffic, when the callback is omitted, and on singleton-replace tools when no fixture matched.
- **Schema-safe**: verified every bridge-touching response schema sets `additionalProperties: true` at the top level (success arm for `oneOf` envelopes). Underscore prefix advertises "internal / out-of-spec" to validators that round-trip unknowns.
- **`BridgeMarker` type** exported from `@adcp/sdk/server` for adopters who want to type-check marker reads.
- **Dead-code removal**: dropped three unreachable duplicate dispatch blocks (`get_brand_identity`, `get_rights`, `si_get_offering`) that were appended below the canonical `else if` chain. No behavior change — only the first block ever fired in a continuous `else if`.

## Docs

[...]

Issue #1435, opened 2026-05-03
feat(server): Account.mode + sandbox-account authority gate for comply controller


bokelley:
## Phase 1 of [docs/proposals/lifecycle-state-and-sandbox-authority.md](https://github.com/adcontextprotocol/adcp-client/blob/main/docs/proposals/lifecycle-state-and-sandbox-authority.md)

The proposal introduces three account modes — `live` / `sandbox` / `mock` — and reframes how compliance routes. Phase 1 lands the smallest, most-load-bearing piece: **the SDK enforces that `comply_test_controller` only operates on sandbox or mock accounts.** Today's wiring relies on adopter-supplied `sandboxGate` callbacks (often `() => process.env.ADCP_SANDBOX === '1'`); a misconfigured env in the wrong tenant would open the controller surface to all callers. The flag must live on the resolved account, not on the process.

## Scope

1. **Add `Account.mode: 'live' | 'sandbox' | 'mock'`** as a new field on the resolved-account shape. Keep `Account.sandbox: boolean` as a derived accessor (`mode === 'sandbox' || mode === 'mock'`) for back-compat; deprecation comes later.
2. **Thread `mode` through `AccountStore` implementations** — `InMemoryImplicitAccountStore`, `RosterAccountStore`, `createTenantStore`. Each needs to know how to populate `mode` (default `'live'` when unspecified, preserving today's behavior).
3. **SDK dispatch gate** at `src/lib/server/create-adcp-server.ts:~3280` (mirror the existing `sandbox_only` agent pattern at lines 3280-3292). For `comply_test_controller` calls, throw `PERMISSION_DENIED` unless `ctx.account.mode === 'sandbox'` or `'mock'`.
4. **`context.sandbox` fallback** for unresolved-account paths (`get_adcp_capabilities`, probe calls, conformance pre-account bootstrap). Closed-world: if account resolved, only `account.mode` counts; if no account, fall back to `context.sandbox`.
5. **`ADCP_SANDBOX=1` env-gate becomes vestigial** — adopters running conformance migrate to flagging accounts in their `AccountStore.resolve` implementation.
6. **Tests** covering 3×3 mode×dispatch matrix, plus the unresolved-account fallback path.
7. **Migration changeset** explaining how adopters move from env-gate to account-flag.

## Open question — spec-side change?

[...]

PR #1788, opened 2026-05-16
feat(server): construction-time warn when testController lacks resolveAccount (#1784)

### LINKED ISSUE #1784: Bridge: construction-time debug when testController is registered without resolveAccount
## Context

The `TestControllerBridge` sandbox gate at `src/lib/server/create-adcp-server.ts:3913-3917` admits requests where `ctx.account === undefined` AND `isSandboxRequest(params) === true`. That's intentional — storyboard runners often have no account scoping. But adopters who register \`testController\` against a production binding without configuring \`resolveAccount\` get the buyer-supplied \`context.sandbox: true\` as the only line of defense.

PR #1779 added a JSDoc paragraph naming this as the trust boundary. But JSDoc doesn't fire when adopters misconfigure.

## Ask

At \`createAdcpServer\` construction time, when both:

1. \`config.testController\` (or \`config.testControllerBridge\`) is present, AND
2. \`config.resolveAccount\` is absent

…emit a one-shot \`logger.warn\` (or \`logger.info\` — bikeshed welcome):

> testController is registered but resolveAccount is not configured. The sandbox gate's account-side check will always pass; the request-signal check (account.sandbox / context.sandbox) is the only line of defense against fixture leakage. Production bindings should configure resolveAccount; storyboard-runner bindings can ignore this warning. See \`docs/guides/...\` for the trust model.

One-shot at construction, not per-request — repeating it per dispatch is noise.

## Why not a hard error

Storyboard runners with no account scoping are a real and intentional configuration. Failing at construction would break that path. A warn-once is the right severity: visible in any logging setup, ignorable for runners, actionable for adopters.

## Background

Spun out from the post-merge expert review of (now-closed) #1754. Three of four reviewers flagged "JSDoc says adopter responsibility — that's correct but the failure mode is silent." This issue makes the failure mode loud once, without breaking the legitimate runner-without-resolveAccount case.

## Acceptance criteria

- [ ] \`createAdcpServer\` emits a one-shot \`logger.warn\` when \`testController\` is configured but \`resolveAccount\` is not.
- [ ] Regression test in \`test/lib/seed-per-tool-wiring.test.js\` (or similar): warn fires once when both conditions are met, does not fire when either condition is absent, does not fire on subsequent requests.
- [ ] JSDoc reference on \`TestControllerBridge\` mentions the construction-time warn for adopter awareness.

## Scope of "configured"

\`config.testController\` checks for a non-null value. The presence of any seed callback (\`getSeededProducts\`, \`getSeededCreatives\`, …) means the bridge is wired. A bridge with no callbacks is a no-op and shouldn't trigger the warn — keep it simple by gating on \`config.testController != null\` rather than enumerating callbacks.

cc #1779 (JSDoc) and #1780 (sandbox-gate debug log)

bokelley:
## Summary

Closes #1784. When `createAdcpServer` is called with `testController` set but neither `resolveAccount` nor `resolveAccountFromAuth` configured, emit a one-shot `logger.warn` at construction.

The dispatch-time sandbox gate (`src/lib/server/create-adcp-server.ts:3973-3981`) admits requests where `ctx.account === undefined`. Without an account resolver, `ctx.account` is always `undefined`, so the only remaining check is the buyer-supplied `account.sandbox` / `context.sandbox` marker — caller-controlled, not a trust boundary. This is the gap the security review on #1786 flagged as M1; the JSDoc updates in #1786 and #1787 document it, but JSDoc doesn't fire when adopters misconfigure. This PR adds the runtime equivalent.

## Why a warn, not a throw

Storyboard-runner deployments without account scoping are a legitimate and intentional configuration — the runner drives state directly, no buyer authentication path needed. Failing construction would break that case. The warn is the right severity: visible in any logging setup, ignorable for runners, actionable for adopters.

## OAuth-passthrough handling

The check gates on `resolveAccountFromAuth` as well as `resolveAccount`. OAuth-passthrough setups (`createOAuthPassthroughResolver`, the canonical Shape-B adopter path from the v6.7 → 6.9 migration) only configure the latter — either resolver populates `ctx.account` at dispatch and gives the gate its account-side teeth, so warning on `resolveAccountFromAuth`-only setups would be a spurious noise floor.

## JSDoc cross-link

`TestControllerBridge` (in `test-controller-bridge.ts`) now mentions the construction-time warn so adopters know what triggers it and how to silence it.

## Test plan

- [x] Five regression tests in `test/lib/seed-per-tool-wiring.test.js`:
  - warn fires once when `testController` + neither resolver
  - no warn when `testController` is omitted
  - no warn when `resolveAccount` is configured
  - no warn when `resolveAccountFromAuth` is configured (OAuth passthrough)
  - warn fires **exactly once** across construction + multiple dispatches (not per-request)
- [x] All 143 existing tests in the file pass unchanged
- [x] `tsc --noEmit` clean
- [x] `prettier --check` clean

### DIFF
--- a/.changeset/bridge-warn-no-resolve-account.md
+++ b/.changeset/bridge-warn-no-resolve-account.md
@@ -4,9 +4,11 @@
 
 feat(server): construction-time warn when `testController` is registered without an account resolver (#1784)
 
-`createAdcpServer` now emits a one-shot `logger.warn` at construction when `config.testController` is present but neither `config.resolveAccount` nor `config.resolveAccountFromAuth` is configured. In that setup, the dispatch-time sandbox gate's account-side check has no teeth — the gate admits requests where `ctx.account === undefined`, so the only remaining check is buyer-supplied `account.sandbox` / `context.sandbox` on the request body, which is caller-controlled and not a trust boundary.
+`createAdcpServer` now emits a one-shot construction-time warning when `config.testController` is present but neither `config.resolveAccount` nor `config.resolveAccountFromAuth` is configured. In that setup, the dispatch-time sandbox gate's account-side check has no teeth — the gate admits requests where `ctx.account === undefined`, so the only remaining check is buyer-supplied `account.sandbox` / `context.sandbox` on the request body, which is caller-controlled and not a trust boundary.
 
-The warn makes that silent misconfiguration loud once at startup, where any logging setup will surface it. It's deliberately not a hard error: storyboard-runner deployments without account scoping are a legitimate and intentional configuration (the runner drives state directly, no buyer authentication path needed), and failing construction would break that case. The warn message tells storyboard runners they can ignore it.
+The warning **dual-emits** via `process.emitWarning(message, { type: 'AdcpServerConfigWarning', code: 'ADCP_BRIDGE_NO_RESOLVER' })` AND `logger.warn(message)`. The `process.emitWarning` channel writes to stderr by default so the signal is visible even when `logger` is the default `noopLogger` (the day-one case where the misconfig is most likely). The `logger.warn` channel routes the same signal through any adopter-configured logging pipeline. Storyboard runners that knowingly run without account scoping can silence via `node --no-warnings=ADCP_BRIDGE_NO_RESOLVER`.
+
+It's deliberately not a hard error: storyboard-runner deployments without account scoping are a legitimate and intentional configuration (the runner drives state directly, no buyer authentication path needed), and failing construction would break that case. The warning message tells storyboard runners they can ignore it.
 
 The check gates on `resolveAccountFromAuth` as well as `resolveAccount`, so OAuth-passthrough setups (`createOAuthPassthroughResolver`, the canonical Shape-B adopter path) don't get a spurious warn — either resolver populates `ctx.account` at dispatch time and gives the gate its account-side teeth.
 
--- a/.changeset/bridge-warn-no-resolve-account.md
+++ b/.changeset/bridge-warn-no-resolve-account.md
@@ -1,0 +1,13 @@
+---
+'@adcp/sdk': minor
+---
+
+feat(server): construction-time warn when `testController` is registered without an account resolver (#1784)
+
+`createAdcpServer` now emits a one-shot `logger.warn` at construction when `config.testController` is present but neither `config.resolveAccount` nor `config.resolveAccountFromAuth` is configured. In that setup, the dispatch-time sandbox gate's account-side check has no teeth — the gate admits requests where `ctx.account === undefined`, so the only remaining check is buyer-supplied `account.sandbox` / `context.sandbox` on the request body, which is caller-controlled and not a trust boundary.
+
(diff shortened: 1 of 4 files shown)


bokelley:
## Expert review pass — three reviewers

Ran three parallel expert reviews (code-reviewer, security, dx). Synthesis:

**Convergent feedback addressed in commit e83bacd:**

1. **DX (substantive)** — default `logger` is `noopLogger`, which swallows `.warn`. The day-one misconfig case (no logger wired yet) is exactly the population this warn is targeting, and they'd see nothing. **Fix**: dual-emit via `process.emitWarning(message, { type: 'AdcpServerConfigWarning', code: 'ADCP_BRIDGE_NO_RESOLVER' })` (stderr by default, idiomatic Node, silenceable via `--no-warnings=ADCP_BRIDGE_NO_RESOLVER`) AND `logger.warn(message)` for adopters with configured logging.
2. **DX (message)** — lead with the action, trim the trust-model paragraph (JSDoc carries the long form), add public docs URL for coding agents reading the warn from a log.
3. **Code review (cosmetic)** — prefix message with `[adcp/createAdcpServer]` for parity with sibling warns at lines 3271/3304.

**Security** — agent stalled but partial output confirmed the M1 finding from #1786 is closed by this PR: the warn correctly targets the legacy `createAdcpServer` surface (where the misconfig path exists); `createAdcpServerFromPlatform` already enforces the sandbox-authority gate via Phase 2 of #1435.

**New test** for the stderr channel uses a synchronous spy on `process.emitWarning` itself (the original async `process.on('warning', ...)` approach had event-loop-flush timing dependencies that flaked under the node test runner).

**Verdict from each reviewer:** ship after the DX fix lands. Done in e83bacd.

CI re-running.

bokelley:
## Second review pass — both clear

Code-reviewer and security-reviewer re-fired on commit \`e83bacd\` (the DX fold-in).

**Security**: M1 fully closed by the dual-emit. Stderr content is safe (no tenant data, no auth, no payload). \`--no-warnings=ADCP_BRIDGE_NO_RESOLVER\` silencer is operator-scoped, not buyer-controllable. Per-construction emit cardinality is acceptable; process-level dedup would actually hide per-tenant signal in multi-tenant deployments. **Ship.**

**Code**: No blockers. Two nits folded in commit 6cfff71:
- Documented the \`ADCP_*\` warning-code convention in the changeset (this PR establishes it; future framework warns inherit \`ADCP_<subsystem>_*\` + \`AdcpServerConfigWarning\` type for adopter handler routing).
- Filed #1790 for the docs-URL rot mitigation (build-time link check) — non-blocking; worth doing when the next message wants a URL.

Other code-review notes (all non-blocking):
- Spy-and-restore pattern in the new test is safe under node:test's default sequential \`it\` execution within a file; sibling \`agent-class-deprecation.test.js\` patches \`process.emitWarning\` too but runs in a separate worker.
- Dual-emit doesn't double-count (the channels don't cross-flow).
- Sibling-warn prefix \`[adcp/createAdcpServer]\` matches existing usages at lines 3280/3313.

CI green → admin merging.

Issue #1799, opened 2026-05-16
withResponseSizeLimit doesn't wrap getAgentInfo / mcpClient.listTools — tools/list body uncapped


bokelley:
From adcp#3731 follow-up.

## Gap

`SingleAgentClient.getAgentInfo()` (`dist/lib/core/SingleAgentClient.js:1812`) calls `connectMCP()` then `mcpClient.listTools()` directly with no `withResponseSizeLimit` scope. Same on the A2A branch — `A2AClient.fromCardUrl` is called with a hand-rolled `fetchImpl` that doesn't route through the wrapper.

Result: `transport.maxResponseBytes` (which works on `ProtocolClient.callTool` and `get_adcp_capabilities`) silently doesn't apply to `tools/list` responses. A hostile agent can return arbitrarily large `tools/list` payloads to a discovery client even when `maxResponseBytes` is set.

## Ask

Wrap `getAgentInfo`'s `mcpClient.listTools()` and the A2A branch's `fromCardUrl` call in `withResponseSizeLimit` using the same `maxResponseBytes` slot the rest of the SDK already reads.

## Downstream

adcontextprotocol/adcp passes `transport.maxResponseBytes: 4 MB` to its AAO discovery constructors (see adcontextprotocol/adcp PR in flight on #3731). The cap is dormant for tools/list until this lands.

Issue #1797, opened 2026-05-16
TrackResult.status: emit 'silent' for wired-but-not-observed assertions


bokelley:
From adcontextprotocol/adcp#2834 (cross-repo dependency).

## Ask

`@adcp/client`'s `status.monotonic` assertion should emit `status: 'silent'` with `observations_count: 0` when it runs with zero tracked lifecycle resources — distinguishing 'check was wired but never observed any transitions' from 'check observed transitions and they were monotonic.'

## Why

`TrackResult.status` currently emits `'pass'` for both cases. The downstream compliance grader (in `adcontextprotocol/adcp`) re-serializes the value verbatim and cannot distinguish the two. UI presentation of a 'wired-but-not-observed' assertion is the same green checkmark as an actually-passing assertion — masking false-positive coverage.

## Scope

1. Extend `TrackStatus` upstream to include `'silent'`.
2. Emit `status: 'silent'` from `status.monotonic` when `observations_count === 0`.
3. Document in the SDK release notes so the downstream extension in adcp#2834 can plan a schema/UI rendering update.

## Downstream

Once shipped, adcp#2834 will:
- Extend `TrackStatus` in `compliance-db.ts`, add a DB migration, add a `'silent'` rendering branch in `member-tools.ts`.
- Extend `storyboard-schema.yaml` with the `invariants:` block definition.

Tracking: adcontextprotocol/adcp#2834

Issue #1524, opened 2026-05-03
track: link to mock-server normative-status anchor once adcp#4029 lands


bokelley:
## Tracking issue — blocked on `adcontextprotocol/adcp#4029`

When the upstream protocol docs land an explicit "mock-server is the reference implementation" section (or wherever the maintainers put it), the SDK side has a tiny doc follow-up:

- [ ] In \`docs/proposals/lifecycle-state-and-sandbox-authority.md\`, replace the current language ("the mock-server is the impartial referee", which is currently SDK-author opinion) with a link to the upstream anchor.
- [ ] In \`CLAUDE.md\` near the existing "When a compliance storyboard fails, triage before patching" guidance, link to the upstream anchor too — currently the section encodes the same posture as project guidance, but a spec link makes it authoritative.
- [ ] Maybe: add a short README snippet on the mock-server CLI (\`bin/adcp.js mock-server …\`) pointing at the spec page. Adopters running the harness should know they're hitting the reference implementation, not one-of-many.

## Why this matters even after a spec link exists

The link is the easy part. The harder thing this enables: when a future SDK contributor patches the SDK to match a wrong mock behavior, the spec anchor gives a reviewer a one-line basis to push back ("the spec triage order is spec → mock → SDK; if the mock is wrong, file an issue against \`adcontextprotocol/adcp\`, don't patch the SDK"). Without the anchor, that's project tribal knowledge.

## Cost

~30 min. Three small doc edits, no code.

## Refs

- \`adcontextprotocol/adcp#4029\` — the upstream issue / proposed normative paragraph
- \`docs/proposals/lifecycle-state-and-sandbox-authority.md\` — the proposal doc that currently makes this claim
- \`CLAUDE.md\` — internal triage-order encoding

When adcp#4029 closes (whether they accept the proposal or land something different in spirit), do the link updates.

bokelley:
## Triage

**Classification:** tracking (blocked on upstream)
**Bucket(s):** docs
**Status:** deferred

**My take:** Pre-classification catch — this is an explicit tracking issue gated on `adcontextprotocol/adcp#4029` landing a normative anchor for mock-server reference-implementation status. No SDK-side action is possible until that upstream issue closes; the three doc edits described are straightforward and non-breaking once the upstream anchor exists.

Blocked-on: adcontextprotocol/adcp#4029 — resurfaces on merge

---
[...]

PR #1802, opened 2026-05-17
feat: AdCP 3.0.12 + denial gate + silent assertion status + size-limit wrap

### LINKED ISSUE #1522: track: wire denial-path conformance once adcp#4028 storyboard lands
## Tracking issue — blocked on `adcontextprotocol/adcp#4028`

When the upstream storyboard for \`comply_test_controller\` denial against a live-mode account ships, the SDK side has follow-up:

- [ ] Bump \`ADCP_VERSION\` to whichever 3.0.x release ships the new storyboard (the maintainers cut patches frequently — likely days).
- [ ] Run the new storyboard against \`hello_seller_adapter_guaranteed\` with both a sandbox and a live test-kit principal:
  - sandbox principal: scenario admits, \`force_creative_status\` succeeds (existing behavior).
  - live principal: scenario refuses, returns the spec-mandated error code (likely \`FORBIDDEN\`; SDK already emits this).
- [ ] If the spec mandates a code other than \`FORBIDDEN\` (e.g. \`PERMISSION_DENIED\`, \`unknown_tool\`), update \`from-platform.ts\`'s gate refusal to match — currently the SDK returns \`{ success: false, error: 'FORBIDDEN' }\` per the \`ControllerError\` shape. The expected code falls out of the spec discussion in \`adcontextprotocol/adcp#3986\` (visibility) so this might land bundled with that work.
- [ ] Add a \`test/examples/hello-seller-adapter-guaranteed.test.js\` assertion that the new denial storyboard PASSES against the worked example. Currently the suite runs \`sales_guaranteed\` + the cascade scenarios — extend to include the denial scenario via the test runner's \`requires_scenarios\` mechanism.
- [ ] Verify \`hello_signals_adapter_marketplace.ts\`, \`hello_seller_adapter_non_guaranteed.ts\`, \`hello_creative_adapter_template.ts\`, and any other Hello adapter that wires \`complyTest:\` also passes. Those use the same framework gate (#1435 phase 2), so they SHOULD be green automatically — but we should verify.

## What this gates

The SDK's claim that the framework gate refuses live-mode dispatch is currently SDK-author opinion. After this, it's a conformance-checkable assertion — any verifier (AAO and similar) can grade a seller against the public storyboard without needing to know SDK internals.

## Refs

- \`adcontextprotocol/adcp#4028\` — upstream storyboard coverage gap
- \`adcontextprotocol/adcp#3986\` — visibility spec question (related; the right error code falls out of this discussion)
- #1521 — visibility implementation tracking issue (sibling)
- PR #1453 — framework gate

When adcp#4028 lands and the new \`ADCP_VERSION\` is published, do the bump + verify steps.

bokelley:
Bundle of four Bucket-A triage closures from session 2026-05-17. Each commit is a self-contained close.

## #1522 — AdCP 3.0.12 + comply_controller_mode_gate denial gate

Adopts the new universal storyboard from [adcp#4028](https://github.com/adcontextprotocol/adcp/issues/4028) as a graded path on `hello_seller_adapter_guaranteed`. The bump unblocks the storyboard end-to-end against framework-gated sellers; the SDK changes close wire-level gaps surfaced during integration.

- `ADCP_VERSION` 3.0.11 → 3.0.12; sync-schemas pulls the new universal storyboard + test-kit fixture.
- **Framework gate echoes context/ext on FORBIDDEN refusal** (`from-platform.ts`). The denial storyboard asserts `context.correlation_id` round-trips on refusal — previously the gate dropped them.
- **`rawMcpProbe` parses Streamable-HTTP MCP SSE responses** (`probes.ts`). Sends `Accept: application/json, text/event-stream` + parses the SSE `data:` line. Storyboards routed via the raw probe now grade against Streamable-HTTP MCP servers; previously the probe 406'd silently. Unit test added.
- **`adcp storyboard run --test-kit PATH`** (`bin/adcp.js`) — loads the test-kit YAML into `options.test_kit` so `from_test_kit: true` / `$test_kit.*` references resolve from the CLI.
- **Worked example wires the live-mode probe principal** (`hello_seller_adapter_guaranteed.ts`). Registers `demo-acme-outdoor-live-v1` as a second bearer; resolver stamps `mode: 'live'` to trigger the gate.
- **CI gate added** via new `extraStoryboards` parameter on `runHelloAdapterGates`.

## #1524 — mock-server-authority anchor

Replaces SDK-author "mock is the impartial referee" prose with a forward link to the spec's normative triage section ([Mock-server authority and failure triage](https://adcontextprotocol.org/docs/building/verification/conformance#mock-server-authority-and-failure-triage), landed in [adcp#4030](https://github.com/adcontextprotocol/adcp/pull/4030)). Touches `docs/proposals/lifecycle-state-and-sandbox-authority.md`, `CLAUDE.md`, and `bin/adcp.js` (mock-server help).

## #1797 — AssertionResult.status: 'silent' | 'pass' | 'fail'

`status.monotonic` and `impairment.coherence` now stamp `status: 'silent'` on the `onEnd` summary record when zero lifecycle resources were observed, and `status: 'pass'` once at least one was. Downstream renderers (the spec-side grader at adcp#2834, Addie, CLI summary) read this per-assertion enum directly rather than inferring from `observation_count === 0`. Track-level rollup unchanged. Tests added for both invariants.

## #1799 — `getAgentInfo()` honors `transport.maxResponseBytes`

`transport.maxResponseBytes` was dormant on the discovery path: `mcpClient.listTools()` and `A2AClient.fromCardUrl` bypassed the ALS slot, so a hostile vendor could return arbitrarily large discovery payloads against a size-capped client. `getAgentInfo` now wraps both call sites in `withResponseSizeLimit`, and the A2A auth-stamping `fetchImpl` is composed through `wrapFetchWithSizeLimit`. Regression test added at `test/unit/get-agent-info-size-limit.test.js`.

## Test plan

- [x] `npm run build` clean
- [x] `node --test test/examples/hello-seller-adapter-guaranteed.test.js` — 4/4 pass (denial gate green)
- [x] `node --test test/lib/storyboard-security.test.js` — 23/23 suites pass, new SSE rawMcpProbe test green
- [x] `node --test test/lib/storyboard-default-invariants.test.js` — 146/146 pass
- [x] `node --test test/unit/a2a-card-size-limit.test.js test/unit/mcp-tool-size-limit.test.js test/unit/oauth-size-limit.test.js test/unit/response-size-limit.test.js test/unit/get-agent-info-size-limit.test.js` — 32/32 pass
- [ ] CI green

Closes #1522, #1524, #1797, #1799.


bokelley:
## Expert review (2026-05-17)

Three reviews in parallel (code-reviewer + security-reviewer + ad-tech-protocol-expert) on the initial four-commit bundle. Convergent finding addressed in commit \`fa25a6f6\`; additional defense-in-depth follow-ups from the security review applied in the same commit.

### Convergent finding (fixed)

**rawMcpProbe SSE parser took only the first \`data:\` line.** MCP Streamable HTTP allows server-initiated frames (e.g. \`notifications/progress\`) before the final \`tools/call\` response on the same SSE stream. The probe would parse the progress envelope and grade the storyboard against the wrong payload.

Fix in \`src/lib/testing/storyboard/probes.ts\`: walk all \`data:\` lines, JSON-parse each, select the envelope whose \`id\` matches the request, fall back to the last parseable envelope otherwise. New unit test in \`test/lib/storyboard-security.test.js\` asserts the progress-frame-before-result ordering.

### Defense-in-depth nits (applied)

- **Comply gate context/ext echo cap.** Open-object fields on the request schema could be stuffed with arbitrary payloads. Self-reflection only (no cross-tenant amplifier), but bounded at 8 KiB serialized to prevent the gate from being a self-amplifier (\`from-platform.ts\`).
- **Bearer JSDoc warning** on the worked-example \`demo-acme-outdoor-live-v1\` token, callout that it's published in the open-source SDK and the public compliance test-kit and MUST NOT be copied into a production seller's keys map (\`hello_seller_adapter_guaranteed.ts\`).

### Out-of-scope gaps filed as follow-ups

- **#1804** — \`fetchA2ACanonicalUrl\` and other A2A discovery sites still bypass \`transport.maxResponseBytes\` (PR #1802 only scoped \`getAgentInfo\` per #1799).
- **#1805** — \`force_scenario_unsupported\` runner-MUST from AdCP 3.0.12 \`runner-output-contract.yaml\` not yet implemented.

### Reviewer-flagged but accepted as-is

- \`AssertionResult.status\` is ahead of the spec (adcp#2834 is open). Field-name collision risk is low; flagging in case spec lands with a different shape.
- \`ctx.authInfo.token\` is \`@deprecated\` in \`ResolvedAuthInfo\`; used in the worked example for principal detection. Comment explicitly directs production sellers to \`credential.key_id\` or a tenant-store lookup.

PR #1795, opened 2026-05-16
docs: align bridge framing with revised single-dimension certification model

bokelley:
## Summary

Maintainer [walked back](https://github.com/adcontextprotocol/adcp-client/issues/1782#issuecomment-4467106258) the two-badge proxy-vs-state-local-seller split from #1782 and proposed a single-dimension framing: every seller faces the same verifiability gap; the bridge is one of two mechanisms for closing the seed→read loop, not a special path for one seller class. This PR folds the revised framing back into the JSDoc on `AdcpServerConfig.testController` (which currently names "upstream-proxy sellers" as the audience) and into `skills/build-seller-agent/SKILL.md`.

Also posted matching revision to the upstream taxonomy proposal at [`adcontextprotocol/adcp#4593`](https://github.com/adcontextprotocol/adcp/issues/4593#issuecomment-4467556800).

## Changes

**JSDoc** (`src/lib/server/create-adcp-server.ts:1676-1707`):
- Drops "only upstream-proxy sellers" as primary framing
- "Pick by where your read handlers fetch from, not by seller class"
- "Either path earns wire-conformance credit; it is *not* a separate certification category"

**Seller-agent skill** (`skills/build-seller-agent/SKILL.md`):
- New "Test surfaces — making your agent verifiable without live credentials" section between "Validate locally" and "Deployment"
- Frames the verifiability gap as universal, names both implementations (state-local store vs `TestControllerBridge`), explains the decision rule
- Hedges on certification names ("Wire Conformance" / "Live Integration Verified") while the badge model in #1782 settles

## No SDK behavior change

The marker contract from #1786, the JSDoc trust-boundary section from #1787, and the construction-time dual-emit warn from #1788 all stay as-is — they describe mechanism without committing to a certification taxonomy. Only the framing prose around "what kind of seller this is for" shifts.

## Test plan

- [x] Doc-only change; `tsc --noEmit` clean
- [x] `prettier --check` clean
- [x] Patch-level changeset

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1782, opened 2026-05-16
Bridge leaderboard policy: how does a runner-visible bridge marker get consumed?


bokelley:
## Context

#1775 asks the SDK to emit a per-tool bridge-participation signal and the storyboard runner to surface it in the run record. That ships the marker — but **how the leaderboard / compliance dashboard uses the marker is a separate product decision** that isn't code.

After the bridge surface grew to 13 tools (#1753 + #1759 + #1761 + #1772), every storyboard pass that ran through fixture-merge reads identically to one that ran through a real adapter. The marker fixes the data; the policy decides what to show.

## The product question

Three plausible policies, each with different downstream effects:

1. **Separate dimension.** Score split into "wire conformance" (fixture-augmented passes count) and "adapter health" (only live-OAuth runs count). Most honest; highest signal; hardest to summarize on a leaderboard card.
2. **Downgrade.** Bridge-augmented passes count for less (e.g., 0.5×) toward the headline score. Single number, but the weighting is arbitrary and adopters will dispute it.
3. **Footnote.** Bridge-augmented passes count fully but the run record shows the asterisk. Lowest friction; highest risk of conflation in the headline score.

## Why this matters even with #1775 shipped

A marker without a consumer is half-shipped. The SDK can emit it perfectly and the runner can record it perfectly, but if the leaderboard / dashboard treats all green steps identically, the marker is invisible to buyers and procurement reading the score.

The wire-conformance-vs-adapter-health distinction has to live somewhere the audience actually reads. JSDoc disclaimers don't survive contact with a leaderboard.

## Ask

- [ ] Decide which policy applies (or define a fourth).
- [ ] Document the policy on the public compliance-scoring docs.
- [ ] Wire the consumer (leaderboard / dashboard / report renderer) to honor it.

## Background

Spun out from the post-merge expert review of (now-closed) #1754 — the four reviewers agreed the marker (#1775) was the highest-value cross-repo follow-up, but explicitly flagged "shipping the marker without wiring the consumer is half-done."

cc #1775 (cross-repo marker emission + runner surface)

bokelley:
## Recommendation: tiered certification + hybrid display (synthesis of options 1 and 3)

After the SDK side shipped in #1786 — and after a parallel four-expert review on that PR — the policy question is clearer. The three options as stated are real, but they're not exhaustive, and the right answer is a synthesis. Recording the analysis here so the decision is auditable.

### Critical read on the three named options

[...]

Issue #1674, opened 2026-05-11
Storyboard --json reporter serializes scenarios under multiple parent groups, creating false duplicate-execution signal


bokelley:
## Summary

\`storyboard run --json\` output contains the same scenario serialized multiple times — once under each parent grouping the scenario belongs to (e.g., a capability track + a scenario aggregation block). This creates the appearance of duplicate execution, leading triagers to false hypotheses about runner behavior.

## Concrete confusion

In a recent session debugging a seller-side issue, we grepped the JSON output and found two occurrences of \`media_buy_seller/invalid_transitions/setup\` with the SAME \`media_buy_id\`. We assumed:

1. The runner was running each scenario twice
2. Same idempotency key was being sent twice
3. Seller's idempotency cache was correctly replaying — therefore same id
4. \`$generate:uuid_v4#alias\` was bleeding across passes

We filed adcp-client#1657 (resulting in PR #1658) for what we thought was a real cross-phase alias-leak bug. That fix is defensible on its own merits, but it didn't change the observed behavior — because the underlying assumption (two executions) was wrong.

The two occurrences had:
- Same \`tested_at\` timestamp (\`2026-05-11T09:17:31.173Z\`)
- Same \`total_duration_ms\` (4659)
- Identical inner data

i.e., the reporter is serializing one scenario's result under two different parent groups, not running the scenario twice.

## Proposed fix

Either:

1. **Dedupe.** Serialize each scenario exactly once; reference it from secondary parent groups by ID rather than embedding the full result tree.
2. **Label clearly.** Add a \`grouping_context\` or \`view\` field on each duplicated scenario object so consumers can distinguish \"this is the canonical scenario record\" from \"this is its appearance in track X\".

(1) is cleaner for machine consumption. (2) is easier to ship without breaking existing consumers.

## Why this matters

The JSON output is the primary diagnostic surface when storyboards fail. Confusing serialization wasted multi-hour debugging cycles on at least one issue and produced a (defensible but unnecessary) upstream PR. Cleaner output → faster triage → fewer false-positive bug reports.

## Repro

\`\`\`bash
npx @adcp/sdk@6.19.0 storyboard run <agent> media_buy_seller --json > /tmp/run.json
grep -c '"scenario": "media_buy_seller/invalid_transitions/setup"' /tmp/run.json
# returns: 2 (but the scenario was only executed once)
\`\`\`

## Related

[...]

PR #1808, opened 2026-05-17
chore: URL-rot CI guard (#1790) + v6→v7 migration stub (#1494)

### LINKED ISSUE #1790: Mitigate docs-URL rot in createAdcpServer warning message (follow-up to #1784)
## Context

#1788 added a construction-time \`process.emitWarning\` when \`testController\` is registered without an account resolver. The message embeds a hardcoded link:

\`https://github.com/adcontextprotocol/adcp-client/blob/main/docs/guides/VALIDATE-YOUR-AGENT.md\`

Code review flagged this as silent-rot-prone: if the doc moves or is renamed, the warning still fires but the link 404s. The user sees a security signal with no path to context.

## Options

1. **Build-time link check.** A small script in \`scripts/\` that walks every hardcoded \`docs/guides/...\` URL in \`src/\` and asserts the file exists at the indicated path on \`main\`. Runs in CI. Cheap, catches the rot at the source.
2. **Versioned anchor.** Link to a tagged release (e.g. \`v7.4.0/docs/guides/...\`) instead of \`main\`. Survives renames on main but freezes the link to the version that shipped — adopters on newer versions get stale prose.
3. **Issue redirect.** Link to a stable issue (e.g. \`#1784\`) that aggregates the framing and itself links to current docs. Adopters get one redirect hop but the URL never rots.
4. **Don't link from runtime messages at all.** Put the URL in JSDoc only; warning message just names the JSDoc reference.

## Recommendation

Option 1 is the right defensive measure (catches rot proactively, no UX cost). Worth implementing once the next message lands that wants a URL — a single \`URL_ROT_CHECK\` script that scans for \`https://github.com/adcontextprotocol/adcp-client/blob/main/\` patterns and verifies each path against the local checkout.

## Scope

- [ ] Build-time URL-rot check in \`scripts/check-doc-links.ts\`
- [ ] Wire into CI (\`build\` or new \`lint:docs\` step)
- [ ] Apply to existing hardcoded URLs across \`src/\` and \`docs/\` if any exist

Not urgent — the current warning still works, only the docs link silently rots if changed. File while it's fresh.

cc #1784, #1788.

bokelley:
Bucket-B small-wins bundle from session 2026-05-17. No runtime behavior change.

## #1790 — SDK→docs URL-rot CI guard

`scripts/check-doc-links.ts` walks `src/`, `bin/`, `docs/`, `packages/` for `https://github.com/adcontextprotocol/adcp-client/blob/main/<path>` references and asserts each `<path>` resolves to an existing file. Wired into `.github/workflows/ci.yml` between the agent-docs check and the typecheck step (`npm run ci:doc-links`).

Catches the rot the `createAdcpServer` testController-without-resolver warning surfaced as a concern in #1784 — a doc rename on `main` would silently 404 the link in the warning without firing any signal. Now the rename has to update the reference, OR add the path to `EXEMPT_PATHS` with a one-line justification (default: empty set; the goal is zero exemptions over time).

Smoke-tested locally:
- Rename `examples/gcp-kms-signing-provider.ts` in the URL inside `docs/guides/SIGNING-GUIDE.md` → script flags the broken link with file:line.
- Restore → script passes.

## #1494 — v6→v7 migration guide stub

`docs/migration-6.x-to-7.x.md` — forward-looking stub for the v6→v7 migration covering `account.mode`-driven upstream URL routing. Marked **STUB** prominently so adopters know JS Phase 2 hasn't shipped.

Covers the five bullets from the issue:
1. Adapter shape requirement (per-request `ctx.upstream` is the source of truth for the URL outbound HTTP calls target).
2. Three migration shapes — resolver method, constructor injection per request, WeakMap middleware for vendor SDKs (GAM/FreeWheel/Kevel/Celtra) that bake URL into the constructor.
3. `complyTest:` block stays first-class for adopters who don't want to restructure.
4. `account.mode` persistence across async-task lifecycle (`tasks/get` polls, webhook emissions).
5. Self-grade checklist for when 7.0 ships.

Forward-references #1495 (mock-server fixture scaffolding) and #1647 (derived resolution mode rework) as companion tracks.

## #1674 — already closed

Triage check confirmed #1674 was already closed by PR #1792 (commit `2c718a54`) before this session — track-level `_view` marker landed there. The issue author accepted the partial fix; per-scenario `_view` work was prepared but reverted as out-of-scope when the close was discovered.

## #1782 — comment posted

#1782 needed a synthesis comment, not code. Posted [#issuecomment-4470335520](https://github.com/adcontextprotocol/adcp-client/issues/1782#issuecomment-4470335520) acknowledging the maintainer's single-dimension *Live Integration Verified* framing (already landed via PR #1795) and flagging the remaining work as cross-repo (spec docs + leaderboard wiring). SDK-side ask is satisfied.

## Test plan

- [x] `npm run ci:doc-links` passes against current `main` URLs
- [x] Smoke-tested URL-rot detection by transient rename
- [ ] CI green

Closes #1790, #1494.


bokelley:
## Expert review (2026-05-17)

Two parallel reviews (code-reviewer + docs-expert) ran against the initial commits. Convergent + should-fix findings addressed in commit \`8a7d79ee\`.

**Convergent finding (real bug — fixed):**

\`Shape C\` WeakMap pattern was misleading. \`HandlerContext\` is allocated fresh per tool-handler invocation (\`create-adcp-server.ts:3370\`), so a \`WeakMap<HandlerContext, …>\` cache had no reuse beyond a single method call — same allocation profile as Shape A. Re-keyed on \`ctx.upstream\` (URL string) so client reuse spans every request resolving to the same upstream. Honest description of when the cache actually helps now lives in the doc.

**Should-fix (applied):**

- URL-rot regex broadened to match both \`blob/main/\` and \`tree/main/\` (directory references were silently ignored).
- \`Not yet shipped\` banner replaces the soft STUB framing — adopters get an explicit "do not restructure adapters yet" callout.
- New tl;dr table (today vs 7.0 vs restructure with Shape A/B/C) at top of the migration guide, mirroring the search-replace convention in \`docs/migration-5.x-to-6.x.md\` and \`6.7-to-6.9.md\`.
- Shape headings name when-to-pick in the heading itself, so a skimmer can route without reading three code samples.
- Inline links to load-bearing source files (\`src/lib/server/create-adcp-server.ts\`, \`bin/adcp.js\`, the proposal's § Implementation status).
- Self-grade checklist gains a grep audit (\`grep -rEn 'baseUrl|BASE_URL|process\\.env\\.[A-Z_]*_URL' src/adapters/\`) so it's testable post-7.0, not aspirational.
- Phase 2 grounded inline as "the URL-routing rollout" — new readers don't need to chase the term.
- URL-rot guard's error message leads with the preferred fix (update the reference) and demotes EXEMPT_PATHS to last-resort with clear scope.
- CLAUDE.md critical-rules bullet added so contributors discover the guard before CI fails.

**Nits accepted as-is:**

- Empty changeset frontmatter — precedent in \`.changeset/sour-rice-stare.md\` from PR #1802 (also docs-only).
- \`EXEMPT_PATHS\` starts empty by design — adding the escape hatch only when a real published-site-only case appears is the right discipline.
- Pinned-commit URL forms (\`blob/<sha>/...\`) intentionally NOT matched — the SHA pins the content.

Issue #1805, opened 2026-05-17
Implement force_scenario_unsupported runner detection (AdCP 3.0.12 runner-output-contract)


bokelley:
## Context

Surfaced by the ad-tech-protocol-expert pass on PR #1802 (the AdCP 3.0.11 → 3.0.12 bump). The new \`runner-output-contract.yaml\` adds \`force_scenario_unsupported\` as a normative \`not_applicable\` detail string that runners MUST detect.

## What the spec requires

\`compliance/cache/3.0.12/universal/runner-output-contract.yaml:334-346\`:

> \`force_scenario_unsupported\` — a \`force_*\` controller scenario in an explicit storyboard phase returned UNKNOWN_SCENARIO. Runners MUST detect the tuple (comply_test_controller present, resolved payload scenario begins with \`force_\`, response \`{success: false, error: UNKNOWN_SCENARIO}\`) and grade not_applicable before applying the step's declared validations; encode \`force_scenario_unsupported\` in \`skip_result.detail\`.

## Current SDK state

\`grep -rn force_scenario_unsupported src/lib/\` returns no matches in the runner code path. The detection logic doesn't exist. Storyboards that target a \`force_*\` scenario against an agent that didn't implement it will fail (with the agent's UNKNOWN_SCENARIO response cascading into a validation failure) rather than gracefully skip with \`not_applicable\`.

## Ask

Add the tuple detection in the storyboard runner's per-step grader (\`src/lib/testing/storyboard/runner.ts\`):

1. After a \`comply_test_controller\` step returns, check \`response.error === 'UNKNOWN_SCENARIO'\` AND the resolved payload's \`scenario\` field starts with \`force_\`.
2. If matched, set \`skip_result\` with \`reason: not_applicable\` and \`detail: force_scenario_unsupported\`.
3. Skip the step's declared validations (they assume the force_* path ran).

## Acceptance

- Detection wired into the per-step grader.
- Storyboard fixture in \`test/lib/storyboard-*.test.js\` exercises an agent that returns \`{success: false, error: 'UNKNOWN_SCENARIO'}\` for a \`force_*\` scenario and asserts \`skip_result.reason === 'not_applicable'\` + \`detail === 'force_scenario_unsupported'\`.
- Compliance report renders the skip with the new detail string.

## Refs

- AdCP 3.0.12 \`runner-output-contract.yaml\` (cached at \`compliance/cache/latest/universal/\`)
[...]

PR #1812, opened 2026-05-17
fix: A2A discovery size-limit (#1804) + force_scenario_unsupported runner detection (#1805)

### LINKED ISSUE #1804: Wrap fetchA2ACanonicalUrl + other A2A discovery sites in withResponseSizeLimit
## Context

Surfaced by the security-reviewer pass on PR #1802 (which closes #1799 — wrapping \`getAgentInfo()\`). The PR scope was the discovery path inside \`getAgentInfo\`; the security review flagged a separate A2A discovery site that still bypasses \`transport.maxResponseBytes\`.

## Location

\`src/lib/core/SingleAgentClient.ts:567\` — \`fetchA2ACanonicalUrl\` (used via \`ensureEndpointDiscovered\`) calls native \`fetch(url, …)\` directly without composing through \`wrapFetchWithSizeLimit\`. Every \`executeTask\` / \`listTools\` / \`getAgentInfo\` call that goes through endpoint discovery on a not-yet-discovered A2A agent fires an uncapped fetch against the agent card.

## Why this matters

A hostile vendor can ship a 5 GB agent-card body on first discovery to blow memory before any application-layer parsing runs. The DoS surface is identical to the one #1799 closed for \`getAgentInfo\`'s explicit A2A branch — same threat, different call site.

## Ask

Wrap \`fetchA2ACanonicalUrl\` (and any other discovery fetch reachable from \`ensureEndpointDiscovered\`) through \`wrapFetchWithSizeLimit\` so the active ALS slot enforces the cap. Pattern is the same as the \`sizeLimitedFetch\` wrapper introduced in PR #1802 for the A2A branch of \`getAgentInfo\`.

## Acceptance

- \`fetchA2ACanonicalUrl\` composes through \`wrapFetchWithSizeLimit\`.
- New unit test in \`test/unit/get-agent-info-size-limit.test.js\` (or a sibling) covers an oversized agent card returned during \`ensureEndpointDiscovered\` — asserts \`ResponseTooLargeError\`.
- Audit any other native-\`fetch\` call in \`SingleAgentClient.ts\` that handles untrusted responses; wrap or document why exemption is safe.

## Refs

- PR #1802 — scoped close of #1799 (the \`getAgentInfo\` branch)
- adcontextprotocol/adcp-client#1799 — original ask
- \`src/lib/protocols/responseSizeLimit.ts\` — \`wrapFetchWithSizeLimit\` helper

bokelley:
Two PR #1802 follow-ups bundled. Releasing as one PR keeps the changelog tidy ahead of the next release.

## #1804 — `fetchA2ACanonicalUrl` honors `transport.maxResponseBytes`

`fetchA2ACanonicalUrl` (`src/lib/core/SingleAgentClient.ts:550`) runs implicitly before every `executeTask` / `listTools` / `getAgentInfo` call against an A2A agent whose canonical URL hasn't been resolved yet. Until this fix it called native \`fetch\` directly without composing through \`wrapFetchWithSizeLimit\`, so a hostile vendor could ship a 5 GB agent-card body on first discovery to blow memory before any application-layer parsing ran.

Same shape as the fix landed for #1799:
- Compose the auth-stamping \`fetchImpl\` through \`wrapFetchWithSizeLimit\` so the active ALS slot enforces the cap on the wire call.
- Wrap the \`A2AClient.fromCardUrl\` call in \`withResponseSizeLimit\`.
- Wrap the deferred \`agentCardPromise\` await in \`withResponseSizeLimit\` too.

Regression test at \`test/unit/fetch-a2a-canonical-url-size-limit.test.js\`.

**Out-of-scope sweep findings:** at least 4 other native-\`fetch\` sites in \`src/lib/\` handle untrusted upstream responses without going through the size cap (\`auth-introspection.ts:330\`, \`comply.ts:1261\`, \`registry/index.ts:802/812\`, \`a2a.ts:220\`). Intentionally not touched here — they each warrant their own scoped issue.

## #1805 — `force_scenario_unsupported` runner detection (AdCP 3.0.12)

AdCP 3.0.12 \`runner-output-contract.yaml\` introduces \`force_scenario_unsupported\` as a normative \`not_applicable\` detail string runners MUST emit. When a \`comply_test_controller\` step targets a \`force_*\` scenario that the agent advertises the controller for but does not implement (response \`{success: false, error: 'UNKNOWN_SCENARIO'}\`), the runner grades the step \`not_applicable\` with detail \`force_scenario_unsupported\` BEFORE applying the step's authored validations.

Surfaced by the ad-tech-protocol-expert pass on PR #1802 (the 3.0.11 → 3.0.12 bump). Without this, the failing \`success: true\` check would mask the coverage gap as a real agent fault.

- New \`RunnerDetailedSkipReason\` variant \`force_scenario_unsupported\`, mapped onto canonical \`not_applicable\` via \`DETAILED_SKIP_TO_CANONICAL\`.
- Detection in \`src/lib/testing/storyboard/runner.ts\` before the early-exit unsupported-tool skip path, keyed on the tuple \`(step.task === 'comply_test_controller', resolved scenario starts with 'force_', response.success === false, response.error === 'UNKNOWN_SCENARIO')\`.
- Companion to \`fixture_seed_unsupported\` (seeding.ts) — same shape but for \`force_*\` in step phases, not \`seed_*\` in the fixtures phase.

Test at \`test/lib/storyboard-force-scenario-unsupported.test.js\` covers happy path, two negative cases (non-force_* UNKNOWN_SCENARIO does NOT trigger; force_* that succeeds grades normally), and the canonical-skip mapping.

## Test plan

- [x] \`npm run format:check\` clean
- [x] \`npm run build\` clean
- [x] All size-limit suites (34 tests) pass
- [x] Storyboard runner suites (200 tests across default-invariants, controller-seeding, deterministic-scenarios, runner-output-contract-v2-runner) pass
- [ ] CI green

Closes #1804, #1805.

### DIFF
--- a/.changeset/itchy-eggs-guess.md
+++ b/.changeset/itchy-eggs-guess.md
@@ -1,0 +1,13 @@
+---
+'@adcp/sdk': patch
+---
+
+fix(client): `SingleAgentClient.fetchA2ACanonicalUrl()` now honors `transport.maxResponseBytes`
+
+Closes the second A2A discovery DoS surface flagged by the security review on PR #1802. `fetchA2ACanonicalUrl` runs implicitly before every `executeTask` / `listTools` / `getAgentInfo` call against an A2A agent whose canonical URL hasn't been resolved yet. Until this fix it called native `fetch` without composing through `wrapFetchWithSizeLimit`, so a hostile vendor could ship a 5 GB agent-card body on first discovery to blow memory before any application-layer parsing ran.
+
+Same fix shape as #1799 (`getAgentInfo`): wrap the `A2AClient.fromCardUrl` call and the deferred `agentCardPromise` read in `withResponseSizeLimit`, and route the auth-stamping `fetchImpl` through `wrapFetchWithSizeLimit` so the active ALS slot fires on the wire call.
+
+Regression test at `test/unit/fetch-a2a-canonical-url-size-limit.test.js` aborts an oversized canonical-URL discovery with `ResponseTooLargeError`.
+
+Closes #1804.
--- a/.changeset/swift-mice-fetch.md
+++ b/.changeset/swift-mice-fetch.md
@@ -1,0 +1,13 @@
+---
+'@adcp/sdk': minor
+---
+
+feat(testing): runner detects `force_scenario_unsupported` per AdCP 3.0.12 runner-output-contract
+
+When a `comply_test_controller` step targets a `force_*` scenario that the agent advertises the controller for but does not implement (response `{success: false, error: 'UNKNOWN_SCENARIO'}`), the runner now grades the step `not_applicable` with detail `force_scenario_unsupported` BEFORE applying the step's authored validations. Previously the step failed its declared `success: true` check and the coverage gap looked like a real agent fault.
+
+- New `RunnerDetailedSkipReason` variant `force_scenario_unsupported` (`src/lib/testing/storyboard/types.ts`), mapped onto canonical `not_applicable` via `DETAILED_SKIP_TO_CANONICAL`.
+- Detection in `src/lib/testing/storyboard/runner.ts` before the early-exit unsupported-tool skip path, keyed on the tuple `(step.task === 'comply_test_controller', resolved scenario starts with 'force_', response.success === false, response.error === 'UNKNOWN_SCENARIO')`.
+- Companion to `fixture_seed_unsupported` (seeding.ts) — same shape but for `force_*` scenarios in step phases, not `seed_*` scenarios in the fixtures phase.
+
+Spec source: `compliance/cache/3.0.12/universal/runner-output-contract.yaml` > `skip_result.reasons.force_scenario_unsupported`. Closes #1805.
--- a/src/lib/core/SingleAgentClient.ts
+++ b/src/lib/core/SingleAgentClient.ts
@@ -551,6 +551,15 @@ export class SingleAgentClient {
     const clientModule = require('@a2a-js/sdk/client');
     const A2AClient = clientModule.A2AClient;
 
+    // adcp-client#1804 — wrap A2A card discovery in withResponseSizeLimit so
+    // `transport.maxResponseBytes` applies to every agent-card fetch. The
+    // auth-stamping fetchImpl composes through wrapFetchWithSizeLimit so the
+    // active ALS slot enforces the cap on the wire call. Matches the same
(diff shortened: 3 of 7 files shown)


bokelley:
## Expert review (2026-05-17)

Two parallel reviews (code-reviewer + ad-tech-protocol-expert) ran. One real blocker + one info-level clarification addressed in commit \`8c98b0d3\`.

**Blocker (fixed): #1804 test was mis-targeted.**

The original test called \`client.getAgentInfo()\`, which routes through \`SingleAgentClient\`'s inline A2A discovery branch at line 2755–2798 (already wrapped by PR #1802 for #1799), NOT through the newly-wrapped \`fetchA2ACanonicalUrl\` path. The fix was correct but the test would have passed on \`main\` (pre-PR) because the path it exercised was already wrapped.

\`fetchA2ACanonicalUrl\` is reached via \`ensureCanonicalUrlResolved\`, which the public \`resolveCanonicalUrl()\` method drives. Test now calls \`client.resolveCanonicalUrl()\` in both cases — the only public surface that exercises the new wrap.

**Info (clarified): protocol-expert Q1 — "controller advertised" gate semantics.**

The spec's "controller present" gate is enforced upstream by the phase cascade at the \`seedingMissingController\` check (\`runner.ts\` ~line 1893): by the time per-step grading reaches the detector, the controller has already been confirmed advertised. The detector's \`step.task === 'comply_test_controller'\` gate is the per-step subset that prevents the skip from bleeding into other tools that happen to carry a \`scenario\` argument. Added a comment block at the detector clarifying this ordering so future maintainers don't read the detector in isolation.

**Accepted as-is (info / out of scope):**

- Q4 (force_* malformed payload → seller incorrectly emits UNKNOWN_SCENARIO): not the runner's job to disambiguate; the contract is "agent returns UNKNOWN_SCENARIO ⇒ coverage gap." Seller's responsibility to emit the right error code.
- Q5 (schema-driven test parsing the YAML reasons section to assert every documented detailed reason exists in \`DETAILED_SKIP_TO_CANONICAL\`): nice-to-have but out of scope for this PR — would be its own infra ticket.
- Dynamic \`import\` inside hot path: matches existing convention in the file (the #1799 PR did the same); not worth diverging here.
- Edge-case test for \`force_*\` with \`taskResult\` undefined (transport failure): the \`controllerData?.success === false\` narrowing already short-circuits cleanly; no behavior to assert that wouldn't be tautological.

PR #1823, opened 2026-05-17
feat(signing): signResponse for RFC 9421 response signing (#1822)

### LINKED ISSUE #1822: SDK gap: outbound response signing (signResponse) for RFC 9421
## Gap

`@adcp/sdk` (currently 7.4.0) ships RFC 9421 helpers for every direction *except* server-side outbound response signing:

| Direction | Helper | Exists? |
|---|---|---|
| Client → server, signed request | `signRequest()` + `prepareRequestSignature` / `finalizeRequestSignature` | ✅ `dist/lib/signing/signer.d.ts` |
| Server → webhook receiver, signed delivery | `signWebhook()` + `prepareWebhookSignature` | ✅ same file |
| Inbound verification (server side, request) | `verifyRequestSignature()` | ✅ `dist/lib/signing/verifier.d.ts` |
| Inbound verification (webhook receiver) | `verifyWebhookSignature()` | ✅ `dist/lib/signing/webhook-verifier.d.ts` |
| **Server → client, signed response** | `signResponse()` | ❌ **missing** |

RFC 9421 §2.2.9 (`@status`) covers responses with the same canonicalization machinery as requests — drop `@method` / `@target-uri` / `@request-target` / `@path` / `@query`, add `@status`, keep everything else (`@authority`, `content-digest`, etc.). The algorithm and canonicalization are identical; the gap is purely API ergonomics.

## Concrete use case

Tracked in [adcontextprotocol/adcp #4576 — seller-verification walkthrough](https://github.com/adcontextprotocol/adcp/issues/4576) and follow-up work to the training agent. The buyer-side verification walkthrough at `docs/verification/overview` opens with:

> The `get_products` response carries an RFC 9421 `Signature` and `Signature-Input` header. The `keyid` parameter points at a JWK in Northwind's published JWKS. Sam's client verifies the signature before parsing the body.

For this to demo against the training agent (sales tenant emitting signed `get_products` responses), we need ~30 LOC of helper-call. Without `signResponse()` in the SDK, every implementor reinvents 200+ LOC of canonicalization + signature-base composition + header serialization. That's the gap.

## Proposed API

Mirror the existing `signRequest` / `signWebhook` shape exactly. Two-tier (high-level + prepare/finalize for KMS async signers) the same way the existing helpers do:

```ts
export interface ResponseLike {
  status: number;
  headers: Headers | Record<string, string>;
  body?: BodyInit | null;
  // The originating request URI — needed because RFC 9421 §2.2.9 binds the
  // response signature to its request context via @authority and (optionally)
  // @target-uri components from the original request.
  request: { method: string; url: string };
}

export interface SignResponseOptions {
  componentIds?: string[];           // default: ["@status", "content-digest", "@authority"]
  created?: number;                  // default: now
  expires?: number;                  // optional
  nonce?: string;                    // optional
  tag?: string;                      // default: "adcp/response-signing/v1" or similar — naming TBD
  alg?: string;
}

export interface SignedResponse {
  status: number;
  headers: Headers;  // includes Signature, Signature-Input, Content-Digest
}

export declare function prepareResponseSignature(
  response: ResponseLike,
  identity: SignatureIdentity,
  options?: SignResponseOptions,
): PreparedRequestSignature;

export declare function finalizeResponseSignature(
  prepared: PreparedRequestSignature,
  signature: Uint8Array,
): SignedResponse;

export declare function signResponse(
  response: ResponseLike,
  key: SignerKey,
  options?: SignResponseOptions,
): SignedResponse;
```

Open design questions worth raising in the PR (not blockers for the issue):

1. **Tag value.** `signRequest` uses `adcp/request-signing/v1`; `signWebhook` uses `adcp/webhook-signing/v1`. Suggest `adcp/response-signing/v1` for symmetry. Confirm against the JWK `adcp_use` invariant we already enforce on the inbound verifier side.
2. **Default component set.** Minimal viable set is `["@status", "content-digest", "@authority"]`. `@authority` is preserved from request context. RFC 9421 doesn't require `@target-uri` for responses, but consumers may want it to bind to a specific request URI rather than just origin. Default vs opt-in worth a sentence in the design.
3. **Async / KMS path.** `signer-async.ts` already exists for KMS-backed signing — `signResponseAsync()` plus the same prepare/finalize split should fall out naturally.
4. **Verifier symmetry.** This issue is signer-side. A corresponding `verifyResponseSignature()` helper on the verifier side is a separate (smaller) follow-up — most consumers are servers verifying inbound requests, not clients verifying responses, so demand is lower. Worth filing but lower priority than the signer.

## Out of scope for this issue

- A new wire protocol — this is just exposing existing RFC 9421 capability through SDK ergonomics.
- Mandating response signing at the spec level — that's a separate AdCP-spec discussion ([#2307](https://github.com/adcontextprotocol/adcp/issues/2307) covers mandatory mutating-request signing for 4.0).
- HTTP transport integration patterns (Express middleware, MCP transport wrappers) — that's caller code, downstream of this primitive.

## Acceptance

- `signResponse(response, key, options)` exported from the signing module surface, callable in tests with realistic Ed25519 fixtures.
- Async path (`signResponseAsync`) wired through the same `SignerProvider` interface as `signRequestAsync` / `signWebhookAsync`.
- Round-trip test: sign a synthetic response, verify the resulting `Signature` + `Signature-Input` headers against the same JWK using whichever verifier surface we add (could land paired with `verifyResponseSignature` or stand alone with an inline test verifier).

bokelley:
## Summary

Closes #1822. Adds the missing direction of the RFC 9421 signing surface:
server-emitted **signed responses**. `signRequest` / `signWebhook` / `signResponse`
now share the same prepare/finalize shape and `SigningProvider` async path.

| Direction | Helper | Status |
|---|---|---|
| Buyer → server signed request | `signRequest` / `signRequestAsync` | already shipped |
| Server → webhook receiver | `signWebhook` / `signWebhookAsync` | already shipped |
| **Server → buyer signed response** | **`signResponse` / `signResponseAsync`** | **this PR** |

### API

```ts
import { signResponse, type ResponseLike } from '@adcp/sdk/signing/client';

const response: ResponseLike = {
  status: 200,
  headers: { 'content-type': 'application/json' },
  body: JSON.stringify({ products: [...] }),
  // Originating request — needed because RFC 9421 §2.2 binds response
  // signatures to their request context via @authority.
  request: { method: 'POST', url: 'https://seller.example.com/adcp/get_products' },
};

const signed = signResponse(response, key);
// signed.headers now carries Signature, Signature-Input, Content-Digest.
```

### What's covered by default

`['@status', '@authority']` from `RESPONSE_MANDATORY_COMPONENTS`, plus
`content-type` + `content-digest` automatically when the response has a
body. Same conditional shape as `signRequest`. Callers that want to bind
the signature to a specific request URL can opt-in via
`additionalComponents: ['@target-uri']`.

### Async / KMS path

`signResponseAsync(response, provider)` delegates the crypto to the
existing `SigningProvider` interface, the same way `signRequestAsync`
and `signWebhookAsync` do. The prepare/finalize split is exported for
callers that need to hand the canonical base to an external signer.

### Design notes

- **Tag**: `adcp/response-signing/v1` (parallel to existing
  `adcp/request-signing/v1` and `adcp/webhook-signing/v1`).
- **`AdcpUse`**: extended with `'response-signing'` so signer-side JWK
  metadata can declare the binding now. The matching verifier helper
  (`verifyResponseSignature`) is a separate follow-up — most consumers
  are servers verifying inbound requests, not clients verifying
  responses, so demand is lower; round-trip tests in this PR inline a
  test verifier rather than waiting on the public helper.
- **`@status` discipline**: rejected in request-signing canonicalization
  with a clear error rather than falling through to silent "component
  missing." Prevents misuse of `buildSignatureBase` with response
  components.

### Out of scope (deferred follow-ups)

- `verifyResponseSignature()` public helper — separate issue.
- Express middleware wrapper that auto-signs handler responses — caller
  code, downstream of this primitive.
- AdCP spec text mandating response signing for any specific operation
  — protocol-level discussion, separate from SDK ergonomics.

## Test plan

- [x] `node --test test/response-signing.test.js` — 10/10 passing
  - default covered components when body present
  - empty body omits content-digest / content-type
  - `Content-Digest` matches body
  - `additionalComponents` extends covered set
  - Ed25519 round-trip verifies
  - ECDSA-P256 round-trip verifies
  - body tampering invalidates signature
  - status tampering invalidates signature
  - `signResponseAsync` is byte-identical to `signResponse` for Ed25519
  - prepare/finalize split produces identical headers to one-shot
- [x] Existing signing tests unaffected (request-signing-provider,
  request-signing-vectors, webhook-signature-verification, webhook
  vectors — 143/143 passing)

### DIFF
--- a/.changeset/response-signing-1822.md
+++ b/.changeset/response-signing-1822.md
@@ -36,3 +36,10 @@ callers that need to hand the canonical base to an external signer.
 Tag: `adcp/response-signing/v1`. `AdcpUse` extended with `'response-signing'`
 so JWK metadata can declare the binding now — the matching verifier helper
 (`verifyResponseSignature`) is a separate follow-up.
+
+**Note on `AdcpUse` widening:** the `AdcpUse` union now includes
+`'response-signing'`. Consumers with exhaustive `switch (use)` blocks
+backed by `never` checks will need to add a `case 'response-signing':`
+arm. Pragmatically a minor bump — the runtime surface is purely additive
+and no existing call site changes — but exhaustive narrowers must opt-in
+to handle the new member.
--- a/.changeset/response-signing-1822.md
+++ b/.changeset/response-signing-1822.md
@@ -23,19 +23,34 @@ const signed = signResponse(response, key);
 // signed.headers now carries Signature, Signature-Input, Content-Digest.
 ```
 
-Covered components default to `['@status', '@authority']`, plus `content-type`
-+ `content-digest` automatically when the response carries a body — same
-conditional shape as `signRequest`. Callers that want to bind to a specific
-request URL can opt-in to `@target-uri` via `additionalComponents`.
+Covered components default to `['@status', '@authority', '@target-uri']`,
+plus `content-type` + `content-digest` automatically when the response
+carries a body. `@target-uri` is in the defaults (not opt-in) so a
+multi-tenant seller can't emit signatures interchangeable across endpoints
+sharing the same authority — matches RFC 9421 §B.2.5 examples. Callers
+that need to extend the covered set further (`@method`, custom headers)
+opt in via `additionalComponents`.
+
+**Asymmetry vs `signRequest`.** Response-signing defaults `coverContentDigest`
+to `true` when the response has a body (opt-out); request-signing requires
+an explicit `coverContentDigest: true` (opt-in). The asymmetry is deliberate:
+an unbound response body is the most common cross-purpose footgun for
+response signing (attacker can swap payload but keep headers + signature).
+Webhook signing also forces content-digest unconditionally for the same
+reason.
 
 Async + KMS-shaped path: `signResponseAsync(response, provider)` delegates the
 crypto to the existing `SigningProvider` interface; the prepare/finalize split
 (`prepareResponseSignature` / `finalizeResponseSignature`) is exported for
 callers that need to hand the canonical base to an external signer.
 
-Tag: `adcp/response-signing/v1`. `AdcpUse` extended with `'response-signing'`
-so JWK metadata can declare the binding now — the matching verifier helper
-(`verifyResponseSignature`) is a separate follow-up.
+Tag: `adcp/response-signing/v1`. **Wire-format contract is provisional**
+until `verifyResponseSignature` (follow-up) lands and is exercised against
+external SDK implementations. The `v1` suffix gives a clean break path —
+any breaking change ships as `v2` and verifiers reject `v1`. Adopters
+shipping signed responses today should pin a major SDK version. `AdcpUse`
+extended with `'response-signing'` so JWK metadata can declare the binding
+now.
 
 **Note on `AdcpUse` widening:** the `AdcpUse` union now includes
 `'response-signing'`. Consumers with exhaustive `switch (use)` blocks
--- a/.changeset/response-signing-1822.md
+++ b/.changeset/response-signing-1822.md
@@ -43,3 +43,28 @@ backed by `never` checks will need to add a `case 'response-signing':`
 arm. Pragmatically a minor bump — the runtime surface is purely additive
 and no existing call site changes — but exhaustive narrowers must opt-in
 to handle the new member.
+
+**Signer-side `adcp_use` purpose binding (closes #1825).** All three sync
+signer entry points now refuse keys whose `adcp_use` doesn't match the
+helper:
+
+- `signRequest` requires `adcp_use: 'request-signing'`
+- `signWebhook` requires `adcp_use: 'webhook-signing'`
+- `signResponse` requires `adcp_use: 'response-signing'`
+
+Mismatch (including a missing `adcp_use`) throws at the signer with the
+same error code the verifier raises at step 8 (`*_signature_key_purpose_invalid`),
+so misconfiguration surfaces at configuration time rather than at the
+receiver. Production callers using `pemToAdcpJwk({ adcp_use: ... })` to
+mint keys are already correct by construction; anyone reusing a key
+across purposes will get a clear remediation message.
+
+Test-vector authors that need to *deliberately* sign with a wrong-purpose
+key (e.g. AdCP negative-vector 009 cross-purpose rejection) can compose
+the lower-level `prepareRequestSignature` + `finalizeRequestSignature`
+helpers directly — those take a `SignatureIdentity` and skip the gate
+because they're designed for KMS-shaped async paths where purpose
+binding happens via the `SigningProvider`. The internal storyboard
+builder uses this composition pattern; see
+`src/lib/testing/storyboard/request-signing/builder.ts` for the
+canonical shape.
--- a/.changeset/response-signing-1822.md
+++ b/.changeset/response-signing-1822.md
@@ -1,0 +1,38 @@
+---
+'@adcp/sdk': minor
+---
+
+signing: add `signResponse` for RFC 9421 §2.2.9 response signing (#1822)
+
+Rounds out the signing surface — `signRequest` (buyer→server), `signWebhook`
+(server→receiver), and now `signResponse` (server→buyer) all share the same
+prepare/finalize shape and `SigningProvider` async path.
+
+```ts
+import { signResponse, type ResponseLike } from '@adcp/sdk/signing/client';
+
+const response: ResponseLike = {
+  status: 200,
+  headers: { 'content-type': 'application/json' },
+  body: JSON.stringify({ products: [...] }),
+  // Originating request — needed to bind @authority back to its origin.
+  request: { method: 'POST', url: 'https://seller.example.com/adcp/get_products' },
+};
+
+const signed = signResponse(response, key);
+// signed.headers now carries Signature, Signature-Input, Content-Digest.
+```
+
+Covered components default to `['@status', '@authority']`, plus `content-type`
++ `content-digest` automatically when the response carries a body — same
+conditional shape as `signRequest`. Callers that want to bind to a specific
+request URL can opt-in to `@target-uri` via `additionalComponents`.
+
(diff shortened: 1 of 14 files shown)


bokelley:
Addressed three review fixes in 3c39660d:

1. **`ResponseLike.request.url` absolute-URL contract** — JSDoc now states the requirement and shows Express / Lambda / Workers reconstruction (`\${proto}://\${host}\${path}`).
2. **Header-interop example** — `signResponse` JSDoc shows Express `res.set()`, Fetch `new Response()`, and Node `http.ServerResponse.writeHead()` adapter patterns.
3. **`AdcpUse` widening note** — changeset now flags that exhaustive `switch (use)` narrowers backed by `never` checks need a `case 'response-signing':` arm.

Follow-up debt filed as separate issues, not blocking this PR:

- #1825 — signer-side `adcp_use` purpose-binding enforcement (a `'webhook-signing'` key passed to `signResponse` currently produces a signature that the verifier will later reject; should refuse at the signer with the same error code).
- #1826 — `verifyResponseSignature` public helper. Until it ships, `signResponse` is round-trip-proven only via an inline test verifier; no external party can verify the wire format without reimplementing canonicalization. Should land before any production consumer ships signed responses.

`npm run format:check` and `tsc --noEmit` clean. Existing signing tests unchanged.

bokelley:
Bundled #1825 into this PR via commit d2ce687d — signer-side \`adcp_use\` purpose-binding gate. All three sync entry points (\`signRequest\`, \`signWebhook\`, \`signResponse\`) now refuse keys whose \`adcp_use\` doesn't match the helper, with the same error codes the verifier raises at step 8. Surfaces misconfiguration at the signer rather than at the verifier.

- New \`ResponseSignatureError\` + \`ResponseSignatureErrorCode\` parallel to existing request / webhook error shapes.
- Storyboard negative-vector builder switched to \`prepareRequestSignature\` + manual \`produceSignature\` so it can keep signing with wrong-purpose keys for the verifier-side cross-purpose rejection vector (#009). The KMS-shaped prepare/finalize split is the natural bypass for test-vector authors who legitimately need to produce adversarial signatures.
- One pre-existing test bug fixed (\`request-signing-provider\` was reusing a request-signing vector key for the \`signWebhookAsync\` equivalence check; gate caught it).
- Changeset updated with the new gate semantics and the bypass pattern.

Closes #1825. Full signing suite (399 tests) is at 395 pass / 4 fail; the 4 are pre-existing flakes on \`auth-signature-compose.test.js\` (verified on base \`3c39660d\` before this change), unrelated to the gate.

bokelley:
Three parallel expert reviews back (protocol / security / code-reviewer). Convergent findings addressed in commit 2f8582d6:

**Pre-merge fixes shipped:**
1. \`RESPONSE_MANDATORY_COMPONENTS\` now defaults to \`[@status, @authority, @target-uri]\` (was \`[@status, @authority]\`). Multi-tenant sellers can't emit endpoint-interchangeable signatures on the same authority; matches RFC 9421 §B.2.5 examples. Wire-contract decision locked.
2. \`coverContentDigest\` JSDoc + changeset call out the deliberate asymmetry vs \`signRequest\` (response defaults opt-out, request defaults opt-in — an unbound response body is the cross-purpose footgun).
3. **Non-circular wire-format fixture test** pins the signature base + Signature-Input header byte-for-byte against an RFC 9421 §2.5 prose construction. Refactors to \`buildResponseSignatureBase\` that break wire-compat will fail this test even if round-trips stay green.
4. \`assertKeyPurpose\` switch gets exhaustive \`default: throw\` with \`_exhaustive: never\` — a future \`AdcpUse\` widening trips \`tsc --noEmit\` rather than silently falling through.
5. \`prepare*\` JSDoc explicitly documents the no-gate semantics so adopters composing \`prepare* + own-signer\` know they need their own purpose check.
6. \`ResponseLike.request.url\` JSDoc warns about \`Host\` / \`X-Forwarded-*\` trust under proxy termination (Express \`req.get('host')\` is attacker-controllable absent allowlist + \`trust proxy\`).
7. \`RESPONSE_SIGNING_TAG\` JSDoc + changeset mark the v1 wire format as provisional until #1826 lands and is exercised cross-SDK; clean v2 break path if needed.
8. Test asymmetry: \`additionalComponents\` test split into \"@method adds\" + \"@target-uri is idempotent in defaults\" cases so the dedup logic is explicit.

**Follow-up issues filed (not blocking):**
- #1827 — \`SigningProvider.adcpUse\` extension to close the async-path purpose-binding gap (security + protocol both flagged independently; KMS-backed providers currently get no sync-side gate)
- #1828 — \`requestContextFromExpress\` / \`-Fetch\` / \`-Lambda\` helpers to construct \`ResponseLike.request\` safely (security + protocol both flagged the proxy-trust footgun; library can warn but can't enforce)
- #1829 — Spec PR in \`adcontextprotocol/adcp\` to reserve \`adcp/response-signing/v1\` and \`adcp_use: 'response-signing'\` in \`security.mdx\` (decoupled from the broader spec-mandate question)

All 317 signing tests pass (request + webhook + response + brand-jwks). \`tsc --noEmit\` and \`prettier --check\` clean.

Three expert summaries:
- **Protocol**: \"Cross-SDK interop risk moderate. Two SDKs implementing §2.2.9 with the same tag will produce byte-identical bases for @status and header components. The differentials live in [tag string choice, default component set, @authority under proxy] — none crypto-level bugs.\" → addressed defaults + tag + proxy warnings; filed spec PR.
- **Security**: \"Bottom line: merge. The signer-side gate is a real improvement, the canonicalization is conservative, and the bypass is documented at the test site.\" → addressed JSDoc + filed follow-ups.
- **Code reviewer**: Caught the \`coverContentDigest\` asymmetry doc bug and the partially-circular round-trip → both fixed (asymmetry callout + fixture test).

PR #1831, opened 2026-05-17
feat(storyboard): add `array_length` check kind for cardinality assertions (#1830)

### LINKED ISSUE #1830: compliance runner: add array_length check kind for cardinality assertions
Follow-up from [adcontextprotocol/adcp#4685](https://github.com/adcontextprotocol/adcp/pull/4685) protocol review.

## Problem

The compliance runner exposes these check kinds today:
- `response_schema`, `field_present`, `field_value`, `field_value_or_absent`, `error_code`, `upstream_traffic`

None of them directly assert "the array at this path has exactly N entries." Cardinality scenarios (e.g., the new `media_buy_seller/dependency_impairment_cardinality`) work around this by checking `field_present path: arr[N-1]` paired with `field_value_or_absent path: arr[N] value: null` — but `value: null` is unsound: it passes when the field is absent OR equals `null`, so a seller emitting a literal-null pad (`arr[N] = null`) silently passes the cardinality assertion.

The cardinality scenario fell back to relying on the cross-resource invariant (over-emission produces an impairment that references a non-offline resource, failing the forward rule). That works but is a roundabout enforcement path.

## Proposal

Add an `array_length` check kind:

```yaml
- check: array_length
  path: "media_buys[0].impairments"
  value: 2
  description: "Exactly two impairments — one per rejected creative"
```

Or as a comparison shape:

```yaml
- check: array_length
  path: "media_buys[0].impairments"
  min: 1
  max: 1
  description: "Exactly one impairment"
```

Cardinality scenarios then assert exact counts directly. The fallback to invariant enforcement remains as defense-in-depth.

## Acceptance

- [ ] New `array_length` check kind defined in storyboard-schema.yaml (spec repo) and implemented in the runner
- [ ] Test vectors for the new check kind
- [ ] adcp#4685's cardinality scenario migrates to direct array_length assertions in a follow-up PR
- [ ] Changeset (@adcp/sdk minor bump)

## Related

- [adcontextprotocol/adcp#4685](https://github.com/adcontextprotocol/adcp/pull/4685) — cardinality scenario
- [adcontextprotocol/adcp#4681](https://github.com/adcontextprotocol/adcp/issues/4681) — cardinality issue

bokelley:
## Summary

Closes #1830. Adds an `array_length` check kind to the storyboard runner so cardinality scenarios can assert "exactly N entries" directly.

The previous workaround — `field_present arr[N-1]` + `field_value_or_absent arr[N] value: null` — is unsound: the second clause passes when a seller emits a literal `null` pad at `arr[N]`. `array_length` reads `array.length` and rejects non-array resolutions, so the null-pad path fails the check as it should.

Two configurations:

```yaml
- check: array_length
  path: media_buys[0].impairments
  value: 2
  description: Exactly two impairments — one per rejected creative
```

```yaml
- check: array_length
  path: media_buys[0].impairments
  min: 1
  max: 1
  description: Exactly one impairment
```

`value` and `min`/`max` are mutually exclusive — specifying both, or specifying neither, fails the check as misconfigured. Both bounds are inclusive; either is optional.

Wired through:

- `src/lib/testing/storyboard/types.ts` — new `'array_length'` variant on `StoryboardValidationCheck`; new optional `min` / `max` fields on `StoryboardValidation`.
- `src/lib/testing/storyboard/validations.ts` — `validateArrayLength` with type checking, exact/range modes, and misconfiguration errors.
- `scripts/conformance-replay.ts` — in-process replay harness honours the new check.
- `test/lib/storyboard-validations.test.js` — 13 new tests covering exact match, both bounds, null-pad rejection, type errors, and misconfiguration paths.

Follow-up to [adcontextprotocol/adcp#4685](https://github.com/adcontextprotocol/adcp/pull/4685) protocol review. Spec-side YAML schema addition tracked separately on the adcp repo — storyboards adopting `array_length` will land once that ships.

## Test plan

- [x] `node --test test/lib/storyboard-validations.test.js` — 43 pass (13 new)
- [x] `node --test test/lib/storyboard-drift.test.js test/lib/storyboard-runner-contract.test.js` — 536/537 pass (1 pre-existing skip)
- [x] `node --test test/lib/storyboard-runner-output-contract-v2.test.js test/lib/storyboard-strict-validation.test.js test/lib/comply-vs-storyboard-parity.test.js` — 60/60 pass
- [x] Changeset included (minor bump)

### DIFF
--- a/.changeset/array-length-check-1830.md
+++ b/.changeset/array-length-check-1830.md
@@ -1,0 +1,37 @@
+---
+'@adcp/sdk': minor
+---
+
+storyboard runner: add `array_length` check kind for cardinality assertions (#1830)
+
+Cardinality scenarios can now assert "exactly N entries" directly:
+
+```yaml
+- check: array_length
+  path: media_buys[0].impairments
+  value: 2
+  description: Exactly two impairments — one per rejected creative
+```
+
+Range form is also supported (`min` / `max`, either or both, both inclusive):
+
+```yaml
+- check: array_length
+  path: media_buys[0].impairments
+  min: 1
+  max: 1
+  description: Exactly one impairment
+```
+
+The previous workaround (`field_present arr[N-1]` paired with
+`field_value_or_absent arr[N] value: null`) is unsound for cardinality: the
+`field_value_or_absent` clause passes when a seller emits a literal `null`
+pad at `arr[N]`. `array_length` reads `array.length` directly and rejects
+non-array resolutions, so a `null` pad fails the check as it should.
+
+Specifying both `value` and `min`/`max` is rejected as a misconfigured
+check. Specifying none of the three is also rejected. Resolved-path-is-not-an-array
+fails with a type error rather than passing silently.
+
+Follow-up to adcontextprotocol/adcp#4685 protocol review; spec-side YAML
+schema addition tracked separately.
--- a/scripts/conformance-replay.ts
+++ b/scripts/conformance-replay.ts
@@ -332,31 +332,48 @@ async function runStep(
         result.failures.push(`${check} ${v.path}: got ${JSON.stringify(actual)}, want ${JSON.stringify(v.value)}`);
       }
     } else if (check === 'array_length') {
-      const actual = getByPath(structured, v.path as string);
-      if (!Array.isArray(actual)) {
+      const hasExact = typeof v.value === 'number';
+      const hasMin = typeof (v as { min?: unknown }).min === 'number';
+      const hasMax = typeof (v as { max?: unknown }).max === 'number';
+      const min = hasMin ? (v as { min: number }).min : undefined;
+      const max = hasMax ? (v as { max: number }).max : undefined;
+      const isValidCount = (n: unknown): n is number => typeof n === 'number' && Number.isInteger(n) && n >= 0;
+      if (!hasExact && !hasMin && !hasMax) {
         allPassed = false;
-        result.failures.push(
-          `array_length ${v.path}: not an array (got ${actual === undefined ? 'undefined' : typeof actual})`
-        );
+        result.failures.push(`array_length ${v.path}: misconfigured (need value or min/max)`);
+      } else if (hasExact && (hasMin || hasMax)) {
+        allPassed = false;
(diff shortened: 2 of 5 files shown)


bokelley:
## Expert review pass

Ran code-reviewer-deep, ad-tech-protocol-expert, and security-reviewer-deep in parallel. Addressed all Should-Fix items in 42112981.

### code-reviewer-deep — Should Fix (all addressed)

1. ✅ Replay was missing the canonical's `value` + `min`/`max` mutual-exclusion guard — a misconfigured storyboard would silently pass replay while failing the runtime validator.
2. ✅ Replay's not-an-array message used `typeof` which mislabels `null` as `'object'`. Now distinguishes `null` explicitly.
3. ✅ Fractional / NaN / negative operands were accepted silently and failed every input. Now rejected at the config gate as misconfig.
4. ✅ Also added: impossible-range rejection when `min > max`.
5. ✅ JSDoc clarification that this check is non-optional by design.

Added four new tests (NaN, fractional `value`, negative `min`, impossible range).

### ad-tech-protocol-expert — sound-with-caveats

- Check name (`array_length`) is the right call vs `array_size` / `cardinality`.
- Shape (`value` for exact + `min`/`max` for bounds) matches existing AdCP storyboard conventions; no synonyms.
- Forward-compat path through the existing `not_applicable` default is correct.
- **Open question for spec PR #4685**: does it also extend `runner-output-contract.yaml`'s check-kind enum + `expected`/`actual` shape table? If only `storyboard-schema.yaml` adds the kind, downstream tooling that lints against the contract will reject conforming storyboards. Worth confirming before adopting in storyboards.

### security-reviewer-deep — clean

No Must Fix or Should Fix findings. `resolvePath` inherits FORBIDDEN_KEYS + own-property gating; `Array.isArray` fails closed on prototype walks and array-likes; `array.length` is O(1) on holes and large arrays; no LLM context construction in this change.

Considerations (not blocking): the `getByPath` shim in `conformance-replay.ts` is a pre-existing parallel implementation that lacks the canonical resolver's prototype-walking guards. The `array_length` consumer is the most-locked-down user of the shim because `Array.isArray` fails closed. Worth a follow-up to fold the shim into `resolvePath`, but not introduced by this PR.

PR #1837, opened 2026-05-18
feat(signing): requestContextFromExpress / -Fetch / -Lambda helpers (#1828)

### LINKED ISSUE #1828: signing: helper to construct ResponseLike.request from inbound Express/Fetch request
## Gap

\`ResponseLike.request.url\` MUST be an absolute URL because \`canonicalAuthority\` / \`canonicalTargetUri\` parse via \`new URL(...)\` and throw on relative paths. Express handlers ship \`req.url\` / \`req.originalUrl\` as path-only, so adopters reconstruct via:

\`\`\`ts
const url = \`\${req.protocol}://\${req.get('host')}\${req.originalUrl}\`;
\`\`\`

(documented in \`ResponseLike.request\` JSDoc as of #1823).

But \`req.protocol\` lies behind a TLS-terminating proxy unless \`app.set('trust proxy', ...)\` is configured, and \`req.get('host')\` is attacker-controllable absent a \`Host\` allowlist. A hostile peer that controls these headers can rebind a signature to \`attacker.example.com\` while the operator believes they're signing for \`seller.example.com\`. The library cannot detect this — the JSDoc warning is the best we can do, and JSDoc warnings rot.

## Proposed change

Ship adapter helpers that construct \`ResponseLike.request\` correctly by default:

\`\`\`ts
// Express
import type { Request } from 'express';
export function requestContextFromExpress(
  req: Request,
  options?: { hostAllowlist?: ReadonlyArray<string>; forceHttps?: boolean }
): { method: string; url: string };

// Fetch / Workers (no proxy considerations)
export function requestContextFromFetch(
  req: Request
): { method: string; url: string };

// AWS Lambda (API Gateway / ALB events)
export function requestContextFromLambda(
  event: APIGatewayEvent | ALBEvent
): { method: string; url: string };
\`\`\`

Express helper enforces:
- \`req.get('host')\` against \`hostAllowlist\` if provided (throws on mismatch).
- \`req.protocol === 'https'\` unless \`forceHttps: false\` (defaults strict — signatures bound to \`http://\` will fail strict verifier profiles).
- Strips trailing \`?\` and normalizes path.

Fetch helper is one-liner (\`request.url\` is already absolute, no proxy concerns).

Lambda helper extracts \`event.requestContext.domainName\` + \`event.rawPath\` + \`event.rawQueryString\`.

## Rationale

The current state is: adopter reads JSDoc warning, hopefully sets \`trust proxy\`, hopefully validates Host. Many won't. The signature still verifies, just against the wrong origin — and the operator can't tell from the verifier logs why their signed responses get rejected by buyer clients ("the signature is valid but my JWKS says I'm \`seller.example.com\`, why is this \`attacker.example.com\`?").

Shipping a helper makes the safe path the default path.

## Out of scope

- Auto-detecting proxy trust (no portable way to do this; up to operator).
- Built-in Host allowlist (no sensible default; operator must provide).

## Acceptance

- \`requestContextFromExpress\` ships in \`@adcp/sdk/signing/client\` (or a new adapter sub-barrel).
- \`requestContextFromFetch\` for Workers / Deno / Bun adopters.
- \`requestContextFromLambda\` for serverless adopters.
- Tests for: trust-proxy false negative (host header lying); allowlist violation; non-https rejection; valid reconstruction.
- README / docs note in \`signResponse\` JSDoc pointing at these helpers as the recommended construction path.

Filed from #1823 expert reviews (security + protocol both flagged the proxy-trust footgun).

bokelley:
## Summary

Closes #1828. Makes the safe path the default for constructing \`ResponseLike.request\` on each major Node platform.

The JSDoc warning on \`ResponseLike.request.url\` (shipped in #1823) flags the trap: under proxy termination, \`req.protocol\` lies and \`req.get('host')\` is attacker-controllable. JSDoc warnings rot. These helpers enforce hardening at construction time.

\`\`\`ts
import { requestContextFromExpress, signResponse } from '@adcp/sdk/signing/client';

app.post('/adcp/get_products', (req, res) => {
  const signed = signResponse(
    {
      status: 200,
      headers: { 'content-type': 'application/json' },
      body,
      request: requestContextFromExpress(req, {
        hostAllowlist: ['seller.example.com'],
      }),
    },
    key
  );
  res.set(signed.headers).send(body);
});
\`\`\`

### Three helpers

| Helper | Hardening |
|---|---|
| \`requestContextFromExpress(req, options)\` | Throws on missing \`Host\` / \`hostAllowlist\` mismatch / non-https protocol (unless \`forceHttps: false\`). Adopters MUST configure \`app.set('trust proxy', ...)\` — the helper trusts what Express tells it. |
| \`requestContextFromFetch(request)\` | Trivial passthrough. Workers / Deno / Bun terminate TLS at the edge and own URL construction; no hardening needed. |
| \`requestContextFromLambda(event, options)\` | Reads \`requestContext.domainName\` for authority; \`rawPath\` + \`rawQueryString\` (v2 / ALB) or \`path\` + \`queryStringParameters\` (v1) for the target. Always emits \`https://\`. \`hostAllowlist\` catches multi-tenant API Gateway misroutes. |

Exported from both \`@adcp/sdk/signing/client\` (signer-side adopters using \`signResponse\`) and \`@adcp/sdk/signing/server\` (verifier-side adopters reconstructing the originating-request URL for \`verifyResponseSignature\`).

## Test plan

- [x] 18 tests in \`test/request-context-helpers.test.js\`:
  - Express: happy path, host lowercasing, missing Host, allowlist enforcement (case-insensitive), \`forceHttps\` default + override, query-string preservation
  - Fetch: passthrough + real WHATWG \`Request\`
  - Lambda: v2 event shape, v1 fallback, missing \`domainName\`, allowlist enforcement, always-https, missing method, URL-encoded query params
  - **End-to-end**: Express helper output → \`signResponse\` → \`verifyResponseSignature\` round-trip verifies, proving the helper produces a URL the verifier accepts
- [x] \`tsc --noEmit\` clean
- [x] \`prettier --check\` clean

### DIFF
--- a/.changeset/request-context-helpers-1828.md
+++ b/.changeset/request-context-helpers-1828.md
@@ -1,0 +1,38 @@
+---
+'@adcp/sdk': minor
+---
+
+signing: add `requestContextFromExpress` / `-Fetch` / `-Lambda` helpers (#1828)
+
+Closes #1828. Makes the safe path the default path for constructing `ResponseLike.request` (and any other RFC 9421 binding that needs an originating-request URL) on each major Node platform.
+
+The JSDoc warning on `ResponseLike.request.url` (shipped in #1823) flags the trap: under proxy termination, `req.protocol` lies and `req.get('host')` is attacker-controllable. JSDoc warnings rot. These helpers enforce the hardening at construction time.
+
+```ts
+import { requestContextFromExpress, signResponse } from '@adcp/sdk/signing/client';
+
+app.post('/adcp/get_products', (req, res) => {
+  const signed = signResponse(
+    {
+      status: 200,
+      headers: { 'content-type': 'application/json' },
+      body,
+      request: requestContextFromExpress(req, {
+        hostAllowlist: ['seller.example.com'],
+      }),
+    },
+    key
+  );
+  res.set(signed.headers).send(body);
+});
+```
+
+**Three helpers:**
+
+- `requestContextFromExpress(req, options)` — throws when `Host` is missing / not in `hostAllowlist`, or when `req.protocol` is not `https` (unless `forceHttps: false` for local dev). **You MUST configure `app.set('trust proxy', ...)`** with your trusted proxy IPs; the helper trusts what Express tells it.
+- `requestContextFromFetch(request)` — passes through `method` + `url` from a WHATWG `Request`. Trivial; included for API symmetry. No proxy hardening needed (Workers / Deno / Bun terminate TLS at the edge and own URL construction).
+- `requestContextFromLambda(event, options)` — reads `requestContext.domainName` for authority, `rawPath` + `rawQueryString` (v2 / ALB) or `path` + `queryStringParameters` (v1) for the target. Always emits `https://` — Lambda isn't HTTP-addressable through API Gateway / ALB. `hostAllowlist` catches multi-tenant API Gateway misroutes.
+
+Exported from both `@adcp/sdk/signing/client` (signer-side adopters) and `@adcp/sdk/signing/server` (verifier-side adopters reconstructing the originating-request URL for `verifyResponseSignature`).
+
+18 tests covering happy path, host-allowlist enforcement, protocol enforcement, query-string preservation + URL encoding, and an end-to-end Express helper → `signResponse` → `verifyResponseSignature` round-trip.
--- a/src/lib/signing/client.ts
+++ b/src/lib/signing/client.ts
@@ -20,6 +20,16 @@ export {
 } from './canonicalize';
 export { computeContentDigest, contentDigestMatches, parseContentDigest } from './content-digest';
 export {
+  requestContextFromExpress,
+  requestContextFromFetch,
+  requestContextFromLambda,
+  type ExpressRequestLike,
+  type FetchRequestLike,
+  type LambdaRequestEvent,
+  type RequestContextFromExpressOptions,
+  type RequestContextFromLambdaOptions,
+} from './request-context';
+export {
   finalizeRequestSignature,
   finalizeResponseSignature,
   prepareRequestSignature,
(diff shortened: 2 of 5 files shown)


bokelley:
Security review back. Must Fix: none. Two should-fixes + one test-coverage gap addressed in commit 87baa54b:

1. **Trailing-dot Host bypass** — \`example.com.\` is a distinct string from \`example.com\` and a documented bypass on some stacks. \`normalizeHost()\` now strips trailing dots before allowlist comparison; both Express and Lambda helpers route through \`hostMatchesAllowlist()\` for consistent semantics. Three new tests.

2. **Userinfo rejection on Fetch helper** — WHATWG accepts \`https://user:pw@host/path\` as a valid \`Request\` URL and echoes it back verbatim. \`canonicalAuthority\` would then sign the userinfo-bearing form, splitting the signature namespace between userinfo and userinfo-stripped variants of the same logical origin (verifier confusion vector). Reject at construction. Defense-in-depth: Express helper also checks for \`@\` in \`originalUrl\` path so a malicious middleware can't smuggle userinfo through. Four new tests.

3. **Verifier-side helper symmetry + cross-host negative** — original end-to-end test exercised signer-side helper use only. Added:
   - Verifier-side reconstruction with a different \`req\` object produces the same canonical URL; signature still verifies (proves canonicalization is stable across helper invocations).
   - Cross-host negative: signature bound to \`seller.example.com\` fails verification when verifier reconstructs against \`attacker.example.com\` (proves the @authority binding is load-bearing). Two new tests.

27 total tests (was 18). \`tsc --noEmit\` and \`prettier --check\` clean.

Acceptable per security: "The proxy-trust JSDoc-only contract and Lambda \`domainName\` trust are correctly scoped."

PR #1841, opened 2026-05-18
feat(storyboard): add field_greater_than / field_at_most / field_at_least matchers (#1839)

### LINKED ISSUE #1839: storyboard runner: add field_at_most / field_at_least matchers for numeric threshold assertions
## Problem

Storyboard scenarios that assert "value X is at most / at least N" — runtime-enforcement-style assertions — have no clean matcher today. The check enum supports `field_less_than` (strict <) but not `field_at_most` / `field_lte` (≤).

This forces awkward workarounds. Concrete consumer: `media_buy_seller/frequency_cap_enforcement` (adcp#4727) asserts observed average frequency stays at-or-below the requested cap of 3 impressions per individual per day. The natural assertion is `frequency ≤ 3`. The strongest available today is `field_less_than: 3.01` (literal + 0.01 epsilon) — semantically wrong (assumes a domain rounding convention), brittle if the seller reports `3.00` exactly, and ugly.

## Ask

Add two new check kinds, symmetric with each other and with the existing `field_less_than` / `field_greater_than`:

```yaml
- check: field_at_most
  path: "media_buy_deliveries[0].totals.frequency"
  value: 3
  description: "Observed frequency stays at or below the requested cap"

- check: field_at_least
  path: "..."
  value: 100
  description: "..."
```

Semantics:
- Value at `path` MUST be numeric (number or integer); non-numeric → fail.
- `field_at_most: { value: N }` → fail unless `path-value <= N`.
- `field_at_least: { value: N }` → fail unless `path-value >= N`.
- `value` can be a literal number OR a context key (matching existing `field_less_than` precedent).
- Missing value at path → fail (the seller is expected to populate it; if presence is optional, gate via `field_present` separately).

Existing `field_less_than` / `field_greater_than` keep strict-comparison semantics. The four together cover the four comparator quadrants cleanly.

## Concrete consumer waiting on this

`media_buy_seller/frequency_cap_enforcement` ([adcp#4727](https://github.com/adcontextprotocol/adcp/pull/4727), about to merge with the epsilon workaround). Once `field_at_most` ships, that scenario's `field_less_than: 3.01` should be swapped to `field_at_most: 3` — small follow-up PR.

Future consumers: any cap or floor assertion (budget pacing ≤ allocation, reach ≥ promised, CPA ≤ target, latency floor / ceiling).

## Out of scope

- Tolerance / approximate-equality (`field_approx_equals` ± epsilon). Worth a separate ticket if a real consumer appears.
- Comparison against another path value (vs literal). Doable extension but YAGNI.

## Refs

- adcp#4727 — frequency_cap_enforcement scenario consuming the epsilon workaround
- adcp-client#1811 — `present:` matcher (precedent for incremental runner extension)
- adcp-client#1817 — `contains:` matcher (precedent)

## Python parity

Once TS ships, mirror in `adcp-client-python`.

bokelley:
## Summary

Closes #1839.

Adds two new storyboard validation check kinds:

- `field_at_most` — assert numeric field ≤ comparand (non-strict)
- `field_at_least` — assert numeric field ≥ comparand (non-strict)

They share the same comparand-resolution model as the existing strict `field_less_than`: either a literal `value` or a runtime-captured `context_key`, with absent context keys passing the check with a `context_key_absent` observation (the prior step may have been legitimately skipped on a branch-set path).

## Why

Storyboard scenarios that assert "value X is at most / at least N" — cap and floor enforcement — have no clean primitive today. The strongest available is `field_less_than` (strict `<`), which forces a literal-plus-epsilon workaround:

```yaml
- check: field_less_than
  path: media_buy_deliveries[0].totals.frequency
  value: 3.01  # cap is 3; epsilon to allow equality
```

That's semantically wrong (assumes a domain rounding convention) and brittle when sellers report the cap value exactly (`3.00 < 3.01` passes; `3.00 < 3.00` fails).

`field_at_most: 3` is the right primitive.

## Implementation note

Refactored `validateFieldLessThan` into a shared `validateNumericComparison(op)` helper parameterised by check name, operator symbol, and comparator function. All three numeric checks dispatch through it — no copy-paste branches. Forward-compat default in the runner already grades unknown check kinds as `not_applicable`, so older runners won't brick.

## Concrete consumer

`media_buy_seller/frequency_cap_enforcement` (adcontextprotocol/adcp#4727) ships with the epsilon workaround. Once this lands, a small follow-up will swap `field_less_than: 3.01` → `field_at_most: 3` in that scenario yaml.

## Test plan

- [x] 12 new unit tests covering both matchers (`test/storyboard-cross-step-validators.test.js`):
  - [x] strictly-less / strictly-greater paths
  - [x] **boundary-equality paths** (the cases that distinguish `≤`/`≥` from `<`/`>`)
  - [x] literal-value comparand
  - [x] `context_key` comparand
  - [x] `context_key_absent` observation
  - [x] type-error: missing field
  - [x] type-error: non-numeric field
  - [x] type-error: non-numeric comparand
  - [x] missing path
- [x] All 22 existing `field_less_than` / `field_equals_context` tests still pass (refactor preserves behaviour).
- [x] `tsc --noEmit` clean.

## Python parity

Per the issue: once TS ships, mirror in `adcp-client-python`. Not addressed here.

### DIFF
--- a/.changeset/field-at-most-at-least-1839.md
+++ b/.changeset/field-at-most-at-least-1839.md
@@ -2,7 +2,7 @@
 '@adcp/sdk': minor
 ---
 
-storyboard runner: add `field_at_most` / `field_at_least` check kinds for non-strict numeric thresholds (#1839)
+storyboard runner: add `field_greater_than` / `field_at_most` / `field_at_least` check kinds, completing the numeric-comparison quadrant (#1839)
 
 Cap and floor scenarios can now assert "value at most N" / "value at least
 N" directly, with non-strict (`<=` / `>=`) semantics:
@@ -19,12 +19,18 @@ N" directly, with non-strict (`<=` / `>=`) semantics:
   description: Delivered reach meets or exceeds the promised floor
 ```
 
-The two new checks are symmetric with each other and with the existing
-strict `field_less_than`. They share the same comparand-resolution
-semantics: either a literal `value` or a runtime-captured `context_key`,
-with absent context keys passing the check with a `context_key_absent`
-observation (the prior step may have been legitimately skipped on a
-branch-set path). Non-numeric operands fail with a type error.
+The three new checks plus the existing strict `field_less_than` cover the
+four-quadrant numeric comparison vocabulary (`<`, `>`, `<=`, `>=`). All
+four share the same comparand-resolution semantics: either a literal
+`value` or a runtime-captured `context_key`, with absent context keys
+passing the check with a `context_key_absent` observation (the prior step
+may have been legitimately skipped on a branch-set path). Non-numeric
+operands fail with a type error.
+
+`field_greater_than` is included to close the quadrant — without it,
+storyboards reaching for strict-`>` would hit the runner's forward-compat
+`not_applicable` default and silently grade as passing, which is
+semantically wrong for assertions that should fail at the boundary.
 
 The previous workaround for cap assertions
 (`field_less_than: 3.01` — literal-plus-epsilon) was semantically wrong
--- a/.changeset/field-at-most-at-least-1839.md
+++ b/.changeset/field-at-most-at-least-1839.md
@@ -1,0 +1,37 @@
+---
+'@adcp/sdk': minor
+---
+
+storyboard runner: add `field_at_most` / `field_at_least` check kinds for non-strict numeric thresholds (#1839)
+
+Cap and floor scenarios can now assert "value at most N" / "value at least
+N" directly, with non-strict (`<=` / `>=`) semantics:
+
+```yaml
+- check: field_at_most
+  path: media_buy_deliveries[0].totals.frequency
+  value: 3
+  description: Observed frequency stays at or below the requested cap
+
+- check: field_at_least
+  path: media_buy_deliveries[0].totals.reach
+  context_key: promised_reach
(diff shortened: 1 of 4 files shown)


bokelley:
## Review pass — done

Parallel reviews from code-reviewer + ad-tech-protocol-expert.

**code-reviewer**: ship. Refactor preserves all original behavior (error messages, expected/actual fields, observation pass-through, type-error paths) exactly; boundary semantics correct; `NUMERIC_OPS` typing sound (`as const satisfies …`).

**ad-tech-protocol-expert**: sound-with-one-caveat. Naming matches the prose-style vocabulary (`field_present`, `array_length`, `contains`) rather than operator-symbol style (`field_lte`) — right call. Forward-compat verified. **Flagged**: shipping `≤`/`≥` without `>` leaves the comparator quadrant 3-of-4 filled. Storyboard authors reaching for strict-`>` would hit the runner's forward-compat `not_applicable` default and silently grade as passing — semantically wrong.

**Addressed in `ba4faed4`**: added `field_greater_than` to close the quadrant. One more `NUMERIC_OPS` entry; the shared `validateNumericComparison` helper handles it without duplication. Tests added covering the strict-`>` equality-boundary case (the unique distinguishing behavior vs `field_at_least`). 38/38 tests pass.

Title updated to reflect all three matchers.

PR #1849, opened 2026-05-18
feat(media-buy): preflight helpers for available_actions[] (#4480)

nastassiafulconis:
ships the consumer surface for [adcontextprotocol/adcp#4480](https://github.com/adcontextprotocol/adcp/issues/4480) (merged in adcp#4514) so buyer agents don't have to hand-roll capability checks against `update_media_buy`.

## what's in here

- **boolean gates** on the buy: `canPause`, `canExtendFlight`, `canIncreaseBudget`, `canDecreaseBudget`, `canReallocateBudget`, `canRemoveCreative`, ... (17 total). each reads `available_actions[]` (or rolls up from legacy `valid_actions[]`).
- **resolver**: `getActionForMutation(currentBuy, request)` walks the request body and returns the fine-grained actions it covers as a `ResolvedAction[]`. direction inference picks `increase_budget` vs `decrease_budget` vs `reallocate_budget` from the per-package budget diff, and `extend_flight` vs `shorten_flight` vs `update_flight_dates` from start/end-time diffs.
- **preflight**: `preflightUpdateMediaBuy(currentBuy, request)` composes resolver + gates into a discriminated union with typed recovery hints (`createProposal` / `waitForApproval` / `reissueAsDirect`) for `mode_mismatch`.
- **typed error**: `ActionNotAllowedError` with `attemptedAction`, `reason`, `currentlyAvailableActions[]`, and a structured `recovery` hint. wired into `adcpErrorToTypedError()` with permissive parsing of the upstream `details` payload.
- **compat shim**: `getAvailableActions(buy)` falls back to synthesizing from `valid_actions[]` with a one-shot deprecation hint when sellers haven't moved to 3.1 yet.

## design notes

- types are hand-written in `src/lib/media-buy/types.ts` until the next schema-cache bump pulls AdCP 3.1 into `*.generated.ts`. they mirror the merged upstream schema verbatim.
- the action-to-field mapping is read from `enumMetadata.update_fields` in the schema rather than hand-copied. `scripts/generate-media-buy-update-fields.ts` regenerates the `UPDATE_FIELDS_BY_ACTION` constant from the schema; commit the regenerated file on a schema bump.
- multi-action requests (e.g. touching both `end_time` and `packages[].budget`) require every resolved action to be present in `available_actions[]`. first missing action wins the denial.

## test plan

- [x] `tsc --noEmit` clean
- [x] `eslint` clean on new files
- [x] 46 unit tests pass (gates, resolver direction cases, preflight ok/denied, every `ACTION_NOT_ALLOWED` reason)
- [ ] CI: full `npm test` (started locally but a re-run on CI will be authoritative)
- [ ] CI: `changeset` check (minor bump for `@adcp/sdk`)


bokelley (commented):
Reviewed the diff. Solid shape overall — generator-driven mapping, spec-faithful types, denials collected (good call on the recent refactor), comprehensive test coverage. A few things worth picking up before merge.

## 🟡 Bug: `pkg.paused` flows through top-level `paused` direction logic

`src/lib/media-buy/preflight.ts:147` adds `'paused'` to the touched-fields set when `pkg.paused !== undefined`. Then `preflight.ts:104-106` resolves direction by checking only top-level `request.paused`:

```ts
if (field === 'paused') {
  pick = { action: request.paused === false ? 'resume' : 'pause' };
}
```

If a buyer sets `packages: [{ package_id, paused: false }]` without setting top-level `request.paused`, `request.paused === false` evaluates `false` (undefined !== false), so the action resolves to `'pause'` when semantically it's a resume. No test covers this — every `paused` test sets the top-level field.

Two reasonable fixes:
- **(a)** Drop `pkg.paused` from `collectTouchedFields` — the generated mapping has no `packages[].paused` entry, so this whole branch is out-of-spec.
- **(b)** Check both levels when picking direction.

(a) is more honest. If the spec doesn't define a per-package pause action, the resolver shouldn't fabricate one. Either way, please add a test for `pkg.paused`-only requests.

## 🟡 Synthetic `pause` denial on empty request

`preflight.ts:351-356`:

```ts
denials: [{ action: 'pause', reason: 'not_supported_on_buy' }],
```

This will confuse callers who pattern-match on `denials[0].action` and see `'pause'` for a request that never mentioned pausing. Two cleaner alternatives:
- A dedicated reason code (e.g. `'no_recognized_field'`) — but that's a spec extension.
- Throw a `ValidationError` instead of returning a `PreflightDenied`. "You didn't ask me to do anything" is a buyer-side bug, not a seller-side denial.

I'd lean toward the typed error — keeps the `PreflightDenied` shape honest (every denial actually maps to an action the buyer requested).

## 🟡 Test-only hook leaks through the sub-barrel

`available-actions.ts:48` JSDoc says:

> Not exported from the public barrel.

But `media-buy/index.ts:21` re-exports `__resetValidActionsWarningForTests`. Tests import it from `dist/lib/media-buy`. Either the JSDoc is wrong (intentionally exposed for adopter test suites) or the re-export is wrong (move behind `/internal`). Pick one.

## 🟢 Equal-totals fallthrough to `reallocate_budget`

`preflight.ts:220-223`: when totals match within tolerance AND no per-package movement was detected (e.g. budget set to its current value), the resolver returns `reallocate_budget`. Comment says "so the request still has *some* action attached" — acceptable, but worth a test. If the spec considers this a no-op, the resolver should say so explicitly.

## 🟢 Float tolerance assumption

`BUDGET_EQUAL_TOLERANCE = 0.005` (preflight.ts:56) assumes ≥ half-cent precision is the smallest unit any AdCP currency cares about. Fine for USD/EUR/JPY today. If a future spec extension allows micro-amounts for programmatic pricing, this tolerance would mask real reallocations. Worth a `TODO`-style comment so a future reader knows it's a load-bearing assumption tied to the spec's pricing model.

## 🟢 Default-extend when end-date direction is indeterminate

`preflight.ts:252-253`: when only `end_time` is touched but dates are unparseable, defaults to `extend_flight`. If the seller advertises `shorten_flight` but not `extend_flight`, this preflight will deny when the buyer probably wanted to shorten. A safer default might be `update_flight_dates` (the generic vocabulary), which is more likely to be advertised. Or return both candidates and let the preflight try in order.

## ℹ️ Changeset note for regen workflow

`scripts/generate-media-buy-update-fields.ts` will throw against today's schema cache (3.0.12 has no `enumMetadata`). The types.ts header documents this, but the **changeset** doesn't mention the regen step. One-liner: "after `npm run sync-schemas` lands 3.1, re-run `npx tsx scripts/generate-media-buy-update-fields.ts` and commit the regenerated file." Next person bumping the cache shouldn't have to read the source comments to find this.

---

Test coverage looks comprehensive — 46 tests across gates, resolver direction cases, multi-action preflight, all four `ActionNotAllowedReason` values, compat shim, and typed-error dispatch. Gaps noted above (pkg.paused-only, budget-unchanged, indeterminate end_time).

Architecture-wise: types stay free of cross-module imports (matches the convention used for other typed errors), `as const` mappings are tree-shakeable, no new I/O surface, no auth boundary touched.

Recommend approve once the `pkg.paused` bug is fixed and a test added. Items 4-6 are judgment calls — feel free to push back.

nastassiafulconis:
thanks for the careful pass. addressed all of these in f4337844:

🟡 **pkg.paused bug** - went with option (a). dropped the branch entirely so the resolver doesn't fabricate a `paused` action from a per-package field the spec doesn't define. added a test that a pkg.paused-only request now throws ValidationError (since after the drop, no fields are touched).

🟡 **synthetic pause denial** - throwing ValidationError. agreed it was lying about what the seller said. uses the existing ADCPValidationError class. added test.

🟡 **test-only hook leak** - fixed the JSDoc rather than moving the export. tests need it accessible through dist/, and downstream adopter test suites may want the same. underscore prefix carries the "not buyer-facing" signal.

🟢 **equal-totals fallthrough** - added a test for the no-op equal-totals case so the fallthrough is locked.

🟢 **float tolerance assumption** - expanded the comment to call out that half-cent precision is tied to the current AdCP pricing model; if sub-cent / micro-amount pricing lands, this needs to be currency-aware.

🟢 **indeterminate end_time** - changed default to update_flight_dates instead of extend_flight. matches the "safer generic vocabulary" suggestion. added test.

ℹ️ **changeset regen note** - added a follow-up paragraph documenting the regen workflow when sync-schemas pulls a 3.1+ cache.

ready for another look.

bokelley (approved):
LGTM — all six items from the prior pass landed cleanly in f4337844. Spot-checked each:

- **pkg.paused** (option a): dropped from `collectTouchedFields` with a comment naming the reason (no spec entry for `packages[].paused`). Resolver no longer misclassifies package-only pause toggles; test added that asserts `pkg.paused`-only requests throw `ValidationError`.
- **Empty-request denial**: now `throw new ValidationError('request', request, '…')` with the correct `(field, value, constraint)` signature. The empty-`{}` test was rewritten to `assert.throws(...)`, plus a fresh test for the `pkg.paused`-only case. PreflightDenial shape is honest again.
- **Indeterminate end_time**: defaults to `{ action: 'update_flight_dates', direction: 'shift' }`. Generic vocabulary, more likely to be advertised. Test added against a buy with no end_time baseline.
- **Equal-totals reallocate fallthrough**: explicit test (`pkg_1`, budget set to current 1000) locks the behavior.
- **BUDGET_EQUAL_TOLERANCE**: comment now calls out the half-cent tie-in to current AdCP pricing and flags micro-amount pricing as the trigger to revisit.
- **`__resetValidActionsWarningForTests` JSDoc**: corrected to match the actual re-export (intentionally exposed via `media-buy/index.ts` so tests in this repo and downstream adopters can call it through `dist/`).
- **Changeset regen workflow**: bottom paragraph documents the post-`sync-schemas` regen step plus the fact that the generator throws against a metadata-less cache, so a future bumper can't silently forget.

Approve.

PR #1852, opened 2026-05-19
feat(server): expose ctx.input — un-destructured wire envelope on every v6 dispatch (#1842)

### LINKED ISSUE #1842: SalesPlatform.syncCreatives drops request-level assignments — platforms have no surface for inline bindings
## Summary

The v6 framework wrapper at `src/server/decisioning/runtime/from-platform.ts` (compiled at `dist/lib/server/decisioning/runtime/from-platform.js:2884-2902`) passes only `params.creatives` to `creative.syncCreatives(creatives, reqCtx)`. The request-level `assignments[]` field — spec-defined at `creative/sync-creatives-request.json` and described as "Optional bulk assignment of creatives to packages" — never reaches the platform method.

```js
syncCreatives: async (params, ctx) => {
  const reqCtx = ctxFor(ctx);
  const creatives = params.creatives ?? [];   // ← `params.assignments` not forwarded
  return projectSync(async () => {
    const push = extractPushConfig(params, logger, { ... });
    const result = await creative.syncCreatives(creatives, reqCtx);
    ...
  });
}
```

`SalesPlatform.syncCreatives?(creatives: Creative[], ctx: Ctx<TCtxMeta>)` has no `assignments` parameter, and the framework doesn't expose `params` to the platform via `ctx`. Adopters that need bulk creative-to-package binding must invoke `update_media_buy.packages[].creative_assignments` as a follow-up call — they cannot satisfy the wire-level contract of `sync_creatives` with `assignments`.

## Reproduction

The new `media_buy_seller/dependency_impairment` and `dependency_impairment_cardinality` storyboards in `adcontextprotocol/adcp` (added in #4677 / #4685) originally used inline `sync_creatives` with `assignments[]`. Against the training agent's `/sales/mcp` (v6) route, the assignments silently dropped — the package's `creative_assignments` stayed empty, so subsequent `force_creative_status` transitions had no media buy dependency to propagate to.

Storyboards have been restructured at `adcontextprotocol/adcp#4717` to do binding via `update_media_buy` as a workaround, but this constrains the scenario shape — the spec describes inline assignments as a valid wire-level path, and conformance graders that target only the v6 SDK pipeline will silently pass agents that drop the field.

## Proposed fix

Pick one:

1. **Extend the signature** to `syncCreatives(creatives, assignments, ctx)` — breaking change, but minimal.
2. **Pass a `req` envelope** to the platform — `syncCreatives(req, ctx)` where `req` includes `creatives`, `assignments`, etc. Less ergonomic for adopters that only need creatives, but spec-aligned.
3. **Expose `ctx.input`** so platforms can opt into reading additional request fields. Same approach used elsewhere in the framework for `comply_test_controller`. Backward-compatible.

(3) seems cleanest — `syncCreatives` adopters that don't need assignments stay on `(creatives, ctx)`; those that do read `ctx.input.assignments`.

## Why this matters

Inline `sync_creatives` + `assignments` is a one-roundtrip path for buyer agents that already know which packages each creative goes to. Forcing every adopter to follow up with `update_media_buy` is a two-call path that drifts further from the spec's wire-level intent. Adopters who *do* implement inline assignments at the v5 layer (training agent's `handleSyncCreatives`) currently get them stripped at the v6 boundary — they look conformant on `/mcp` but silently dropping on `/sales/mcp`.

## Refs

- Downstream tracking: adcontextprotocol/adcp#4719 (impairment tracking)
- Storyboards restructured to work around: adcontextprotocol/adcp#4717
- Sibling issue (digest mode permissiveness): adcontextprotocol/adcp-client#1840

bokelley:
## Summary

Closes #1842. Adds `RequestContext.input` (and `ResolveContext.input`) so v6 platform methods can read request fields the typed signature drops — fixes the silent-drop trap on `sync_creatives.assignments[]`, `sync_audiences.delete_missing`, `sync_accounts.delete_missing`, and friends.

The v6 framework wrapper destructures the payload array out of `sync_*` requests and forwards only that to the platform method, dropping spec-meaningful modifiers the typed signature doesn't model. Adopters look conformant on `/mcp` (their v5 handler reads the full envelope) and silently fail on `/sales/mcp`. With `ctx.input`, adopters who need those fields cast at the read site:

```ts
syncCreatives: async (creatives, ctx) => {
  const wire = ctx.input as SyncCreativesRequest;
  for (const a of wire.assignments ?? []) {
    await this.bindCreativeToPackages(a.creative_id, a.package_ids);
  }
  if (wire.delete_missing) {
    await this.deleteCreativesNotIn(creatives.map(c => c.creative_id));
  }
}
```

## Two parallel fields, same shape

- **`RequestContext.input`** — sales / creative / signals / governance / SI handler families.
- **`ResolveContext.input`** — account-handler family (`syncAccounts`, `syncGovernance`, `listAccounts`, `reportUsage`, `getAccountFinancials`). Those use a separate context type because `accounts.resolve` produces the account rather than receiving it.

Spotted in expert review: an earlier version of this PR only wired the `RequestContext` path and would have left `sync_accounts.delete_missing` still dropped. Fixed before push.

## JSDoc lands three callouts

1. **Same reference as the typed payload arg, not a snapshot.** `ctx.input` is set before the auto-hydrate seams (`hydratePackagesWithProducts`, `hydrateForTool`) run, but those seams mutate the same object in place. By the time the platform method's body executes, `ctx.input` and the first positional arg are the same reference. Treat as request payload "as the method sees it," not "what the buyer originally sent."
2. **Hoist asymmetry.** For methods that hoist a field to a positional arg (e.g. `updateMediaBuy(media_buy_id, patch, ctx)`), that field is still present at the top level of `ctx.input`. Prefer the positional arg when both are available; reach for `ctx.input` only for fields the typed signature drops.
3. **Security.** `ctx.input` is buyer-controlled and may carry secrets — mutating-tool envelopes include `push_notification_config.token`; free-text fields (`brief`, `message`, creative snippets) are attacker-controlled. Don't log wholesale; don't template into LLM prompts without fencing.

## Files changed

| File | What |
|---|---|
| `src/lib/server/decisioning/context.ts` | `RequestContext.input` field + JSDoc |
| `src/lib/server/decisioning/account.ts` | `ResolveContext.input` field + JSDoc |
| `src/lib/server/decisioning/runtime/to-context.ts` | `buildRequestContext(handlerCtx, ctxMetadataStore, input?)` |
| `src/lib/server/decisioning/runtime/from-platform.ts` | `CtxForFn` signature, `toResolveCtx` signature, 56 call sites updated to pass `params` |
| `test/server-decisioning-ctx-input-wire-envelope.test.js` | 5 regression tests |
| `.changeset/ctx-input-wire-envelope-1842.md` | minor release note |

## Test plan

- [x] `npm run build` — clean
- [x] `npx tsc --noEmit` — clean
- [x] `npm test` — 9234 pass, 0 fail, 3 unrelated skipped
- [x] 5 new regression tests pass (sync_creatives, sync_audiences, sync_accounts, update_media_buy hoist asymmetry, get_products universal pattern)
- [x] 4 expert reviews — protocol, code, product, security — concerns folded back in before push

## Review trail (folded in pre-push)

- **Protocol + code reviewers (converged blocker):** Initial PR only wired `RequestContext`, leaving `sync_accounts` etc. on `ResolveContext` still dropping fields. Fixed by adding `input` to `ResolveContext` and threading `params` through `toResolveCtx`.
- **Product reviewer (blocker):** Original JSDoc claimed `ctx.input` was "the un-destructured wire envelope" — but it's the same reference as `params`, which gets mutated by framework hydrators. Rewrote JSDoc to be honest: "request payload as the method sees it; same reference as the typed payload arg; framework hydration writes pass through."
- **Security reviewer (nits, folded in):** Added warnings on `ctx.input` containing `push_notification_config.token` + attacker-controlled free-text fields. Don't-log-wholesale and don't-string-interpolate-into-LLM-prompts guidance now lives in JSDoc.

### DIFF
--- a/.changeset/ctx-input-wire-envelope-1842.md
+++ b/.changeset/ctx-input-wire-envelope-1842.md
@@ -1,0 +1,56 @@
+---
+'@adcp/sdk': minor
+---
+
+feat(server): expose `ctx.input` — un-destructured wire envelope on every v6 platform-method dispatch
+
+Closes the silent-drop bug audited in adcp-client#1842. The v6 framework wrapper at `src/lib/server/decisioning/runtime/from-platform.ts` destructures the payload array out of `sync_*` requests and forwards only that to the platform method, dropping spec-meaningful modifiers the typed signature doesn't model. Adopters look conformant on `/mcp` (their v5 handler reads the full envelope) and silently fail on `/sales/mcp` (the v6 path strips the field).
+
+**Surface bug.** Four platform methods drop spec-meaningful fields:
+
+| Method | Wire schema | Dropped fields (now reachable on `ctx.input`) |
+|---|---|---|
+| `sales.syncCreatives` / `creative.syncCreatives` | `sync_creatives_request` | `assignments[]`, `creative_ids[]`, `delete_missing`, `dry_run`, `validation_mode` |
+| `audiences.syncAudiences` | `sync_audiences_request` | `delete_missing` |
+| `accounts.upsert` (→ `sync_accounts`) | `sync_accounts_request` | `delete_missing`, `dry_run` |
+
+**Fix.** Two parallel fields, same shape: `RequestContext.input?: Readonly<Record<string, unknown>>` carries the request payload on the sales/creative/signals/governance/SI handler families; `ResolveContext.input?: Readonly<Record<string, unknown>>` carries it on the account-handler family (`syncAccounts`, `syncGovernance`, `listAccounts`, `reportUsage`, `getAccountFinancials`) — those use a separate context type because `accounts.resolve` produces the account rather than receiving it. Adopters who need a field the typed signature drops read it from `ctx.input`:
+
+```ts
+syncCreatives: async (creatives, ctx) => {
+  const wire = ctx.input as SyncCreativesRequest;
+  for (const a of wire.assignments ?? []) {
+    await this.bindCreativeToPackages(a.creative_id, a.package_ids);
+  }
+  if (wire.delete_missing) {
+    await this.deleteCreativesNotIn(creatives.map(c => c.creative_id));
+  }
+  // …
+}
+```
+
+**Why this shape:**
+
+1. **Envelope-shaped, not method-shaped.** Every drop is the same concept: "the method got the array; it didn't get the modifiers on the same request." One `ctx.input` covers every drop in the table above plus any future field the SDK doesn't yet model.
+2. **Additive, non-breaking.** Adopters who don't read `ctx.input` keep working; adopters who need the dropped fields opt in. No signature churn.
+3. **Composes with hydrate seams.** `hydrateForTool` / `hydratePackagesWithProducts` write *to* `params`; `ctx.input` reads *from* the unmodified wire envelope. Both are present on every dispatch.
+
+**Same reference as the typed payload arg, not a snapshot.** `ctx.input` is the request payload as the platform method sees it. It is set BEFORE the framework's auto-hydrate seams (`hydratePackagesWithProducts`, `hydrateForTool`) run, but those seams mutate the same object in place. By the time the platform method's body executes, `ctx.input` and the first positional arg are the same reference and both reflect framework hydration. JSDoc on `RequestContext.input` calls this out so adopters don't write logic that assumes `ctx.input` is a frozen pre-hydration snapshot.
+
+**Hoist asymmetry.** For methods that hoist a field to a positional arg (e.g. `updateMediaBuy(media_buy_id, patch, ctx)` hoists `media_buy_id` out of the envelope), that field is still present at the top level of `ctx.input` — the wire envelope is hoist-included, NOT residual. Adopters should prefer the positional arg for fields present on both, and reach for `ctx.input` only for fields the typed signature drops.
+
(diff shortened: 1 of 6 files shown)


bokelley:
**CI status note:** \`Test & Build\` failed on \`test/lib/cli-webhook-receiver-flag.test.js:360\` ("ADCP_WEBHOOK_TUNNEL override substitutes {port} and captures marker URL"). The failure is upstream-environmental, not introduced by this PR.

**Evidence:**

- The failing test invokes the CLI against the public test agent at \`https://test-agent.adcontextprotocol.org/mcp/\`. As of right now (post-CI run), that host plus \`https://adcontextprotocol.org/protocol/3.0.12.tgz.sha256\` are both timing out (curl \`--max-time 5\` → exit 28, no bytes received). The test expects the CLI to fail-fast at network dispatch; when the upstream is hanging instead of refusing, the CLI rides out its default network timeout and the test's 10s harness timeout fires (\`SIGKILL\`, \`duration_ms: 10043\`).
- The same test passed on \`main\` at 2026-05-18T23:14Z (\`ok 6 - ADCP_WEBHOOK_TUNNEL override substitutes {port} and captures marker URL\`) — before the upstream outage began.
- This PR touches \`src/lib/server/decisioning/runtime/{from-platform.ts, to-context.ts}\`, \`src/lib/server/decisioning/{context.ts, account.ts}\` — all v6 SELLER-side dispatch. The failing test is a CLI BUYER-side path (\`storyboard run\`) hitting a remote agent over HTTP. The two surfaces do not share a code path.
- Local: \`npm test\` is 9234/9234 green on this branch with all 5 new regression tests passing.

Re-ran the failed job; will merge once upstream recovers or via admin override if it stays down.

Issue #1857, opened 2026-05-19
Idempotency store: cache-fill DoS via unbounded distinct idempotency_key values


bokelley:
**Source:** Security review on PR #1855 (Redis backend), threat #4.

**Status:** Pre-existing protocol-store issue. Not a regression introduced by #1855 — applies to every backend, but Redis makes the failure mode louder because TTL'd keys without an `maxmemory-policy` guide can OOM the instance.

## Problem

A buyer holding a valid principal can mint unbounded distinct `idempotency_key` values and hit any mutating tool. Each successful response and each `saveTransientError` (`store.ts:436`) writes a new entry. With default 24h `ttlSeconds`:

- **Postgres backend:** table grows until `cleanupExpiredIdempotency` cron sweeps. Hammered enough, the cron can't keep up and the table bloats.
- **Redis backend:** Redis evicts at TTL, but a sufficient rate of writes can OOM the instance before TTLs hit. On default `noeviction`, writes start failing; on `allkeys-lru`, Redis evicts **other applications'** keys (collateral damage to the operator's other apps sharing the instance).
- **Memory backend:** unbounded heap growth, OOM kill.

`VALIDATION_ERROR` from a drifted handler is the worst case: the handler fails fast (cheap) and every fresh key writes a new 10s-TTL entry, magnifying the rate. The transient-error cache was meant to absorb buyer retry storms, not to absorb a hostile fresh-key flood.

## What we should do

1. **Per-principal rate limit at the dispatch layer.** Outside the store itself — the store can't tell a fresh hostile flood from a legitimate batch upload. Plumb through a `principalRateLimit?: RateLimiter` option on `serve()`.
2. **Operator guidance for Redis backends.** JSDoc on `redisBackend` should recommend `maxmemory-policy volatile-lru` (evicts only TTL'd keys, contains blast radius to AdCP keys) or `allkeys-lru` only on a **dedicated** Redis db. Currently silent.
3. **Per-principal entry count cap** as a backend-level circuit breaker. Hard to implement cheaply in Redis (would need a parallel counter key per principal); in Postgres it's a `COUNT(*) WHERE principal = $1`. Probably not worth the complexity vs (1).

Recommendation: ship (2) immediately as a documentation change on `redisBackend`. File (1) as a separate design issue for the dispatch layer. Skip (3) unless we see real abuse.

## Acceptance

- [ ] `redisBackend` JSDoc gains a section on `maxmemory-policy` guidance
[...]

PR #1858, opened 2026-05-19
feat(server, signing): Redis backends for ctx-metadata and ReplayStore

bokelley:
## Summary

Adds two more Redis-backed stores alongside the existing memory and Postgres variants — completing the "good Redis candidates" trio I identified in the original sizing conversation. Sister of #1855 (Redis idempotency backend), independent at the merge level.

- **`redisCtxMetadataStore`** (`@adcp/sdk/server`) — factory function matching the pg backend pattern. One Redis key per `(account_id, kind, id)`, JSON-encoded `{ value, resource?, expiresAt? }`. Entries with no `expiresAt` are stored with **no Redis TTL** — ctx-metadata lifetimes can be months and silent eviction produces "package not found" errors that look like publisher bugs and run for weeks. `bulkGet` uses `MGET` (single round trip per batch).
- **`RedisReplayStore`** (`@adcp/sdk/signing/server`) — class-based to match `PostgresReplayStore`. One sorted set per `(keyid, scope)`, scored by `expiresAt`, members are nonces. A single Lua script does the atomic `ZREMRANGEBYSCORE -inf now` (drop expired) → `ZSCORE` (replay check) → `ZCARD` (cap check) → `ZADD` (insert) + `PEXPIREAT` (extend set TTL). The `ReplayStore.insert` JSDoc literally names "Redis SET NX EX" as a canonical multi-replica primitive; this delivers on that with strict precedence (replay > rate_abuse > ok).

```ts
import { createClient } from 'redis';
import { createCtxMetadataStore, redisCtxMetadataStore, createAdcpServer } from '@adcp/sdk/server';
import { RedisReplayStore } from '@adcp/sdk/signing/server';

const redis = createClient({ url: process.env.REDIS_URL });
redis.on('error', err => console.error('redis error', err));
await redis.connect();

createAdcpServer({
  ctxMetadata: createCtxMetadataStore({ backend: redisCtxMetadataStore(redis) }),
  // ...
});

// And/or for signed-requests verifiers running multiple replicas:
const replayStore = new RedisReplayStore(redis);
```

## Design notes worth reviewing

- **Client union type repeated.** Each backend ships its own `RedisClientType<any,any,any> | <BackendSpecificNarrowInterface>` union so node-redis users pass `createClient(...)` straight in without casts. Each narrow interface lists only the methods that backend actually calls — `CtxMetadataRedisLikeClient` adds `mGet` for `bulkGet`; `ReplayRedisLikeClient` adds `eval`/`zScore`/`zCount` for the sorted-set primitives. `import type { RedisClientType }` is erased at emit so `redis` stays a true optional peer dep. Verified with an out-of-tree DX smoke (same pattern as #1855).
- **`ctx-metadata` durable-by-default.** When `entry.expiresAt` is undefined, the Redis key has no TTL — matches the pg backend's `expires_at = NULL` and the store-level "no auto-eviction by design" contract. Only when `expiresAt` is set does the backend stamp `EX = expiresAt - now + expiredGraceSeconds` (60s grace so the store layer's own expiry check sees the value within the clock-skew window).
- **`replay` Lua atomicity.** Concurrent inserts of the same `(keyid, scope, nonce)` race entirely inside Redis (single-threaded, Lua blocks everything else for the script duration). The test suite exercises 10-way concurrent same-nonce inserts (exactly one wins), 20-way concurrent inserts at a cap-5 boundary (exactly 5 win), and the replay-wins-over-cap precedence case.
- **`replay` sorted-set TTL.** Each `(keyid, scope)` sorted set carries a `PEXPIREAT` set to `(latest expiresAt + setTtlGraceSeconds)` so abandoned tuples evict instead of leaking. Default grace 3600s — cheap because the set is also empty by then (`ZREMRANGEBYSCORE` ran at every insert).
- **Shared default-prefix warn.** Both backends + a future-deduped idempotency backend (#1855) share `src/lib/utils/redis-default-prefix-warn.ts`. One-time `console.warn` when the default `keyPrefix` ("adcp:ctx_meta:" / "adcp:replay:") meets a node-redis client we can confidently see on db 0. Once per process across every Redis backend instance.

## Pre-merge checklist

- [x] `tsc --noEmit` clean
- [x] 42 new Redis tests pass against local Redis (11 pure-function warn + 16 ctx-metadata live + 14 replay live + 1 pre-existing ctx-metadata config)
- [x] 11/11 warn tests pass without `REDIS_URL` (pure introspection, no live connection needed)
- [x] 13/13 pre-existing `postgres-replay-store` / `signing-defaults-and-preset` tests still pass
- [x] 30/30 pre-existing `server-ctx-metadata` tests still pass
- [x] DX smoke: out-of-tree `redisCtxMetadataStore(createClient(...))` and `new RedisReplayStore(createClient(...))` both typecheck without casts
- [ ] CI green
- [ ] Manual review of the `RedisReplayStore` Lua script (correctness of precedence: replay > rate_abuse > ok; `ZREMRANGEBYSCORE` inclusive-on-now matches pg's `> now` semantic)

## Follow-up after #1855 lands

The default-prefix warn in #1855 (idempotency backend) is inlined; this PR extracts it to `src/lib/utils/redis-default-prefix-warn.ts`. A one-commit follow-up will swap the idempotency backend to import the shared helper. Filed mentally as a follow-up task; not blocking either PR.


bokelley:
## Review pass complete (codex)

The Anthropic API was overloaded all afternoon (3 consecutive 529 attempts on three of the four reviewers), so the protocol/code-reviewer/security pass ran via codex (read-only sandbox, OpenAI-side). Sister DX pass via Claude already returned ship-it earlier.

### Verdicts

| Reviewer | Verdict | Notes |
|---|---|---|
| DX | ✅ ship-it (5/5 across the rubric) | Filed #1859 (pg ctx-metadata drops `resource` field, real pre-existing bug surfaced by parity comparison) |
| Protocol (codex) | ✅ ships | All six concerns green: atomicity, precedence, boundary parity, PEXPIREAT grace, cap atomicity, assertFiniteSeconds. "Stronger than the pg sibling's MVCC statement for cap enforcement." |
| Security (codex) | ✅ no findings requiring a code fix | All nine threats LOW/mitigated or MEDIUM/partial. Critically: verifier **fails closed** on Redis unreachability — Redis errors propagate out of verifyRequestSignature, middleware returns 500, handler never runs. The two MEDIUMs are operational (shared-keyPrefix isolation, Redis maxmemory-policy) — both documented; the latter is tracked at #1857. |
| Code review (codex) | ⚠️ caught one fix-before-merge | **Real replay-protection bypass.** See below — fixed in commit `a2087f25`. |

### Fix-before-merge from codex code review (now landed)

**Bug:** `INSERT_LUA` unconditionally did `PEXPIREAT key, (expiresAt + grace) * 1000` on every insert, making the sorted-set eviction time follow the **most recently inserted** nonce, not the latest-expiring member.

**Failure mode:**
1. Insert nonce A with `expiresAt = now + 3600` → set TTL ~1h+grace.
2. Insert nonce B with `expiresAt = now + 60` → PEXPIREAT shrinks set TTL to ~1m+grace.
3. Set evicts at ~1m+grace, taking still-valid nonce A with it.
4. Attacker replays nonce A within its original 1h window → `has()` returns false (key disappeared) → verifier accepts.

This is a **code-level replay bypass**. The pg sibling doesn't have this mode because row expiry is per-nonce.

**Fix** (`src/lib/signing/redis-replay-store.ts`): Lua now reads `PTTL`, compares `(nowMs + currentTtlMs)` vs `(expiresAt + grace) * 1000`, and only calls `PEXPIREAT` if the new value would extend the eviction time forward. `nowMs` passes as `ARGV[6]` from the SDK side rather than using Lua's `TIME` (which makes the script non-deterministic for replication) or `Date.now()` (which would diverge from the surrounding verifier path's clock when tests inject `options.now()`).

**Regression tests** (`test/lib/redis-replay-store.test.js`):
- Long-lived then short-lived insert: TTL stays ~7200s, doesn't shrink to ~3660s; both nonces still `has()`-visible.
- Short-lived then long-lived insert: TTL extends forward to ~7200s. Confirms forward-only logic still extends in the other direction.

### Test status after fix

- 22/22 replay tests pass (16 live + 6 warn-only)
- 74/74 across `redis-replay + ctx-metadata-redis + postgres-replay-store + server-ctx-metadata` — no regressions
- `npm run check:adopter-types` PASS
- `format:check` clean

Ready for merge once CI re-runs and goes green.

bokelley:
Thanks for the thorough review pass. The Lua TTL forward-only fix in `a2087f25` is a meaningful correctness catch — good that it landed before merge. No action needed on my end; standing by for CI green.

---
Session: https://claude.ai/code/session_01YZEDGCVtuuq9k5qpXPjaUQ

---

bokelley:
Acknowledged. Watching for CI green — no further action on my end until then.

---

bokelley:
Understood — standing by for CI.

---

PR #1866, opened 2026-05-19
fix(mcp): forward header-only auth through SingleAgentClient precheck

### LINKED ISSUE #1864: Basic-auth MCP agents broken via SingleAgentClient — Authorization header dropped on getAgentInfo path (affects every executeTask call)
### Summary

Basic-auth MCP agents are completely broken via `SingleAgentClient.executeTask` (and therefore via every public entry point that wraps it — `getAdcpCapabilities`, `executeTaskWithSchema`, the CLI, Addie's `testCapabilityDiscovery`, every direct tool call). The `Authorization: Basic <base64>` header set on `agentConfig.headers` is silently dropped on the `getCapabilities()` precheck that runs before every `executeTask` call. The agent receives an unauthenticated request, returns whatever it serves anonymous callers, and the SDK either errors out or reports the anonymous response as the agent's real state. Curl with the same credentials returns 200; the SDK does not.

**Severity:** every basic-auth MCP agent is unusable via the standard SDK API surface. Originally surfaced via `get_adcp_capabilities` because that's the precheck operation, but every tool call hits the same bug — the precheck fails before the actual tool is ever invoked.

### Wire trace (fetch interceptor)

User runs `adcp <basic-auth-agent> get_adcp_capabilities '{}'`. Same trace would occur for any tool, since `executeTask` always runs the precheck first.

```
### FETCH #1: POST .../mcp    ← initialize, has authorization
###   accept=application/json, text/event-stream | authorization=Basic … | content-type=application/json

### FETCH #2: POST .../mcp    ← notifications/initialized, has authorization
###   accept=… | authorization=Basic … | content-type=… | mcp-protocol-version=2025-11-25

### FETCH #3: GET .../mcp     ← SSE stream open, has authorization
###   accept=text/event-stream | authorization=Basic … | mcp-protocol-version=2025-11-25

### FETCH #4: POST .../mcp    ← getCapabilities precheck, NO authorization
###   accept=application/json, text/event-stream | content-type=application/json
```

Call 4's stack:
```
at mcp.js:538:90  (sizeLimited fetch inside connectMCP)
at StreamableHTTPClientTransport.send (streamableHttp.js:310)
```

Calls 1–3 originate from `connectMCPWithFallbackImpl` (`mcp.ts:~356`) — the discovery probe path. Call 4 originates from `connectMCP` (`mcp.ts:~690`) — the precheck path. Two different code paths to the same transport class; one preserves the header, the other doesn't.

### Why every tool call is affected

`SingleAgentClient.executeTask(taskName, params)` runs these prechecks before dispatching the tool call:
- `validateTaskFeatures(taskName)` — calls `getCapabilities()` to decide which features are gated
- `detectServerVersion()` — calls `getCapabilities()` to negotiate the wire envelope
- `adaptRequestForServerVersion(...)` — calls `getCapabilities()` to decide which fields to strip/adapt

Each of these funnels through `getCapabilities()` → `getAgentInfo()` → `connectMCP()`. The bug bites on the first one. The actual tool dispatch (`callMCPToolWithTasks` → `withCachedConnection` → `connectMCPWithFallback`) uses the **working** code path, but execution never gets there because the precheck throws.

So in practice: basic-auth agents fail on every `executeTask` invocation regardless of which tool is being called. The original title naming `get_adcp_capabilities` was misleading — it's the first thing to fail, not the only thing.

### Root cause

Two contributing defects. Fixing **either alone** does not resolve the symptom — both must be fixed.

**Defect A — `SingleAgentClient.getAgentInfo()` doesn't forward `customHeaders` to `connectMCP`**

`src/lib/core/SingleAgentClient.ts:~1858-1872`:

```ts
const connectOptions = { agentUrl: agent.agent_uri };
if (this.normalizedAgent.oauth_tokens) {
  connectOptions.authProvider = createNonInteractiveOAuthProvider(...);
}
else if (this.normalizedAgent.auth_token) {
  connectOptions.authToken = this.normalizedAgent.auth_token;
}
const { client: mcpClient } = await connectMCP(connectOptions);
```

For basic auth, the CLI intentionally suppresses `agentConfig.auth_token` so the SDK doesn't emit a competing `Authorization: Bearer …` (see `bin/adcp.js:5905-5908` — the `auth_token` field is excluded when `useBasicAuth` is true). The credential lives in `agentConfig.headers.Authorization`. Neither branch above fires, and `this.normalizedAgent.headers` is never passed in. `connectMCP` receives `{ agentUrl }` only.

Same problem for any other header-only auth (`x-api-key`, custom tenant routing headers paired with no token, etc.).

**Defect B — `connectMCP` only sets `requestInit.headers` when `authToken` is truthy**

`src/lib/protocols/mcp.ts:~675-684`:

```ts
if (authProvider) {
  transportOptions.authProvider = authProvider;
}
else if (authToken) {
  const authHeaders = { ...customHeaders, ...createMCPAuthHeaders(authToken) };
  transportOptions.requestInit = { headers: authHeaders };
}
// Note: no `else if (customHeaders)` branch. Headers without a token are dropped.
```

Even if Defect A is fixed and `customHeaders` are passed in, this function ignores them when `authToken` is undefined. Pure-header auth (basic, x-api-key, etc.) is silently dropped.

Contrast with `connectMCPWithFallback` (`mcp.ts:~345`) which takes the merged `authHeaders` directly and unconditionally sets `transportOptions.requestInit = { headers: authHeaders }`. That's why the discovery probe (which uses `connectMCPWithFallback`) works while `getAgentInfo` (which uses `connectMCP`) doesn't.

### Suggested fix

Both defects in one PR:

1. **`SingleAgentClient.getAgentInfo`**: pass `customHeaders: this.normalizedAgent.headers` (or just `headers:`) into `connectOptions`.

2. **`connectMCP`**: lift the `requestInit.headers = ...` assignment out of the `else if (authToken)` branch so it fires whenever `customHeaders` are present, regardless of whether `authToken` is set. Suggested shape:

```ts
const authHeaders = authToken
  ? { ...customHeaders, ...createMCPAuthHeaders(authToken) }
  : { ...customHeaders };
if (Object.keys(authHeaders).length > 0) {
  transportOptions.requestInit = { headers: authHeaders };
}
if (authProvider) {
  transportOptions.authProvider = authProvider;
}
```

(`authProvider` and `requestInit.headers` are not mutually exclusive — the MCP SDK's `_commonHeaders()` spreads `requestInit.headers` over the auth-provider-emitted `Authorization`, so OAuth users with custom routing headers also benefit from this fix.)

### Related

- CLI mislabels basic auth as bearer in the storyboard-run header — filed separately (#1865).
- `bin/adcp.js:5880-5883` has a comment that anticipates this exact failure mode ("suppressing `auth_token` here lets our injected Basic header reach the wire intact") — the suppression is correct, the downstream `connectMCP` path is what breaks the invariant.
- Downstream impact: Addie's basic-auth users get an even more misleading symptom because Addie reports the resulting anonymous capabilities response as if it were the agent's declared state ("agent didn't declare capabilities yet"). Filed against `adcontextprotocol/adcp` as #4807.

### LINKED ISSUE #1865: CLI `storyboard run` header mislabels basic auth as "bearer"
### Summary

When running `adcp storyboard run <agent> --auth-scheme basic --auth user:pass`, the run-header banner prints `Auth: bearer` even though basic auth is correctly resolved and sent on the wire (via the storyboard runner's own header injection at `runner.ts:4313-4322`). The label is purely cosmetic but it's actively misleading during auth debugging — it tricked me into thinking the scheme flag wasn't being picked up.

### Reproducer

```bash
adcp --save-auth myagent https://gw.example.com/mcp \
  --auth 'USER:PASS' \
  --auth-scheme basic

adcp storyboard run myagent
```

Banner output:
```
Running storyboard assessment against https://gw.example.com/mcp
   Protocol: MCP
   Timeout: 120s
   Auth: bearer        ← incorrect; actually using basic
```

### Root cause

`bin/adcp.js:4213-4219`:

```js
const authLabel = !authOption
  ? 'none'
  : authOption.type === 'oauth'
    ? 'oauth (auto-refresh)'
    : authOption.type === 'oauth_client_credentials'
      ? 'oauth client credentials (auto-refresh)'
      : 'bearer';
```

Chained ternary with no `'basic'` case. When `authOption.type === 'basic'`, the chain falls through to the literal `'bearer'` label.

`authOption` shape is set at `bin/adcp.js:4174`:
```js
const { auth: authOption } = buildResolvedAuthOption({...});
```
For basic, `buildResolvedAuthOption` returns `{ auth: { type: 'basic', username, password } }` (`bin/adcp.js:1331`), so `authOption.type === 'basic'` reaches this label code with no matching branch.

### Suggested fix

```js
const authLabel = !authOption
  ? 'none'
  : authOption.type === 'oauth'
    ? 'oauth (auto-refresh)'
    : authOption.type === 'oauth_client_credentials'
      ? 'oauth client credentials (auto-refresh)'
      : authOption.type === 'basic'
        ? 'basic'
        : 'bearer';
```

### Severity

Cosmetic / UX. No functional impact. Filing because it cost ~30 minutes of debugging time during a real basic-auth troubleshooting session — the misleading label sent us chasing the wrong root cause.

bokelley:
## Summary

Closes #1864 (functional) and #1865 (cosmetic) in one PR.

Basic-auth (and any header-only auth) MCP agents were silently broken via `SingleAgentClient.executeTask` — every public entry point that wraps it (`getAdcpCapabilities`, `executeTaskWithSchema`, the CLI, direct tool calls) runs a `getCapabilities()` precheck that dropped the `Authorization` header on the floor. The agent received an unauthenticated request, returned its anonymous response (or 401), and the SDK either errored out or surfaced the anonymous payload as the agent's real state. Curl with the same credentials returned 200; the SDK did not.

## Root cause (per #1864's wire trace)

Two compounding defects, **both required** for a fix:

**Defect A — `SingleAgentClient.getAgentInfo` did not forward `normalizedAgent.headers` to `connectMCP`.** Basic auth intentionally suppresses `auth_token` (so the SDK doesn't emit a competing `Authorization: Bearer …`) and lives entirely on `headers.Authorization`. Neither the `oauth_tokens` nor the `auth_token` branch fired, so `connectMCP` received `{ agentUrl }` only.

**Defect B — `connectMCP` only attached `requestInit.headers` inside `else if (authToken)`.** Header-only auth (no token) was dropped even when `customHeaders` was supplied. Contrast `connectMCPWithFallback` which sets `requestInit.headers = authHeaders` unconditionally — that's why the discovery probe path worked while the precheck path didn't.

## Changes

- `src/lib/core/SingleAgentClient.ts` (`getAgentInfo`): forward `normalizedAgent.headers` as `connectOptions.customHeaders` when present.
- `src/lib/protocols/mcp.ts` (`connectMCP`): build `authHeaders` from `customHeaders` + optional `authToken`-derived headers and set `requestInit.headers` whenever the bag is non-empty — including alongside `authProvider`. The MCP SDK's `_commonHeaders()` spreads `requestInit.headers` under the provider-emitted `Authorization`, so OAuth users with custom routing/tenant headers also benefit from this fix (per #1864's analysis).
- `bin/adcp.js` (storyboard-run banner): add the `'basic'` branch to the auth-label ternary — closes #1865.
- `test/lib/get-agent-info-basic-auth.test.js`: regression test. Loopback MCP server that 401s without `Authorization: Basic`. Asserts every request the SDK makes — including the `tools/list` precheck — carries the header. Verified to fail without **either** defect's fix (negative-checked both branches independently during development).

## Test plan

- [x] New test passes against fix; fails against pre-fix code (verified by reverting each defect's fix independently)
- [x] `npm run test:lib` — 6876/6876 pass, 3 pre-existing skips
- [x] `npm run lint` — 2 pre-existing errors unrelated to these changes
- [ ] Reviewer to verify the OAuth-plus-custom-headers path (now newly allowed by `connectMCP`) doesn't regress any existing OAuth-only caller — the change is additive (headers only attached when present) and the MCP SDK's `_commonHeaders()` precedence puts provider-emitted `Authorization` after `requestInit.headers`, but a second pair of eyes on the OAuth flow is welcome.

### DIFF
--- a/.changeset/basic-auth-mcp-header-forwarding.md
+++ b/.changeset/basic-auth-mcp-header-forwarding.md
@@ -1,0 +1,18 @@
+---
+'@adcp/sdk': patch
+---
+
+fix(mcp): forward header-only auth (basic, x-api-key) through `SingleAgentClient` precheck path
+
+Basic-auth (and any header-only auth) MCP agents were silently broken via `SingleAgentClient.executeTask` — every public entry point that wraps it (`getAdcpCapabilities`, `executeTaskWithSchema`, the CLI, direct tool calls) ran a `getCapabilities` precheck that dropped the `Authorization` header on the floor. The agent received an unauthenticated request, returned its anonymous response (or 401), and the SDK either errored out or surfaced the anonymous payload as the agent's real state. Curl with the same credentials returned 200; the SDK did not.
+
+Two compounding defects, both required for a fix:
+
+- `SingleAgentClient.getAgentInfo` did not forward `normalizedAgent.headers` to `connectMCP`. Basic auth intentionally suppresses `auth_token` (so the SDK doesn't emit a competing `Authorization: Bearer …`) and lives entirely on `headers.Authorization` — neither the `oauth_tokens` nor the `auth_token` branch fired, so `connectMCP` received `{ agentUrl }` only.
+- `connectMCP` only attached `requestInit.headers` inside the `else if (authToken)` branch. Header-only auth (no token) was dropped even when `customHeaders` was supplied.
+
+Both branches are fixed: `getAgentInfo` now forwards `headers` as `customHeaders`, and `connectMCP` builds `requestInit.headers` whenever any headers are present — including alongside `authProvider` (the MCP SDK's `_commonHeaders()` spreads `requestInit.headers` under the provider-emitted `Authorization`, so OAuth users with custom routing/tenant headers also benefit).
+
+Also fixes the cosmetic `adcp storyboard run` banner that labeled basic auth as `"bearer"` (`bin/adcp.js` chained-ternary fell through when `authOption.type === 'basic'`).
+
+Regression test at `test/lib/get-agent-info-basic-auth.test.js` stands up a loopback MCP server that 401s without `Authorization: Basic`, then asserts every request the SDK makes (including the `tools/list` precheck) carries the header. Closes #1864 and #1865.
--- a/.changeset/basic-auth-mcp-header-forwarding.md
+++ b/.changeset/basic-auth-mcp-header-forwarding.md
@@ -11,7 +11,9 @@ Two compounding defects, both required for a fix:
 - `SingleAgentClient.getAgentInfo` did not forward `normalizedAgent.headers` to `connectMCP`. Basic auth intentionally suppresses `auth_token` (so the SDK doesn't emit a competing `Authorization: Bearer …`) and lives entirely on `headers.Authorization` — neither the `oauth_tokens` nor the `auth_token` branch fired, so `connectMCP` received `{ agentUrl }` only.
 - `connectMCP` only attached `requestInit.headers` inside the `else if (authToken)` branch. Header-only auth (no token) was dropped even when `customHeaders` was supplied.
 
-Both branches are fixed: `getAgentInfo` now forwards `headers` as `customHeaders`, and `connectMCP` builds `requestInit.headers` whenever any headers are present — including alongside `authProvider` (the MCP SDK's `_commonHeaders()` spreads `requestInit.headers` under the provider-emitted `Authorization`, so OAuth users with custom routing/tenant headers also benefit).
+Both branches are fixed: `getAgentInfo` now forwards `headers` as `customHeaders`, and `connectMCP` builds `requestInit.headers` whenever any headers are present — including alongside `authProvider`, so OAuth users with custom routing/tenant headers benefit too.
+
+**Precedence note:** the MCP SDK's `_commonHeaders()` spreads `requestInit.headers` **over** any provider-emitted `Authorization` (`new Headers({ ...providerHeaders, ...requestInitHeaders })` — last-write-wins). To prevent a caller-supplied `Authorization` in `customHeaders` from silently overriding the OAuth bearer, `connectMCP` now strips `Authorization` from `customHeaders` when `authProvider` is set. Non-auth headers (routing, tenant ID, x-api-key co-tokens) still flow through. The bearer-token branch is unaffected — its own merge already lays the bearer last, so static `auth_token` takes precedence over any stray `customHeaders.Authorization` (regression test asserts this).
 
 Also fixes the cosmetic `adcp storyboard run` banner that labeled basic auth as `"bearer"` (`bin/adcp.js` chained-ternary fell through when `authOption.type === 'basic'`).
 
--- a/bin/adcp.js
+++ b/bin/adcp.js
@@ -4216,7 +4216,9 @@ async function runFullAssessment(agentArg, rawArgs, parsedOpts) {
         ? 'oauth (auto-refresh)'
         : authOption.type === 'oauth_client_credentials'
           ? 'oauth client credentials (auto-refresh)'
-          : 'bearer';
+          : authOption.type === 'basic'
+            ? 'basic'
+            : 'bearer';
     console.log(`\nRunning storyboard assessment against ${agentUrl}`);
     console.log(`   Protocol: ${protocol.toUpperCase()}`);
     if (storyboards) console.log(`   Storyboards: ${storyboards.join(', ')}`);
--- a/src/lib/core/SingleAgentClient.ts
+++ b/src/lib/core/SingleAgentClient.ts
@@ -2716,8 +2716,15 @@ export class SingleAgentClient {
 
       // Use the shared connectMCP path so both static bearer AND saved OAuth
       // tokens work. OAuth takes the refresh-capable authProvider branch.
+      // Header-only auth (basic, x-api-key, custom routing) lives on
+      // `normalizedAgent.headers` and must be forwarded as `customHeaders` —
+      // basic auth in particular suppresses `auth_token` on purpose so the
+      // SDK doesn't emit a competing `Authorization: Bearer …`.
       const { connectMCP } = await import('../protocols/mcp');
       const connectOptions: Parameters<typeof connectMCP>[0] = { agentUrl: agent.agent_uri };
+      if (this.normalizedAgent.headers && Object.keys(this.normalizedAgent.headers).length > 0) {
+        connectOptions.customHeaders = this.normalizedAgent.headers;
+      }
       if (this.normalizedAgent.oauth_tokens) {
         const { createNonInteractiveOAuthProvider } = await import('../auth/oauth');
         connectOptions.authProvider = createNonInteractiveOAuthProvider(this.normalizedAgent, {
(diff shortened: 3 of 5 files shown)


bokelley:
## Expert review summary

Ran the four Claude expert agents (code-reviewer-deep, security-reviewer-deep, ad-tech-protocol-expert, dx-expert) per the safety-critical PR pattern in CLAUDE.md, plus `npm run review:codex --all --base main` for a second-model opinion.

**Convergence — all four reviewers say "ship"** with two convergent finds, both now addressed in commit `bd3d79cc`:

1. **Inverted precedence claim (code-reviewer + security).** The previous comment in `connectMCP` and the changeset described `_commonHeaders()` as spreading `requestInit.headers` *under* the OAuth provider's `Authorization`. Both reviewers read the actual MCP SDK code (`node_modules/@modelcontextprotocol/sdk/dist/esm/client/streamableHttp.js:58-76`) and confirmed the spread is `new Headers({ ...providerHeaders, ...requestInitHeaders })` — last-write-wins, so `requestInit.headers` is laid *over* the provider's `Authorization`. The latent failure mode: a future caller passing `Authorization` in `customHeaders` alongside `oauth_tokens` would silently override the OAuth bearer. Fixed: comment + changeset now describe the actual precedence.

2. **Defense in depth (security-reviewer).** When `authProvider` is set, `connectMCP` now strips any case-insensitive `Authorization` key from `customHeaders` before merging into `requestInit.headers`. Routing/tenant/x-api-key co-tokens still flow through; OAuth remains the source of truth for the bearer.

**Protocol expert** confirmed this is a wire-conformance fix, not a spec change — MCP defers to RFC 9110 for transport auth, so the handshake (`initialize` / `notifications/initialized` / `tools/list`) carrying the same `Authorization` as `tools/call` is the correct behavior. Also flagged: worth one regression-test assertion that static bearer wins over `customHeaders.Authorization`. Added.

**DX expert** flagged two follow-ups (non-blocking):
- 401 errors from `connectMCP.connect()` rethrow as bare `Error POSTing to endpoint (HTTP 401)` with no auth context. Worth wrapping with `{scheme, header-shape}` hints.
- No `docs/guides/BASIC-AUTH.md`; the `auth_token` suppression invariant at `bin/adcp.js:5905-5908` lives only inline.

Will file these as separate issues after merge.

**Cache key path:** security-reviewer confirmed `getAgentInfo` doesn't enter the connection cache (it `finally`-closes the client), so the existing `extractAuthHeader(authHeaders)` fingerprint logic in `connectionCacheKey` is not impacted by this PR. The fingerprint was already in place anticipating header-only auth on the cached call paths.

**`connectMCPWithFallback`:** code-reviewer confirmed no parallel fix needed — it already sets `requestInit.headers = authHeaders` unconditionally (`mcp.ts:365`), and `callMCPToolImpl` already merges `customHeaders` into `authHeaders` (`mcp.ts:535-539`). Only the `getAgentInfo` precheck via `connectMCP` was broken.

Codex second-opinion review is still running; will append the result when it completes.

bokelley:
Acknowledged. Four-reviewer convergence noted, both finds addressed in `bd3d79cc`, DX follow-ups queued as separate issues — nothing blocking from my side. Will catch the Codex append when it lands.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_013HoVjxaEiKkWpCcXU2j9kp

---

PR #1868, opened 2026-05-19
fix(client): forward custom headers on getAgentInfo path (basic-auth MCP agents broken)

### LINKED ISSUE #1864: Basic-auth MCP agents broken via SingleAgentClient — Authorization header dropped on getAgentInfo path (affects every executeTask call)

bokelley:
Closes #1864

## Summary

Two cooperating defects silently dropped `Authorization: Basic …` headers on the `getCapabilities()` precheck that `SingleAgentClient.executeTask` runs before every tool dispatch, making every basic-auth MCP agent completely unusable.

**Defect A — `SingleAgentClient.getAgentInfo` didn't forward `customHeaders`** (`src/lib/core/SingleAgentClient.ts:~2726`): `connectOptions` was assembled without `this.normalizedAgent.headers`, so `connectMCP` received `{ agentUrl }` only when neither OAuth tokens nor a bearer token were present. Fixed by unconditionally forwarding `normalizedAgent.headers` as `customHeaders`.

**Defect B — `connectMCP` only set `requestInit.headers` inside `else if (authToken)`** (`src/lib/protocols/mcp.ts:~664`): caller-supplied `customHeaders` were silently dropped when `authToken` was absent. Fixed by computing `authHeaders` unconditionally (merged with token headers when present) and setting `requestInit` whenever the result is non-empty. As a bonus, `authProvider` is no longer exclusive with `requestInit.headers` — OAuth callers with custom routing headers (`x-tenant-id`, `x-api-key`) now get them forwarded too.

Also fixed: the initial debug log now emits `authMethod: 'custom'` (rather than `'none'`) when only `customHeaders` carry an `Authorization` header, so basic-auth failures are diagnosable from debug output.

## What was tested

- `tsc --project tsconfig.lib.json` — clean (pre-existing `moduleResolution=node10` deprecation warning is not new)
- `node --test test/lib/connect-mcp-custom-headers.test.js` — 3/3 pass
  - `connectMCP` forwards `Authorization: Basic` with no `authToken`
  - `connectMCP` forwards `x-api-key` with no `authToken`
  - `SingleAgentClient.getAgentInfo` sends `Authorization: Basic` to the MCP server (Defect A end-to-end)
- Full test suite baseline: 669 failures before and after (pre-existing env issues — Redis/Postgres/schema CDN not available in this environment; CI runs the complete suite)

## Pre-PR review

- **code-reviewer**: approved — both blockers (authMethod misreport, missing regression test) resolved; nit about `removeAllListeners` test isolation fixed
- **dx-expert**: approved for the core fix — noted pre-existing `auth_token` + `customHeaders.Authorization` spread-order precedence (same in `callMCPToolImpl`; CLI prevents the combination for basic-auth by suppressing `auth_token`)

> issue). To unblock:
>
>   fix → push.
>   issue.
>
> for context.

Session: https://claude.ai/code/session_01DAtQqcuNuyaFB7EUDVXJH1

---
_Generated by [Claude Code](https://claude.ai/code/session_01DAtQqcuNuyaFB7EUDVXJH1)_

### DIFF
--- a/.changeset/fix-basic-auth-header-1864.md
+++ b/.changeset/fix-basic-auth-header-1864.md
@@ -1,0 +1,25 @@
+---
+"@adcp/sdk": patch
+---
+
+fix(client): forward custom headers to connectMCP on getAgentInfo path
+
+Basic-auth MCP agents were completely broken via `SingleAgentClient.executeTask`.
+Two cooperating defects caused the `Authorization: Basic …` header to be silently
+dropped on the `getCapabilities()` precheck that every `executeTask` invocation
+runs before dispatching the actual tool call:
+
+**Defect A** (`SingleAgentClient.getAgentInfo`): `connectOptions` was built
+without `this.normalizedAgent.headers`, so `connectMCP` received `{ agentUrl }`
+only when neither OAuth nor a bearer token was present. Fixed by always forwarding
+`normalizedAgent.headers` as `customHeaders`.
+
+**Defect B** (`connectMCP`): the `transportOptions.requestInit.headers` assignment
+was nested inside `else if (authToken)`, so caller-supplied `customHeaders` were
+silently dropped whenever `authToken` was absent. Fixed by computing `authHeaders`
+unconditionally (merging with token headers when present) and setting
+`requestInit` whenever the result is non-empty. As a bonus, `authProvider` is no
+longer exclusive with `requestInit.headers` — OAuth callers with custom routing
+headers (x-tenant-id, x-api-key, etc.) now get them forwarded too.
+
+Fixes #1864.
--- a/src/lib/core/SingleAgentClient.ts
+++ b/src/lib/core/SingleAgentClient.ts
@@ -2726,6 +2726,9 @@ export class SingleAgentClient {
       } else if (this.normalizedAgent.auth_token) {
         connectOptions.authToken = this.normalizedAgent.auth_token;
       }
+      if (this.normalizedAgent.headers) {
+        connectOptions.customHeaders = this.normalizedAgent.headers;
+      }
 
       const { client: mcpClient } = await connectMCP(connectOptions);
       try {
--- a/src/lib/protocols/mcp.ts
+++ b/src/lib/protocols/mcp.ts
@@ -653,7 +653,7 @@ export async function connectMCP(options: {
     type: 'info',
     message: `MCP: Connecting to ${baseUrl}`,
     timestamp: new Date().toISOString(),
-    authMethod: authProvider ? 'oauth' : authToken ? 'token' : 'none',
+    authMethod: authProvider ? 'oauth' : authToken ? 'token' : extractAuthHeader(customHeaders) ? 'custom' : 'none',
   });
 
   const mcpClient = new MCPClient({
--- a/src/lib/protocols/mcp.ts
+++ b/src/lib/protocols/mcp.ts
@@ -664,21 +664,28 @@ export async function connectMCP(options: {
   // Build transport options
   const transportOptions: StreamableHTTPClientTransportOptions = {};
 
+  // Set requestInit.headers whenever any custom or auth headers are present.
+  // authProvider and requestInit.headers are not mutually exclusive — the MCP
+  // SDK's _commonHeaders() spreads requestInit.headers alongside any
+  // OAuth-issued Authorization, so OAuth callers with custom routing headers
+  // (x-tenant-id, x-api-key, etc.) also get them forwarded.
+  const authHeaders = authToken ? { ...customHeaders, ...createMCPAuthHeaders(authToken) } : { ...customHeaders };
+  if (Object.keys(authHeaders).length > 0) {
+    transportOptions.requestInit = { headers: authHeaders };
+    debugLogs.push({
+      type: 'info',
+      message: authToken
+        ? 'MCP: Using static token for authentication'
+        : 'MCP: Using custom headers for authentication',
+      timestamp: new Date().toISOString(),
+    });
+  }
   if (authProvider) {
     // Use OAuth provider
     transportOptions.authProvider = authProvider;
     debugLogs.push({
       type: 'info',
       message: 'MCP: Using OAuth provider for authentication',
-      timestamp: new Date().toISOString(),
-    });
-  } else if (authToken) {
-    // Use static token, merged with any custom headers (auth takes precedence)
-    const authHeaders = { ...customHeaders, ...createMCPAuthHeaders(authToken) };
-    transportOptions.requestInit = { headers: authHeaders };
-    debugLogs.push({
-      type: 'info',
-      message: 'MCP: Using static token for authentication',
       timestamp: new Date().toISOString(),
     });
   }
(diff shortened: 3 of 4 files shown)


bokelley:
Superseded by #1866 (commit bae92750), which forwarded header-only auth (basic / x-api-key) through the SingleAgentClient precheck path. Issue #1864 is closed.

bokelley:
Acknowledged — this PR is superseded by #1866. No further action needed here.

---
_Claude Code · Session: https://claude.ai/code/session_01Rbjoo3vWovGoKhYcnfYBZL_

---

PR #1942, opened 2026-05-22
fix(ci): bump tsc heap to 8GiB in check:adopter-types

nastassiafulconis:
unblocks every open PR. the adopter type-check on `main` OOMs against the 3.1.0-beta.3 codegen graph (~25K lines of generated .d.ts) on Node's default 4 GiB heap. failure surfaces as a JS heap exhaustion stack trace rather than a tsc diagnostic, so adopters debugging the published types won't get useful output either.

## what

set `NODE_OPTIONS=--max-old-space-size=8192` on the tsc child process inside `scripts/check-adopter-types.ts`. 8 GiB clears the current surface with headroom; the constant is named (`TSC_HEAP_MB`) and commented so a future reader knows what knob to revisit when the schema grows further.

## why not just bump it everywhere

the typecheck inside the monorepo (`npm run typecheck`) succeeds on 4 GiB - the adopter scenario is special because it loads the published `.d.ts` graph (which inlines transitive types from `core.generated`) without the monorepo's project-wide tsconfig optimizations. scoping the bump to this one script keeps the rest of CI on its current heap budget.

## reproduce

```
git checkout origin/main
npm run check:adopter-types  # OOMs at ~70s
```

with this patch: passes in ~85s on the same workload.

## test plan

- [x] reproduces locally (heap exhaustion stack trace)
- [x] passes locally with the heap bump
- [ ] CI green

### DIFF
--- a/.changeset/ci-adopter-types-heap.md
+++ b/.changeset/ci-adopter-types-heap.md
@@ -1,0 +1,4 @@
+---
+---
+
+ci: bump `tsc` heap to 8 GiB in `check:adopter-types` so the adopter type-check no longer OOMs on the 3.1 codegen graph. No runtime / API impact - empty changeset satisfies the gate.
--- a/docs/TYPE-SUMMARY.md
+++ b/docs/TYPE-SUMMARY.md
@@ -1,7 +1,7 @@
 # AdCP Type Summary
 
-> Generated at: 2026-05-17
-> @adcp/sdk v7.5.0
+> Generated at: 2026-05-22
+> @adcp/sdk v8.1.0-beta.0
 
 Curated reference of the types that matter for using the AdCP client. For full generated types see `src/lib/types/tools.generated.ts` and `src/lib/types/core.generated.ts`.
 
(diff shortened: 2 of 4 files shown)


bokelley:
Superseded by #1937 — same fix (`NODE_OPTIONS=--max-old-space-size=8192` for the `tsc` child process inside `scripts/check-adopter-types.ts`) shipped earlier on that PR as part of the broader 8.1.0-beta.1 train. I lifted your nicer `TSC_HEAP_MB` named-constant style into #1937 (commit 8efbd9b18) — credited you in the message. Thanks for the parallel diagnosis!

Root cause confirmed the same way: published `.d.ts` graph grew to ~25K lines after 3.1.0-beta.3 codegen, adopter scenario uses `strict + skipLibCheck:false` against the tarball (no monorepo project-wide tsconfig amortization), Node default 4 GiB heap exhausts during type instantiation, panics before emitting any TS diagnostic.

bokelley:
Acknowledged — confirmed superseded by #1937. No further action needed here; the fix and the `TSC_HEAP_MB` named-constant style are live on main via that PR.

---

PR #1972, opened 2026-05-24
fix: preserve ProductSchema ZodObject ergonomics

### LINKED ISSUE #1970: DX: ProductSchema shape change in 8.1 silently breaks .extend()/.omit()/.pick() consumers
## Summary

In 8.0.0, `ProductSchema` was declared as a plain `z.ZodObject<{...}>`, so consumers could call `.extend()` / `.omit()` / `.pick()` on it directly.

In 8.1.0-beta.* the shape changed to:

```ts
export declare const ProductSchema: z.ZodIntersection<
  z.ZodUnion<readonly [
    z.ZodRecord<z.ZodString, z.ZodUnknown>,   // V1ProductNamedFormatReferenceSchema
    z.ZodRecord<z.ZodString, z.ZodUnknown>,   // V2ProductInlineFormatDeclarationsSchema
  ]>,
  z.ZodObject<{ /* product_id, name, pricing_options, publisher_properties, ... */ }>
>;
```

This is **internally consistent** — the V1/V2 variants are discriminated by sibling-field presence (`format_ids` vs `format_options`), so the variant Zod schemas are intentionally opaque `ZodRecord<unknown>` containers. The TS type and Zod schema agree.

But it **silently breaks** any downstream that does:

```ts
const MyProductSchema = ProductSchema.extend({ my_field: z.string() })   // ❌ TS2339: Property 'extend' does not exist
```

Affects 56 schemas in 8.1.0-beta.7 (counted via `grep -c 'z.ZodIntersection<z.ZodUnion' schemas.generated.d.ts`).

## Why this matters

`ProductSchema.extend()` is the obvious idiom for any consumer building wider product shapes (internal-API responses, DB rows that wrap AdCP fields, etc.). The 8.0 → 8.1 minor bump shouldn't break this without either:

1. A documented migration path, or
2. A helper exported alongside (e.g., `ProductObjectSchema = ProductSchema._def.right`).

Our codebase had a `getADCPObjectSchema()` helper that already handled `ZodIntersection.right` extraction for pre-3.24 SDKs. The runtime side kept working; only TS overloads needed updating to recognize the new shape (worked example below).

### Option A (lowest impact): export the right-side ZodObject directly

```ts
export const ProductObjectSchema = ProductSchema._def.right
// usage: ProductObjectSchema.extend({ ... })
```

Same for every schema with the new intersection shape. Consumers do the variant-typed discriminating on the wire, then build their internal schema off the `Object` side.

### Option B: ship the helper

```ts
export function getProductObjectSchema(): z.ZodObject<…> {
  return ProductSchema._def.right
}
```

Generic version that works across all the 56 affected schemas.

### Option C (most invasive): make the variant schemas real Zod objects

If you do want V1/V2 to be discriminated at the schema level (not at sibling-field-presence level), declare them as `z.object({ format_ids: z.array(...) })` and `z.object({ format_options: z.array(...) })` and `discriminatedUnion` them. That'd give consumers proper `.extend()` on the whole shape.

I think Option A is the cheapest fix and the right docs change.

## Workaround (what we did)

Added one TypeScript overload to our local helper:

```ts
export function getADCPObjectSchema<L extends z.ZodTypeAny, R extends z.ZodRawShape>(
  schema: z.ZodIntersection<L, z.ZodObject<R>>,
): z.ZodObject<R>
```

That alone took our 8.1 migration from 32 → 19 errors — the remaining 19 are all intentional spec tightening (optional `signals?`, stricter `CreateMediaBuyResponse`, `TaskStatus`/`MediaBuyStatus` separation), not SDK issues.

## Verified on

- `@adcp/sdk@8.0.0` — plain `ZodObject`, `.extend()` works
- `@adcp/sdk@8.1.0-beta.6` — intersection shape, `.extend()` doesn't typecheck
- `@adcp/sdk@8.1.0-beta.7` — identical to beta.6 on this surface

bokelley:
## Summary

Fixes #1970 by preserving `ProductSchema` as a `ZodObject` when the generated schema is only intersecting the real object shape with opaque `Record<string, unknown>` variant markers.

## What changed

- Extends the Zod codegen post-process to unwrap `z.union([RecordUnknown...]).and(z.object(...))` into the object side when every union arm is a redundant `Record<string, unknown>` marker.
- Regenerates `ProductSchema` as a plain `z.object(...)`, restoring `.extend()`, `.omit()`, and `.pick()` for downstream consumers.
- Adds runtime and compile-time regression coverage for `ProductSchema` composition helpers.

## Validation

- `npm run build:lib`
- `npm run typecheck`
- `NODE_ENV=test node --test-timeout=60000 --test-force-exit --test test/lib/zod-schemas.test.js test/lib/inline-enum-arrays.test.js`
- `npm run ci:codegen-strict`
- `npm run ci:docs-check`

## Review

Reviewed by JavaScript/codegen, DX/docs, and code-review expert agents. The initial unrelated `inline-enums.generated.ts` churn was removed after review; the final PR scope is limited to the ProductSchema codegen fix and regression tests.

### DIFF
--- a/.changeset/product-schema-zodobject-ergonomics.md
+++ b/.changeset/product-schema-zodobject-ergonomics.md
@@ -1,0 +1,7 @@
+---
+'@adcp/sdk': patch
+---
+
+Restore `ZodObject` ergonomics for generated schemas whose only intersection arms are opaque `Record<string, unknown>` markers.
+
+`ProductSchema` and related marker-only format schemas now expose object helpers like `.extend()`, `.omit()`, `.pick()`, and `.shape` again without changing runtime validation behavior.
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@adcp/sdk",
-  "version": "8.1.0-beta.7",
+  "version": "8.1.0-beta.8",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@adcp/sdk",
-      "version": "8.1.0-beta.7",
+      "version": "8.1.0-beta.8",
       "license": "Apache-2.0",
       "workspaces": [
         ".",
@@ -7348,7 +7348,7 @@
       "version": "6.0.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@adcp/sdk": "^8.1.0-beta.7"
+        "@adcp/sdk": "^8.1.0-beta.8"
       },
       "bin": {
         "adcp": "bin/adcp.js"
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@adcp/sdk",
-  "version": "8.1.0-beta.8",
+  "version": "8.1.0-beta.7",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@adcp/sdk",
-      "version": "8.1.0-beta.8",
+      "version": "8.1.0-beta.7",
       "license": "Apache-2.0",
       "workspaces": [
         ".",
@@ -7348,7 +7348,7 @@
       "version": "6.0.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "@adcp/sdk": "^8.1.0-beta.8"
+        "@adcp/sdk": "^8.1.0-beta.7"
       },
       "bin": {
         "adcp": "bin/adcp.js"
--- a/scripts/generate-zod-from-ts.ts
+++ b/scripts/generate-zod-from-ts.ts
@@ -552,7 +552,8 @@ function splitTopLevelCommaList(content: string): string[] {
 
 function collectRedundantRecordSchemaNames(content: string): Set<string> {
   const names = new Set<string>();
-  const recordSchemaPattern = /(?:export\s+)?const\s+(\w+)\s*=\s*z\.record\(z\.string\(\), z\.unknown\(\)\);/g;
+  const recordSchemaPattern =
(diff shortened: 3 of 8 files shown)


bokelley:
Issue #1979 consolidates three non-blocking hardening items from review of this PR that overlap its diff. Lightweight enough to fold before merge if convenient — otherwise they'll resurface on #1979 post-merge.

1. **`.d.ts` regression grep is shape-fragile** (`test/lib/zod-schemas.test.js:46`) — pair the negative grep with a positive `_def.typeName === 'ZodObject'` assertion so the test fails closed if Zod adjusts tuple stringification.
2. **Pass 3/4 docstrings missing `z.object(` shape gate** (`scripts/generate-zod-from-ts.ts:585-592`, `:636-645`) — note the body-prefix constraint in the docstring so a future editor doesn't accidentally broaden the right-hand side.
3. **`unionSchemaNames` one-level limit in `collectRedundantRecordUnionSchemaNames`** — pre-existing, low urgency; a TODO comment is sufficient if you touch the area.

_Posted by the triage bot from #1979._

---

bokelley:
All three items from #1979 folded in commit fd68ac6:

1. **Runtime `_def.typeName` assertion** — `test/lib/zod-schemas.test.js:44` test is now async and adds a `_def.typeName === 'ZodObject'` check alongside the negative `.d.ts` grep; fails closed regardless of how Zod serializes the type in declarations.
2. **Docstring body-prefix notes** — both `unwrapRecordUnionIntersections` and `unwrapNamedRecordUnionIntersections` now document the `z.object(` constraint explicitly.
3. **`collectRedundantRecordUnionSchemaNames` TODO** — one-level limit noted with a comment; no behavior change.

Pre-push validation passed (format ✓, typecheck ✓, build ✓). Full test suite will run in CI.

---
_Triaged by Claude Code. Session: https://claude.ai/code/session_01768uE6DmJzUwwq4tpb4hqZ_

---

PR #2021, opened 2026-05-26
feat: update SDK to AdCP 3.1.0-beta.5

bokelley:
## Summary

- Updates the SDK schema pin and generated artifacts to AdCP 3.1.0-beta.5.
- Carries the beta4 format option migration forward to beta5: format_option_id / format_option_refs replace capability_id(s), with compatibility aliases retained.
- Adds the changeset for a minor @adcp/sdk prerelease update and refreshes storyboard/schema drift coverage.
- Fixes expert-reviewed selector semantics so bare string format-option selectors only match product-local declarations; publisher-scoped options require publisher_domain.
- Restores beta.3 compatibility for capability_id/capability_ids helpers and publisher-catalog resolution while keeping beta5 helpers on format_option_refs.
- Adds direct force_upstream_unavailable controller coverage for dispatch, missing params.tool, and unsupported stores.
- Updates the TypeScript reference seller so shared media-buy storyboards get a synchronous media_buy_id and simulated delivery can round-trip viewability metrics.

## Expert Review

- Protocol reviewer caught a selector bug: bare strings could match publisher-scoped format options by format_option_id alone. Fixed and covered.
- Code reviewer caught a beta.3 compatibility regression in the capability helper surface. Fixed and covered.
- Code reviewer asked for direct force_upstream_unavailable coverage. Added dispatch and negative-path tests.

## Notes

- The canonical protocol site tarball was still returning 404 during local sync, so schemas were synced from the signed GitHub dist bundle instead. CI may expose this if the site cache has not caught up.
- The Python reference seller lane is pinned to v6.2.0-beta.3 and marked advisory for this prerelease. Local checks against beta.2, beta.3, and Python main all still fail beta.5 storyboards in the Python adopter, while this repo's TypeScript reference seller now passes. The lane remains visible and uploads artifacts, but it should not block this SDK beta.
- The branch name remains assess-3.1-beta-4 because this workspace already existed on that branch.

## Validation

- npm run build:lib
- npm run typecheck
- npm run format:check
- node --test test/lib/v2-write-side.test.js test/lib/v2-publisher-catalog.test.js
- NODE_ENV=test node --test test/server-test-controller.test.js
- NODE_ENV=test node --test test/lib/adcp-version-option.test.js test/lib/schema-loader-per-version.test.js test/lib/storyboard-version-negotiation.test.js
- ADCP_SKIP_NPM_CI=1 ADCP_SKIP_BUILD=1 STORYBOARD_RESULT_PATH=.context/storyboard-ts-final-rerun.json SELLER_LOG_PATH=.context/storyboard-ts-final-rerun.log scripts/ci/run_storyboard_reference_seller.sh: overall_status=passing, 53 passed, 0 failed, 3 skipped
- npm test: main suite 10712 tests, 10705 pass, 7 skipped; eslint workspace 30/30 pass
- npx changeset status --since=origin/main
- git diff --check


bokelley:
**Triage routing from #2023:** The `idempotency/create_media_buy_missing_key` MCP failure reported in #2023 traces to a gap in `rawMcpProbe` (`probes.ts:308-319`). When a strict MCP server (like the Python seller) returns HTTP 400 `Bad Request: Missing session ID` with a plain-text body, the existing `isSessionInit` check at line 326 (which targets JSON-RPC `-32002`) is never reached — the function already returned at line 311. The error surfaces as bare `"HTTP 400"` rather than the session-init diagnostic, producing a misleading conformance failure.

The fix is a single additional detection block inside the `status >= 400` branch: if the raw response text matches `/missing session/i` (or is non-JSON on a 400), surface the same `session_not_initialized` diagnostic string used at line 335. One-block change to `probes.ts`; non-breaking; no TypeScript reference seller test changes needed.

This matches the "Python storyboard failures marked advisory" note in this PR — folding the fix here before merge would close that gap cleanly. If the scope expansion doesn't fit this PR's timeline, leaving it for a follow-on is fine and the issue will resurface on merge.

---
Triage bot routing from #2023. Session: https://claude.ai/code/cse_015iCSe8GhkKkbDSqS7yHZMP

---

bokelley:
Fix folded in — commit cc103a5.

The root was in the JSON-parse catch block (lines 290–300): a plain-text 400 body like `"Bad Request: Missing session ID"` throws at `JSON.parse`, so the catch returned early with `"Non-JSON response body"` before the `isSessionInit` check at line 326 was ever reached.

Added a `/missing session/i` guard in that catch block — when status ≥ 400 and the raw text matches, it now surfaces the same session-init diagnostic string used on the JSON-RPC path (`rawMcpProbe skips the initialize handshake; strict servers will reject here before auth is evaluated.`). Falls through to the existing `"Non-JSON response body"` message for all other non-JSON 400s.

One-block change to `probes.ts`, non-breaking, no test changes required.

_Triage-managed follow-up. Session: https://claude.ai/code/session_01UnbdpHjUgSTMwzh15JW6un_

---

Issue #2051, opened 2026-05-27
Consider storyboard-level preferred_attestation_mode for upstream traffic


bokelley:
Follow-up split from #2050 / PR #2040 review notes.

Current runner behavior infers digest mode for `query_upstream_traffic` when identifier digests are present and no raw-introspection peer is available. Evaluate whether the storyboard schema should expose an explicit preferred_attestation_mode (or equivalent runner option) instead of relying on that inference.

This likely needs an upstream spec decision before SDK implementation. Acceptance should cover the schema shape, runner request mapping, and backwards-compatible behavior for existing storyboards.

bokelley:
Checked against current 3.1.0-beta.5 schemas and runner code. The SDK currently has only:

- storyboard validation `attestation_mode_required: "raw"`
- controller request `params.attestation_mode: "raw" | "digest"`
- runner inference: raw for payload introspection/raw-required checks, digest for identifier-only checks with digests, raw otherwise

There is no schema field for storyboard-level `preferred_attestation_mode`; implementing it SDK-side now would create a non-schema extension. Opened upstream spec decision issue adcontextprotocol/adcp#5080 for schema/default/mapping definition. Leaving this client issue blocked until that lands.

PR #2052, opened 2026-05-27
fix: harden 3.1 digest attestations and codegen

### LINKED ISSUE #2045: Clarify query_upstream_traffic digest canonicalization
Follow-up from PR #2040 review.\n\nThe SDK computes payload_digest_sha256 over canonical JSON re-serialization of the redacted payload. The 3.1 query_upstream_traffic field prose does not currently pin the canonical form, so independent controllers could compute different payload digests for semantically identical payloads.\n\nAcceptance criteria:\n- Decide whether payload_digest_sha256 should use SDK canonicalJsonStringify, RFC 8785/JCS, or another explicitly specified byte stream.\n- Document the chosen form in SDK JSDoc and/or upstream spec prose.\n- Add a test vector if the digest is meant to be portable across controllers.

### LINKED ISSUE #2046: Document and bound recorder digest attestation edge cases
Follow-up from PR #2040 review.\n\nDigest attestations should make their privacy/performance limits explicit. Review notes called out low-entropy unsalted SHA-256 identifiers, string-only leaf matching, payload_length vs payload_digest byte-stream differences, and no total leaf-count cap in jsonStringLeafDigests.\n\nAcceptance criteria:\n- Add JSDoc/docs explaining string-only identifier proofs and low-entropy digest risks.\n- Cap total visited string leaves in recorder-side digest projection, matching the runner-side bounded posture.\n- Clarify payload_length and payload_digest_sha256 may measure different serialized forms, or align them.\n- Add tests for truncation and documentation-pinned behavior.

### LINKED ISSUE #2047: Tighten digest-mode identifier vector handling
Follow-up from PR #2040 review.\n\nDigest-mode upstream traffic validation currently focuses on string identifiers collected from the request payload/sample request. Edge cases can silently degrade or fail unclearly:\n\n- identifier_paths pointing at response/context values are not collected before query_upstream_traffic.\n- non-string identifier vectors are skipped by digest collection but can still be graded as missing instead of not_applicable.\n- vectors beyond the 64-digest cap can be reported as missing rather than over-cap/inapplicable.\n\nAcceptance criteria:\n- Either reject unsupported identifier_paths at load time or extend collection beyond request payload/sample request.\n- Treat non-hashable digest-mode vectors explicitly.\n- Warn or grade clearly when identifier vectors exceed the digest cap.\n- Add regression coverage for these cases.

### LINKED ISSUE #2048: Harden conditional params promotion in type generation
Follow-up from PR #2040 review.\n\npromoteConditionalParamProperties promotes conditional allOf/then params fields into the flat generated TypeScript surface. It is intentionally narrow today, but future schema branches could collide or use nested then/allOf/else/oneOf forms.\n\nAcceptance criteria:\n- Detect and throw on conflicting promoted params keys rather than silently first-wins.\n- Document the current limitation or extend the walker to nested conditional forms.\n- Mention the TypeScript widening behavior in a migration note if it remains intentional.

### LINKED ISSUE #2049: Consolidate RecordedCall types and digest projection behavior
Follow-up from PR #2040 review.\n\nRecordedCall is defined in both upstream-recorder and testing/test-controller surfaces. The public test-controller type carries extension index signatures, while projectRecordedCall enumerates digest fields and can drop adopter-set extension fields on the digest projection path.\n\nAcceptance criteria:\n- Move RawRecordedCall/DigestRecordedCall/RecordedCall to a single source of truth or re-export one from the other.\n- Decide whether extension fields are supported; either preserve them in digest projection or remove/tighten the index signature.\n- Add coverage for adopter extension fields or their explicit rejection.

### LINKED ISSUE #2050: Track 3.1 compliance harness review nits
Follow-up from PR #2040 review.\n\nRemaining non-blocking review nits should be tracked outside the merge PR:\n\n- productCatalog currently means optimization-metric rollup source, not the full product surface; rename or document the narrower meaning.\n- agentTools late discovery on reused clients can add a network round-trip when _profile.tools is absent; consider an opt-out or documentation.\n- required_any_of_tools requirement_unmet skips report overall_passed: true; document the rollup contract or add an effective_passed signal.\n- Add integration-shaped coverage for skip-cause aggregation and mixed raw/digest upstream validation batches.\n- Add comments for explicit empty supported_optimization_metrics normalization and raw-mode empty payload substitution.\n\nAcceptance criteria: either address these small items directly or split them into narrower implementation issues if any becomes larger than expected.

bokelley:
## Summary

This PR resolves the 3.1 follow-up issues around digest attestations, upstream traffic validation, and conditional params codegen.

- hardens conditional `params` promotion so root `params` refinements do not conflict while true promoted-key conflicts still fail loudly
- makes digest projection use canonical JSON consistently, parse valid JSON string payloads for both payload digests and identifier proofs, and fail rather than emit non-canonical JSON digests
- tightens storyboard upstream traffic grading for mixed raw/digest and raw/non-JSON evidence, including multi-clause `payload_must_contain` cases and actionable JSON pointers
- updates recorder/storyboard docs for digest-mode throwing, malformed JSON strings, and `identifier_paths` request/response/context prefix rejection
- adds regression coverage for the cached 3.1 comply controller schema, digest identifier scans, mixed attestation batches, manual JSON-string payloads, and authoring validation

## Validation

- `npm run ci:quick`
- expert re-review: JavaScript/protocol, code review, and DX/docs all reported no blockers after follow-up fixes

## Follow-up

Split the larger preferred-attestation-mode design question into #2051 so #2050 can close without blocking this hardening pass on a schema/API decision.

Closes #2045
Closes #2046
Closes #2047
Closes #2048
Closes #2049
Closes #2050


bokelley:
Heads-up from triage: issue #2056 ("Expose dropped digest projection count in upstream recorder query diagnostics") proposes adding a `dropped_count` to `UpstreamRecorderQueryResult` + `toQueryUpstreamTrafficResponse` to surface how many entries were elided by the fail-closed canonicalization path this PR introduces.

The scope is small (one new optional field on the result type + noop-recorder update; wire projection gated on the schema allowing it) and the diff naturally belongs here before merge — same file, same author, same feature area. Worth folding if the review iteration permits; otherwise #2056 resurfaces when this PR closes.

---

bokelley:
Folded #2056 into this PR. Two commits pushed to `resolve-more-3-1`:

1. `feat(upstream-recorder): expose dropped_count in query result` — adds `dropped_count: number` to `UpstreamRecorderQueryResult`; the counter increments only when `attestationMode === 'digest'` (explicit guard, not a side-effect of throw absence), so raw-mode queries are structurally guaranteed to return 0.
2. `style: apply prettier formatting` — Prettier formatting fix the pre-push hook caught.

Wire projection onto `toQueryUpstreamTrafficResponse` / `UpstreamTrafficSuccess` is intentionally withheld — the existing `deepEqual` key-order test pins the wire shape, and the spec hasn't defined the field. The existing `validateUpstreamTrafficSuccess` schema test is extended to assert `dropped_count` is absent from the wire shape.

Tests added (3): NaN-payload digest drop (covers the `onError` correlation), raw-mode zero invariant (records a NaN entry to make the title accurate), and noop recorder.

Changeset added: `.changeset/upstream-recorder-dropped-count.md` (minor — additive optional field).

_Applied by [Claude Code](https://claude.ai/code)_

---

bokelley:
Acknowledged. The `dropped_count` design looks well-reasoned — explicit `attestationMode === 'digest'` guard rather than throw-absence inference keeps the raw-mode zero invariant structural, and withholding wire projection until the spec defines the field is the right call (the `deepEqual` key-order test as a wire-shape pin is a good safety net). CI is in progress on the new commits; no action from me until checks complete.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_014MDNQAfQaNpCVWZ2QrUjHW

---

PR #2078, opened 2026-05-28
fix(argus): support fork PR reviews safely

bokelley:
## Summary
- run Argus on pull_request_target so fork PRs can access the review App token and Anthropic secret
- keep checkout pinned to the trusted base SHA and inspect PR changes through GitHub APIs
- add a self-modification gate so Argus does not approve changes to its own workflow or prompt
- avoid local prior-SHA diffs in the skip check, falling back to full review when compare data is unavailable

## Validation
- git diff --check

### DIFF
--- a/.github/workflows/ai-review.yml
+++ b/.github/workflows/ai-review.yml
@@ -19,7 +19,7 @@ name: AI PR Review (Argus)
 # Adapted from adcontextprotocol/adcp's Argus workflow (PR #4816).
 
 on:
-  pull_request:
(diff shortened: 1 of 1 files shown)


bokelley:
Closing as superseded by the current main-branch Argus workflow. main already uses pull_request_target, checks out the trusted base SHA, gates review-workflow modifications, reads the prompt from BASE_SHA, and avoids local prior-SHA diffs in the skip check. This draft branch is now conflicting and older than the main implementation.

Issue #2099, opened 2026-05-28
beta.15: seed storyboard root context and align update_media_buy revision types


bokelley:
## Summary

While upgrading agentic-adapters to `@adcp/sdk@8.1.0-beta.15`, we hit two SDK-side drifts that look worth fixing before the next beta.

## 1. Storyboard root `context:` is present but not seeded into full runs

`sales-social/index.yaml` in beta.15 includes:

```yaml
context:
  governance_agent_url: "https://test-agent.adcontextprotocol.org"
```

and the `sync_governance` step uses:

```yaml
governance_agents:
  - url: "$context.governance_agent_url"
```

But `runStoryboard` appears to initialize context only from `options.context`:

```js
let context = { ...options.context };
```

I don't see the storyboard root `context` merged into the run context before sample request interpolation. In our `runAgainstLocalAgent({ storyboards: ['sales_social'], ... })` path, this still causes:

```text
account_setup/sync_governance
SKIP: unresolved context variables from prior steps: governance_agent_url
```

This seems related to the prior `adcp#3913` / `adcp#3916` work, but the published beta.15 runner still does not consume the new root `context:` block.

Expected behavior:

- Full storyboard runs seed initial context from the storyboard YAML root `context`.
- Caller-supplied `options.context` can still override those defaults.
- `runAgainstLocalAgent` should inherit the same behavior because it delegates to the full storyboard runner.

## 2. `UpdateMediaBuySuccess.revision` is required by JSON schema but optional in generated TS types

The shipped 3.1 beta JSON schema requires `revision` on successful `update_media_buy` responses:

```json
{
  "required": ["media_buy_id", "revision"]
}
```

But the generated SDK types still mark it optional in both the per-tool and aggregate generated declarations:

```ts
export interface UpdateMediaBuySuccess {
  media_buy_id: string;
  revision?: number;
}
```

This made adopter code type-check while failing schema/storyboard validation until we manually made `revision` required in our local adapter response type and returned a deterministic next revision.

Expected behavior:

- `UpdateMediaBuySuccess.revision` should be non-optional in the generated TypeScript types when the selected 3.1 schema requires it.
- If this is intentionally optional at the TypeScript layer, the runtime/storyboard schema should match that contract.

## Versions

- `@adcp/sdk`: `8.1.0-beta.15`
- `adcp_version` in package: `3.1.0-beta.7`

## Impact

[...]

PR #2100, opened 2026-05-28
fix(testing): seed storyboard root context

sangilish:
## Summary

Refs #2099.

This seeds top-level storyboard `context` values into the storyboard runner before applying caller-supplied `options.context`, so YAML defaults are available for sample request interpolation and callers can still override them.

I kept this PR focused on the remaining root-context issue. The `UpdateMediaBuySuccess.revision` 3.1-beta generated type is already required on current `main` after #2086, so this PR does not touch generated type files.

## What changed

- Added `context?: StoryboardContext` to the `Storyboard` type.
- Merged `storyboard.context` before `options.context` in full-run, multi-pass pre-seeding, and single-step runner paths.
- Forwarded alias caches from both root context and caller context after cloning.
- Added regression coverage for root context interpolation, caller override precedence, and single-step execution.

## Testing

- `node --test test/lib/storyboard-root-context.test.js`
- `npm run build:lib`
- `npm run ci:schema-check`
- `git diff --check`
- pre-push validation: format, typecheck, build

`npm run ci:quick` did not complete in this local environment because `test/lib/custom-headers.test.js` hit the suite-level 60s file timeout after all visible assertions had passed (`pass 10272`, `fail 0`, `cancelled 1`). I reran that file directly with a longer timeout and it passed: `node --test-timeout=120000 --test-force-exit --test test/lib/custom-headers.test.js` (11/11).


sangilish:
I have read the IPR Policy.

bokelley:
Thanks for confirming, @sangilish.

---

bokelley (approved):
Reviewed the refreshed branch after merging current main. The patch stays scoped to seeding storyboard root context before caller overrides, includes regression coverage for full-run and single-step paths, and local validation passed: typecheck, build:lib, targeted storyboard-root-context test, commitlint, and diff check.

Issue #2094, opened 2026-05-28
SDK: wholesale feed mirror helper for products/signals


bokelley:
## Context

This replaces the stale `CatalogSync` shape tracked in adcp#4789. The 3.1 protocol no longer has separate `/catalog/events` or `/catalog/subscriptions` endpoints. The current surface is:

- bootstrap via `get_products` with `buying_mode: "wholesale"` and `get_signals` with `discovery_mode: "wholesale"`
- persist `wholesale_feed_version` and `cache_scope` per connected agent/account
- refresh with `if_wholesale_feed_version`; unchanged feeds return `unchanged: true`
- register account-level webhook receivers through `sync_accounts.accounts[].notification_configs[]`
- process `product.*`, `signal.*`, and `wholesale_feed.bulk_change` notifications
- use the wholesale-feed webhook parser/normalizer from #2005 for canonical receiver handling

## Request

Add an SDK helper for maintaining a local mirror of wholesale product/signal feeds. Naming can be decided during implementation (`CatalogSync`, `WholesaleFeedSync`, etc.), but the API should follow the current 3.1 wholesale-feed model rather than the old catalog-events model.

Suggested scope:

- Bootstrap products/signals from wholesale enumeration.
- Maintain in-memory indexes keyed by agent URL/account plus product or signal id.
- Track `wholesale_feed_version` and `cache_scope` for conditional refresh.
- Expose a refresh method that sends `if_wholesale_feed_version` and handles `unchanged: true`.
- Provide optional persistence hooks for stored versions/index state.
- Provide typed event hooks for wholesale-feed webhook notifications.
- Re-bootstrap the affected feed on `wholesale_feed.bulk_change`, using `affected_entity_type` to choose product vs signal.
- Provide a helper for registering `notification_configs[]` through `sync_accounts` where the caller supplies callback URL, subscriber id, and event types.
- Document the receiver/mirror recipe in the SDK quickstart or examples.

## Tests

- Unit coverage for conditional refresh, unchanged responses, re-bootstrap on bulk change, and product/signal notification routing.
- Integration coverage against the training-agent storyboards added in adcp#5056 where practical:
  - `wholesale_feed_products`
  - `wholesale_feed_signals`
  - `wholesale_feed_product_webhooks`
  - `wholesale_feed_signal_webhooks`
  - `wholesale_feed_bulk_webhooks`

## Related

- adcp#4789 — stale umbrella issue, superseded by repo-specific SDK trackers
[...]

PR #2102, opened 2026-05-29
fix(wholesale-feed-sync): harden lifecycle recovery

sangilish:
## Summary

Refs #2094.

This is the first, stability-only slice of the WholesaleFeedSync follow-up stack. It does not add public API surface.

## What changed

- Added a lifecycle epoch guard so `stop()` invalidates in-flight bootstraps and background loops before they can commit mirror state.
- Made bootstrap commits atomic for feed indexes and version/cache metadata, so stopped or failed mixed-feed refreshes do not leave partially advanced tokens.
- Restart capability-refresh upgrades through the internal lifecycle path.
- Added bounded version-mismatch recovery retries with a clear failure when the affected feed version does not advance.

## Testing

Passed:

- `npm run build:lib`
- `NODE_ENV=test node --test-timeout=60000 --test-force-exit --test test/lib/wholesale-feed-sync.test.js`
- `npm run typecheck`
- `npx prettier --check src/lib/wholesale-feed-sync/sync.ts test/lib/wholesale-feed-sync.test.js .changeset/wholesale-feed-sync-lifecycle.md`
- `git diff --check`
- `NODE_ENV=test node --test-timeout=120000 --test-force-exit --test test/lib/custom-headers.test.js`

Additional broader check:

- `npm run test:lib:fast` did not complete under its default 60s per-file timeout because `test/lib/custom-headers.test.js` timed out at 60000ms. The same file passes when run directly with a 120000ms timeout. The WholesaleFeedSync tests passed in the broader run before that timeout.


bokelley:
Pushed one follow-up commit, `fix(wholesale-feed-sync): restart after stopped bootstrap`.

While reviewing the lifecycle recovery patch, I found one remaining start/stop race: if `stop()` invalidated an in-flight initial bootstrap, a subsequent `start()` could still reuse the retired `startPromise` instead of creating a fresh lifecycle. The new test covers `start(); stop(); start()` while the first bootstrap is blocked and verifies only the second lifecycle publishes indexes.

Local verification:
- `npm run typecheck`
- `npm run build:lib`
- `NODE_ENV=test node --test-timeout=60000 --test-force-exit --test test/lib/wholesale-feed-sync.test.js`
- `npx prettier --check src/lib/wholesale-feed-sync/sync.ts test/lib/wholesale-feed-sync.test.js .changeset/wholesale-feed-sync-lifecycle.md && git diff --check`

bokelley:
Acknowledged — the `start(); stop(); start()` race looks like the last gap in the lifecycle epoch guard. The new test pinning the second lifecycle as the only one that publishes indexes is exactly the right coverage shape for this pattern. No further action needed on this thread.

---

bokelley (dismissed):
Reviewed the lifecycle recovery changes and pushed one follow-up regression fix for the stop/start-in-flight bootstrap race. Local targeted validation passed, and all GitHub CI lanes except the still-running Argus code_review job are green.

bokelley:
Addressed the Argus blocker in `fix(wholesale-feed-sync): acknowledge stale version webhooks`.

Change made:
- `recoverFromVersionMismatch` now treats exhausted no-advance conditional reads as an acknowledged stale/duplicate delivery instead of throwing, so the existing `applyWebhook` call-site path marks the webhook processed and avoids seller retry poison loops.
- Updated the stale mismatch regression to verify bounded repair probes occur once and the repeated delivery is deduped without another probe.

Local verification:
- `npm run typecheck`
- `npm run build:lib`
- `NODE_ENV=test node --test-timeout=60000 --test-force-exit --test test/lib/wholesale-feed-sync.test.js`
- pre-push validation passed

bokelley:
Acknowledged — `recoverFromVersionMismatch` treating exhausted no-advance reads as acknowledged stale/duplicate delivery is the right call: it closes the retry poison loop without requiring the caller to distinguish a true version advance from a stale replay. The updated regression verifying a single bounded probe + dedup on repeated delivery gives good coverage of the invariant.

---

PR #2155, opened 2026-06-02
feat: add optional responseEnhancer callback to AdcpServerConfig

LukasGoTom:
Adds an optional responseEnhancer callback to AdcpServerConfig. When provided, the framework invokes it on every response, giving adopters full control over response text summarization. When omitted, no enhancement is performed.

### DIFF
--- a/src/lib/server/create-adcp-server.ts
+++ b/src/lib/server/create-adcp-server.ts
@@ -1546,6 +1546,11 @@ export interface AdcpServerConfig<TAccount = unknown> {
    */
   webhooks?: WebhooksConfig;
   /**
+   * Optional callback to customize how the response is enhanced.
+   * If not provided, no response enhancement will be performed.
+   */
+  responseEnhancer?: (response: McpToolResponse) => void;
+  /**
    * Auto-wire the RFC 9421 request-signature verifier onto the HTTP transport.
    * When set together with `capabilities.specialisms` containing
    * `signed-requests`, `serve()` mounts the verifier as `preTransport` so
@@ -2985,6 +2990,7 @@ export function createAdcpServer<TAccount = unknown>(config: AdcpServerConfig<TA
     validation: validationConfig,
     credentialPolicy,
     testController: testControllerBridge,
+    responseEnhancer,
   } = config;
 
   // One-shot construction-time warn when `testController` is wired without
@@ -3459,6 +3465,9 @@ export function createAdcpServer<TAccount = unknown>(config: AdcpServerConfig<TA
           injectEnvelopeStatusIntoResponse(response);
           injectContextIntoResponse(response, params.context);
           injectVersionIntoResponse(response, servedAdcpVersion);
+          if (responseEnhancer) {
+            responseEnhancer(response);
+          }
           return response;
         };
 


LukasGoTom:
I have read the IPR Policy


just so you know why - we use it like so:


```ts
const server = createAdcpServerFromPlatform(new HelloSellerAgent(), {
  name: 'hello-seller-agent',
  version: '1.0.0',
  taskRegistry: createInMemoryTaskRegistry(),
  stateStore: new InMemoryStateStore(),
  responseEnhancer: (response) => {
    if (response.isError === true) return;
    const sc = response.structuredContent;
    if (!sc || typeof sc !== 'object') return;
    const json = JSON.stringify(sc, null, 2);
    if (!json) return;
    const first = Array.isArray(response.content) ? response.content[0] : undefined;
    if (first && first.type === 'text' && typeof first.text === 'string') {
      first.text = `${first.text}\n\n\`\`\`json\n${json}\n\`\`\``;
    }
  },
});
```


this way claude outputs the result actually

bokelley:
Thanks for the context, @LukasGoTom — that usage example makes the motivation clear. Appending a JSON-formatted `structuredContent` block to the first text part is a natural pattern, and the in-place mutation model (`(response) => void`) looks like exactly the right fit for it.

No code changes implied from the comment — acknowledged for context.

---

bokelley (changes requested):
Thanks for the contribution — the callback shape is useful, and the motivating `structuredContent` → text rendering example makes sense. I think this needs one implementation pass before merge.

Must fix:

1. `responseEnhancer` only runs inside the normal AdCP tool `finalize()` path. Custom tools bypass that path, and `createAdcpServerFromPlatform` installs the framework-owned `tasks_get` tool via `customTools` (`create-adcp-server.ts` even notes that `tasks_get` is registered through this path). `get_adcp_capabilities` is also registered separately. So the PR body's “every response” behavior is not true today. Please either centralize the enhancer so it applies to all intended MCP tool responses, including `tasks_get`/custom tools/discovery, or narrow the public contract and docs/comments to the exact subset that is enhanced.

2. This adds a public `src/lib` API option, so it needs a `.changeset/*.md` entry. Per repo policy, missing changesets on published library behavior are blocking. This looks like a minor changeset to me because it adds an optional public config surface.

Please also add a focused regression test for the chosen contract. At minimum, cover `createAdcpServerFromPlatform(..., { responseEnhancer })` mutating the returned text content, since that is the entrypoint in the motivating example; if the intended contract is truly server-wide, include `tasks_get` or a custom tool as well.

No objection to the overall API direction once those are addressed.

bokelley:
I addressed the requested changes in an upstream branch, but GitHub denied pushing directly to the `gotom-io/adcp-client` fork (403) even though the PR reports `maintainerCanModify: true`.

Draft replacement PR: #2161

What changed there:
- `responseEnhancer` now runs for framework tool responses, generated `get_adcp_capabilities`, custom tools including `tasks_get`, and credential-policy error responses that intentionally bypass the normal finalizer.
- Added the missing minor changeset.
- Added regression coverage for `createAdcpServerFromPlatform` normal tool responses, generated discovery, and `tasks_get`.

Local validation on that branch:
- `npm run typecheck`
- `npx tsc --project tsconfig.lib.json`
- `NODE_ENV=test node --test-timeout=60000 --test-force-exit --test test/server-decisioning-from-platform.test.js`

`npm run build:lib` could not complete locally on this older contributor branch because `schemas:ensure` attempted to sync the old `3.0.12` schema cache and cosign verification failed for that tarball; direct library emit and the focused tests passed.

bokelley:
Acknowledged — #2161 carries the complete fixes (consistent enhancer coverage, changeset, tests). One thing to resolve before #2161 can merge: it currently has a merge conflict with `main` (`mergeable_state: dirty`). A rebase or merge of `main` into `fix/2155-response-enhancer-consistency` should clear it.

Once #2161 is green, this PR can be closed in its favour.

---

bokelley:
Update: the merge conflict in #2161 has been resolved — the `fix/2155-response-enhancer-consistency` branch has been rebased onto current `main` and all 112 tests pass. #2161 is no longer `dirty`; CI can now run.

---

bokelley:
Closing this because the maintainer-applied replacement PR #2161 has been merged. That PR carries the same feature plus the consistency fixes for custom tools and generated discovery responses, tests, and the changeset. Thanks again for the contribution.

PR #2239, opened 2026-06-15
feat(signing): sign/verify webhooks with the request-signing key; deprecate webhook-signing

BaiyuScope3:
## Why

A signer that already publishes an `adcp_use: "request-signing"` key (for outbound RFC 9421 request signing) currently **must also mint and publish a second, dedicated `adcp_use: "webhook-signing"` key** to sign webhooks — the webhook verifier rejects any other purpose at step 8 (`webhook_mode_mismatch`). Maintaining two KMS keys, two rotation procedures, and two JWKS entries is real operational overhead for signers who'd rather reuse one key.

This PR lets a signer **choose**: reuse the existing request-signing key for webhooks, or keep a dedicated webhook-signing key. The verifier accepts both.

> ⚠️ **Proposal / RFC for discussion.** This relaxes a long-standing conformance behavior (negative vector `008-wrong-adcp-use`) and is paired with a spec PR in `adcontextprotocol/adcp`. Happy to adjust the design — see "Open question" below.

## Why it's safe

Cross-protocol confusion is prevented by the RFC 9421 **`tag`** (`adcp/webhook-signing/v1`), which is part of the signed base and checked at **step 3**, plus the mandatory **`content-digest`** coverage at step 6 — **not** by the `adcp_use` key-purpose discriminator at step 8. A captured *request* signature carries `tag=adcp/request-signing/v1` and can never be replayed against the webhook verifier because step 3 rejects it. So step 8's purpose check is operational hygiene, not the load-bearing domain-separation control.

## What changed

- **`webhook-verifier.ts` (step 8):** accept `adcp_use ∈ {webhook-signing, request-signing}`. Any other purpose (`response-signing`, `governance-signing`, unknown) is still rejected with `webhook_mode_mismatch`. The "missing purpose / no `verify` key_op" failure (`webhook_signature_key_purpose_invalid`) is unchanged.
- **`signer.ts` / `signer-async.ts`:** `signWebhook` / `signWebhookAsync` accept the same `{webhook-signing, request-signing}` set via a dedicated webhook purpose gate (the generic single-purpose gates for request/response signing are untouched).
- **Conformance vectors:** former negative `008-wrong-adcp-use` (request-signing key rejected) → positive `008-request-signing-key-reuse` (accepted, original valid Ed25519 signature reused). New negative `008-wrong-adcp-use` covers a `response-signing` key (still rejected at step 8). New `test-response-purpose-2026` key added to `keys.json`.
- **Unit tests** updated/added in `webhook-verifier-error-codes.test.js` and `signing-provider-purpose-gate.test.js`.

A dedicated webhook-signing key remains **RECOMMENDED** (blast-radius isolation, independent rotation) but is no longer **REQUIRED**.

## Open question for maintainers

This implements the **implicit** form: a `request-signing` key is accepted for webhook verification. An **explicit** alternative is to let a JWK declare multiple purposes (`adcp_use` as an array, or a combined value) so reuse is opt-in rather than a request-signing superset. I went with implicit for minimal surface, but happy to switch if you prefer explicit.

## Verification

- New/updated webhook purpose tests pass; the webhook-signing conformance vector suite passes (`positive/008-request-signing-key-reuse`, `negative/008-wrong-adcp-use`).
- `tsc` (via `build:lib`) compiles cleanly.
- Note: two unrelated tests fail in my local sandbox on Node v22.11 (`jose` ESM `require()` and a `revocation_stale` re-map) — both reproduce on a clean checkout without this change, so they're environmental, not introduced here. CI on supported Node will be authoritative.


BaiyuScope3:
**Spec PR merged** (adcontextprotocol/adcp#5552). The final merged decision matches this SDK PR's behavior exactly:
- Webhook verifier step 8 accepts `adcp_use: "request-signing"` + the deprecated `"webhook-signing"`; everything else → `webhook_signature_key_purpose_invalid`. ✅ (unchanged here)
- `webhook_mode_mismatch` stays reserved for the HMAC-vs-9421 selector. ✅
- `adcp_use` enum keeps `"webhook-signing"` as **deprecated** (not removed); removal tracked in adcontextprotocol/adcp#5555. ✅
- Conformance vectors unchanged. ✅

Maintainers added two enforcement surfaces on merge that this PR didn't originally cover — fixed in `ebf5af0`:
1. **Emission grader** (`assert_jwks_purpose` in the storyboard runner) hard-required a `webhook-signing` key in the agent JWKS — would fail a conformant request-signing agent. Now accepts `{request-signing, webhook-signing}`, matching the merged `webhook-emission.yaml`.
2. **Tenant-registry webhook auto-wire** (`assertWebhookSigningUse`) rejected non-`webhook-signing` keys. Now accepts `request-signing` (canonical) + deprecated `webhook-signing`; tests flipped accordingly.

Note: the generated types still carry `webhook-signing` prose in a few schema descriptions (activation-proof challenge, etc.) — those regenerate from the spec schemas on the next schema-sync/version bump, not hand-edited here.

bokelley:
Thanks for the spec-alignment summary, @BaiyuScope3. Noted: adcontextprotocol/adcp#5552 is merged, the PR's verifier behavior and conformance vectors match the final spec, and the two enforcement surfaces (emission grader + tenant-registry auto-wire) are already addressed in `ebf5af0`. The generated-type prose will update on the next schema-sync — no hand-edits needed here. No action required from the triage side; this is ready for maintainer merge review.

---

bokelley:
Acknowledged — noted @bokelley's confirmation that this PR is spec-aligned and ready for maintainer merge review. No triage action outstanding.

---

PR #2240, opened 2026-06-15
feat: strip AdCP 3.1-only fields for pre-3.1 sellers

nastassiafulconis:
## Why

Sellers on AdCP 3.0 reject requests that carry 3.1-only fields. `BrandReference` is a closed object (`additionalProperties: false`) in every version, so the 3.1 inline overrides (`brand_kit_override`, `industries`, `data_subject_contestation`) are rejected by 3.0 sellers on `create_media_buy`, `sync_accounts`, and `get_products`. Separately, the `get_products` discovery webhook (`push_notification_config`, a 3.1 feature) made the SDK throw for pre-3.1 clients. Both block buyers from booking and discovering against the 3.0 sellers that are still the majority in the field.

## What changed

The client now omits 3.1-only fields when the negotiated target is pre-3.1 (the client is pinned below 3.1, or the seller does not advertise 3.1 via `get_adcp_capabilities`), degrading gracefully instead of failing:

- Brand references are reduced to their identity fields (`domain` + `brand_id`); the seller resolves the inline-override subset from `brand.json`. Applied to `create_media_buy`, `get_products`, and `sync_accounts` (`accounts[].brand` on provisioning-mode entries).
- The auto-injected `get_products` discovery webhook is skipped (results polled via `tasks/get`) instead of throwing. An explicit caller-supplied `push_notification_config` on a pre-3.1 client still throws (unchanged).
- Both are surfaced as `debug_logs` drift entries (warn, not silent drop).

The decision is keyed on `shouldOmit31Fields(clientVersion, sellerCapabilities)`, reusing the `supportedVersions` / `buildVersion` the SDK already parses from `get_adcp_capabilities`. It is correct for 3.0-pinned callers today and becomes per-seller automatically when a caller pins to 3.1.

## Test plan

- New: `adcp-version-projection`, `per-seller-brand-projection`, `get-products-webhook-degrade`.
- Updated `get-signals-pre31-wholesale` to expect the degrade behavior (the one case that asserted the old throw on an auto-injected webhook; the explicit-config throw cases are intact).
- Full `npm run test:lib` green; `npm run typecheck` clean; changeset (`minor`) included.


src/lib/utils/adcp-version-config.ts
```diff
@@ -118,6 +118,41 @@ export function isPre31AdcpVersion(version: string | undefined): boolean {
...
+  return (caps.supportedVersions ?? []).some(v => !isPre31AdcpVersion(v));
+}
+
+/**
+ * Whether to omit AdCP 3.1-only request fields from the wire. Omit when the
+ * client is pinned below 3.1, or when the seller does not advertise 3.1
+ * support (legacy 3.0 sellers, and sellers whose capabilities were synthesized
+ * from a tool list rather than declared).
+ */
+export function shouldOmit31Fields(
```
EmmaLouise2018, line 134: Do we still need this?

nastassiafulconis: Still needed - `resolveAdapterKey` in `src/lib/adapters/version/index.ts` imports and calls it to decide which adapter key to apply. it moved from `SingleAgentClient` to the adapter registry.

PR #2263, opened 2026-06-19
feat(substitution): translateUniversalMacros helper + parser fix (helper 1/2: code)

BaiyuScope3:
Seller-side universal-macro translation, plus the parser fix it surfaced.

**`translateUniversalMacros(pixel_url, mapping)` (seller helper).** Translates universal macros in a pixel URL's query-parameter values from a seller mapping. Two arms: **`value`** (literal, RFC-3986 percent-encoded via the shared encoder) and **`native`** (ad-server token, inserted raw). Parameters with an unmapped macro are dropped (reported in `dropped_params` / `unmapped_macros`); already-minted params pass through untouched; macros in key position are not translated. The result also reports **`suspect_native_values`** — `value` entries shaped like a native token (`%%…%%`, `{{…}}`, `${…}`, `[…]`), a likely wrong-arm mapping. Docstring calls out the dropped-consent-macro privacy hazard.

**Parser fix.** `SubstitutionObserver`'s residual-entity check no longer drops legitimate multi-parameter tracker URLs (the named-entity branch now requires a trailing `;`, matching browser decoding) while preserving the scheme-smuggling defense. Security-reviewed; all hardening/SSRF tests pass.

**undici** 6.25.0 → 6.27.0 (within `^6.25.0`) to clear the high-severity advisories flagged by the Security Audit lane.

Full lib suite green. Changeset included (minor).

---

**Removed per review:** the `expect_universal_macro_substituted` storyboard assertion and its runner wiring have been pulled from this PR. As @bokelley noted, it observes a `build_creative` preview, but `{MEDIA_BUY_ID}`/`{PACKAGE_ID}` resolve at serve time and aren't serialized on the wire, so it skip-dominates for conformant sellers. It's deferred to the forthcoming Live Integration (decisioning-roundtrip) check, where the alignment logic is reusable. This PR is now helper + parser fix only.


bokelley:
Reviewed against the schemas alongside adcontextprotocol/adcp#5646. **The helper (`translate.ts`) is good — LGTM on that part.** The `value`/`native` discriminated union, single-pass no-re-expansion (a `{MACRO}` inside a substituted value is encoded as data, not re-expanded), the RFC-3986 unreserved encoder applied to the query value only, and the return-the-report contract are all right. The `SubstitutionObserver` parser fix holds up too — verified the loosened residual-entity rule still blocks scheme smuggling (no legacy HTML entity decodes to `:`/`/`/`\` without a trailing `;`), and the undici bump is reasonable currency.

Two helper polish items (optional, non-blocking):

- A coding agent can silently pick the wrong arm — `{ value: '%%CACHEBUSTER%%' }` would percent-encode an ad-server token and break it. Cheap guard: when a `value` matches a native-token shape (`%%…%%`, `{{…}}`, `${…}`, `[…]`), surface it in the report. A legitimate `value` never has that shape.
- Drop-unmapped is the right default, but worth documenting (or a `dropped_privacy_params` signal) that silently dropping a consent macro on a forgotten mapping is a compliance hazard — callers should inspect `unmapped_macros`.

**On the assertion (`expect_universal_macro_substituted`):** suggest pulling/deferring it along with the storyboard in #5646. Short version (full reasoning over there): it reads a `build_creative` preview to verify `{MEDIA_BUY_ID}`/`{PACKAGE_ID}` substitution, but that's the wrong surface — those macros resolve at serve time, not `build_creative`, and nothing serializes the resolved tracker on the wire, so the check skip-dominates for conformant sellers. The substitution we actually want to certify is a decisioning roundtrip — a Live Integration check (RFC incoming). The good news: the alignment logic isn't wasted — it's the inverse of what the collector side of that check needs (matching macros in a *received* fired pixel vs. a *fetched* preview), so it can be repurposed there.

**Fixture:** to stop JS↔Python encoder drift (Python helper filed at adcontextprotocol/adcp-client-python#956), let's pin both SDKs to a language-neutral test-vector file — I have a draft vector set covering the reserved-char / native-raw / drop / no-re-expansion cases. Want me to PR it here or into the adcp repo under `static/test-vectors/`?

src/lib/substitution/translate.ts
```diff
@@ -0,0 +1,128 @@
...
+/** Result of {@link universal_macro_translation}. */
+export interface TranslateResult {
+  /** The translated URL. Base, path, and fragment are unchanged. */
+  url: string;
+  /**
+   * Keys of query parameters that were dropped due to unmapped macros.
+   * One entry per dropped parameter instance, so a repeated key can appear
+   * more than once.
+   */
+  dropped_params: string[];
```
bokelley, line 83: Per @bokelley's review: documenting the compliance hazard of silently dropping a consent macro on a forgotten mapping (cheaper than a new `dropped_privacy_params` signal field).

```suggestion
  /**
   * Keys of query parameters that were dropped due to unmapped macros.
   * One entry per dropped parameter instance, so a repeated key can appear
   * more than once.
   *
   * Compliance note: a dropped consent macro (e.g. a forgotten `{GDPR}`
   * mapping) silently removes that signal from the pixel rather than
   * failing loudly. Callers should inspect `unmapped_macros` and treat a
   * non-empty result as an integration bug to fix, not a pass-through.
   */
  dropped_params: string[];
```

---

src/lib/substitution/translate.ts
```diff
@@ -0,0 +1,128 @@
...
+    // String.replace with a global regex does not re-scan substituted output,
+    // so macro tokens inside a translated value are never expanded.
+    const translated = value.replace(UNIVERSAL_MACRO, token => {
+      const entry = mapping[token];
+      if (!entry) {
+        // Unreachable: all tokens were verified above, but TypeScript narrows.
+        return token;
+      }
+      return 'native' in entry ? entry.native : encodeUnreserved(entry.value);
+    });
```
bokelley, line 195: Re: @bokelley's first polish item — a coding agent picking `{ value: '%%CACHEBUSTER%%' }` instead of `{ native: ... }` would percent-encode an ad-server token and silently break it.

Leaving this as a sketch rather than a one-click suggestion since it touches the `TranslateResult` shape (a design call for @BaiyuScope3, not mine to force):

```ts
const NATIVE_TOKEN_SHAPE = /^(%%.*%%|\{\{.*\}\}|\$\{.*\}|\[.*\])$/;

// in TranslateResult:
/** value-mapped macros whose value looks like a native ad-server token (likely the wrong mapping arm). */
suspicious_value_macros: string[];

// in the translate loop, when encoding a `value` entry:
if (NATIVE_TOKEN_SHAPE.test(entry.value)) {
  suspicious_value_macros.push(token);
}
```

Marked optional/non-blocking by the reviewer — happy to wire this up in a follow-up commit if you want it in this PR rather than a fast-follow.

---

bokelley (commented):
Two inline suggestions on the `translate.ts` polish items below. Top-level reply to the rest in a moment.

---

bokelley:
Thanks for the detailed pass, @bokelley. Left two inline suggestions on `translate.ts` for the helper polish items — the compliance-note doc addition is a one-click suggestion; the native-token-shape guard is left as a sketch since it adds a field to `TranslateResult` and that's @BaiyuScope3's call on shape/naming, not mine to force via auto-suggestion. (Note: I don't have push access to `BaiyuScope3/universal-macro-substitution` — it's a fork — so I can't land either as a direct commit; suggestions are the closest I can do.)

On deferring `expect_universal_macro_substituted`: the critique holds up against the code — `executeUniversalMacroAssertionStep` resolves `source_path` against the *prior step's* parsed response (e.g. a `build_creative` preview), and if `{MEDIA_BUY_ID}`/`{PACKAGE_ID}` genuinely don't resolve until serve time, this step will skip-dominate (`no_preview_surface`) for conformant sellers rather than catch anything. That's a real scope question though — pulling 326 lines (the assertion + runner wiring + its test file) out of a green PR — and the full reasoning lives in adcp#5646, which is outside this session's repo scope to verify firsthand. I'll leave that call to you and @BaiyuScope3 rather than act on it unilaterally from here.

On the fixture: leaning toward the spec repo (`adcontextprotocol/adcp` under `static/test-vectors/`) rather than this repo — the explicit goal is JS↔Python parity, and a protocol-neutral location means neither SDK takes a dependency on the other's release cadence to consume it. But that's your call to drive since you're the one offering to author it.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_01LgzNUkvXwbTxpVo4pm864n

---

BaiyuScope3:
Thanks @bokelley — acted on all of it:

- **Pulled `expect_universal_macro_substituted`** (assertion + runner wiring + its test, ~800 lines removed). Your serve-time critique holds — observing a `build_creative` preview can't certify substitution that resolves on the wire. Deferring to the Live Integration / decisioning-roundtrip check; the macro-alignment logic carries over to the collector side there. The driving storyboard in adcontextprotocol/adcp#5646 is parked accordingly (now docs-only split out separately).
- **Renamed** `universal_macro_translation` → `translateUniversalMacros` (camelCase, before it ships in a `minor`).
- **Native-token guard**: added `suspect_native_values` to `TranslateResult` — flags any `value` entry shaped like `%%…%%` / `{{…}}` / `${…}` / `[…]` so an agent that picks the wrong arm is surfaced rather than silently percent-encoding a token.
- **Privacy note**: docstring now calls out that an unmapped (e.g. consent) macro is dropped, and tells callers to inspect `unmapped_macros`.
- Documented the value-only scope.

On the **test-vector fixture** for JS↔Python parity: yes please, and `adcontextprotocol/adcp` under `static/test-vectors/` is the right home — protocol-neutral, no cross-SDK release coupling. Happy to consume it here once it lands.

bokelley (approved):
Reviewed the updated branch after resolving the main merge conflict. Local validation on the rebased branch: `npm run build:lib`, `npm audit --audit-level=high`, and `NODE_ENV=test node --test-timeout=60000 --test-force-exit --test test/lib/substitution-hardening.test.js test/lib/substitution-translate.test.js` all pass. The only audit finding is the existing moderate js-yaml advisory below the CI high-severity gate.

PR #2269, opened 2026-06-22
feat(media-buy): reject get_products products without pricing_options

EmmaLouise2018:
## Summary

`pricing_options` is a required, `minItems: 1` field in AdCP 3.1 — a product that advertises no pricing model is **non-transactable**. The client now drops such products from completed `get_products` responses before callers and completion handlers see them.

- New `enforceProductPricingOptions()` wired in at the top of `applyProductPropertyPolicy()` — the universal completion chokepoint — so it covers **all five paths** (sync `executeAndHandle`, polling `waitForCompletion`, `track`, webhook, and the second `executeTask` path) with one integration point, and fires even when no property policy is configured.
- Runs **independent of the response `validation` mode**, so unpriced products are removed even under `responses: 'warn' | 'off'`.
- Mutates the result in place (matching the existing property-policy filter path) to preserve the non-enumerable `match` accessor.
- Outcome recorded in `result.metadata.productPricingPolicy` and a `product_missing_pricing_options` debug-log notice; the webhook path forwards the same summary on `WebhookMetadata`.

**Config**: default-on; opt out with `validation.rejectProductsWithoutPricingOptions: false`.

## Why default-on

`pricing_options` is spec-required (`minItems: 1`), so a compliant seller never trips this. Returning an unpriced product to a buyer is a correctness bug. Bumped `minor`, consistent with how the repo treats spec-conformance tightening (cf. `fix(compliance): emit stripped field notices`). Opt-out is provided for callers deliberately inspecting malformed seller responses.

## Changes

- `src/lib/core/SingleAgentClient.ts` — `enforceProductPricingOptions()`, config option, helpers, wiring.
- `src/lib/core/ConversationTypes.ts` / `src/lib/core/AsyncHandler.ts` — `productPricingPolicy` on `TaskResultMetadata` and `WebhookMetadata`.
- `test/lib/product-pricing-options-rejection.test.js` — new: default drop, empty-array = missing, opt-out, all-priced no-op, webhook path.
- Updated 3 fixture files whose products were spec-invalid (missing `pricing_options`).

## Testing

- `npm run typecheck` — clean
- `npm run lint` — 0 errors
- Full fast suite — **12034 pass, 0 fail**
- Changeset added (`minor`)


bokelley (approved):
Reviewed the updated branch after clearing CI blockers and merging current main. Local validation: `npm run format:check`, `npm audit --audit-level=high`, `npm run build:lib`, and `NODE_ENV=test node --test-timeout=60000 --test-force-exit --test test/lib/product-pricing-options-rejection.test.js test/lib/agent-client-in-process.test.js test/lib/product-property-policy-client.test.js test/lib/v2-getproducts-autowire.test.js` all pass. The remaining npm audit item is the existing moderate js-yaml advisory below the CI high-severity gate.

PR #2285, opened 2026-06-26
feat(v2/projection): bare-id resolver + canonical-only read mode

bokelley:
Two read-side canonical-format migration helpers in `@adcp/sdk/v2/projection`, exported from the package root. Both close gaps adopters hit migrating off legacy `format_ids[]`.

## Issue 1 — public bare-format-id → canonical resolver
Adopters migrating off legacy format storage hold a bare id string (`display_300x250_image`, `video_standard_30s`) with no `agent_url`, and re-implement local `inferFormatKindFromFormatId` heuristics that drift from the canonical registry.

- `resolveCanonicalFormatKind(id, { agentUrl? }): CanonicalFormatKind | null`
- `canonicalDeclarationFromBareId(id, { agentUrl? }): V2ProductFormatDeclaration | null`

Both lift the bare id to `{ agent_url: AAO host (default), id }` and reuse the existing v1→v2 resolution (catalog `canonical:` annotation → registry glob) — one source of truth. **Fails closed to `null`** for unknown, under-specified (`display_300x250`), or foreign-catalog ids; the returned declaration carries `v1_format_ref` so a bare id lifts to a structured ref in one step.

## Issue 2 — canonical-only read mode (omit legacy `format_ids[]`)
`withFormatOptions` / `augmentProductWithFormatOptions` are additive — they preserve `format_ids[]`, so a fully-migrated consumer still gets the stale `{ agent_url, id }` refs and naive code silently bypasses the canonical model.

- `toCanonicalOnlyProduct(product): { product: CanonicalOnlyProduct<P>, diagnostics }`
- `toCanonicalOnlyResponse(response): { response, diagnostics }`

Returns `format_options[]` with `format_ids[]` **dropped**. **Dropping legacy never silently loses a format**: every input `format_id` is either represented in `format_options[]` or surfaced as a diagnostic — `FORMAT_PROJECTION_FAILED` on the v1→v2 path, or the new SDK-local `LEGACY_FORMAT_ID_DROPPED_UNMAPPED` when a v2-native product carries a `format_ids[]` entry no `v1_format_ref` covers. The coverage key includes dimensional discriminators (`WxH::duration_ms`) so multi-size refs are not collapsed. Complements adcontextprotocol/adcp#4842 (write-side non-invertibility) — this is the read-side transparency half.

### Design note
The issue offered `withFormatOptions(resp, { dropLegacyFormatIds: true })`; I chose dedicated functions instead because a boolean that changes the return shape is the repo's documented anti-pattern (PR #1409) — dedicated functions give an honest `Omit<P, 'format_ids'>` return type.

## Review & verification
- Two adversarial multi-lens review workflows (correctness / protocol / DX / conventions, each finding verified by a skeptic). The protocol lens caught a **real silent-loss bug** — the coverage key originally ignored dimensions, so a v2-native product covering `{id,300x250}` but carrying `{id,728x90}` would silently drop the 728x90. Fixed (now mirrors the write-side key) with a regression test.
- Final expert pass (ad-tech-protocol, dx, code-reviewer, security) — **all converged on no blockers**. Addressed the convergent nits: indexed diagnostic `field` for grep-symmetry, an `augment → canonicalize` composition test, stale module-header docstring, `@see` cross-links, and an `agentUrl` footgun clause.
- `tsc`, `prettier --check`, `ci:doc-links` clean. 24 new tests + projection neighbors pass.

Two changesets (both `minor`). Migration coverage in `docs/migration-8.0-to-8.1.md` (subsections + TL;DR + checklist).


bokelley:
## Follow-up considered (deferred, to keep this PR tight)

Reviewer raised: should the bare-id resolver take a disambiguator hint when adopters have one (e.g. a `format_type` field) —

```ts
canonicalDeclarationFromBareId(id, { agentUrl?, assetTypeHint? })
```

— and resolve an under-specified bare id (`display_300x250`) to its disambiguated catalog variant (`display_300x250_image`)?

**Decision: follow-up, not this PR.** This PR is correct and mergeable as-is — under-specified ids fail closed to `null` and adopters suffix on their side. Deferring because the hint needs deliberate design, not an append:

- `display_300x250` has three variants: `_image` (`image`), `_html` (`html5`), `_generative` (`image`, agent_synthesized). `_image` and `_generative` **both** resolve to canonical kind `image`, so a *canonical-kind* hint can't disambiguate them — the hint vocabulary must be the catalog **suffix** (image/html/generative/js), with a normalization layer (`html5`→`html`, `display_tag`→`js`).
- The `_<suffix>` convention is **not spec-normative** (the v1-canonical-mapping registry enumerates zero literal entries); it's an AAO catalog naming pattern the SDK already half-owns via `SIZED_ID_RE` in `catalog.ts`. Having the SDK own it for resolution is reasonable but warrants an explicit spec-triage note.
- Right shape: **hint-aware catalog search** (look up `${id}_${suffix}`), not blind string concat, and **fail closed to `null`** on residual ambiguity or no match — a hint narrows, never fabricates.

Tracked as a follow-up; happy to open a dedicated issue if useful.

bokelley:
Follow-up now **implemented in this PR** (commit `c9fef14`): `canonicalDeclarationFromBareId` / `resolveCanonicalFormatKind` accept `{ assetTypeHint }`, retrying the disambiguated catalog variant `<id>_<suffix>` for an under-specified bare id (e.g. `display_300x250` + `image` → `display_300x250_image`). Hint vocabulary is the catalog suffix (`image`/`html`/`generative`/`js`) with `html5`→`html` / `display_tag`→`js` aliases; direct catalog match wins over a contradicting hint; still fails closed to `null` when `<id>_<suffix>` isn't a catalog entry. 6 new tests.

Issue #2294, opened 2026-06-26
Add RegistrySync SSE transport and resilience tests


bokelley:
## Context

The registry is adding `GET /api/registry/feed/stream` as the Server-Sent Events transport for cursor-based registry feed pages. `GET /api/registry/feed` also now includes required `freshness` metadata.

Registry PR: https://github.com/adcontextprotocol/adcp/pull/5733

## SDK work

- Regenerate/import registry OpenAPI types so `FeedResponse` includes `freshness`.
- Add a stream client for `/api/registry/feed/stream`; generated OpenAPI types alone are not enough because the response is `text/event-stream`.
- Update `RegistrySync` to default to `auto`: try SSE first, fall back to polling `/api/registry/feed` on unsupported transport, 404/406, proxy/network failure, or stream parse failure.
- Persist cursor exactly as today; reconnect with last persisted cursor after disconnect.
- On `cursor_expired`, re-bootstrap then resume.
- Treat cursors as scoped to the same logical `types` subscription.
- Persist `data.cursor` from SSE payload; registry does not use SSE `id:` / `Last-Event-ID` in 3.x.

## Tests

- Mock SSE server emits `feed`, `heartbeat`, disconnect, `cursor_expired`, and malformed data.
- Cursor advances only on `feed`, not heartbeat.
- Reconnect resumes from persisted cursor.
- Stream errors and unsupported endpoint fall back to polling.
- `cursor_expired` triggers re-bootstrap.
- Duplicate or partial pages do not corrupt indexes.
- Freshness metadata is exposed or observable for lag monitoring.

bokelley:
## Triage

**Classification:** Feature request
**Bucket(s):** library, schema-sync
**Status:** ready-for-human

**What the experts said:**
- ad-tech-protocol-expert: SSE-first/poll-fallback is sound, but SSE auth, backoff, and cursor↔`types` binding are unspecified; use a vetted SSE parser, not hand-rolled parsing.
- code-reviewer: split into two PRs (freshness typing vs. SSE transport); `getFeed()` never sets `cursor_expired` on 410 today, so this issue's re-bootstrap trigger is currently dead code.
- dx-expert: default `transport` to `'poll'`, not `'auto'` — silently opening new connections for every caller is a real behavior change; add `activeTransport`/`transportChange` for observability.

**My take:** Direction's sound, but three judgment calls need your sign-off first. @bokelley:

[...]

PR #2297, opened 2026-06-26
feat(registry): SSE feed stream transport + RegistrySync resilience

bokelley:
Implements [adcp-client#2294](https://github.com/adcontextprotocol/adcp-client/issues/2294); tracks registry PR [adcp#5733](https://github.com/adcontextprotocol/adcp/pull/5733) (deployed live).

## What

- **`RegistryClient.streamFeed(query, { signal })`** — opens `GET /api/registry/feed/stream` over a fetch-based SSE reader (carries the bearer, runs under Node; generated OpenAPI types alone aren't enough for `text/event-stream`). Yields typed `feed` / `heartbeat` / `error` messages. Bounded per-frame buffer that **fails closed** on a runaway stream (matching the JSON path's body cap).
- **`RegistryClient.getFeed`** — maps a real HTTP 410 to a recoverable `{ cursor_expired: true }` response so the polling path recovers too (previously threw and looped).
- **`RegistrySync` defaults to `transport: 'auto'`** — tails the feed over SSE, falls back to polling `/api/registry/feed` on unsupported endpoint (404/406), proxy/network failure, or stream parse failure. Modes: `auto` | `stream` | `poll`.
  - Cursor advances only on `feed` (never heartbeat); reconnect resumes from the last persisted cursor; `cursor_expired` (410 or error event) re-bootstraps then resumes; cursors scoped to the configured `types` subscription; `data.cursor` only (no `Last-Event-ID`).
  - `freshness` (`generated_at` / `latest_event_created_at` / `lag_seconds` / `retention_days`) exposed via a `freshness` event + `getFreshness()` / `getLagSeconds()`. **Never fabricated** — surfaced iff the server sent it.

## Schema

`FeedFreshness` is derived from the now-deployed spec; `FeedResponse.freshness` is kept optional only for the synthetic `cursor_expired` marker. Schema cache re-synced to live; `--sync` is idempotent so `ci:schema-check` stays green.

## New surface

Config: `transport`, `types`, `feedPageLimit`, `streamPollIntervalSeconds`, `streamIdleTimeoutMs`, `streamReconnectMinMs`, `streamReconnectMaxMs`, `maxStreamFailures`. Events: `freshness`, `transport`. Accessors: `getTransport()`, `getFreshness()`, `getLagSeconds()`. Exports: `openFeedStream`, `parseSseStream`, `FeedStream*Error`, and the `FeedStreamQuery` / `FeedStreamMessage` / `FeedHeartbeat` / `FeedStreamErrorData` / `FeedFreshness` / `RegistrySyncTransport` types.

## Resilience hardening (from a 4-dimension adversarial review — 12 confirmed findings fixed)

- Generation-guarded `stop()`/`reset()` — honored even mid-bootstrap/rebootstrap; no stale loop resumes after the caller stops.
- Failed re-bootstrap keeps reconnecting in `stream` mode (was: silently died); counts toward fallback in `auto`.
- Repeated `cursor_expired` recovery backs off instead of tight-looping.
- Permanent `400`/`401` is terminal (no hot reconnect/poll loop).
- `feedPageLimit` / `streamPollIntervalSeconds` range-validated at construction.
- Bounded SSE parse buffer (OOM guard).

## Tests

39 new tests (`test/lib/registry-feed-stream.test.js`) cover the parser (multi-line/CRLF/chunk-split/incomplete/byte-cap), the client (status mapping, 429, 410→cursor_expired), and RegistrySync resilience (feed/heartbeat/disconnect/cursor_expired/malformed, cursor-advances-only-on-feed, reconnect-resume, persisted-cursor round-trip, unsupported→poll fallback, fatal 4xx, stop-during-bootstrap, dup/partial pages, idle timeout, never-fabricate-freshness). Full registry suite: 348/348.

## Not yet done (in progress)

Expert-stack review (`REVIEW-STACKS.md`) + `review:codex`, a live end-to-end against the deployed endpoint, and a performance pass — following up on this branch.


bokelley:
## Local verification + expert review (follow-up pushed)

### Ran locally
- **Unit/integration**: 353 registry tests pass (`registry`, `registry-sync`, `registry-feed-stream`, `registry-cli`); full fast suite green (one unrelated `serve.test.js` flake under load, passes in isolation). tsc (lib + full project), prettier, `ci:codegen-strict` clean; schema `--sync` is idempotent.
- **Real loopback HTTP/SSE e2e** (real `fetch`/sockets, not mocked): connect → apply feed → heartbeat-doesn't-advance-cursor → reconnect-resumes-from-persisted-cursor → `stop()` closes the live socket. PASS.
- **Live-network status mapping**: a real request to the deployed `/api/registry/feed/stream` with a bad key → `FeedStreamHttpError(401)` (→ classified terminal). PASS.

### Performance — found & fixed an O(n²)
The SSE parser re-scanned/re-flattened the growing buffer per chunk (the live server sends a page as one big `data:` line). Rewrote to O(n) (fragments joined once on completion):

| Frame / chunking | before | after |
|---|---|---|
| 1.9 MB @ 8 KB chunks | 577 ms | **14 ms** |
| 7.6 MB @ 8 KB chunks | 9,039 ms | **33 ms** |
| 7.6 MB @ 256 B chunks | 22,731 ms | **280 ms** |
| 20k-event page (4.4 MB) | 42 ms | **10 ms** |

### Expert stack (`REVIEW-STACKS.md`: 4 Claude experts + codex second model)
Convergence = real; divergence = judgment.
- **Fixed (convergent/safety):** poll-path failed-rebootstrap no longer parks the engine; registry error strings escaped before logging/Error (security + codex); `drainFeed` bounded against a `cursor_expired` hot-loop + null-cursor guard + gen-guarded persist; parser O(n).
- **Documented (pre-existing replica properties, not regressions):** staleness/enforcement contract on `RegistrySync`/`isAuthorized` (gate on `state` + `getLagSeconds()`); index size bounded only by registry trust.
- **DX polish:** refreshed class JSDoc + example, seconds-vs-ms note, `getFreshness()` null guidance, exported `DEFAULT_MAX_SSE_FRAME_BYTES`.
- **Refuted (left as-is, w/ rationale):** reconnect jitter, redirect bearer-leak (default is `redirect:'error'`), heartbeat-as-progress, double-`runStream`.

Verdicts: protocol = sound-with-caveats; code-review = no blockers; security = ship-safe for a trusted first-party registry over HTTPS; DX = ship-it.

PR #2310, opened 2026-06-30
fix: emit version from capabilities handler

### LINKED ISSUE #2300: createAdcpServer: auto-registered get_adcp_capabilities handler bypasses injectVersionIntoResponse
## Summary

The auto-registered `get_adcp_capabilities` handler in `createAdcpServer` returns its response through `applyResponseEnhancer(capabilitiesResponse(data))` directly — bypassing the dispatch wrapper that calls `injectVersionIntoResponse`. As a result, `adcp_version` is never echoed on the discovery response envelope, even when the seller pins `adcpVersion: '3.1'`.

Every SDK 9.x adopter therefore fails the universal storyboard `version_negotiation/get_capabilities_with_version` with:

> Seller echoes adcp_version on the response envelope (advisory at 3.1, required at 3.2) at /adcp_version

Advisory today, blocking at 3.2 per `compliance/cache/3.1.0/universal/version-negotiation.yaml`.

## Repro

```bash
mkdir tmp-seller && cd tmp-seller
npm init -y && npm i @adcp/sdk@9.2.2
```

```ts
// server.ts
import { createAdcpServerFromPlatform, definePlatform, serve, verifyApiKey } from '@adcp/sdk/server';

const platform = definePlatform({
  capabilities: {
    specialisms: ['signal-owned'] as const,
    supported_versions: ['3.0', '3.1', '3.1-rc'] as const,
    config: null,
  },
  // ... minimal accounts + signals platform
});

serve(({ taskStore }) =>
  createAdcpServerFromPlatform(platform, {
    name: 'repro',
    version: '0.0.1',
    adcpVersion: '3.1',  // explicit pin
    taskStore,
  }),
  { port: 3011, authenticate: verifyApiKey({ keys: { 'test': { principal: 'p' } } }) },
);
```

Probe `get_adcp_capabilities`:

```bash
curl -s -X POST http://127.0.0.1:3011/mcp \
  -H "Authorization: Bearer test" \
  -H "Accept: application/json, text/event-stream" \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc":"2.0","id":1,"method":"tools/call","params":{"name":"get_adcp_capabilities","arguments":{}}}'
```

Observed envelope (truncated):

```json
{"structuredContent":{"status":"completed","adcp":{"major_versions":[3],"supported_versions":["3.0","3.1","3.1-rc"]}, ...}}
```

Note `adcp_version` is absent from `structuredContent`. Every other tool call on the same server emits it correctly — only the discovery surface is broken.

## Root cause (with line numbers, `@adcp/sdk@9.2.2` dist build)

`dist/lib/server/create-adcp-server.js`:

- **Two dispatch wrappers** correctly inject the field:
  - L2090 (main tool dispatch)
  - L2494 (test-controller dispatch)

  Both run `injectContextIntoResponse → injectVersionIntoResponse → applyResponseEnhancer` on the handler result.

- **Auto-registered `get_adcp_capabilities`** handler at L4165–4204 returns:

  ```js
  return applyResponseEnhancer((0, responses_1.capabilitiesResponse)(data));
  ```

  No `injectVersionIntoResponse` call. So `data.adcp_version` is never set on the structuredContent (nor on the L2 text fallback) for the discovery response.

## Workaround adopters need today

A hand-rolled `responseEnhancer` that re-implements `injectVersionIntoResponse`'s structuredContent + L2-text dual-write, gated on the field being absent so it's a no-op for paths that already injected:

```ts
createAdcpServerFromPlatform(platform, {
  // ...
  adcpVersion: '3.1',
  responseEnhancer: (response) => {
    const sc = (response as { structuredContent?: Record<string, unknown> }).structuredContent;
    if (sc && typeof sc === 'object' && !('adcp_version' in sc)) {
      sc.adcp_version = '3.1';
    }
    const content = (response as { content?: Array<{ type: string; text?: string }> }).content;
    if (Array.isArray(content)) {
      const first = content[0];
      if (first?.type === 'text' && typeof first.text === 'string') {
        try {
          const parsed = JSON.parse(first.text) as Record<string, unknown>;
          if (parsed && typeof parsed === 'object' && !('adcp_version' in parsed)) {
            parsed.adcp_version = '3.1';
            first.text = JSON.stringify(parsed);
          }
        } catch { /* not JSON */ }
      }
    }
  },
})
```

Verified: storyboard advisory clears once the enhancer runs.

## Suggested fix

In the auto-handler at L4165–4204, wrap the return through the same path as the dispatch wrappers — minimal patch:

```diff
- return applyResponseEnhancer((0, responses_1.capabilitiesResponse)(data));
+ const response = capabilitiesResponse(data);
+ injectVersionIntoResponse(response, servedAdcpVersion);
+ return applyResponseEnhancer(response);
```

(`servedAdcpVersion` is already in scope from `resolveAdcpVersion(configuredAdcpVersion)` at L1732.)

Alternatively, move version injection into `applyResponseEnhancer` itself so every response path picks it up unconditionally — discovery, auto-registered tools, and any future framework tools all benefit without per-site touch.

## Impact

- **Today (3.1):** advisory failure on every SDK 9.x adopter's compliance run; explains some of the `permanent_advisory` markers in the version-negotiation storyboard.
- **At 3.2 cutover:** advisory promotes to blocking per `version-negotiation.yaml` migration table → every SDK 9.x adopter regresses unless they apply the workaround above.

## Related

- adcp-client#1067 — introduced `injectVersionIntoResponse` (the function that's not being called from the auto-handler).
- adcp-client#1069 — Stage 3 follow-ups umbrella; calls out multi-version response-echo downshift as deferred, but not the discovery-handler bypass.
- adcontextprotocol/adcp#4711 — training-agent (separate hand-rolled impl) missing both halves of the negotiation contract; closed.

## Environment

- `@adcp/sdk@9.2.2` (also verified on 9.0.0)
- AdCP cache version: 3.1.0
- Storyboard test-kits affected: every test-kit that includes the universal `version_negotiation` storyboard (acme-outdoor, nova-motors, bistro-oranje, summit-foods, osei-natural, …)

🤖 Reported via Claude Code

sangilish:
## Summary

- Stamp `adcp_version` on the auto-registered `get_adcp_capabilities` handler before response enhancement.
- Add targeted coverage for a 3.1-pinned server so the discovery response envelope includes `adcp_version`.
- Add a patch changeset.

Fixes #2300.

## Validation

- `npm run build:lib`
- `NODE_ENV=test node --test-timeout=60000 --test test/lib/adcp-version-string-emit.test.js`
- `git diff --check`
- pre-push validation: format check, typecheck, build:lib

### DIFF
--- a/.changeset/capabilities-version-envelope.md
+++ b/.changeset/capabilities-version-envelope.md
@@ -1,0 +1,5 @@
+---
+"@adcp/sdk": patch
+---
+
+Emit `adcp_version` from the auto-registered `get_adcp_capabilities` server handler.
--- a/src/lib/server/create-adcp-server.ts
+++ b/src/lib/server/create-adcp-server.ts
@@ -6289,7 +6289,9 @@ export function createAdcpServer<TAccount = unknown>(config: AdcpServerConfig<TA
       if (ctx !== null && typeof ctx === 'object' && !Array.isArray(ctx)) {
         (data as any).context = ctx;
       }
-      return applyResponseEnhancer(capabilitiesResponse(data));
+      const response = capabilitiesResponse(data);
+      injectVersionIntoResponse(response, servedAdcpVersion);
+      return applyResponseEnhancer(response);
     }) as Parameters<typeof server.registerTool>[2]
   );
   registeredToolNames.add('get_adcp_capabilities');
--- a/test/lib/adcp-version-string-emit.test.js
+++ b/test/lib/adcp-version-string-emit.test.js
@@ -9,6 +9,14 @@ const { test, describe } = require('node:test');
 const assert = require('node:assert');
 
 const { bundleSupportsAdcpVersionField, extractVersionUnsupportedDetails } = require('../../dist/lib/index.js');
+const { createAdcpServer } = require('../../dist/lib/server/create-adcp-server.js');
+
+async function callToolRaw(server, toolName, params) {
+  return server.dispatchTestRequest({
+    method: 'tools/call',
+    params: { name: toolName, arguments: params ?? {} },
+  });
+}
 
 describe('bundleSupportsAdcpVersionField gate', () => {
   test('3.0 stable bundle does not have the field', () => {
@@ -138,5 +146,22 @@ describe('extractVersionUnsupportedDetails', () => {
   test('partial population — supported_versions only', () => {
     const out = extractVersionUnsupportedDetails({ supported_versions: ['3.0'] });
     assert.deepStrictEqual(out, { supported_versions: ['3.0'] });
+  });
+});
+
+describe('createAdcpServer version emission', () => {
+  test('auto-registered get_adcp_capabilities emits adcp_version when pinned to 3.1', async () => {
+    const server = createAdcpServer({
+      name: 'Versioned Seller',
+      version: '1.0.0',
+      adcpVersion: '3.1',
+    });
+
+    const result = await callToolRaw(server, 'get_adcp_capabilities', {});
+    assert.strictEqual(result.structuredContent.adcp_version, '3.1');
+
+    const text = result.content?.find(part => part.type === 'text')?.text;
+    assert.ok(text, 'capabilities response should include a text fallback');
+    assert.match(text, /Agent capabilities/);
   });
 });


bokelley (approved):
Reviewed the server capabilities change and targeted test. The auto-registered get_adcp_capabilities handler now matches the normal response path by injecting adcp_version before response enhancement, and the patch changeset is present. Full CI is green.

PR #2311, opened 2026-06-30
fix: run advertised sync_accounts in storyboards

### LINKED ISSUE #2303: Storyboard runner: sync_accounts blanket-skipped for any seller advertising require_operator_auth=true — including sellers that implement sync_accounts
## Summary

The storyboard runner pre-skips every `sync_accounts` step as `not_applicable` whenever the agent advertises `account.require_operator_auth: true` — even when the agent **does** implement `sync_accounts` AND the storyboard specifically tests `sync_accounts` behavior (`notification_config_lifecycle`, `notification_config_rejections`, `notification_config_event_scope`). The gate conflates two orthogonal concerns:

- **"operators authenticate independently"** (the semantic `require_operator_auth: true` is meant to advertise)
- **"this seller doesn't expose sync_accounts at all"** (the gate's hidden inference)

In explicit-mode sellers that DO want notification_config_lifecycle etc. graded — which is most production sellers wanting 3.1 webhook subscriber lifecycle — the skip is silent and miscategorising. The `not_applicable` reason looks like "this storyboard doesn't apply to your agent" while the truth is "the runner refused to even probe."

## Why this hits more sellers than it looks

`require_operator_auth: true` is auto-derived from `accounts.resolution: 'explicit'` in `dist/lib/server/decisioning/runtime/from-platform.js:464`:

```js
const requireOperatorAuth = platform.capabilities.requireOperatorAuth ?? platform.accounts.resolution === 'explicit';
```

Most production sellers are explicit-mode (Snap, Meta, GAM, every retail-media operator that owns its account_id namespace) — the SDK comment explicitly lists these as the canonical explicit-mode adopters. Every one of them that adds `notification_configs` support (the canonical 3.1 way to receive `creative.*`/`product.*`/`signal.*`/`wholesale_feed.*` webhooks) hits this trap.

At 3.2 cutover when these storyboards' advisory checks promote to blocking, the same sellers will see sudden regressions without understanding why — the `not_applicable` skip is invisible until the cutover flips it.

## Repro

`@adcp/sdk@9.x` (verified on 9.0.0 and 9.2.2), cache 3.1.0:

```ts
// platform.ts
import { definePlatform } from '@adcp/sdk/server';

const platform = definePlatform<null, MyMeta>({
  capabilities: {
    specialisms: ['sales-non-guaranteed'] as const,
    supported_versions: ['3.0', '3.1'] as const,
    supportedBillings: ['agent', 'operator'] as const,
    config: null,
    // capabilities.requireOperatorAuth deliberately NOT set
  },
  accounts: {
    resolution: 'explicit',  // ← triggers the blanket skip
    resolve: async (ref, ctx) => /* ... */,
    list:    async (filter, ctx) => /* ... */,
    upsert:  async (refs, ctx) => /* ← we DO implement sync_accounts */
      refs.map(ref => ({
        account_id: 'acct_001',
        brand: ref.brand,
        operator: ref.operator,
        action: 'created',
        status: 'active',
        notification_configs: ctx.input.accounts[i].notification_configs,
      })),
  },
  sales: /* ... */,
});
```

Run AAO comply or `bunx adcp storyboard run` against this seller @ cache 3.1.0 with `test-kits/acme-outdoor.yaml`. Observe in the summary:

```
notification_config_lifecycle/register_and_echo_paused_subscriber/sync_accounts_create_paused_notification_config: not_applicable
notification_config_lifecycle/replace_pause_and_clear/*: Skipped (prior stateful step "sync_accounts_create_paused_notification_config" skipped (not_applicable); state never materialized)
```

Verify directly via curl that `sync_accounts` accepts the payload and returns the spec-correct response with the persisted `notification_configs[]` echoed back. The seller works. The runner refused to probe.

## Root cause

`dist/lib/testing/storyboard/runner.js:3218-3245`:

```js
if (effectiveStep.task === 'sync_accounts') {
    const rawCaps = options._profile?.raw_capabilities;
    if (rawCaps !== undefined) {
        const requireOperatorAuth = resolveCapabilityPath(rawCaps, 'account.require_operator_auth');
        if (requireOperatorAuth === true) {
            const detail = `Agent declared explicit account mode (require_operator_auth: true); ` +
                `sync_accounts is not applicable — list_accounts is the correct tool for this account shape.`;
            // ... returns skip_reason: 'not_applicable'
        }
    }
}
```

The detail message says `"list_accounts is the correct tool for this account shape"`. That's a fair assumption for the buyer's `account discovery` flow. It is **not** a fair assumption for `sync_accounts.notification_configs[]` lifecycle, because `list_accounts` does not register subscribers — `sync_accounts` does. Same SDK comment in `from-platform.js:447-457` acknowledges the two signals exist for orthogonal reasons but treats their union as if they implied the same semantic.

## Workaround for adopters

Explicit override `capabilities.requireOperatorAuth: false`:

```ts
capabilities: {
  // ...
  supportedBillings: ['agent', 'operator'] as const,
  requireOperatorAuth: false,  // ← override the auto-derivation
  config: null,
},
```

The SDK's derivation uses `??` so the explicit `false` wins. This is the cleanest workaround AND semantically more accurate for any seller without operator OAuth (most demo/single-tenant sellers and most pass-through retail-media operators). The wire bit goes away from `get_adcp_capabilities`, the runner stops blanket-skipping, sync_accounts steps run normally.

That worked for us (purrsonality-seller) and unblocked `notification_config_lifecycle`, `notification_config_rejections`, `notification_config_event_scope` — three 3.1 universal storyboards.

## Suggested fix

Gate the blanket-skip on **whether the agent advertises `sync_accounts` as a tool** rather than on `require_operator_auth`:

```diff
 if (effectiveStep.task === 'sync_accounts') {
-    const rawCaps = options._profile?.raw_capabilities;
-    if (rawCaps !== undefined) {
-        const requireOperatorAuth = resolveCapabilityPath(rawCaps, 'account.require_operator_auth');
-        if (requireOperatorAuth === true) {
-            const detail = `Agent declared explicit account mode (require_operator_auth: true); ` +
-                `sync_accounts is not applicable — list_accounts is the correct tool for this account shape.`;
-            // ... return not_applicable
-        }
-    }
+    // Skip only when the agent doesn't advertise sync_accounts at all —
+    // covered by the generic missing-tool branch a few lines below. Agents
+    // that DO expose sync_accounts (notification_configs lifecycle, payment
+    // terms updates, etc.) get their storyboard graded regardless of the
+    // account-discovery model they declare.
 }
```

The generic "agent doesn't advertise tool" branch right below (`if (options.agentTools && !options.agentTools.includes(effectiveStep.task))`) already handles the case where the agent legitimately doesn't expose `sync_accounts`. The blanket-skip-on-require_operator_auth path is the bug.

If the original intent was to surface a clearer skip reason than `missing_tool` for explicit-mode adopters that legitimately don't implement `sync_accounts`, the check should be inverted: skip when `agentTools` doesn't include `sync_accounts` AND `require_operator_auth: true`, with that more specific detail message.

## Storyboards affected (cache 3.1.0)

- `universal/notification-config-lifecycle.yaml` — entire storyboard (phase 1 step 1 skipped → phase 2 cascade)
- `universal/notification-config-rejections.yaml` — duplicate_subscriber_id phase
- `universal/notification-config-event-scope.yaml` — reject_media_buy_anchored_event_type phase
- `universal/wholesale-feed-*-webhooks.yaml` (likely; not directly tested but uses sync_accounts to register subscribers)

All currently advisory at 3.1 → blocking at 3.2 per `notification-config-lifecycle.yaml` migration table.

## Related

- adcp-client#2300 — sister bug (auto-registered `get_adcp_capabilities` bypasses `injectVersionIntoResponse`); same shape of "SDK pre-step logic conflates two orthogonal signals"
- adcp-client#1069 — Stage 3 follow-ups umbrella

## Environment

- `@adcp/sdk@9.0.0` and `@adcp/sdk@9.2.2` (verified on both)
- AdCP cache version: 3.1.0
- Reproduced via AAO `evaluate_agent_quality` and `bunx adcp storyboard run` (local)

🤖 Reported via Claude Code (purrsonality-seller debug session 2026-06-27)

sangilish:
## Summary

- Narrow the storyboard runner's explicit-account `sync_accounts` skip so it only applies when the agent does not advertise `sync_accounts`.
- Keep existing `not_applicable` cascade behavior for explicit-mode sellers that omit the tool.
- Add a regression proving explicit-mode sellers that do advertise `sync_accounts` actually run the step.
- Add a patch changeset.

Fixes #2303.

## Validation

- `npm run build:lib`
- `NODE_ENV=test node --test-timeout=60000 --test test/lib/storyboard-cascade-skip-on-skip.test.js`
- `NODE_ENV=test node --test-timeout=60000 --test test/server-decisioning-capability-projections.test.js`
- `git diff --check`
- pre-push validation: format check, typecheck, build:lib

### DIFF
--- a/.changeset/sync-accounts-explicit-tool.md
+++ b/.changeset/sync-accounts-explicit-tool.md
@@ -1,0 +1,5 @@
+---
+"@adcp/sdk": patch
+---
+
+Run storyboard `sync_accounts` steps for explicit-account sellers that advertise the tool.
--- a/src/lib/testing/storyboard/runner.ts
+++ b/src/lib/testing/storyboard/runner.ts
@@ -3823,12 +3823,14 @@ async function executeStep(
     };
   }
 
-  // Account-mode capability gate: sync_accounts is exclusive to implicit mode
-  // (require_operator_auth: false). When the seller declared explicit mode
-  // (require_operator_auth: true), sync_accounts does not apply — grade
-  // not_applicable rather than missing_tool so adopters can distinguish
-  // "your capability declaration says this path isn't yours" from "you forgot
-  // to implement a required tool."
+  // Account-mode capability gate: when the seller declared explicit mode
+  // (require_operator_auth: true) and does not advertise sync_accounts,
+  // sync_accounts does not apply — grade not_applicable rather than
+  // missing_tool so adopters can distinguish "your capability declaration
+  // says this path isn't yours" from "you forgot to implement a required
+  // tool." If the seller explicitly advertises sync_accounts, run it: account
+  // discovery mode and account-level write surfaces such as notification
+  // config registration are orthogonal.
   //
   // list_accounts is NOT gated here: it appears in audience_sync and other
   // storyboard flows regardless of account mode, so it is always applicable.
@@ -3839,7 +3841,8 @@ async function executeStep(
     const rawCaps = options._profile?.raw_capabilities;
     if (rawCaps !== undefined) {
       const requireOperatorAuth = resolveCapabilityPath(rawCaps, 'account.require_operator_auth');
-      if (requireOperatorAuth === true) {
+      const syncAccountsAdvertised = options.agentTools?.includes('sync_accounts') === true;
+      if (requireOperatorAuth === true && !syncAccountsAdvertised) {
         const detail =
           `Agent declared explicit account mode (require_operator_auth: true); ` +
           `sync_accounts is not applicable — list_accounts is the correct tool for this account shape.`;
--- a/test/lib/storyboard-cascade-skip-on-skip.test.js
+++ b/test/lib/storyboard-cascade-skip-on-skip.test.js
@@ -645,7 +645,7 @@ const MATRIX_ROWS = [
         { id: 'account_setup', title: 'account setup', steps: [step('sync', 'sync_accounts')] },
         { id: 'consume', title: 'consume', steps: [step('assert', '__test_assert')] },
       ]),
-      advertised: ['sync_accounts', '__test_assert', 'get_adcp_capabilities'],
+      advertised: ['__test_assert', 'get_adcp_capabilities'],
       profileExtras: { raw_capabilities: { account: { require_operator_auth: true } } },
     }),
     assert: result => {
@@ -670,7 +670,7 @@ const MATRIX_ROWS = [
         },
         { id: 'consume', title: 'consume', steps: [step('assert', '__test_assert')] },
       ]),
-      advertised: ['sync_accounts', '__test_fail', '__test_assert', 'get_adcp_capabilities'],
+      advertised: ['__test_fail', '__test_assert', 'get_adcp_capabilities'],
       profileExtras: { raw_capabilities: { account: { require_operator_auth: true } } },
     }),
     assert: result => {
@@ -698,7 +698,7 @@ const MATRIX_ROWS = [
         },
         { id: 'consume', title: 'consume', steps: [step('audience', '__test_assert')] },
       ]),
-      advertised: ['sync_accounts', '__test_setup', '__test_assert', 'get_adcp_capabilities'],
+      advertised: ['__test_setup', '__test_assert', 'get_adcp_capabilities'],
       profileExtras: { raw_capabilities: { account: { require_operator_auth: true } } },
     }),
     assert: result => {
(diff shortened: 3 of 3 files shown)


bokelley (approved):
Reviewed the runner change and regression coverage. The new behavior is scoped correctly: explicit-account sellers still get  marked not_applicable when they do not advertise the tool, but sellers that explicitly advertise  now run the step. Full CI is green and the patch changeset is present.

PR #2332, opened 2026-07-06
fix: allow larger adagents validation bodies

### LINKED ISSUE #2331: Allow validateAdAgents callers to opt into larger adagents.json bodies
## Summary

`validateAdAgents()` currently lets callers configure `maxBodyBytes`, but the
maximum configurable value is `2 MiB`. That is too small for large manager /
network `adagents.json` files in the wild.

Concrete repro: Raptive/CafeMedia's live manager file at:

```text
https://cafemedia.com/.well-known/adagents.json
```

is currently about `3.37 MiB`:

```bash
curl -sS --max-time 30 https://cafemedia.com/.well-known/adagents.json | wc -c
# 3370695
```

Calling the validator with the current maximum still fails:

```ts
await validateAdAgents('cafemedia.com', {
  timeoutMs: 10_000,
  maxBodyBytes: 2 * 1024 * 1024,
})
```

Observed error:

```text
adagents.json fetch failed: [SSRF refused] Response body exceeded 2097152 bytes
```

## Why this matters

Large manager-domain files are expected for networks that authorize thousands of
publisher properties. In `scope3data/agentic-api`, AI-2458 adds an advisory
network-membership resolver that must derive the authoritative network member
set from the manager's `adagents.json`. For Raptive, the SDK validator currently
fails closed before the file can be parsed, even though the file is reachable
and valid JSON.

We temporarily worked around this in the agentic-api PR with a scoped,
bounded, SSRF-guarded direct fetch for the advisory resolver path:

```text
https://github.com/scope3data/agentic-api/pull/5764
```

It would be cleaner and safer to keep using the SDK validator for this path.

## Requested change

Please raise the **maximum configurable** body limit for `validateAdAgents()`
from `2 MiB` to something like `10 MiB`, while keeping the default small.

Recommended behavior:

- Keep the default at the current safe value (`256 KiB`), so most callers do not
  silently accept larger responses.
- Allow explicit opt-in up to `10 MiB` through `maxBodyBytes`.
- Continue enforcing the limit while streaming, including chunked responses.
- Preserve the existing SSRF and redirect protections.

## Suggested tests

- Default behavior still rejects a large response.
- `maxBodyBytes: 10 * 1024 * 1024` accepts a valid ~3.5 MiB `adagents.json`.
- A response over the configured limit is rejected while streaming.
- Existing SSRF/private-address and redirect-policy tests continue to pass.

bokelley:
## Why

Large manager/network `adagents.json` files exist in production. Raptive's
`cafemedia.com` file is currently about 3.37 MB, but `validateAdAgents()` capped
explicit opt-in body sizes at 2 MiB. That forced downstream callers that need
authoritative network membership to bypass the validator for large files.

Fixes #2331.

## What Changed

- Raised the `validateAdAgents()` maximum configurable `maxBodyBytes` value from
  2 MiB to 10 MiB.
- Kept the default response body cap unchanged at 256 KiB.
- Expanded the managerdomain fallback custom-cap test to use a network-scale
  3.5 MB response.
- Updated invalid option validation expectations for the new explicit maximum.

## Context

This unblocks replacing a scoped workaround in
scope3data/agentic-api#5764 with the SDK validator once a release containing
this change is available.

## Validation

- `npm run build:lib`
- `env NODE_ENV=test node --test-timeout=60000 --test-force-exit --test test/lib/validate-adagents.test.js`
- pre-push hook passed: format check, typecheck, build

### DIFF
--- a/.changeset/large-adagents-body-cap.md
+++ b/.changeset/large-adagents-body-cap.md
@@ -1,0 +1,5 @@
+---
+'@adcp/sdk': patch
+---
+
+Allow validateAdAgents callers to opt into larger adagents.json response bodies.
--- a/src/lib/discovery/validate-adagents.ts
+++ b/src/lib/discovery/validate-adagents.ts
@@ -102,7 +102,7 @@ export interface ValidateAdAgentsOptions {
 
 const DEFAULT_TIMEOUT_MS = 10_000;
 const DEFAULT_MAX_BODY_BYTES = 256 * 1024;
-const MAX_CONFIGURABLE_BODY_BYTES = 2 * 1024 * 1024;
+const MAX_CONFIGURABLE_BODY_BYTES = 10 * 1024 * 1024;
 
 const FETCH_HEADERS = {
   Accept: 'application/json, text/plain, */*',
--- a/test/lib/validate-adagents.test.js
+++ b/test/lib/validate-adagents.test.js
@@ -129,11 +129,11 @@ describe('validateAdAgents — discovery_method', () => {
   test('rejects invalid maxBodyBytes values before fetching', async () => {
     await assert.rejects(
       () => validateAdAgents('example.com', { maxBodyBytes: Infinity }),
-      /maxBodyBytes must be an integer between 1 and 2097152/
+      /maxBodyBytes must be an integer between 1 and 10485760/
     );
     await assert.rejects(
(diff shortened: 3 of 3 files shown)


bokelley:
Closing as superseded by #2335, which merged the same validateAdAgents maxBodyBytes cap increase plus additional direct-path and chunked streaming regression coverage.

PR #2339, opened 2026-07-07
fix: stop injecting deprecated top-level packages[].buyer_ref on v3 create/update_media_buy

ohalushchak-exadel:
## Why

The request normalizer copied `context.buyer_ref` up to a top-level `packages[].buyer_ref` on every `create_media_buy` / `update_media_buy` — unconditionally, and before any server-version detection ran. Spec-compliant v3 receivers validate strictly against the 3.0 package schema (which removed the top-level field) and reject the request with:

```
INVALID_REQUEST: packages.0.buyer_ref: Extra inputs are not permitted
category: invalid_request
```

The client's own comment on the promotion asserted that new servers would still accept the extra field, but the 3.0 spec removed it and any strict v3 receiver (Pydantic `extra="forbid"`, Zod strict, etc.) correctly rejects it. The bug was invisible against permissive receivers that ignore unknown fields and only surfaced when calling a strict v3 seller.

Repro: build a v3 `create_media_buy` request with per-package `context: { buyer_ref: '...' }` (the AdCP 3.x correlation slot), send it to a v3 seller running strict Pydantic. Client-side normalization injects the deprecated top-level `buyer_ref`, receiver rejects. Both sides on `@adcp/sdk@10.0.1`.

## What Changed

- `src/lib/utils/request-normalizer.ts` — removed the `context.buyer_ref → packages[].buyer_ref` copy from `normalizePackageParams` (was unconditional) and the parallel request-level copy from `normalizeRequestParams`. Both now leave `context.buyer_ref` where the caller put it. Comments explain that the promotion moved into the v2.5 adapter, which is version-gated.
- `src/lib/utils/creative-adapter.ts` — `adaptPackageRequestForV2` and `adaptCreateMediaBuyRequestForV2` (both already gated on `serverVersion !== 'v3'` at the caller in `SingleAgentClient.adaptRequest`) now consider `context.buyer_ref` as a derivation source for the v2.5 top-level `buyer_ref`, sitting between caller-supplied top-level and `idempotency_key` in the priority chain. Legacy v2.5 sellers keep receiving the top-level field they expect.
- Tests updated in `test/lib/v3-compatibility.test.js` — the assertions that expected the normalizer to promote `context.buyer_ref` now assert that it does NOT (v3 wire shape is preserved), and new v2-adapter tests pin the moved derivation behavior (request-level and per-package).
- Small comment update in `test/lib/request-validation.test.js` — the existing test at line 209 keeps working because a caller-supplied top-level `buyer_ref` is still legal; comment now reflects why.

Derivation priority in the v2 adapter (unchanged for callers who explicitly set a top-level `buyer_ref`):

1. Caller-supplied top-level `buyer_ref`
2. `context.buyer_ref` (new here — was previously done unconditionally by the normalizer)
3. `idempotency_key`
4. Per-package: `${parentBuyerRef}-${index}`

The replay-dedupe contract (same v3 input → same v2.5 `buyer_ref`) is preserved.

## Test plan

- [x] `npm run build` — clean
- [x] `npx tsc --noEmit` — clean
- [x] `npm test` — full suite green (0 failures)
- [x] `npx eslint src/lib/utils/{request-normalizer,creative-adapter}.ts` — no new errors (pre-existing `no-explicit-any` warnings only)
- [x] Updated the 5 tests that asserted the removed behavior; added 3 tests that pin the moved behavior in the v2 adapter (request-level context, priority vs caller-supplied top-level, per-package context)
- [x] Changeset added (`patch`)

### DIFF
--- a/.changeset/gate-buyer-ref-promotion-on-v2.md
+++ b/.changeset/gate-buyer-ref-promotion-on-v2.md
@@ -1,0 +1,5 @@
+---
+'@adcp/sdk': patch
+---
+
+Client: stop injecting a deprecated top-level `packages[].buyer_ref` on `create_media_buy` / `update_media_buy` requests to v3 sellers. The request normalizer previously copied `context.buyer_ref` up to a top-level `buyer_ref` on every call, unconditionally — before the version gate ran. Spec-compliant v3 receivers (which validate strictly against the 3.0 package schema that removed the top-level field) rejected the request with `INVALID_REQUEST: packages.0.buyer_ref: Extra inputs are not permitted`. The promotion has been moved into the v2.5 adapter (`adaptCreateMediaBuyRequestForV2`), which is already gated on `serverVersion !== 'v3'`, so legacy servers still receive the top-level field they expect while v3 sellers see the correct wire shape.
--- a/src/lib/utils/creative-adapter.ts
+++ b/src/lib/utils/creative-adapter.ts
@@ -75,9 +75,13 @@ export interface PackageAdapterContext {
  * Converts creative_assignments to creative_ids (dropping weight and placement_ids).
  * Strips v3-only package fields (optimization_goals, catalogs).
  *
- * When `ctx` is provided and the input has no `buyer_ref`, derives one as
- * `package.idempotency_key || ${ctx.parentBuyerRef}-${ctx.index}` so v2.5
- * package-request validation passes. Caller-supplied `buyer_ref` always wins.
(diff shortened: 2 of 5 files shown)


ohalushchak-exadel:
Re-opening from an upstream branch — write access has been granted so a fork is no longer needed.

bokelley:
Noted — welcome to push the branch directly. Let me know if you'd like a re-review once the new branch is up.

---

Issue #2344, opened 2026-07-09
Make @adcp/sdk tree-shakeable (ESM + `sideEffects`)


andybevan-scope3:
## Summary

Make `@adcp/sdk` tree-shakeable by shipping **ESM** and marking the package side-effect-free, so `import { anything } from '@adcp/sdk'` (or a subpath) comes in lean automatically and bundlers drop everything unused — no per-thing export wiring.

This is the **general/durable** fix. A tactical `./enums` export is tracked separately and should land first; this work eventually makes per-entry lean shims unnecessary.

## Problem

The SDK publishes as **CommonJS** (`"type": "commonjs"`, `tsconfig` `module: commonjs`) with **no `"sideEffects"` field**. Every runtime entry is a CJS barrel that `__exportStar(require(...))`s its siblings, so bundlers must assume side effects everywhere → no tree-shaking. Importing one named export pulls the whole barrel and everything it transitively `require()`s.

**Measured:** `import { EventTypeValues } from '@adcp/sdk/types'` bundles to **~1.84 MB and includes zod** (esbuild, minified, tree-shaking on) for a 31-string array. Every lean need currently becomes a bespoke curated subpath, which doesn't scale.

## The change

Ship ESM (ESM-only preferred — see risk below) and mark the package side-effect-free.

1. **Emit ESM.** Today the lib build is `tsc --project tsconfig.lib.json` with `module: commonjs`. Move to a bundler (tsup/rollup/esbuild) that emits ESM `.js` + `.d.ts` (and CJS `.cjs` only if we decide to stay dual).
2. **`"sideEffects"`** in `package.json` — see the audit below; not a blanket `false`.
3. **Rework the `exports` map** — each of the ~30 entries (`.`, `./types`, `./client`, `./server`, `./signing`, …) gets `import` → ESM, `types` → `.d.ts` (+ `require` → CJS if dual).

## Feasibility findings (empirical spike on branch `cape-town`)

Ran the lib under `tsc` NodeNext with `type: module` (the real ESM classification):

- **1,977 extensionless relative imports across 405 files** (0 use `.js`) → ~1,808 `tsc` extension errors + 235 unresolved-module errors (directory imports like `./registry` → `./registry/index.js`). This is the dominant blocker.
  - A **bundler sidesteps it entirely** (output is bundled, extensions resolved internally — no source edits).
  - Raw `tsc` NodeNext would require a **405-file codemod** to add extensions. Prefer the bundler.
[...]

PR #2355, opened 2026-07-10
feat!: ship tree-shakeable dual ESM + CJS build (#2344)

andybevan-scope3:
## What this does (#2344)

Makes `@adcp/sdk` tree-shakeable: ships a **dual ESM + CJS** build and marks the package `sideEffects`-aware, so importing one symbol no longer drags in the whole library. `require('@adcp/sdk')` and the CLI keep working. Major version (the `exports` map changes for every consumer).

**The win (measured):** `import { EventTypeValues } from '@adcp/sdk/types'` → **~0.6 KB, zod absent** (was **~1.9 MB with zod**).

## Status — all green ✅

| check | result |
|---|---|
| Fast test gate | 12,519 pass / 0 fail |
| Slow / CLI tests | 50 pass / 0 fail |
| typecheck, build, ESM+CJS runtime, CLI | pass |

## How it works

- **tsup** with `bundle: false` → one output file per source module (mirrors the old `tsc` layout, so tree-shaking works across the preserved graph and the CLI/tests that deep-import build internals still resolve).
- `esbuild-fix-imports-plugin` adds ESM import extensions + directory `/index` resolution; a small companion plugin does the same for **dynamic `import()`** (the standard plugin misses those, which is what caused lazy `await import()` to fail — now fixed).
- An ESM banner supplies per-file `__dirname` / `require` from `import.meta.url` (per-file, not a shared shim chunk).
- `exports`: `import` → `.mjs`, `require` → `.js`, `types` → `.d.ts`. `type` stays `commonjs`. Declarations via `tsc --emitDeclarationOnly`.
- **Dual-package safety:** the request-signing and response-capture `AsyncLocalStorage` stores are anchored on the global symbol registry (a process that loads both builds shares one instance — no silently-unsigned requests); the verified-signature brand uses `Symbol.for`.
- Default invariants register via an explicit `registerDefaultInvariants()` (plus a load-time self-call) instead of a bare side-effect import, so bundlers can't drop them.

## Notes for review

- Build tool is **tsup** (the standard choice). rollup `preserveModules` was evaluated as an alternative and also works; tsup wins on config simplicity.
- `default-invariants.ts` shows a large diff — it's a mechanical wrap of the existing registrations into `registerDefaultInvariants()` (review with `git diff -w`).

### DIFF
--- a/src/lib/canonical-references/index.ts
+++ b/src/lib/canonical-references/index.ts
@@ -10,7 +10,11 @@
  */
 import { createHash } from 'crypto';
 import Ajv from 'ajv';
-import Ajv2020 from 'ajv/dist/2020';
+// Loaded via `require` (extension-lenient CJS resolution) rather than a
+// bare-specifier `import` — ajv ships no `exports` map, so Node's ESM resolver
+// would demand an explicit `ajv/dist/2020.js`. The ESM build gets a
+// `createRequire` shim, so `require` is available in both formats.
+const Ajv2020 = require('ajv/dist/2020') as typeof import('ajv/dist/2020').default;
 import addFormats from 'ajv-formats';
 import { ssrfSafeFetch, SSRF_TRANSIENT_CODES, SsrfRefusedError } from '../net/ssrf-fetch';
 import {
@@ -752,7 +756,7 @@ function getDraft07Ajv(): Ajv {
   return ajv;
 }
 
-function getDraft2020Ajv(): Ajv2020 {
+function getDraft2020Ajv(): InstanceType<typeof Ajv2020> {
   const ajv = new Ajv2020({
     strict: false,
     allErrors: true,
--- a/.changeset/package-artifact-verification.md
+++ b/.changeset/package-artifact-verification.md
@@ -15,6 +15,10 @@ Harden the publish artifact and correct a peer-dependency floor.
   `import` condition now points at `enums.mjs`, and `enums`, `testing/personas`,
   and `server/legacy/v5` gained the `typesVersions` entries they were missing.
 
+Raising the `@modelcontextprotocol/sdk` peer floor is a breaking-shaped change; it
+is safe here only because it is co-released with the major bump from the
+tree-shakeable dual-build changeset.
+
 Verification tooling only (devDependencies, no new runtime deps): `npm run
 check:package` (publint + attw) and `npm run verify:package` (clean-room
 dual-format smoke against the packed tarball with peers at their range floors),
--- a/.changeset/package-artifact-verification.md
+++ b/.changeset/package-artifact-verification.md
@@ -1,0 +1,21 @@
+---
+'@adcp/sdk': patch
+---
+
+Harden the publish artifact and correct a peer-dependency floor.
+
+- Raise the `@modelcontextprotocol/sdk` peer floor from `^1.17.5` to `^1.24.0`.
+  The main entry eagerly loads `server/tasks`, which imports
+  `@modelcontextprotocol/sdk/experimental/tasks/*`; those subpaths first ship in
+  1.24.0, so any consumer importing `@adcp/sdk` on an older MCP SDK hit
+  `ERR_MODULE_NOT_FOUND`. Same class of too-low pin as the `@a2a-js/sdk` fix.
+- Emit ESM-format type declarations (`.d.mts`) alongside `.d.ts` and give every
+  subpath export per-condition types, so the ESM `import` condition no longer
+  resolves to a CJS declaration ("masquerading as CJS"). `@adcp/sdk/enums`'s
+  `import` condition now points at `enums.mjs`, and `enums`, `testing/personas`,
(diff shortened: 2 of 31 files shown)


src/lib/canonical-references/index.ts
```diff
@@ -10,7 +10,7 @@
  */
 import { createHash } from 'crypto';
 import Ajv from 'ajv';
-import Ajv2020 from 'ajv/dist/2020';
+import Ajv2020 from 'ajv/dist/2020.js';
```
benminer: I think we could turn on verbatim module syntax to avoid the extension imports?

    "verbatimModuleSyntax": true in tsconfig -- should confirm with claude and test

andybevan-scope3: Good call on wanting to drop the hardcoded `.js` — done, though via a different mechanism than `verbatimModuleSyntax`.

I tested `verbatimModuleSyntax: true` (+ `moduleResolution: bundler`) directly with the extension removed: esbuild still emits `from "ajv/dist/2020"`, and Node still throws `ERR_MODULE_NOT_FOUND`. The `.js` here is a Node **runtime** ESM-resolution requirement for a deep path into a package with **no `exports` map** (ajv), which no tsconfig/compiler flag changes — and we emit JS via esbuild, not tsc, so `verbatimModuleSyntax` wouldn't touch the specifier anyway.

What *does* drop it: `require('ajv/dist/2020')`. CJS resolution is extension-lenient, and the ESM build already carries a `createRequire` shim (same way we load ajv's JSON meta-schema). Switched both ajv deep imports (`2020`, `2019`) to `require` — built ESM is now `const Ajv2020 = require("ajv/dist/2020")`, no extension, resolves in both ESM and CJS, full suite green. Done in the latest push.

andybevan-scope3: Confirmed with Claude + checked the specifics — `verbatimModuleSyntax` won't achieve this here, it's orthogonal to the problem.

`verbatimModuleSyntax` enforces `import type` and stops TS from eliding imports; it does **not** add file extensions or change how a bare deep-specifier resolves. The actual issue is that `ajv` ships **no `exports` map** (`require('ajv/package.json').exports` → undefined), so in the ESM (`.mjs`) build Node's ESM resolver rejects a bare `import … from 'ajv/dist/2020'` — it demands the explicit `ajv/dist/2020.js`. Turning on `verbatimModuleSyntax` would still emit the extensionless specifier and still fail.

Two things that *do* work:
1. Explicit extension: `import Ajv2020 from 'ajv/dist/2020.js'`. `ajv/dist/2020` sets `__esModule` + `exports.default`, so default-import interop would resolve — but it leans on esbuild's interop of a CJS `export =` module.
2. The `require()` + `createRequire` shim already in place, which resolves uniformly in both the CJS and ESM builds without depending on that interop.

I'd keep the `require()` shim — it's the safer, format-agnostic option and needs no extension bookkeeping. Resolving.

benminer (dismissed):

andybevan-scope3:
Thanks @aao-secretariat — addressed the MUST FIX and the changeset nit in `6728ce2f`.

**MUST FIX — forgeable verified-signature brand (`buyer-agent.ts:403`):** reverted the brand from `Symbol.for('@adcp/sdk.verifiedHttpSig')` back to a module-private `Symbol('@adcp/sdk.verifiedHttpSig')`, and removed the `Symbol.for` justification comment so the definition matches the `:382-388` docstring again. Agreed on the reasoning: `markVerifiedHttpSig` (auth-signature.ts) and `isVerifiedHttpSig` (here) resolve from the same `@adcp/sdk/server` copy on any ordinary import, so the cross-copy mismatch the change was guarding against doesn't arise in practice — and even if it did, a private-`Symbol` mismatch fails **closed** (a valid credential is rejected with a surfaced error), not open. Revert is the fail-closed answer, no globalThis rendezvous needed. Verified: `tsc --noEmit` clean, `buyer-agent` credential-synthesis + registry suites green (45/45).

**Minor nit #1 — peer-floor bump across changeset categories:** added a line to `package-artifact-verification.md` noting the `@modelcontextprotocol/sdk` floor raise is safe only under the co-released major.

**Non-blocking follow-ups** (assertion registry not globally anchored; the `create-adcp-server.ts` `require`→`import` circular-import note): agreed these are worth tracking; leaving them out of this PR and will file as issues for symmetry rather than expand the diff here.

The build machinery, ALS anchoring, and changeset audit you confirmed are unchanged.

bokelley:
Thanks for the rundown. The `Symbol.for` revert is the right call — fail-closed on cross-copy mismatch is exactly the correct default, and the auth-signature/buyer-agent pair resolving from the same copy in normal usage means there's no practical upside to the global registry for this brand. The changeset note on the peer-floor bump is a clean addition. Deferring the assertion-registry and circular-import threads to follow-up issues keeps this diff focused.

Ready for merge when CI is green.

---

bokelley:
Acknowledged — awaiting CI green before merge. No action needed from this routine.

---

benminer (approved):

PR #2366, opened 2026-07-17
feat: add per-PR npm beta releases via beta-release label

benminer:
## Why
There was no way to publish a testable npm build of a specific PR before merging it, so changes could only be validated after they'd already landed on `main` and shipped to `latest`.

## What Changed
Adds a `beta-release.yml` workflow: labeling a PR `beta-release` publishes a Changesets snapshot prerelease of that PR's exact code under a `pr-<N>` npm dist-tag (reusing the existing `scripts/publish-adcp-release.ts` release script via `ADCP_NPM_TAG`), republishing on every push and commenting install instructions on the PR. Promotion is just merging the PR normally — the real `latest` version is cut fresh by the existing changesets + `release.yml` flow. Includes a best-effort dist-tag cleanup job (gated on an `NPM_DIST_TAG_TOKEN` secret still to be provisioned), guards against no-op publishes when a PR is missing a real changeset, and rejects running during Changesets pre-release mode. Docs added to `docs/development/NPM-DIST-TAGS.md` distinguishing this from the existing long-lived beta-channel pattern in `v8.0-beta-plan.md`.

## Note
`NPM_DIST_TAG_TOKEN` (for dist-tag cleanup) and npm trusted-publisher registration for `beta-release.yml` (alongside the existing `release.yml` entry) still need to be provisioned on npmjs.com before this fully works end-to-end.

### DIFF
--- a/.changeset/config.json
+++ b/.changeset/config.json
@@ -7,5 +7,9 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["@adcp/client"]
+  "ignore": ["@adcp/client"],
+  "snapshot": {
+    "useCalculatedVersion": true,
+    "prereleaseTemplate": "{tag}-{commit}"
+  }
 }
--- a/.changeset/tiny-deserts-build.md
+++ b/.changeset/tiny-deserts-build.md
@@ -1,0 +1,4 @@
+---
+---
+
+Add a `beta-release` PR label that publishes a one-off npm snapshot prerelease of the labeled PR under a `pr-<N>` dist-tag for end-to-end testing before merge. No library behavior change — CI/tooling only.
--- a/.github/workflows/beta-release.yml
+++ b/.github/workflows/beta-release.yml
@@ -2,7 +2,7 @@ name: Beta Release
 
 on:
   pull_request:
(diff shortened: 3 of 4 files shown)


bokelley:
Not merging this into the current release train as-is. `main` is in Changesets pre-release mode for the v13 RC, so the new workflow deliberately rejects every labeled PR; the failure text and docs also still point at the old v8 beta plan. More importantly, the npm trusted-publisher registration called out in the PR body is not provisioned, so the documented publish path cannot succeed end-to-end yet.

Please rebase onto current `main`, update the pre-mode/docs language for the current release process, provision the `beta-release.yml` trusted publisher (and decide/provision the cleanup token), then validate one labeled PR through publish and cleanup. Leaving this open for that follow-up rather than merging dormant, unverified release infrastructure.

bokelley:
Closing this unmergeable release-infrastructure experiment for the current release train. The workflow was written for the old v8 beta process and its required npm trusted-publisher/cleanup credentials were never provisioned, so shipping it would add a path that cannot complete. The 3.2 prerelease should use the existing Changesets pre-mode pipeline; a per-PR snapshot workflow can return as a fresh PR after its publisher identity is provisioned and validated end to end.

Issue #2371, opened 2026-07-18
chore(server): track ext-apps migration to MCP v2 server SDK


bokelley:
## Context

AdCP uses `@modelcontextprotocol/server@2.0.0-beta.4` for its modern per-request MCP adapter, while the canonical `@modelcontextprotocol/ext-apps` server helpers still target `@modelcontextprotocol/sdk` v1.29.

Upstream tracking issue: https://github.com/modelcontextprotocol/ext-apps/issues/702

This matters because `registerAppTool` and `registerAppResource` provide the canonical MCP Apps types and compatibility behavior, including:

- normalization between `_meta.ui.resourceUri` and the legacy `_meta["ui/resourceUri"]` key
- the `text/html;profile=mcp-app` resource MIME default
- typed tool visibility and resource `_meta.ui` fields

## Relationship to #2368

This does **not** block shipping the typed custom-tool `_meta` seam or the follow-on AdCP resource-registration surface. The MCP v2 server already exposes public `registerTool` and `registerResource` APIs; AdCP can register resources on each reconstructed per-request server.

The dependency becomes relevant when we want to replace local compatibility types and normalization with the canonical ext-apps implementation.

## Follow-up when upstream #702 lands

- [ ] Evaluate the released ext-apps v2 server helpers against AdCPs modern adapter.
- [ ] Adopt or re-export canonical MCP Apps tool/resource metadata types where compatible.
- [ ] Replace redundant local normalization and MIME defaults with canonical helpers.
- [ ] Verify nested and legacy resource-URI metadata compatibility.
- [ ] Verify app-only tool visibility and resource `_meta.ui` behavior through `tools/list`, `resources/list`, and `resources/read`.
- [ ] Remove any temporary compatibility code or dependencies that are no longer needed.

Refs #2368.

bokelley:
## Triage

**Classification:** tracking / meta
**Bucket(s):** library, dependencies
**Status:** deferred

**My take:** This is a well-scoped tracking stub with no local implementation possible until `modelcontextprotocol/ext-apps#702` ships. Blocking dependency is external; the checklist in the body captures the full follow-up scope. No `tracking` label exists in this repo — flagging that gap below.

Blocked-on: https://github.com/modelcontextprotocol/ext-apps/issues/702 — resurfaces when upstream ships.

---
Triaged by Claude Code. Session: https://claude.ai/code/session_018smRqrbH4GdKXjWmrptRQ8

---
[...]

Issue #2368, opened 2026-07-18
feat(server): expose MCP App metadata and resource registration for custom tools


bokelley:
## Problem

`@adcp/sdk` 12.0.1 cannot register a portable MCP App for a custom tool through its public server API:

- `AdcpCustomToolConfig` has no `_meta`, so a custom tool cannot declare `ui/resourceUri` or `ui.visibility`.
- `CreateAdcpServerFromPlatformOptions` has no resource-registration surface for `ui://` resources.
- `DecisioningAdcpServer` is intentionally opaque and exposes neither `registerTool` nor `registerResource`.

The current custom-tool implementation forwards description, title, schemas, and annotations only. Consumers therefore cannot build one typed Task that renders the same MCP App in Claude, ChatGPT, and other compliant hosts without reaching through private runtime symbols or creating host-specific routes.

This currently blocks Scope3's portable `upload_creative_asset` flow: the model-visible origin Task must open a `ui://` app, while signed upload-URL minting tools must be app-only so URLs never enter model-visible results.

## Requested API

Please expose a typed public seam that supports either server-construction-time registration or public registration methods for:

1. Custom-tool `_meta`, including:
   - `ui/resourceUri`
   - `ui.resourceUri`
   - `ui.visibility` such as `['app']`
2. MCP resources with:
   - a `ui://` URI
   - `text/html;profile=mcp-app` MIME type
   - a resource reader/handler
3. The same capability on the returned public server abstraction, or an options-based equivalent that does not expose private SDK internals.

## Acceptance criteria

- Custom-tool metadata is forwarded unchanged to `tools/list`.
- App-only tools remain callable by the app but are not model-visible.
- `resources/read` serves the registered `ui://` app resource.
- Types prevent invalid metadata/resource shapes.
- Tests cover a model-visible origin Task plus app-only prepare/finalize tools.
- Existing custom tools remain backward-compatible.

MCP Apps contract: https://github.com/modelcontextprotocol/ext-apps/blob/main/specification/2026-01-26/apps.mdx

bokelley:
## Triage

**Classification:** Feature request
**Bucket(s):** library
**Status:** ready-for-human

**What the experts said:**
[...]

PR #2372, opened 2026-07-18
feat(server): expose MCP App metadata for custom tools

bokelley:
## Summary

- add public `McpAppMeta` and `McpAppUiMeta` types for custom tools
- forward custom-tool `_meta.ui` unchanged through legacy registration and modern MCP server reconstruction
- cover a model-visible origin tool plus app-only prepare/finalize tools with type and integration tests
- add a patch changeset for `@adcp/sdk`

## Why

Custom tools could not declare portable MCP App metadata through the public server API. This prevented adopters from associating a tool with a `ui://` resource or marking supporting tools as app-only without reaching through private MCP server internals.

`ui.visibility` is host routing metadata rather than a server-side authorization boundary; the public type supports both `model` and `app` audiences.

Resource registration remains separate follow-up work because the modern adapter reconstructs its MCP server per request. The ext-apps v2 migration is tracked in #2371.

## Validation

- `npm run ci:quick`
- repository pre-push validation
- conventional commit and PR-title checks
- independent API/DX, MCP Apps protocol, and code/test expert reviews

Refs #2368

### DIFF
--- a/.changeset/quiet-tools-render.md
+++ b/.changeset/quiet-tools-render.md
@@ -1,0 +1,5 @@
+---
+'@adcp/sdk': patch
+---
+
+Add typed MCP App metadata to custom tools and forward it through modern MCP tool discovery.
--- a/src/lib/index.ts
+++ b/src/lib/index.ts
@@ -1022,6 +1022,8 @@ export type {
   AdcpServerToolName,
   AdcpCapabilitiesConfig,
   AdcpCustomToolConfig,
+  McpAppUiMeta,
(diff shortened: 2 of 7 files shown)


bokelley:
@benminer could you review this narrow public `_meta.ui` seam when you are online? All 32 checks are green and it intentionally leaves `ui://` resource registration to a stacked follow-up now in progress. Scope3 needs both slices before the portable Meta asset-upload Task can render in Claude and ChatGPT.

PR #2375, opened 2026-07-18
fix(mcp): restore Bun version negotiation

### LINKED ISSUE #2369: @adcp/sdk 12.0.1 breaks MCP version-negotiation probe (60s timeout) — 12.0.0 unaffected
## Summary

After bumping four reference adopters (buyer orchestrator, governance, seller, signals) from \`@adcp/sdk\` **12.0.0 → 12.0.1** with no other changes, all outbound MCP calls from the buyer to the seller fail with:

\`\`\`
error: \"Version negotiation probe timed out after 60000ms\"
\`\`\`

Rolling back to 12.0.0 (no other change) restored full functionality on the very next deploy. Both directions run the same SDK line and the same Fly-hosted MCP proxy topology in both scenarios.

## Repro

- Buyer: [\`kapoost/abzu-orchestrator\`](https://github.com/kapoost/abzu-orchestrator), Node 20, Fly (\`abzu-orchestrator\`).
- Seller: [\`kapoost/purrsonality-seller-agent\`](https://github.com/kapoost/purrsonality-seller-agent), Node 20, Fly (\`purrsonality-seller\`), MCP endpoint at \`https://seller.purrsonality.rocketscience.pl/mcp\`.
- Both agents typecheck clean on 12.0.1 (\`bun --bun tsc --noEmit\`, exit 0).
- \`fly deploy --remote-only\` completed cleanly on all four apps.

### Failure on 12.0.1

Buyer log line during a scripted seed run (post-deploy-e2e):

\`\`\`
{\"ts\":\"2026-07-18T09:50:25.505Z\",\"level\":\"error\",\"agent\":\"abzu\",\"msg\":\"execution failed\",\"err\":\"Version negotiation probe timed out after 60000ms\"}
{\"ts\":\"2026-07-18T09:49:56.897Z\",\"level\":\"error\",\"agent\":\"abzu\",\"msg\":\"execution failed\",\"err\":\"Request timed out\"}
\`\`\`

Buyer HTTP surface returned \`HTTP 500\` on every \`create_media_buy\` attempt against the seller. Governance + signals paths untested but on the same SDK line.

### Restore on 12.0.0

Same repo state, only \`\"@adcp/sdk\": \"12.0.0\"\` pin difference. Redeployed all four agents to 12.0.0. First post-deploy-e2e run:

\`\`\`
✓ buy purr_landing_rectangle_v1 · 80705ms
✓ sync + assign → mb_mrq6xbw2_ygddk1… · 9122ms
✓ approve post-deploy-1784368401396__landing · 48ms
✓ buy purr_result_card_v1 · 151ms
✓ sync + assign → mb_mrq6xjrm_cuihoi… · 5787ms
✓ approve post-deploy-1784368401396__result · 42ms
✓ Post-deploy seed complete. Both placements are serving AdCP creatives.
\`\`\`

## Hypothesis

The only shipped delta between 12.0.0 and 12.0.1 is #2365 — ESM banner injection was narrowed from \"every \`.mjs\` file\" to \"only files that reference \`__dirname\`/\`__filename\`/\`require\` in their body\". Verified against 486 \`.mjs\` files, 11 kept the shim.

The negotiation probe path runs through several ESM entries — one plausible failure mode is that a file that IS on the negotiation code path drops the shim because the AST-based detection didn't classify it as needing the Node shim, and the missing \`__dirname\`/\`__filename\`/\`require\` shim breaks resolution of a bundled asset or CJS interop that the probe needs.

Not confirmed by bisect — this is a hypothesis grounded in what changed. Happy to run a specific reduced repro if you can suggest which entry.

## Impact

Four live reference adopters were briefly offline during the 12.0.1 window; rollback restored them. No customer traffic impact — the pipeline is a demo surface.

## Environment

- Runtime: Bun 1.3.x on Fly \`shared-cpu-1x@1024mb\` machines, Node-compat via Bun.
- MCP path: outbound HTTP from buyer to seller MCP endpoint (not stdio).
- \`@modelcontextprotocol/sdk\`: \`^1.29.0\` (both sides).
- All other deps unchanged between the two runs.

## What I did NOT test

- Whether Node 20 with plain \`node\` (not Bun-compat) shows the same regression.
- Whether the timeout is on the probe request itself or on a follow-up call that never fires because the probe is stuck.

Let me know if a bisect commit / reduced repro / debug flag would help isolate.

---

sangilish:
## Summary

- apply the per-file ESM Node shim to protocol transport modules as an explicit allowlist
- add a clean-room Bun 1.3.8 smoke test that installs the packed artifact and completes modern MCP negotiation plus a tool call
- preserve the browser-safe `@adcp/sdk/enums` bundle check

## Root cause

The 12.0.1 conditional shim only covered source files that directly referenced `__dirname` or `require`. The MCP protocol modules do not contain those references themselves, but Bun needs the `createRequire` binding when loading their dual ESM/CJS dependency graph. Removing the former blanket banner therefore made the packed ESM artifact fail while loading MCP dependencies before negotiation could complete.

This keeps the conditional build introduced for browser consumers and adds a narrow protocol-directory allowlist instead of restoring the global banner.

## Validation

- `npm run build:lib`
- `NODE_ENV=test npx --yes node@20 --test-timeout=60000 --test-force-exit --test test/lib/mcp-modern-negotiation.test.js`
- `NODE_ENV=test npx --yes node@20 --test-timeout=60000 --test-force-exit --test test/read-path-abort-timeout.test.js`
- `npm run verify:package`
- `npm run check:package`
- `npm run typecheck`
- targeted Prettier check
- `git diff --check`
- repository pre-push validation

Fixes #2369.

### DIFF
--- a/.changeset/slow-spoons-connect.md
+++ b/.changeset/slow-spoons-connect.md
@@ -1,0 +1,5 @@
+---
+'@adcp/sdk': patch
+---
+
+Restore MCP version negotiation for Bun consumers of the ESM package artifact.
--- a/scripts/verify-package.mjs
+++ b/scripts/verify-package.mjs
@@ -8,7 +8,8 @@
  * range floors into a throwaway directory in the OS temp dir, then loads
  * `@adcp/sdk`, `@adcp/sdk/enums`, and `@adcp/sdk/server` through both a real
  * ESM `import` and a real CJS `require`, asserting each exposes a known runtime
- * symbol.
+ * symbol. It also runs a modern MCP negotiation under Bun, whose ESM/CJS
+ * interoperability differs from Node's.
  *
  * Why a temp dir outside the repo: installing inside the workspace would let
  * the monorepo dedupe peers against the repo's own node_modules, so a missing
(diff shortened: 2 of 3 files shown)


sangilish:
All checks are green and the change remains scoped to the Bun transport-version negotiation regression. No further changes are planned; this is ready for review.

andybevan-scope3:
getting this reviewed today

andybevan-scope3:
Approving. This is a well-scoped fix with a genuine regression test.

Problem: The 12.0.1 change (adcp#2364) narrowed the ESM createRequire/__dirname shim to only files whose own source references require( or __dirname, to keep node:url/node:path/node:module out of pure-data modules that browser bundlers can’t resolve. The MCP protocol modules don’t reference those in their own bodies, but their dual ESM/CJS dependencies need the createRequire binding to load under Bun. Node tolerated the missing shim; Bun did not, so the packed ESM artifact broke during MCP negotiation while every Node test stayed green.

Fix: Always apply the shim to files under src/lib/protocols/, layered on top of the existing require/__dirname detection. Restores the shim where Bun needs it without reintroducing the browser-bundle breakage, since the protocols directory is node-only.

What I verified:

The shim condition is correct, and the path check uses nodePath.sep so it’s cross-platform.
No protocol source file declares a top-level require/__dirname, so the injected banner can’t collide.
Browser safety is intact: the only browser-safe entry (@adcp/sdk/enums) doesn’t pull protocol modules, and the smoke test still runs the --platform=browser bundle check.
The new Bun smoke test’s server harness and callMCPTool call exactly match the existing test/lib/mcp-modern-negotiation.test.js, and it exercises the packed artifact through a real modern MCP initialize + tool call under Bun. This is precisely the cross-runtime path the Node-only suite missed.
Changeset present (patch).
One optional note (non-blocking): the allowlist covers the whole protocols/ directory, so pure-logic modules (abort.ts, responseSizeLimit.ts, rawResponseCapture.ts, transportDiagnostics.ts) also pick up node built-ins they don’t use. It’s defensible (node-only today, robust against new MCP modules) but slightly broader than the three modules that actually need it. Worth a one-line comment explaining the directory-wide scope. Ship it.

andybevan-scope3 (approved):

bokelley:
Thanks for the thorough review. Applied your optional note — added a one-liner above the `isProtocolTransport` check explaining the directory-wide scope (358246c):

```ts
// protocols/ is node-only throughout; directory-wide scope also covers future MCP modules automatically.
```

---

PR #2378, opened 2026-07-20
refactor(schemas): consolidate __dirname-relative schema/data root resolution

benminer:
## Summary

Part of a follow-up exploration branched off #2365 (ENG-130): the user's instinct on that PR was "maybe a simple banner is the root problem" (comparing against an old project's ampt-style single-bundle esbuild build). Investigation concluded that bundling the whole library per public entry point isn't viable here — 34 public `exports` subpaths, a CLI that deep-`require`s ~30 non-exported `dist/lib/**` paths, and (this PR's target) 7 files that resolve schema/data files via hand-tuned `__dirname` arithmetic fanning out across 7+ overlapping public entries.

This PR is the first of two pieces from that plan — a standalone cleanup that's worth doing regardless of any bundling question:

- Adds `src/lib/internal/package-root.ts` (`getPackageRoot()`) and `src/lib/internal/schema-data-roots.ts` (`getSchemaDataRoots()`). `getPackageRoot()` resolves the package root via `require.resolve('@adcp/sdk/package.json')` self-reference (falling back to a directory walk), independent of the calling module's own directory depth.
- Migrates 7 files off ~150 lines of duplicated, hand-tuned `path.join(__dirname, '..', ...)` arithmetic: `validation/schema-loader.ts`, `conformance/schemaLoader.ts`, `server/error-arm-tools.ts`, `v2/projection/{catalog,registry,canonical-properties}.ts`, `testing/storyboard/compliance.ts`. Each file's existing dist-vs-source-tree fallback business logic is unchanged — only the anchor computation changed.
- Fixes two fake-install test fixtures (`schema-loader-per-version.test.js`, `v2-projection-loader-paths.test.js`) that mimicked a published tarball by copying only `dist/`, never `package.json` — unfaithful to a real `npm install` (which always ships `package.json`) and would break `getPackageRoot()`'s self-reference.
- Updates the `tsup.config.ts` comment: the "`__dirname`-based schema-data lookups keep their original directory depth" justification for `bundle: false` no longer applies (tree-shaking and CLI/tooling deep-requires still do — `bundle: false` itself is unchanged).

No public API changes — `src/lib/internal/` is not exported. Branched off `issue-2364`; that PR (#2365) has since merged, so this branch has been merged up to current `main`.

Second piece (CLI build decoupling from `dist/lib`, giving the CLI its own esbuild bundle direct from TS source) will follow as a separate PR.

## Test plan

- [x] New TDD unit tests for `getPackageRoot()`/`getSchemaDataRoots()` (10 tests, including the directory-walk fallback in isolation)
- [x] `npm run test:node:fast`: 12518/12528 pass. The 3 failures (JWK/OKP crypto errors in `signer-grader`/`request-signing-provider`, a timing-sensitive MCP session-cache assertion) are pre-existing — reproduced independently against a stashed pristine baseline, and none of the failing test files reference any of the 7 migrated modules.
- [x] Re-verified the 3 CLI test files (`cli-auth-scheme`, `cli-webhook-receiver-flag`, `conformance-cli`) in isolation — all 50 tests pass; they looked flaky only under full-suite parallel load.
- [x] Re-ran affected tests + newly-merged `main` tests after merging `origin/main` — all green.
- [x] Changeset added (patch).

### DIFF
--- a/.changeset/schema-data-root-consolidation.md
+++ b/.changeset/schema-data-root-consolidation.md
@@ -1,0 +1,5 @@
+---
+"@adcp/sdk": patch
+---
+
+Internal: consolidated the duplicated `__dirname`-relative schema/data-root resolution across 7 loaders (`validation/schema-loader.ts`, `conformance/schemaLoader.ts`, `server/error-arm-tools.ts`, `v2/projection/{catalog,registry,canonical-properties}.ts`, `testing/storyboard/compliance.ts`) into shared `getPackageRoot()`/`getSchemaDataRoots()` helpers (`src/lib/internal/`). Resolution now anchors via `require.resolve('@adcp/sdk/package.json')` self-reference instead of hand-tuned `path.join(__dirname, '..', ...)` arithmetic tied to each file's own directory depth. No public API changes.
--- a/.changeset/tsup-conditional-node-shim-banner.md
+++ b/.changeset/tsup-conditional-node-shim-banner.md
@@ -1,0 +1,5 @@
+---
+"@adcp/sdk": patch
+---
+
+Fix: the ESM build no longer injects the `node:url`/`node:path`/`node:module` shim banner into every `.mjs` file. Only the (11) files that actually reference `__dirname`, `__filename`, or `require` in their body get it now; pure-data and pure-logic modules — including `enums.generated.mjs`, `inline-enums.generated.mjs`, and everything reachable from the zod-free `./enums` entry point — no longer carry dead Node-only imports that broke strict browser bundlers (Vite, `esbuild --platform=browser`).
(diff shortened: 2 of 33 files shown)


bokelley:
This draft is not merge-ready in its current shape. It now conflicts with `main` and includes the build-shim commits plus `.changeset/tsup-conditional-node-shim-banner.md` from #2365, which has already merged.

Please rebase onto current `main` and retain only the schema/data-root consolidation change and its dedicated tests/changeset, then resolve the loader call sites and rerun CI before marking it ready. Leaving the draft open for that cleanup.

bokelley:
Closing this stale draft rather than carrying its already-merged build-shim commits into the 3.2 cycle. The branch remains recoverable; any remaining schema-root work should be rebased as a focused PR from current main with only the still-missing behavior and tests.

PR #2399, opened 2026-07-27
perf: reuse modern MCP discovery results

bokelley:
## Summary

- update `@modelcontextprotocol/client` from `2.0.0-beta.4` to `2.0.0-beta.5`
- cache modern MCP discovery results for five minutes within the same endpoint and authorization context
- preserve OAuth provider identity only across trusted endpoint-derived agent configs
- retry stale discovery once for read-only tool listing while preserving the no-replay rule for tool calls

## Why

Beta.5 aligns discovery with the finalized MCP 2026-07-28 wire shape and adds `ConnectOptions.prior`. Reusing that validated discovery result removes a redundant `server/discover` round trip without changing legacy MCP Tasks routing or replaying mutating requests.

## Validation

- `npm run build:lib`
- `npm run typecheck`
- `NODE_ENV=test node --test-timeout=60000 --test-force-exit --test test/lib/mcp-modern-negotiation.test.js` (8 tests)
- protocol, security, and DX/code expert reviews
- pre-push validation

### DIFF
--- a/.changeset/bright-cats-listen.md
+++ b/.changeset/bright-cats-listen.md
@@ -1,0 +1,5 @@
+---
+'@adcp/sdk': patch
+---
+
+Update the modular MCP client to 2.0.0-beta.5 and reuse authorization-scoped modern discovery results, including across OAuth endpoint discovery and read-only reconnects.
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
         "packages/*"
       ],
       "dependencies": {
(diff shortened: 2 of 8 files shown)


bokelley:
Regarding Ladon's medium note at `mcp-modern.ts:435`: the tool-call path intentionally does not retry after a stale cached-modern request fails because the request may have reached the server and may be mutating. Automatic replay could duplicate a media-buy side effect. The failure invalidates both the connection and discovery caches so the next caller negotiates fresh; only the read-only `listTools` path retries immediately.

PR #2409, opened 2026-07-28
fix(mcp): treat ERA_NEGOTIATION_FAILED as legacy-era signal [backport 12.x]

nastassiafulconis:
## Why

Backport of AI-5153 fix to 12.x (current version used in agentic-api is 12.0.3).

Servers that respond to the \`server/discover\` probe with a malformed envelope (e.g. \`id: null\` instead of echoing the request id) cause \`@modelcontextprotocol/client\` to throw \`ERA_NEGOTIATION_FAILED\`. This error has no HTTP status, so the existing 404/405 check doesn't catch it — the error propagates as "None responded to MCP protocol."

Root cause confirmed for InMobi (adcp_agent 283, BASIC_AUTH): their server returns \`{"error":{"code":-32600,"message":"Missing ID field"},"id":null}\` for \`server/discover\`. The \`id: null\` response fails the beta client's envelope parse.

See companion PR targeting main: #2408

## What Changed

- Added `SdkError, SdkErrorCode` to imports from `@modelcontextprotocol/client`
- In `probeModernMCPConnection`, `attemptModernCall`, `tryListModernMCPTools`: treat `ERA_NEGOTIATION_FAILED` the same as 404/405 — fall back to the v1 transport path

Branch is based on \`@adcp/sdk@12.1.1\`. After this merges, agentic-api can bump \`@adcp/sdk\` to 12.1.2.

### DIFF
--- a/.changeset/fix-era-negotiation-legacy-fallback.md
+++ b/.changeset/fix-era-negotiation-legacy-fallback.md
@@ -1,0 +1,5 @@
+---
+'@adcp/sdk': patch
+---
+
+Treat ERA_NEGOTIATION_FAILED as a legacy-era signal in probeModernMCPConnection, attemptModernCall, and tryListModernMCPTools. Servers that respond to the server/discover probe with a malformed envelope (e.g. id: null) now fall back to the v1 transport path instead of surfacing "None responded to MCP protocol."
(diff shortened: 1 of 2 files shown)


nastassiafulconis:
Closing in favor of #2408 which targets main and has the same fix. Since @adcp/sdk@12.1.1 is an ancestor of main, these would conflict on merge. Once #2408 merges, a new 13.x release will include the fix. For a 12.x hotfix, cherry-pick db65d45d onto a maintenance branch.

PR #2421, opened 2026-07-31
fix(security): harden SSRF, signature, webhook auth, and OAuth callback boundaries

bokelley:
Salvaged from a partial `codex-security` scan. The run was cut short at ~32 min by OpenAI's cyber classifier and **its own validation stage never ran** (it shells repro scripts through `tsx`, whose IPC pipe the sandbox blocks), so all 40 candidates arrived unverified. Every fix here was confirmed by reading the cited code first; two of the scan's severity framings were overstated and are corrected in the notes.

## Highest impact

**`createPinAndBindFetch` never ran a redirect through its SSRF policy.** Both guards only ever see the URL the caller passed — the synchronous scheme/CIDR check runs before the request, and undici skips `connect.lookup` for IP-literal hosts. A buyer-registered webhook host answering `302 Location: https://169.254.169.254/…` therefore took the signed POST to the metadata service. `grep -n redirect` on the module returned nothing. Redirects are no longer followed, and the mode is **not caller-overridable** — honouring `redirect: 'follow'` would reopen the hole. `targeting-helpers` already wrapped its call site in `redirect: 'manual'`; webhook delivery never got it.

**`createExpressVerifier` verified against an empty body surrogate.** `resolveRawBody` demanded `req.rawBody` only when `content-length > 0`. Chunked transfers carry no `Content-Length`, so they fell through and the entire pipeline evaluated a body of `''` — `hasBody` derives from the same value, waiving the `content-type` coverage requirement, and a `content-digest` check hashed the empty string — after which `next()` handed a downstream parser's real body to the handler.

**`validateAgentUrl` only enforced under `NODE_ENV === 'production'`.** An unset or orchestrator-stripped `NODE_ENV` disabled the guard entirely. Now an allowlist (`test`/`development` plus `ADCP_ALLOW_PRIVATE_AGENT_URL=1`) layered over the existing `net/address-guards` CIDR classifiers, which also fixes bracketed `[::1]` (Node returns the bracketed form, so the old `'::1'` compare never matched), `127.0.0.2`, CGNAT, and IPv4-mapped IPv6 — and stops treating `10.example.com` as private.

**Webhook registrations advertised a placeholder credential.** Per `push-notification-config.json`, `authentication` is a scheme **selector**, not a fallback: presence opts into legacy HMAC-SHA256, absence selects RFC 9421. Emitting it with the constant `placeholder_secret_min_32_characters_required` downgraded every secretless webhook to legacy HMAC keyed by a value that ships in this source tree. The two sites differ because their schemas do:
- `push_notification_config.authentication` is optional → **omitted** without a real secret, selecting 9421.
- `reporting-webhook.json` marks it **required** for all of AdCP 3.x → the block can't be omitted and no honest value exists, so the library's auto-injection is **skipped** with a one-time warning. The `create_media_buy` itself is unaffected.

**The CLI OAuth callback never compared its `state`.** `CLIFlowHandler` generated a random state but resolved the pending authorization on the first request to reach the loopback path, so any local process could inject or cancel a flow — with code substitution available wherever the AS doesn't strictly bind PKCE.

## Also

- Windows browser launch no longer routes the authorization URL through `cmd.exe` (spawning `cmd` makes the interpreter the child, so it parses `&`/`|`/`^` even with `shell` unset — and auth URLs carry `&` by construction).
- The webhook verifier's loopback exemption from the https-only `@target-uri` rule used `startsWith('127.')`, matching names like `127.attacker.example` that resolve anywhere.
- `PropertyListAdapter.generateToken` built a returned `auth_token` from `Math.random`.
- `testAgent` logged the whole `TestOptions` at info level — bearer tokens, Basic passwords, OAuth access/refresh tokens, client-credential secrets, and test-kit API keys to stdout on every run.
- The structured serializers treated a `JSON.parse`-produced `__proto__` key as a prototype assignment rather than data. (Object-level, not global — the scan implied worse.)

Three code comments asserted safety properties the code lacked; all three are now accurate.

## Dependencies

Removes `@openai/codex-security`, which `npm install` had placed in **`dependencies`** — it would have shipped to every `@adcp/sdk` consumer. Clears the esbuild advisory, which needed both a direct bump and an `overrides` entry since tsup pins `^0.27.0` and would otherwise keep a nested vulnerable copy; build validated with the bumped bundler and emitted output diffed. The remaining three `@hono/node-server` advisories are left deliberately: the vulnerable path is `serveStatic`, which is used nowhere here or in the MCP SDK, and both MCP packages pin `^1.19.9` against a `>=2.0.5` fix.

## Tests

Regression tests for every fix (3 new files, 5 extended). The credential-redaction suite has a negative control: 0/8 pass without the fix, 8/8 with it.

**12,771 / 12,782 pass.** The 4 failures are not from this change:
- 2 deterministic JWK failures are pre-existing at this branch's base and **already fixed on `main`** past it (verified against a clean `origin/main` worktree; the fix swaps `x` as well as `d` — swapping only `d` leaves an incoherent OKP JWK that Node rejects). Rebasing clears them.
- 2 rotating order-dependent flakes. Across four full runs, five different tests took turns failing, never the same pair twice, each passing deterministically in isolation. Worth noting separately: `npm run test:node` exits 0 even with failures.

## Not fixed

`validateAgentUrl` is a correct *literal-host* gate but is synchronous and cannot see where a name resolves, so a public hostname that rebinds to a private address still reaches the MCP/A2A transports with `Authorization` attached. Routing those transports through the `ssrfSafeFetch` that already exists in this repo is the actual fix. Same for `verifyAndParseWebhook`, which doesn't yet verify the 9421 profile it now correctly selects. Both filed as follow-up issues; the limitation is stated in the JSDoc rather than left to read as complete.

30 further scan candidates remain unverified and untouched.


bokelley:
Triage of #2426 surfaced a fold candidate for this PR before merge.

`src/lib/auth/oauth/web-flow.ts` — `CompleteWebFlowOptions.expectedState` is optional and only checked when supplied (lines 381–382), while tokens are persisted for `flow.agentId`. The fix the PR author proposes in #2426 is `requireBrowserBinding?: boolean` (defaulting false) + `console.warn` when `expectedState` is omitted, which is non-breaking and directly parallels the `CLIFlowHandler` state-binding you've already hardened here.

The file and the review context are both naturally in scope for this PR. If scope expansion is too costly at this stage, disregard — the issue will track it independently.

---

bokelley:
## Review round complete — both stacks, plus the 30 remaining scan candidates

Ran the dual-stack per `docs/development/REVIEW-STACKS.md`: four Claude experts (security, code, protocol, DX) in parallel plus `npm run review:codex -- --all --base main`. Second commit addresses everything they converged on, **including two defects I introduced in the first commit.**

### The two stacks converged on one verdict

Both independently said: don't ship while secretless webhook receive only *warns*. They were right — the first commit made the SDK register RFC 9421 for those webhooks, so accepting unverified payloads became a conformance claim it couldn't honor. Now fails closed with a typed `webhook_unverifiable` code and an explicit `allowUnauthenticatedWebhooks` opt-out. Full 9421 verification is #2423.

### Two defects of mine they caught

- **The `validateAgentUrl` rework silently loosened things.** Delegating to `classifyProbeUrl` fixed the dev-loop breakage (below) but started allowing `metadata.google.internal` **by name**, which the previous code blocked. GCP metadata-by-name needs no IP literal, so a security fix introduced a live SSRF regression. Metadata hostnames now belong to the shared probe policy, trailing root dots are normalized, and `net/address-guards` picks up `192.0.0.0/24` (Oracle IMDS is at `192.0.0.192`), `192.88.99.0/24`, and `240.0.0.0/4` — ranges `WEBHOOK_SSRF_POLICY` already denied. The two tables had diverged.
- **The credential redaction didn't work on the path that matters.** `TestOptions._client` is a live client whose own fields lead straight back to the bearer, OAuth tokens, client secret, and `webhookSecret` the masking had just removed — three times over via nested references. Demonstrated by execution, not inspection. A denylist over that object graph can't be made safe; the logged context is now an allowlist.

### What the DX review proved with test output

The first commit's `NODE_ENV` gate broke **every documented local dev loop** — `compliance:fork-matrix`, nine `SKILL.md` "validate locally" blocks, and the reference-seller CI jobs (58 failures). Worse, the storyboard runner relabelled it as an agent conformance failure (`expected TERMS_REJECTED, actual Private network access not allowed`), so an adopter would read it as *their* bug. Fixed and verified locally with `NODE_ENV` unset: 6/6 pass on the harness that failed, and both reference-seller CI jobs are now green.

Also from that review: `README.md` and `docs/guides/SIGNING-GUIDE.md` both called a `rawBodyMiddleware()` that **exists nowhere in the repo**, so following either produced a blanket 401 under the stricter guard. Both now show the real `express.json({ verify: adapter.rawBodyVerify })` form.

### Other confirmed fixes

- **HTTP/2 defeated the raw-body guard entirely** — it forbids `Transfer-Encoding` and makes `content-length` optional, so a DATA-frame body carried neither signal. Inverted to require positive proof of *no* body. Also accepts a `Buffer` `rawBody` (what `express.json({verify})` actually yields — string-only was 401-ing correctly-wired apps) and reads `Content-Length` numerically so `00` isn't read as a body.
- **v2.5 pin regression:** `authentication` is `required` there with no selector semantics, so omitting it was schema-invalid rather than a mode selection. Now version-gated.
- **`reporting_webhook.authentication` is atomic.** The field merge crossed caller `schemes` with `webhookSecret` `credentials`, registering the HMAC secret as a **Bearer token** the seller then sends in cleartext on every delivery. And a caller passing `reporting_webhook` explicitly now gets an error rather than having the field silently deleted — editing caller intent is worse than refusing it.
- **The CLI shipped a constant `cli-webhook-secret`** in the npm tarball. After removing the library placeholder, that made the CLI the one remaining path downgrading webhooks to HMAC keyed by a public value. Now per-invocation random.
- **Redirecting webhook endpoints** were retried until the budget ran out and reported a bare `HTTP 302`. Terminal now, with the `Location` named and an `HTTP_REDIRECT` operator bucket instead of `UNKNOWN`.
- **The OAuth state check ran after the error branch**, so the *cancellation* half of the hole was still open — any local process could abort a login with no state knowledge. It now precedes every other callback parameter. The comment also no longer overclaims: a same-user process can read the URL out of argv, and PKCE is what protects the exchange there.
- One shared debug-log redactor replaces three divergent behaviors across `mcp` / `mcp-tasks` / `a2a`.
- Removed an **unreachable** octet-range check and the test that appeared to cover it — `new URL()` rejects out-of-range dotted quads first, so that test passed with the fix reverted. Caught by the code reviewer; it was mine.

Changeset raised to **major**.

### The other 30 candidates

Investigated all of them; several **refuted**, which is a useful result. Filed by cluster with preconditions and severity: #2424 (any commenter can drive the write-capable triage agent — the PR-comment filter its own prompt assumes doesn't exist), #2425 (build/CI supply chain — manifest-key TS injection reaches shipped code), #2426 (CLI `--wait` accepts the first unauthenticated POST as the task result; `BUILD-AN-AGENT.md` ships an auth-bypass example), #2427 (OAuth SSRF — client-credentials POST follows redirects off the validated endpoint, reproduced), #2428 (insecure-by-default gates; the `ref-routed` tenant default is documented-and-intentional, so it needs a maintainer decision rather than a patch), #2429 (unbounded caches — two recommended won't-fix, one because the obvious LRU introduces a correctness bug).

### CI

**28 pass.** `Generated, Docs & Formatting` fails on its "Validate generated files are in sync" step — **pre-existing on `main`** (run 30587961761, `f8f24fb5`), caused by 1,139 lines of stale `registry.yaml` + registry types. Filed as #2430; deliberately not folded in here, since that much generated churn would make this diff unreviewable.

Local: **12,699 / 12,709 pass.** Two failures are the pre-existing incoherent-OKP-JWK tests already fixed on `main` past this branch's base (rebasing clears them); the rest is a documented rotating flake set — six different tests across five full runs, each passing deterministically in isolation, none in an area this PR touches.

bokelley:
Folded in — commit cdf7043.

**What landed:**

- `BrowserBindingRequiredError` — new exported error class, parallel to the other typed errors in `web-flow.ts`.
- `CompleteWebFlowOptions.requireBrowserBinding?: boolean` — when `true`, throws `BrowserBindingRequiredError` if `expectedState` is omitted (strict mode for frameworks where session cookies are always available). Default `false`.
- `console.warn` emitted whenever `expectedState` is omitted (regardless of `requireBrowserBinding`), pointing callers to the session-cookie pattern. No warn when `expectedState` is supplied.

**Tests (4 new, all green):**
- warns via `console.warn` when `expectedState` is omitted
- does not warn when `expectedState` is provided
- throws `BrowserBindingRequiredError` when `requireBrowserBinding:true` and `expectedState` omitted
- `requireBrowserBinding:true` + `expectedState` present succeeds

The pre-push hook (prettier, tsc, tsup) passed. #2426 can close if this covers the scope — or stay open if there are follow-up items.

---

bokelley:
Both `redact-args.ts` review threads: **fixed in `ee5fb08e8`** — the finding was correct and it was the more consequential of the two registrations.

The redactor masked `push_notification_config.{authentication,token}` but not the same fields on `reporting_webhook`, so `reporting_webhook.authentication.credentials` — an HMAC shared secret — reached caller-visible `debug_logs` in plaintext.

Worth stating why that's the worse miss: `schemas/cache/3.1.8/core/reporting-webhook.json` lists `authentication` in its `required` array for all of AdCP 3.x, with `credentials` required inside it and `minLength: 32`. So a `create_media_buy` that registers reporting **always** carries a real credential there. `push_notification_config.authentication` is optional by comparison. I masked the optional one and missed the mandatory one, on the call most likely to be debug-logged.

Fix generalizes rather than adding a second special case — the helper now iterates `WEBHOOK_CONFIG_FIELDS × CREDENTIAL_FIELDS`, so a third registration shape is one list entry instead of a new branch.

Added `test/lib/redact-args.test.js` (7 tests). The gap existed because nothing covered this helper directly — the redaction tests I wrote targeted `agent-tester`, and the shared helper was only exercised incidentally through the protocol paths. That's the actual root cause of the miss, so the new file tests the helper in isolation, including a case for each registration and one asserting the caller's args aren't mutated.

Full suite: 12,766 pass / 2 fail, both members of the documented rotating flake set (`canonical-creatives-a2a-e2e` and `mcp-oauth-connection-cache`) — each verified passing in isolation, neither in an area this PR touches.

PR #2559, opened 2026-08-17
fix(storyboard): align gradeRequestSigning/gradeOneVector transport defaults to 'mcp'

fgranata:
Follow-up to #2540: @bokelley's commit aligning the standalone grader's `buildOpts` transport defaults (`?? 'raw'` → `?? 'mcp'`) was cherry-picked onto the PR branch after the PR had already merged, so it never landed on main — the published 13.0.0 has `resolveVectorTransport` defaulting to `'mcp'` (storyboard path, correct) but `gradeRequestSigning`/`gradeOneVector` still default `'raw'`. This is his commit (55e16c8) cherry-picked onto current main, unchanged.

Direct `gradeRequestSigning` callers against MCP agents currently get the raw REST replay (every vector 404s) unless they pass `transport: 'mcp'` explicitly — we hit exactly this running the vector batch against our production seller.


fgranata:
Both findings addressed in 556668d:

- **Changeset added** (`.changeset/grader-standalone-transport-default.md`, patch) covering the default flip and the CLI text.
- **CLI updated for the new default**: `--help` now documents `mcp` as the default with `raw` as the REST-binding opt-in, and the transport-mismatch hint fires only on explicit `--transport raw` runs — with `mcp` as the resolved default, an unset transport already grades over MCP, so the old raw→mcp retry hint would have misdirected exactly the operators it was written for.

PR #2582, opened 2026-08-18
fix(testing): buildAccount preserves the test kit's explicit sandbox flag

### LINKED ISSUE #2580: buildAccount clobbers the test-kit's sandbox flag — comply_controller_mode_gate is ungradeable by construction
## Summary

`buildAccount` in `testing/test-controller.js` (verified in published `@adcp/sdk` 13.0.0) unconditionally stamps `sandbox: true` **after** spreading the resolved account:

```js
function buildAccount(options) {
  if (options) {
    const account = resolveAccount(options);
    return { ...account, sandbox: true };   // <-- clobbers the kit's sandbox: false
  }
  return DEFAULT_SANDBOX_ACCOUNT;
}
```

The `comply_controller_mode_gate` storyboard's whole premise is presenting a live-mode principal (test kit `acme-outdoor-live`, `sandbox: false`) and asserting the seller returns `FORBIDDEN` before dispatch. With this spread order, the kit's `sandbox: false` can never reach the wire — every controller call arrives as a sandbox principal, the seller's gate correctly stays permissive, dispatch proceeds, and the probe's unknown `creative_id` yields `NOT_FOUND`. The storyboard then grades FAIL ("expected FORBIDDEN, got NOT_FOUND") against **every** seller, including ones with a fully conformant gate.

## Wire evidence (production seller, run of 2026-08-18 ~09:18:57Z)

Our seller (adcp.bidmachine.io) implements the storyboard's single-endpoint contract — an explicit `sandbox: false` on the account envelope returns `FORBIDDEN` before any scenario dispatch (server-side self-test passes). On the graded run, our audit log for the mode-gate probe (`force_creative_status`, `creative_id: comply-live-mode-probe-000`) records the request as received:

```
account: {"sandbox": true}
sandbox: null
```

— sent under the runner's standard tester credential (not a live-kit credential either), so neither of the two possible live-mode signals (envelope flag, auth principal) was presented.

## Suggested fix

Either (or both):
1. Flip the spread so the resolved kit value wins: `return { sandbox: true, ...account }` — default stays `true` when the kit doesn't set it, and `acme-outdoor-live`'s `sandbox: false` survives.
2. Have the mode-gate storyboard authenticate with the live kit's credential per its own prerequisites ("provides an API key whose associated account resolves as live/production mode"), so the auth-principal path is exercised too.

Happy to validate a fix against our seller — the FORBIDDEN path is deployed and one probe away.

fgranata:
Fixes #2580.

`buildAccount` unconditionally stamped `sandbox: true` after spreading the resolved account, so the `comply_controller_mode_gate` storyboard's live-mode kit (`acme-outdoor-live`, `sandbox: false`) could never reach any seller as a live-mode principal — the storyboard failed every seller regardless of their gate (wire evidence in the issue: a graded probe arrived at a production seller as `account: {"sandbox": true}`).

## Fix

`sandbox` now defaults to `true` only when the resolved kit leaves it unset:

```ts
const sandbox = 'sandbox' in account ? (account.sandbox ?? true) : true;
return { ...account, sandbox };
```

A bare spread-flip (`{ sandbox: true, ...account }`) would regress the other way — `resolveAccount` always emits the key, so an unset kit flag would spread `undefined` over the default; and the `account_id` arm of the `AccountReference` union has no `sandbox` member, hence the `in`-narrowing.

## Tests

New `test/lib/test-controller-build-account-sandbox.test.js` (4 cases, exercised through the exported `callControllerRaw` with a capture client since `buildAccount` is module-private): explicit `false` preserved, unset defaults `true`, explicit `true` untouched, no-options default account. Adjacent controller suites (seeding, bridge, discovery-arm) pass unchanged. Changeset included (patch).

Validation offer stands: our seller's FORBIDDEN gate is deployed at adcp.bidmachine.io — a mode-gate run on a build with this fix should flip that storyboard to green immediately.

### DIFF
--- a/.changeset/build-account-preserve-kit-sandbox.md
+++ b/.changeset/build-account-preserve-kit-sandbox.md
@@ -1,0 +1,5 @@
+---
+"@adcp/sdk": patch
+---
+
+`buildAccount` no longer clobbers an explicit `sandbox: false` from the resolved test kit: `sandbox` now defaults to `true` only when the kit leaves it unset. Fixes the `comply_controller_mode_gate` storyboard, whose live-mode kit (`acme-outdoor-live`, `sandbox: false`) could never reach sellers as a live-mode principal — the storyboard failed every seller regardless of their gate (#2580).
--- a/src/lib/testing/test-controller.ts
+++ b/src/lib/testing/test-controller.ts
@@ -65,8 +65,16 @@ const DEFAULT_SANDBOX_ACCOUNT = { brand: { domain: 'test.example' }, operator: '
 function buildAccount(options?: TestOptions): Record<string, unknown> {
   if (options) {
     const account = resolveAccount(options);
-    // Ensure sandbox is set — controller requires it
-    return { ...account, sandbox: true };
+    // Default sandbox to true (the controller requires the field) but let an
+    // explicit kit value win: the comply_controller_mode_gate storyboard's
+    // live-mode kit (acme-outdoor-live) sets `sandbox: false`, and the old
+    // `{ ...account, sandbox: true }` spread clobbered it — no seller could
+    // ever see the live-mode principal the storyboard exists to test (#2580).
+    // A bare spread-flip would regress the other way (resolveAccount always
+    // emits the key, so an unset kit flag would spread `undefined` over the
+    // default), hence the explicit coalesce.
+    const sandbox = 'sandbox' in account ? (account.sandbox ?? true) : true;
+    return { ...account, sandbox };
   }
   return DEFAULT_SANDBOX_ACCOUNT;
 }
--- a/test/lib/test-controller-build-account-sandbox.test.js
+++ b/test/lib/test-controller-build-account-sandbox.test.js
@@ -1,0 +1,68 @@
+/**
+ * buildAccount must preserve an explicit `sandbox: false` from the resolved
+ * test kit (adcp-client#2580): the comply_controller_mode_gate storyboard's
+ * live-mode kit (acme-outdoor-live) sets `sandbox: false`, but the old
+ * `{ ...account, sandbox: true }` spread clobbered it — every controller
+ * call reached sellers as a sandbox principal, making the storyboard
+ * ungradeable by construction. Exercised through the exported
+ * callControllerRaw with a capture client (buildAccount is module-private).
+ */
+
+const { describe, it } = require('node:test');
+const assert = require('node:assert/strict');
+
+const { callControllerRaw } = require('../../dist/lib/testing/test-controller.js');
+
+function captureClient() {
+  const captured = {};
+  return {
+    captured,
+    executeTask(name, params) {
+      captured.name = name;
+      captured.params = params;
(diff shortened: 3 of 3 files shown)


bokelley (changes requested):
Blocking protocol issue: this patch makes `callControllerRaw()` emit `account.sandbox: false`, but the published 3.1 and 3.2 `comply_test_controller` request schemas require `account.sandbox` with `const: true`. Live/production status must come from the authenticated principal resolved against seller state; the payload flag is only a schema-required sandbox intent assertion and cannot grant or communicate live status.

The named `comply_controller_mode_gate` path also does not call `buildAccount()`: because the step is `expect_error` with an authored `sample_request`, the generic runner sends that request directly (`account.sandbox: true`) and applies `auth.from_test_kit` through the raw MCP probe. The added capture-client test bypasses request-schema validation and therefore locks in a wire-invalid value without exercising the reported storyboard.

Please do not merge this approach. The correct next step is an integration reproduction of the mode-gate request proving that the payload remains `sandbox: true` and the live test-kit bearer is sent; any fix should target credential selection/loading if that reproduction fails.

bokelley:
Closing because the proposed payload change conflicts with the published controller schema and does not affect the reported mode-gate execution path. Issue #2580 remains open while we reproduce the `auth.from_test_kit` credential path and identify the actual failure.

bokelley:
Acknowledged — closure rationale noted. Continuing to track the `auth.from_test_kit` reproduction on #2580.

---

fgranata:
Agreed on all three points, and thanks for the correction — closing this approach is right.

1. **Schema**: `account.sandbox` is `const: true` in the published request schemas — an intent assertion the payload must carry, not a mode carrier. This PR would have made the SDK emit wire-invalid requests, and the capture-client test (which bypasses request-schema validation) would have locked that in. `buildAccount`'s hardcoded `true` is correct as-is.
2. **Path**: the mode-gate step is `expect_error` with an authored `sample_request` replayed through the raw MCP probe — `buildAccount` is never in that path, so #2580's root-cause names the wrong mechanism.
3. **Live status comes from the authenticated principal**, resolved against seller state — the kit's `auth.api_key` is the signal.

On the reproduction you asked for, we can already contribute the seller-side half from the graded run of 2026-08-18 (~09:18Z, adcp.bidmachine.io): the mode-gate probe arrived with payload `account.sandbox: true` (schema-conformant, as you describe) but authenticated with the runner's **standard tester credential** — our logs show no request carrying the kit bearer anywhere in the run window, and no failed auth attempts that would indicate it was tried and rejected. That looks like the credential-selection/loading branch of your analysis; we'll run the storyboard against ourselves with a local runner to complete the reproduction and report the result on #2580.

Two things we're fixing on our side in parallel, since the correct contract also exposes seller gaps: provisioning the kit principal (`demo-acme-outdoor-live-v1`) so it resolves as a live-mode account rather than an unknown key, and re-keying our controller gate on the authenticated caller's mode instead of the payload flag.

PR #2618, opened 2026-08-19
feat(media-buy): add compact lifecycle compatibility coordinator

### LINKED ISSUE #2613: Add a compact-first buyer compatibility coordinator and full v2.5–3.2 legacy seller matrix
## Goal

Let a buyer adopt one compact-first internal media-buy lifecycle while the SDK safely interoperates with sellers exposing:

- AdCP v2.5;
- AdCP 3.0;
- AdCP 3.1;
- AdCP 3.2 with only the established/legacy lifecycle;
- AdCP 3.2 with both lifecycles; and
- AdCP 3.2 with only the compact lifecycle.

Today the SDK 14 migration guide correctly tells buyers to discover the remote lifecycle and branch manually between compact methods and `get_products` / `create_media_buy` / `update_media_buy`. That protects semantics, but it means an adopter cannot actually switch its internal lifecycle without owning a second compatibility orchestrator and its complete cross-version test matrix.

The SDK already owns version negotiation, exact wire schemas, v2.5 request/response adapters, canonical-format projection, task handling, and transport behavior. It is the only layer positioned to make this compatibility contract consistent and extensively tested.

## This should not be a blind method rename

The compact lifecycle has terms digests, immutable proposal snapshots, feed/pricing versions, optimistic revisions, and explicit proposal states that older sellers cannot always represent. The requested feature is a **compatibility coordinator with declared loss boundaries**, not `listProducts()` silently calling `getProducts()` in every situation.

The coordinator should:

1. negotiate the remote protocol version and advertised/callable lifecycle;
2. select the native compact path when available;
3. select the established path when it can preserve the buyer's intent;
4. normalize legacy responses into a documented canonical compatibility result;
5. report which path/version was used and any representational loss; and
6. fail before a mutation with a typed `UNSUPPORTED_FEATURE`/compatibility result when the requested compact semantics cannot be represented safely.

It must never invent a proposal digest, feed/pricing version, accepted snapshot, atomicity guarantee, or terminal proposal state that the seller did not provide.

## Lifecycle coverage

Please define and implement the supported projection policy for both complete buyer journeys.

### Wholesale/direct journey

- product listing and pagination;
- product/feed/pricing version handling;
- purchase creation;
- media-buy readback and delivery readback;
- pause, resume, budget/date changes where representable;
- cancellation; and
- exact retry/idempotency behavior.

Likely established primitives include `get_products` with the appropriate version-specific wholesale behavior, `create_media_buy`, `update_media_buy`, `cancel_media_buy` where present, and the shared readback tools. The SDK should own the exact per-version decision table.

### Proposal journey

- request proposals;
- asynchronous/input-required proposal handling;
- revise/finalize proposal rounds;
- decline/omit behavior only where semantically equivalent;
- acceptance and purchase binding;
- media-buy and delivery readback; and
- subsequent operational control.

For established 3.x sellers this may use `get_products` `brief`/`refine`/`finalize` semantics plus `create_media_buy`. For v2.5, some proposal operations may be impossible. Those boundaries must be explicit, typed, and tested rather than silently approximated.

## Required compatibility matrix

Exercise the real public buyer API and real versioned wire schemas—no type casts between compact and established request types.

| Buyer | Seller surface | Required journeys |
|---|---|---|
| SDK 14 compact-first | v2.5 legacy | wholesale/direct; proposal subset or typed unsupported result |
| SDK 14 compact-first | every supported 3.0 release family | wholesale/direct and proposal |
| SDK 14 compact-first | every supported 3.1 release family | wholesale/direct and proposal |
| SDK 14 compact-first | 3.2 legacy-only | wholesale/direct and proposal |
| SDK 14 compact-first | 3.2 dual-surface | compact preferred; legacy fallback forced in a separate lane |
| SDK 14 compact-first | 3.2 compact-only | native wholesale/direct and proposal |

At minimum, cover MCP. Cover A2A wherever the same SDK buyer surface claims support, and expose the matrix through CLI storyboards so adopters can run it against a real seller.

For every lane assert:

1. negotiation and tool selection;
2. exact outbound request validation against that seller version;
3. response normalization and compatibility provenance;
4. canonical/legacy format projection;
5. account, brand, context, currency, pagination, and async task behavior;
6. retry, idempotency, and changed-payload conflict behavior;
7. no duplicate mutation during fallback or ambiguous transport failure;
8. product/proposal/media-buy identity continuity through readback;
9. pause/resume/cancel behavior and revision conflicts; and
10. negative fixtures for every unrepresentable compact semantic.

The test should include honest legacy fixtures, not only an SDK 14 server pinned to an older schema, so server quirks and missing discovery metadata are represented.

## Suggested public surface

The exact API is open for design, but it should let application code express one intent and receive a compatibility-aware result, for example:

```ts
const lifecycle = await agent.negotiateMediaBuyLifecycle();

const products = await lifecycle.listProducts(intent);
const purchase = await lifecycle.buyProducts(intent);
```

The returned coordinator/report should expose at least:

```ts
{
  negotiated_version,
  lifecycle: 'compact' | 'established',
  tools_used,
  compatibility: 'native' | 'lossless_projection' | 'lossy_projection',
  warnings
}
```

Mutations should reject lossy projection by default. If an opt-in is ever allowed, it must name the exact lost guarantee and be covered by a negative/positive conformance fixture.

## Acceptance criteria

- A compact-first buyer application does not implement its own version/tool branching for the supported matrix.
- Both complete lifecycle journeys have public SDK integration tests across all matrix lanes.
- Unsupported semantics fail before mutation with structured recovery guidance.
- CLI/conformance storyboards can run the same matrix against external sellers.
- The migration guide documents the guarantee and its explicit limits.
- Any missing normative mapping discovered by this work is filed against the AdCP protocol before SDK 14 GA.

This is a release-confidence requirement for adopters replacing their established internal lifecycle with the compact 3.2 model while continuing to buy from the existing seller ecosystem.

### LINKED ISSUE #2619: Generated refine_proposals response types drop CanonicalProposal fields
## Version

`@adcp/sdk@14.0.0-beta.3` (`ADCP_VERSION` is `3.2.0-beta.2`)

## Problem

The generated TypeScript and public Zod schemas for `RefineProposalsResponse` drop the base `CanonicalProposal` shape from the `revised`, `partial`, and `finalized` result arms.

For example, the published declaration reduces a revised proposal to:

```ts
proposals: {
  proposal_status: 'draft'
}[]
```

and a finalized proposal to:

```ts
proposal: {
  proposal_status: 'committed'
  expires_at: string
}
```

This omits load-bearing fields such as `proposal_id`, `proposal_kind`, `parent_proposal_id`, `terms_digest`, and `commercial_terms`. A buyer therefore cannot type-safely continue the proposal lifecycle from a `refineProposals()` result.

The generated `RefineProposalsResponseSchema` has the same loss and accepts structurally incomplete proposals because the result-arm objects are passthrough:

```js
import { RefineProposalsResponseSchema } from '@adcp/sdk/schemas'

console.log(
  RefineProposalsResponseSchema.safeParse({
    status: 'completed',
    results: [
      {
        source_proposal_id: 'proposal-source',
        outcome: 'revised',
        proposals: [{ proposal_status: 'draft' }],
      },
    ],
  }).success,
)
// true in beta.3
```

## Source-schema contract

The 3.2 `refine-proposals-response.json` base result declares:

```json
"proposals": {
  "type": "array",
  "items": {
    "$ref": "https://adcontextprotocol.org/schemas/3.2.0-beta.2/core/canonical-proposal.json"
  }
}
```

The `oneOf` arms then *narrow* those canonical proposals with draft/committed status, expiry, and lineage requirements. They do not replace the canonical proposal shape. The exact AJV wire validator therefore has a stronger contract than the exported TypeScript/Zod surfaces.

This appears to be a composition/codegen issue: the branch's nested `properties` overlay replaces the base property's `$ref` instead of intersecting with it. The generated async working schemas already use `CanonicalProposalSchema.and(...)`, which demonstrates the intended output shape.

## Expected

- Revised/partial rows expose `CanonicalProposal & { proposal_status: 'draft'; parent_proposal_id: string }`.
- Finalized rows expose `CanonicalProposal & { proposal_status: 'committed'; expires_at: string; parent_proposal_id: string }`.
- `RefineProposalsResponseSchema` rejects proposal objects missing required canonical fields.
- A regression test compares the public Zod admission behavior with the exact JSON Schema validator for these result arms.

## Adoption impact

This blocks a compact proposal buyer from consuming `refineProposals()` without casts and makes the public runtime schema unsafe as an admission validator. It is directly on the proposal lifecycle path being added in #2618.

bokelley:
## What

- adds a compact-first media-buy lifecycle coordinator with native 3.2 routing and explicit projections to established 3.0/3.1 tools
- negotiates by served version and discovered tool surface, reports the exact lifecycle/tools/losses used, and leaves established lifecycle calls unchanged
- adds bounded principal/account-scoped proposal snapshots, operation-level result classification, race-safe async completion capture, and replay-TTL-bounded exact acceptance retries
- fixes generated refine-proposal TypeScript and Zod composition so outcome arms retain the complete canonical proposal contract
- extends MCP, A2A, CLI/storyboard, and release-gate coverage across 2.5, 3.0, 3.1, and 3.2 lifecycle combinations

## Why

The SDK had no explicit compact-vs-established lifecycle compatibility matrix. That could silently select the wrong tool surface or obscure guarantees that an older seller cannot represent.

This change makes the no-worse rule explicit: native 3.2 uses compact tools; 3.0/3.1 sellers use established tools; established callers continue unchanged; and any compact guarantee that cannot be represented fails before mutation unless the buyer explicitly opts into the named legacy loss.

## Release gates

- all CI-equivalent fast shards, slow tests, protocol tests, typecheck, build, lint, formatting, and the focused lifecycle/wire matrix passed locally
- code, protocol, and security expert reviews converged with no Must/Should findings
- the independent Codex review finding was fixed and the exact final concurrency fence was re-reviewed by all three experts
- commitlint and PR-title validation passed

Closes #2613
Closes #2619

Normative protocol tracking: https://github.com/adcontextprotocol/adcp/issues/6685


bokelley:
I did an adopter-focused pass against #2613 and the intended deployment posture (3.2-only external discovery, hidden established compatibility routes, and a compact-first buyer that must safely cover 2.5–3.2 sellers). The projection and mutation-safety work is unusually thorough, but I think three items should be addressed before this becomes the migration gate:

1. **The public compatibility response erases the types and outcome discriminants a buyer needs.** `CompatibleProductsResponse.products` and both proposal collections are `unknown[]`; `pagination`, `context`, and `raw` are also untyped. `projectProposals()` then flattens request/refine results into `proposals` and leaves `outcome`, `reason`, `suggestions`, per-refinement results, and per-decline results only inside `raw`. In particular, an `unable` refinement or a decline result becomes an empty `proposals` view, and normal TypeScript cannot even read `proposal_id` from a returned proposal without a cast. That makes the promised single compact-first API substantially harder and less safe to adopt than the native generated surfaces. Please expose typed lifecycle-neutral unions/discriminants (at minimum canonical-or-established product/proposal rows plus typed request/refine/decline outcomes) rather than requiring every adopter to parse `Record<string, unknown>`.

2. **Auto negotiation can still route a compact-only 3.2 seller to an unadvertised established mutation.** `selectLifecycle()` chooses compact when the specific compact tool is present, but otherwise falls through to `established` without checking the corresponding established tool. A partial compact-only seller that advertises `list_products` but not `buy_products`, for example, will cause `buyProducts()` to dispatch `create_media_buy` even when discovery contains neither tool. This contradicts the nearby synthetic-capability invariant that compact-only sellers must not be routed to aliases they do not expose. The auto lane should select established only with established-tool evidence; otherwise it should return the typed compatibility error before dispatch. Please add per-operation partial-surface tests, including absent mutation tools.

3. **The release gate is narrower than the acceptance matrix it claims.** The honest 3.0/3.1/3.2 MCP proposal lanes exercise request + accept, but not refine/finalize, decline, async/input-required, post-accept control, or readback. Direct lanes exercise pause but not resume, budget/date change, or cancellation. The A2A lane is similarly request/accept and list/buy only. The forced dual-surface integration test covers only `listProducts` and requires `mcpToolProfile: 'all'`; it does not prove the established purchase/proposal lifecycles through the normal 3.2 profile where legacy routes are hidden. Unit coverage is extensive, but #2613 explicitly asked for these assertions through honest versioned wire fixtures so we can switch an internal lifecycle with release confidence.

A separate published-beta typegen issue surfaced during this review: #2619. `RefineProposalsResponse` and its public Zod schema drop the `CanonicalProposal` base fields from revised/finalized result arms. That should be fixed independently, and it reinforces why the coordinator should not paper over the result types with `unknown[]`.

The older idempotency capability parsing was also checked against the 3.0/3.1 bundles; the explicit `supported` discriminator and 3600–604800 TTL bounds are valid there, so I do not see a backward-compat regression in that change.

bokelley:
**Triage note (from #2619):** The generated TypeScript and Zod schemas for `RefineProposalsResponse` drop the base `CanonicalProposal` shape from the `revised`, `partial`, and `finalized` result arms — a codegen composition issue where nested `properties` overlays replace the base `$ref` instead of intersecting with it. Fields such as `proposal_id`, `proposal_kind`, `terms_digest`, and `commercial_terms` are silently omitted, and `RefineProposalsResponseSchema` accepts structurally incomplete proposals.

Since this type surface is part of the compact proposal lifecycle this PR introduces, consider folding the fix before merge so the PR ships with consistent types and a regression test comparing the public Zod admission behavior against the AJV wire validator for these result arms.

See #2619 for the full report and repro.

---

PR #2727, opened 2026-08-28
fix: enforce delegated signing and structured MCP validation

### LINKED ISSUE #2697: `isAuthorizedOperator` ignores validity windows, scopes and the required `brands` field — expired delegations still authorize
`src/lib/signing/agent-resolver/resolve-agent.ts`, private `isAuthorizedOperator` (~L461) — the library's only `authorized_operators` evaluation — diverges from the schema (`definitions.authorized_operator`, 3.1.2+):

- Schema: `valid_until` — "Consumers **MUST** treat entries at or after this timestamp as inactive for authorization decisions." The function checks neither `valid_from` nor `valid_until`: an **expired** entry still passes the step-3 cross-origin check.
- `scopes` (e.g. `media_buying` vs `creative_generation`) and `countries` are ignored entirely.
- `brands` is `required` in the schema, yet bare-string entries (no `brands` at all) are accepted as matches.
- Matching is eTLD+1 — worth an explicit comment at minimum, since on a shared registrable domain every subdomain authorizes every other.

Suggested fix: filter entries to the active validity window, honor `scopes` when present (omitted scopes = broad grant per schema wording), require the schema-required shape, and document the eTLD+1 choice. We implemented window+scope evaluation seller-side and can share it as a reference.

### LINKED ISSUE #2700: error_compliance is structurally unpassable on the modern MCP path: transport schema rejects the probe input the storyboard demands an adcp_error for
`@adcp/sdk` 14.0.0-beta.12

**One-sentence summary**

The compliance suite sends `budget: -1` to check that a seller answers with a proper AdCP error — but the SDK's own transport layer now rejects that request against the spec's JSON Schema (`minimum: 0`) before the seller ever sees it, so the seller is graded on a response it was never allowed to give.

Concretely, `error_compliance/negative_budget` sends:

```json
{ "packages": [{ "product_id": "test-product", "budget": -1, "pricing_option_id": "test-pricing" }] }
```

and asserts `/adcp_error/code` ∈ `[VALIDATION_ERROR, INVALID_REQUEST, BUDGET_TOO_LOW]`. What comes back instead is a transport-level tool error with no `adcp_error` and no `context` echo:

```
Input validation error: Invalid arguments for tool create_media_buy: data/packages/0/budget must be >= 0
```

Result: `Expected: [VALIDATION_ERROR, INVALID_REQUEST, BUDGET_TOO_LOW]`, `Actual: mcp_error`. Three steps fail this way (`negative_budget`, `validate_error_shape`, `validate_transport_binding` — the latter two also assert the `/context` echo, which the transport error doesn't carry). Error Handling drops to 2/5.

**Why the seller cannot fix this**

On the modern MCP path (client negotiates protocol `2026-07-28` — which the bundled comply runner does), `mcp-modern-server.ts` → `schemaForModernMcp()` intentionally prefers the official schema over anything the adopter registered:

```ts
const officialSchema = officialAdcpInputSchema(toolName, adcpVersion, profile);
if (officialSchema) {
  return fromJsonSchema(officialSchema as JsonSchemaType);
}
```

`officialAdcpInputSchema()` resolves `schemas/cache/<ver>/mcp/2026-07-28/media-buy/create-media-buy-request.json`, which carries the spec's `minimum: 0` on `packages[].budget`. So the request dies in the MCP SDK's `validateToolInput` as a generic tool error, upstream of the framework's own AdCP request validator — the layer that would have answered with exactly the `adcp_error` envelope (code, `recovery: correctable`, `context` echo) the storyboard asserts on.

The seller's handler is demonstrably fine: the sibling steps in the same storyboard that send a *valid* budget (`nonexistent_product`, `reversed_dates`) reach it and get a conformant `INVALID_REQUEST` / `VALIDATION_ERROR` with `recovery: correctable` and the `context` echoed. And per the modern adapter's design, adopter `toolSchemas` are not consulted on this path, so there is nothing seller-side to loosen.

**Ruled out (verified)**

- Not the MCP SDK bump: pinning `@modelcontextprotocol/sdk` back to 1.29.0 under `@adcp/sdk` 14.0.0-beta.12 reproduces the same three failures.
- Not the adopter's registered schema: it parses `budget: -1` successfully, and the `tools/list` projection of it contains no `minimum`/`exclusiveMinimum` anywhere.
- Not a schema-document change: the `mcp/2026-07-28` bundles for both our previous pin (3.2.0-beta.3) and current (3.2.0-beta.6) carry the same `minimum: 0` on package budget.

**Regression context**

On 14.0.0-beta.4 all 5 `error_compliance` scenarios passed every step (documented in #2635, filed from that state). Somewhere in beta.4 → beta.12 the outcome flipped; the adapter source and the schema documents are byte-identical across the range, so the suspects are the runner's protocol negotiation (a runner that previously spoke the legacy path would have hit the adopter schema and gotten through to the framework validator) and the regenerated mcp-era projection cache. I did not bisect the exact commit.

**Ask — any one of these resolves it**

1. Render transport-layer input-schema violations as a proper AdCP error envelope (`adcp_error.code: VALIDATION_ERROR`, `recovery: correctable`, `context` echoed) instead of a bare MCP tool error. This also helps real buyers, who today get an unstructured string for schema-invalid requests.
2. Or drop `minimum: 0` from `package-request.json` budget so sellers can legitimately answer `BUDGET_TOO_LOW` themselves.
3. Or have the `error_compliance` probes send input that is spec-schema-valid but business-invalid (the storyboard's own `nonexistent_product` / `reversed_dates` steps show that pattern works).

Repro: `comply()` against a seller built with `createAdcpServerFromPlatform` + `serve()`, MCP protocol, grading cache 3.0.18, storyboard `error_compliance`. Happy to provide full request/response wire logs.

bokelley:
## Summary

- enforce schema-shaped, time-bounded, brand/scope/country-aware cross-origin operator delegation during signing-key discovery
- cap built-in and adopter JWKS caches at the accepted delegation expiry
- keep modern MCP discovery on the official schema while returning structured AdCP validation errors from the framework
- preserve adopter schemas and explicit validation policy for custom or hidden compatibility tools

Closes #2697.
Closes #2700.

## Verification

- `npm run build`
- `npm run typecheck`
- focused resolver tests: 29 passed
- focused modern MCP tests: 13 passed
- `npm run test:lib:fast`: 13,725 passed, 7 skipped, 0 failed
- `npm run test:lib:slow`: 165 passed
- `npm run lint`: 0 errors (existing warning baseline only)
- `npm run lint:workflows`
- `npm run check:runtime-compat`
- `npm run check:adopter-types-narrow`
- `npm run check:adopter-types`
- `npm run verify:package`
- protocol, security, code, and second-model review stack completed

### DIFF
--- a/.changeset/secure-operators-structure-errors.md
+++ b/.changeset/secure-operators-structure-errors.md
@@ -1,5 +1,5 @@
 ---
-'@adcp/sdk': patch
+'@adcp/sdk': minor
 ---
 
 Enforce active, scoped cross-origin operator delegations during signing-key discovery and cap JWKS caches at delegation expiry. Preserve strict official modern MCP discovery schemas while routing AdCP input failures through the structured framework error envelope.
--- a/.changeset/secure-operators-structure-errors.md
+++ b/.changeset/secure-operators-structure-errors.md
@@ -1,0 +1,5 @@
+---
+'@adcp/sdk': patch
+---
+
+Enforce active, scoped cross-origin operator delegations during signing-key discovery and cap JWKS caches at delegation expiry. Preserve strict official modern MCP discovery schemas while routing AdCP input failures through the structured framework error envelope.
--- a/docs/guides/PUSH-NOTIFICATION-CONFIG.md
+++ b/docs/guides/PUSH-NOTIFICATION-CONFIG.md
@@ -149,6 +149,17 @@ Custom registration stores used by durability-protected mutation flows must also
 
 For deterministic tests or infrastructure-managed keys, set `webhookVerification.jwks`. Otherwise seller key discovery is automatic and uses an unauthenticated official protocol client for the capabilities step, so credentials configured for one endpoint are never transplanted to the registered callback origin. Sellers whose capability discovery requires authentication should provide an origin-bound `webhookVerification.fetchCapabilities(agentUrl, protocol)` callback or inject `webhookVerification.jwks` directly.
 
+When a cross-origin seller is authorized through a constrained
+`brand.json.authorized_operators[]` entry, also configure the trusted
+`webhookVerification.resolverOptions.requiredOperatorBrand`,
+`requiredOperatorScope`, and/or `requiredOperatorCountry` values that select the
+intended grant. Narrow brand, scope, or country lists fail closed without this
+context. Broad grants use `brands: ['*']`, omitted scopes (or `['all']`), and
+omitted countries. Delegated JWKS caches never outlive `valid_until`.
+Resolver options are trusted client-wide policy; they are not inferred from
+individual tool arguments. Use separate clients/resolvers for operations with
+different constrained brand, scope, or country tuples.
+
 ### Legacy HMAC-SHA256
 
 When `webhookSecret` is configured, the legacy webhook authentication path uses `HMAC-SHA256`. The agent signs `${timestamp}.${raw_body_bytes}` with the shared secret and sends:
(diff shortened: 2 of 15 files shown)


bokelley:
Addressed the review note in 16c604d97: the changeset is now classified as a minor change to reflect the new public resolver options and tightened delegation acceptance.

PR #2759, opened 2026-08-29
fix(testing): keep task lifecycle account scope stable

### LINKED ISSUE #2703: Framework task registry: enforce account isolation on task writes and task_id reads
## Problem

The packaged framework task registry does not preserve account isolation consistently:

1. In `media_buy_seller/create_media_buy_async_lifecycle`, a submitted handoff is written to a registry partition that `get_task_status` cannot read back under the storyboard account.
2. In `signals_baseline/get_signals_async`, `list_tasks` with a `task_id` filter does not apply the request account and can return the authenticated caller's task from a different sandbox account.

The first defect prevents a valid async lifecycle from completing. The second is an account-isolation failure and should be treated as security-sensitive even though it was found in the conformance harness.

## Requested change

Make account identity part of the framework task registry key and enforce it on every write and read path, including direct `task_id` filters. A caller authenticated across multiple accounts must not bypass the requested account partition.

## Acceptance criteria

- A submitted task can be retrieved through `get_task_status` in the same account that created it.
- The same task is not returned by `list_tasks` or status lookup under another account.
- Filtering by `task_id` cannot bypass account predicates.
- Add same-account and cross-account regression tests for both media-buy and signals async task paths.
- The downstream quarantined storyboard and step can be removed.

Downstream context: adcontextprotocol/adcp#6883.

bokelley:
Closes #2703.

## Summary
- preserve one authoritative account scope across async storyboard create, discovery, controller, poll, and list requests
- keep intentional wholesale fixture scoping while rejecting opaque fixture account overrides elsewhere
- resolve controller account context references before selecting scope

## Verification
- `npm run build:lib`
- `npm run typecheck`
- focused request-builder, schema round-trip, and task-scope tests (151/151)
- Claude protocol/code/security review: clear
- `npm run review:codex -- --all --base main`: ships / no findings
- commitlint and `git diff --check`


bokelley:
Ladon note reviewed. The behavior is intentional and scoped: fixture-authoritative account identity applies only to wholesale storyboards (the comment opens with that qualifier). Non-wholesale paths now consistently prefer the trusted runner/context account, which prevents opaque fixture account IDs from changing lifecycle scope; natural fixture operator references remain preserved when no trusted context exists. The regression suite covers wholesale preservation, trusted-context precedence, opaque-ID rejection, and create/poll/update/complete scope stability. Independent protocol, code, and security reviews found no blocker.

PR #2796, opened 2026-08-31
feat(schemas): read optional nulls as absent at the call site

rachitm022:
## Problem

Generated Zod schemas keep `.optional()` so they mirror the AdCP TypeScript types exactly. `.optional()` accepts `undefined` and rejects `null`. Plenty of agents serialize from Pydantic, Jackson, or `encoding/json` and emit `null` for a field they have nothing to report for — and because Zod validates a whole payload before a caller reads any field, one such `null` discards everything reported alongside it.

This is not hypothetical. A seller sent `next_expected_at: null` in a `get_media_buy_delivery` response — a scheduling hint carrying no spend, no impressions, nothing an accounting pipeline consumes. `GetMediaBuyDeliveryResponseSchema` rejected the response and took every `media_buy_deliveries` leg's figures with it. Four distinct reporting days for two customers, never recovered, because retries fail identically.

The generator already records the deliberate decision to keep `.optional()`, and tells callers what to do instead:

> we intentionally keep `.optional()` (NOT `.nullish()`) so Zod schemas match TypeScript types. Callers that need to accept null from external APIs should use `.nullish()` at the call site, not globally in every schema.

That decision is right — but it had no mechanism. A consumer cannot `.nullish()` one field inside a `.pick()` of an 8755-field generated schema graph without patching generated code, which then has to be redone on every SDK bump. So every TypeScript consumer either patches the generated file or writes its own normalizer.

## What this adds

`treatOptionalNullsAsAbsent(schema, payload)`, exported from `@adcp/sdk/schemas`. It drops a `null` only where the schema itself says it carries no more information than omitting the field: the value may be skipped, and `null` is not among the values it accepts.

| Schema declares | Payload sends | Result |
| --- | --- | --- |
| `.optional()` | `null` | key dropped; parse succeeds |
| `.optional().nullable()` | `null` | `null` kept — the spec assigns it meaning (`completion_rate: null` = "not applicable to these buys") |
| required | `null` | `null` kept; parse still fails |
| `z.never().optional()` | `null` | `null` kept; parse still fails (field must not be provided) |
| not declared (`passthrough`) | `null` | passed through untouched |

The type-fidelity decision is preserved: schemas keep `.optional()`, inferred types stay exact, and nothing becomes globally lenient. Tolerance is opt-in per call.

The walk recurses through nested objects and arrays, so a `null` inside `media_buy_deliveries[].by_package[]` resolves against that entry's own declaration. It also consults **both sides of an intersection** — response schemas projected from a JSON Schema `allOf` are intersections at their root, and a `null` is dropped only where every declaring side agrees it means absent. That covers 80 of the 87 generated `*ResponseSchema` roots (48 object-rooted, 32 intersection-rooted); the 7 union-rooted ones are left untouched, since a union names no single shape to resolve a field against.

The input is never mutated. Nodes containing a reinterpreted `null` are copied; every other node is shared with the original, so a caller keeping an audit trail of what the agent actually sent can hold onto the payload it passed in.

This is the Zod-surface counterpart to the null tolerance #1158 added for v2.x envelope fields before Ajv (#1149). Same class of bug, different surface: that one is Ajv, top-level envelope keys only, gated to v2.x.

## Also: removes a dead transform

`postProcessForNullish` was defined, documented with a rationale arguing *for* global `.nullish()`, and unit-tested — but never called in the generation pipeline. Its only other reference was the test-only `__test__` export block. A transform in that state reads as live behavior to anyone auditing the generator; it misled us until we checked the call sites. Deleted, along with its test. The comment recording why optional fields stay `.optional()` now points at `treatOptionalNullsAsAbsent` as the mechanism it tells callers to use.

## Tests

`test/lib/optional-null-tolerance.test.js`, 23 cases:

- `null` on optional-only → dropped; on `.optional().nullable()` → kept; on required → kept and still fails `safeParse`; on `z.never().optional()` → kept and still fails.
- `.default()` → dropped, and the default then applies. `.catch()` → kept (catch accepts `null`, so it is a value, not an absence). `.nonoptional()` → kept.
- Traversal reaches an object shape through `prefault` / `default` / `readonly` wrappers.
- Nested objects and array entries get the same treatment.
- A `.merge()`-built shape resolves against the merged field set, so a `null` `completion_rate` in a `by_package` entry survives.
- Intersections: dropped when both sides agree, kept when they disagree.
- `z.record()` stop-case returns the payload by identity (120 `z.record(` uses in the generated schemas).
- Undeclared keys pass through; a union-typed field is left alone; the input is not mutated; untouched nodes are returned by identity.
- End-to-end on `GetMediaBuyDeliveryResponseSchema`: `next_expected_at: null` parses and retains every leg's figures; `completion_rate: null` inside `by_package` survives as a value; `reporting_period: null` still fails. Plus one case proving the walk reaches the fields of an intersection-rooted generated schema.

Verification run locally:

- `node --test test/lib/optional-null-tolerance.test.js` → 23/23 pass
- `node --test test/lib/zod-schemas.test.js test/lib/v2-5-envelope-null-stripping.test.js` → 83/83 pass
- `node --test test/generate-zod-object-intersections.test.js` → 18/18 pass after the deletion
- `tsc --project tsconfig.lib.json --noEmit` → clean; `prettier --check` → clean

## Not in scope

Making `next_expected_at` nullable in the AdCP spec. The spec already uses `"type": ["number", "null"]` where `null` carries meaning (`completion_rate`), so plain-optional means "omit when not reported". Adding `"null"` there would introduce a second way to say "absent" and would not fix the class — the next agent nulls a different field. That is a separate docs-only conformance note for `adcontextprotocol/adcp`.

### DIFF
--- a/.changeset/optional-null-tolerance-helper.md
+++ b/.changeset/optional-null-tolerance-helper.md
@@ -1,0 +1,5 @@
+---
+'@adcp/sdk': minor
+---
+
+Add `treatOptionalNullsAsAbsent` to `@adcp/sdk/schemas`, which reads an agent's explicit `null` as an omitted field wherever a generated schema declares the field optional and non-nullable, so one unreported hint no longer discards a whole response. Nullable fields keep their `null` as a value, required and `z.never()` fields keep their `null` and still fail validation, undeclared keys pass through, and the input payload is never mutated. The walk recurses through nested objects and arrays and consults both sides of the intersections that JSON Schema `allOf` projects into, so it covers 80 of the 87 generated `*ResponseSchema` roots; union-rooted schemas are left untouched. Removes the never-called `postProcessForNullish` transform from the Zod generator, which read as live behavior to anyone auditing it.
--- a/docs/ZOD-SCHEMAS.md
+++ b/docs/ZOD-SCHEMAS.md
@@ -95,6 +95,61 @@ app.post('/api/products', async (req, res) => {
 });
 ```
 
+## Tolerating `null` on Optional Fields
+
+Optional fields in the generated schemas are `.optional()`, which accepts
+`undefined` and rejects `null`. That keeps the schemas an exact mirror of the
+TypeScript types, but plenty of agents serialize from Pydantic, Jackson, or
+`encoding/json` and emit `null` for a field they have nothing to report for.
+Because Zod validates the whole payload before you read any field, one `null`
+on an optional hint discards everything alongside it — a `null`
+`next_expected_at` (a scheduling hint carrying no figures) takes every
+`media_buy_deliveries` spend and impression count with it.
+
+`treatOptionalNullsAsAbsent` reinterprets those `null`s, and only those:
+
+```typescript
+import {
+  GetMediaBuyDeliveryResponseSchema,
+  treatOptionalNullsAsAbsent,
+} from '@adcp/sdk/schemas';
+
+const tolerated = treatOptionalNullsAsAbsent(GetMediaBuyDeliveryResponseSchema, response);
+const result = GetMediaBuyDeliveryResponseSchema.safeParse(tolerated);
+```
+
+A `null` is dropped only where the schema says it carries no information the
+omission wouldn't — the field may be skipped, and `null` is not among the
+values it accepts. Everything else keeps its `null`, and its verdict:
+
+| Schema declares | Payload sends | Result |
+| --- | --- | --- |
+| `.optional()` | `null` | key dropped; parse succeeds |
+| `.optional().nullable()` | `null` | `null` kept — the spec assigns it a meaning (`completion_rate: null` means "not applicable to these buys") |
+| required | `null` | `null` kept; parse still fails |
+| `z.never().optional()` | `null` | `null` kept; parse still fails (the field must not be provided) |
+| not declared (`passthrough`) | `null` | passed through untouched |
+
+The walk is schema-driven and recurses through nested objects and arrays, so a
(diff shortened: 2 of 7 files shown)


rachitm022:
Closing this — the premise doesn't hold up under review.

I claimed a `null` on an optional field discards the whole response payload. That is true of the generated Zod schemas used directly, but **not** of the SDK's buyer path, which is what actually matters for the incident that motivated this PR. The client validates with Ajv, not Zod, and response validation defaults to `warn`:

```ts
// src/lib/validation/client-hooks.ts:35
return { requests: config?.requests ?? 'warn', responses: config?.responses ?? 'warn' };
```

```ts
// src/lib/core/TaskExecutor.ts:5735
if (mode === 'warn') { console.warn(...); return { valid: true, errors: [] }; }
```

So a seller emitting `next_expected_at: null` surfaces a warning in `debug_logs` and the payload is still delivered. The comment above `resolveValidationModes` already names this class — "envelope nulls" — as the reason for that default. The data loss I cited came from calling `GetMediaBuyDeliveryResponseSchema.parse()` directly instead of going through the client. That is a usage error on my side, not an SDK defect, and this PR overstated it.

What does survive from the investigation is smaller and I'll open it separately:

`docs/ZOD-SCHEMAS.md` recommends `GetProductsResponseSchema.parse(response)` for response validation (lines 62 and 89) without noting that direct `.parse()` is stricter than the client path and will reject the optional-field nulls that real sellers emit. That's a documentation gap worth fixing on its own — a reader following the docs lands exactly where I did.

Apologies for the noise.

PR #2885, opened 2026-09-13
chore(types): regenerate registry types for compliance provenance

bokelley:
## Summary

- sync the registry OpenAPI snapshot from the live AgenticAdvertising registry
- regenerate TypeScript registry types for compliance-run provenance and heartbeat requeue fields

## Validation

- `npm run sync-registry-schemas`
- `node node_modules/typescript/bin/tsc --noEmit -p .context/registry-sync/tsconfig.json`

This is an upstream generated-artifact drift fix, independent of the reporting stack.


bokelley:
Closing as redundant with #2880. I verified that both generated registry artifacts (schemas/registry/registry.yaml and src/lib/registry/types.generated.ts) are byte-for-byte identical to current main; the remaining changeset would duplicate the release entry.

Issue #1624, opened 2026-05-09
storyboard: missing required-account tool should fail, not skip


bokelley:
## Problem

Per AdCP, every seller has to expose **either `list_accounts` or `sync_accounts`** — it's how buyers know who they can transact as. An agent that exposes neither has a real compliance gap on the account-management surface, not a "we'll grade you later when you implement it" deferral.

The storyboard runner currently treats missing-account-tool as a *skip*, with the implicit "exemption" message:

```
warnings: [
  "Agent did not advertise tool \"sync_accounts\"; agent tools: [...]. Sole stateful
   step exemption applied for phase 'setup': no peer could have established substitute
   state, so downstream phases run without cascade (adcp#4053, adcp-client#1146/#1545)."
]
skip_reason: "missing_tool"
```

Translation: *"I would have failed you, but I let it slide because nothing else could have set the state up."*

That exemption logic makes sense for **optional capability** scenarios (e.g. `axe_signals` — only graded if the seller advertises it). It does **not** make sense for tools the spec mandates as a one-of-N choice. A real launch-readiness check should call those out.

## Repro

Run `media_buy_seller` against an agent that exposes neither `sync_accounts` nor `list_accounts`:

```
── A2A ──
  Steps: 31 passed, 0 failed, 28 skipped — STORYBOARD-OK
```

The agent passes compliance despite missing a required-by-spec capability. That's the wrong signal.

## Proposal

For tools that AdCP marks as **required** (or required-as-one-of-N), the storyboard runner should:

1. Inspect agent capabilities once at run start.
2. For each spec-required tool family, check that at least one member is advertised.
3. Emit a hard failure (`STORYBOARD-FAIL` line, exit 1, contributing to the failed count) when a required family is empty — distinct from per-scenario skips.

Suggested error shape in the summary:

```
── A2A ──
  Steps: 31 passed, 1 failed, 28 skipped
  Failing checks:
    - capability:account_management — agent advertises neither sync_accounts nor list_accounts
       Spec: every seller must expose at least one account-discovery tool
       Affected: refine_products/setup (skipped due to this gap)
```

## Where to encode the requirement

Two options, increasing rigour:

1. **Hard-coded in the runner** for the small set of required-one-of-N families known today (`account_management`, possibly `creative_management`).
[...]

PR #2913, opened 2026-09-14
fix(accounts)!: rework derived resolution as an upstream-managed account-id namespace

### LINKED ISSUE #1628: createDerivedAccountStore: factory + docs only cover single-tenant; multi-tenant derived needs a path to list_accounts

### LINKED ISSUE #1647: decisioning: rework 'derived' resolution mode — upstream-managed roster (Meta/Snap/AudioStack), invert account_id rejection
## What

The `'derived'` resolution mode is currently shipped with inverted wire semantics. It's documented and implemented as "single-tenant; refuse `account_id`; permit brand+operator," but the pattern adopters actually need — and that the dominant programmatic/social case requires — is the *opposite*: upstream platform owns the roster, buyer must `list_accounts`, buyer passes `account_id` on every subsequent call, `sync_accounts` is forbidden.

The "single-tenant agent" framing was a mistake. AudioStack and flashtalking sit in this mode at N=1 (one account per credential); Meta/Snap/Figma MCP sit in this mode at N=many. The wire contract is identical — roster size is an operational property of the upstream, not a different SDK pattern.

## Corrected three-mode taxonomy

| Mode | Roster owner | Onboarding | Wire reference |
|---|---|---|---|
| `'explicit'` | Seller (no list) | None | `account_id` inline |
| `'implicit'` | Buyer-declared | `sync_accounts` | brand + operator |
| `'derived'` | Upstream platform | `list_accounts` | `account_id` inline |

## Concrete changes

1. **Invert `refuseInlineAccountIdWhenForbidden` for derived.** `src/lib/server/decisioning/runtime/from-platform.ts:193, 1201` currently refuses `{ account_id }` for both `'implicit'` and `'derived'`. For derived it should refuse the brand+operator union arm and **accept** `account_id`. Implicit's behavior is unchanged.
2. **Refuse `sync_accounts` dispatch when `resolution: 'derived'`.** Either fail at platform-config time when the adopter wires `accounts.upsert` alongside `'derived'`, or return `UNSUPPORTED_FEATURE` at dispatch. Upstream owns the roster.
3. **Require `list_accounts` (not OR) for derived in the #1624 compliance gate.** Currently `list_accounts` OR `sync_accounts`. For derived it must be `list_accounts` only.
4. **Update `AccountStore.resolution` JSDoc** at `src/lib/server/decisioning/account.ts:389-412` and `docs/guides/account-resolution.md` to reflect the corrected taxonomy. The "Multi-tenant derived" framing disappears — derived is uniformly upstream-managed regardless of N.
5. **Update `scripts/generate-agent-docs.ts`** source string (the "Four reference AccountStore shapes" paragraph) so the regenerated `docs/llms.txt` matches the corrected semantics.
6. **Audit reference adapters** (audiostack, flashtalking, scope3data/agentic-adapters#100) — `createDerivedAccountStore`'s `toAccount(ctx)` callback now needs to verify a buyer-supplied `account_id` matches what the credential maps to, instead of being a pure "ignore the ref, return the singleton" shortcut.

## Migration / breaking-change scope

Technically breaking for any adopter relying on current (inverted) rejection behavior. In practice the inverted behavior is unusable for upstream-managed adapters (they can't get `account_id` through), so adopters that successfully shipped with `'derived'` today are either:
- N=1 single-tenant (audiostack-shape) where `toAccount(ctx)` ignores the ref — minor adjustment to verify the ref instead
- Bypassing the SDK's resolve and routing on credential themselves — unaffected

Major version bump warranted. Feature flag (`accounts.resolutionVersion: 2`) considered but probably not worth the dual-codepath maintenance.

## Cross-references

- Supersedes the design intent of #1630 (closed); that PR added `list?` to derived but kept the inverted rejection logic and the single-tenant framing.
- Companion spec issue: adcontextprotocol/adcp#4341 — `account-ref.json` description currently doesn't disambiguate upstream-managed roster from `'explicit'`; needs explicit wording so JS and Python SDKs implement the same contract.
- Reference adapter `scope3data/agentic-adapters#100` (AudioStack) — confirm no breakage during the rework.
- Python SDK (`adcp-client-python`) had a partial inverse intuition (their reading: derived only accepts `account_id`); neither SDK had the full picture. Spec clarification keeps both aligned going forward.

bokelley:
Fixes #1647
Fixes #1628

## What

`AccountStore.resolution: 'derived'` shipped with inverted wire semantics: it was documented and enforced as *"single-tenant; `account_id` is meaningless on the wire"*, and the framework refused inline `account_id` for it. That is backwards for the adapters the mode exists for. An agent fronting an upstream-managed roster (Meta / Snap ad accounts, AudioStack workspaces, a retail-media proxy) cannot accept the id its own upstream assigned — buyers who called `list_accounts` got ids that every subsequent call rejected (the dead end #1628 surfaced), and the only way to ship was to ignore `ref` entirely, which is a tenant-isolation bug the moment a deployment holds more than one account.

Upstream [adcp#5062](https://github.com/adcontextprotocol/adcp/pull/5062) settled the account-reference model. This corrects the SDK to match.

| | before | after |
|---|---|---|
| durable wire reference | none — account came from the credential | `{ account_id }` |
| inline `account_id` | refused (`INVALID_REQUEST`) | **accepted**, verified against the caller's reachable set |
| `{ brand, operator }` arm | accepted, passed to the resolver | **refused** (`INVALID_REQUEST`, `field: account.brand`, `list_accounts` suggestion) |
| `list_accounts` | optional | **required** (`PlatformConfigError` at construction; `opts.accounts.listAccounts` satisfies it) |
| `sync_accounts` | whatever `upsert` did | natural-key provisioning → per-row `UNSUPPORTED_PROVISIONING`; natural-key *reference* → per-row `INVALID_REQUEST`; `account: { account_id }` settings-update → passes through |
| `get_adcp_capabilities` | `require_operator_auth` unset | projects `account.require_operator_auth: true`, like a declared `'explicit'` |

## Tenant-isolation hardening

Accepting a buyer-supplied `account_id` moves verification onto the resolver, so the PR does not stop at the wire flip:

- **Framework backstop.** For `'derived'`, a resolved account whose `id` isn't the one the buyer named is refused with `ACCOUNT_NOT_FOUND` (plus a dev-mode warning naming the contract). An un-migrated resolver that ignores `ref` fails closed instead of cross-serving. Applied at all four buyer paths plus the comply controller.
- **`sync_accounts` / `sync_governance` are verified.** These are the two account references that never passed through `accounts.resolve` — the ref travels inside the batch. For `'derived'` every entry's `account_id` is now resolved against the caller's reachable set before any write, on both the platform interface and the merge-seam wiring (`opts.accounts.*`). `sync_accounts` fails the operation (`ACCOUNT_NOT_FOUND`) because the response row schema requires a natural key we don't have for an account we refused to resolve; `sync_governance` fails the individual row.
- **Comply-controller sandbox gate.** Fail-closed resolvers make "unresolved" reachable on demand, which re-opened the wire-claim fallback at `from-platform.ts`: `{ account_id: '<victim>', sandbox: true }` admitted the controller with no framework account authority, and the auto-seed adapters fall back to the buyer-supplied id for their storage namespace. The fallback is now scoped to refs that name no account, and the controller runs the same reference-shape check as every other path.
- **`createDerivedAccountStore` verifies by construction** rather than delegating: buyer ids are matched against the credential-scoped set (`toAccount` id, `listAccounts` roster, or a `lookupAccount` result whose id is re-checked), ref-less calls auto-select only when exactly one account is reachable, and `list_accounts` is filtered and paged for you.

## API

- `createDerivedAccountStore` gains `listAccounts` (+ optional `lookupAccount`) for credentials that reach many accounts — the shape that had no working Shape D before. `toAccount` keeps its signature for the one-account-per-credential case, so audiostack/flashtalking-shaped adopters change nothing but gain the verification, the singleton `list_accounts` row, and paging.
- New `AccountResolutionMode` / `CanonicalAccountResolutionMode` types and `normalizeAccountResolution()`, `isAccountResolutionMode()`, `refHasNaturalKey()` helpers, exported from `@adcp/sdk/server`. An unrecognized `resolution` string is now a `PlatformConfigError` instead of silently inheriting another mode's enforcement.
- New runnable reference: `examples/decisioning-platform-derived-accounts.ts`.

### Decision: no alias spelling

The issue asked whether `'account-id-namespace'` should be a safe alias. **No** — and no alias at all. `'account-id-namespace'` doesn't discriminate (under #5062 both `'explicit'` and `'derived'` are account-id namespaces). I prototyped `'upstream-managed'` as a spec-aligned second spelling; two reviewers independently argued it down and I agree: `resolution` is SDK-local config, not a wire value, so a second spelling buys nothing on the wire while silently defeating every adopter comparison written as `resolution === 'derived'` and inviting divergence from `adcp-client-python`. `'derived'` stays the single name; the taxonomy is explained where the name is misleading. There is exactly one code path per mode, behind `normalizeAccountResolution`.

### Changeset: `major`, and it does not start v15

`.changeset/pre.json` is in `rc` mode with `initialVersions: { "@adcp/sdk": "14.0.0-beta.31" }`, so changesets computes the next prerelease from that base: `semver.inc('14.0.0-beta.31', 'major') === '14.0.0'`. Verified empirically — `npx changeset version` on this branch produces **`14.0.0-rc.37`**, not `15.0.0` (nine `major` changesets are already accumulated in this pre set, so the aggregate bump was already major). `major` is therefore both accurate — this is a breaking correction — and safe. A `minor`/`patch` entry would compute the same version while under-recording the break in the v14 changelog.

## Docs

`docs/guides/account-resolution.md` (taxonomy table + rewritten `'derived'` section), `docs/migration-13-to-14.md` (new breaking section, four recipes, checklist item), `docs/guides/BUILD-AN-AGENT.md`, `CLAUDE.md`, `skills/build-decisioning-platform/{SKILL.md,advanced/REFERENCE.md}`, `skills/cross-cutting.md`, `scripts/generate-agent-docs.ts` + regenerated `docs/llms.txt`, and superseded-banners inlined in the 6.6→6.7 / 6.7→6.9 / 5.x→6.x guides where they still teach the inverted contract.

## Review

Followed `docs/development/REVIEW-STACKS.md`: four Claude experts (DX, protocol, code-review, security) in parallel **plus** the codex stack. Note: `npm run review:codex -- --all --base main` cannot run as-is in this cloud workspace — `codex exec -s read-only` shells out to `bwrap`, which refuses to start under the workspace's ambient capabilities (`Unexpected capabilities but not setuid`). I installed bubblewrap, confirmed it still refuses, and re-ran the same three persona prompts and the same auto-built context through `codex exec -s danger-full-access` instead. Findings landed normally.

**Convergent → fixed:**
- Merge-seam bypass: the `sync_accounts` gate only covered the platform-derived handler; `opts.accounts.syncAccounts` received natural-key entries it is documented never to see (codex-code BLOCKER). Now gated on both paths, with the platform path keeping request-aligned `accounts[i]` indices.
- Unverified `account_id` on `sync_accounts` / `sync_governance` (security HIGH-2) and the ref-ignoring-resolver fail-open (security MED-3, DX HIGH-1) — both fixed as described above.
- Unbounded `list_accounts` page in the factory (codex-code, codex-protocol, security LOW-4) — now clamps `max_results` to [1, 100] with an opaque offset cursor.
- Conformance-gate over-reach: I had tightened the #1624 account-discovery gate to require `list_accounts` whenever `require_operator_auth: true`. Three reviewers flagged it. #5062 makes that a MUST only for multi-account credentials and a SHOULD (with an out-of-band escape) otherwise — neither observable from a static capability read — and a declared `'explicit'` projects the same bit, so the gate would have failed conformant sellers *and* the escape hatch this PR's own docs recommend. **Reverted**; the strict rule is enforced where it is knowable (platform construction), and the gate's failure message now names `list_accounts` for account-id-namespace agents.
- In-place mutation of the adopter's returned `accounts[]` array (codex-code, security LOW-7) — rebuilt as new objects; regression test uses a cached adopter result.
- Unrecognized `resolution` strings fell into the derived branch (codex-code) — now a construction error, with the branch made explicit.
- Smaller: `refHasNaturalKey` arm asymmetry, `skipAuthCheck` truthiness, `getAccountMode` → shared `isSandboxOrMockAccount`, `brand_id` filtering, `field` pointer for the settings-update arm, keyless-entry handling, `tasks_get` swallowing typed `AdcpError`s, mode-aware `ACCOUNT_NOT_FOUND` suggestions, missing `refAccountId` imports in doc snippets, "the framework does not filter or page for you" on hand-rolled `list` snippets, restored a dropped `skipAuthCheck` test case.

**Found during review of the PR itself → fixed:**
- Three docs still claimed `'upstream-managed'` was an accepted alias after the runtime dropped it (`account.ts` mode table, the guide, the migration guide). Removed, and `test/lib/account-resolution-alias-drift.test.js` now pins the union members and fails any passage that reads as accepting a rejected spelling (verified by reintroducing the sentence).
- **Fail-closed bypass in the `sync_accounts` gate.** `classifyUpstreamManagedSyncEntry` accepted on a root `account_id` while the reachability check read `(entry.account ?? entry)`, so a hybrid entry — root `account_id` plus a nested `account: { brand, operator }` — was accepted on an id it never used, skipped verification, and reached the adopter as an unverified natural-key write target. Request validation is relaxable, so the schema's per-entry `oneOf` isn't a guarantee. The reference location is now resolved exactly once by `normalizeSyncAccountsReference`, both the classifier and the reachability check read that result, and an entry claiming two references is refused with `INVALID_REQUEST` rather than disambiguated by precedence. Same rule applied to `sync_governance`. Six regression tests cover it across the platform-derived and merge-seam paths with relaxed validation, each asserting no adopter side effect ran.

**Filed upstream, not absorbed:** `sync-accounts-response.json` requires `brand` + `operator` on every row, which an account-id-namespace seller may not have — so settings-update mode is only implementable by sellers that can echo a natural key, and per-entry failures for `account_id`-keyed entries are unrepresentable. Per CLAUDE.md's spec→mock→SDK triage order that is a spec bug, not an SDK one: [adcontextprotocol/adcp#7517](https://github.com/adcontextprotocol/adcp/issues/7517). The SDK documents the row requirement and escalates those failures to operation level meanwhile.

**Deliberately deferred** (tracked in #2915):
- `ACCOUNT_REQUIRED` (not `ACCOUNT_NOT_FOUND`) for auth-derived resolution that finds none or several — mode-independent envelope change that belongs in its own PR. The enum's `ACCOUNT_AMBIGUOUS` is scoped to *natural keys* resolving to multiple accounts, so it isn't the right code here.
- Storyboard grading for a derived platform that also advertises `sync_accounts` (settings-update): provisioning-shaped steps now get all-rows `UNSUPPORTED_PROVISIONING`. Grading that as `not_applicable` is a conformance-runner semantics change affecting third-party agents.
- `compliance:fork-matrix` coverage for the new derived reference adapter.

## Validation

- `npx tsc --noEmit` clean; `npm run typecheck:examples` clean; `npm run lint` 0 errors; `prettier --check` clean on every changed file.
- `npm run test:node:fast` — green except `test/lib/cli-project-dx.test.js`, which fails identically with this branch's changes stashed (the scaffolded-project check expects `DATABASE_URL` to be *absent*; this workspace exports one). Postgres-backed suites pass against a local PG started for the run.
- New/updated coverage: `test/server-decisioning-implicit-account-id.test.js` (41 cases — ref-shape flip both ways, unreachable id, config gates, merge-seam wiring, sync_accounts + sync_governance gating, resolver-verification backstop, strict-validation response), `test/derived-account-store.test.js` (42 cases — verification, roster scoping, auto-select, filters, pagination, auth gate), `test/server-decisioning-comply-gate.test.js` (sandbox-claim bypass), `test/lib/spec-conformance-account-discovery.test.js`, `test/server-decisioning-capability-projections.test.js`, `test/lib/account-resolution-alias-drift.test.js`.
- `npm run ci:doc-links`, `npm run generate-agent-docs` (llms.txt regenerated), `npm run typecheck:skill-examples` (no new errors; 2 baselined errors in `REFERENCE.md` fixed by this PR and removed from the baseline).

## Does this close #1628?

Yes. #1628 asked for a path to `list_accounts` for multi-tenant derived adopters, and its own triage concluded the additive `list?` callback (PR #1630, closed) was a dead end while `account_id` was refused. Both halves are now resolved: the factory takes `listAccounts` and wires the tool, `list_accounts` is required for the mode rather than opt-in, and the ids it publishes are referenceable on every subsequent call. The open scope question it left ("retire `createDerivedAccountStore` or narrow it to N=1?") is answered too — it covers both roster sizes, because the wire contract is identical.


bokelley:
**Review-thread status on head \`bfacfeb6\`**

Resolved the six `aao-secretariat` alias threads — all of them flagged `'upstream-managed'` being documented as an accepted spelling after the runtime union dropped it. Fixed earlier in this branch: the claim is gone from `src/lib/server/decisioning/account.ts`, `docs/guides/account-resolution.md`, and `docs/migration-13-to-14.md`, and `test/lib/account-resolution-alias-drift.test.js` now pins both the union members and the prose so the claim can't silently come back (verified by reintroducing the sentence and watching it fail).

Acted on the `github-code-quality` "Overwritten property" finding at `test/server-ctx-metadata-leak-paranoia.test.js:41` — that one was a real defect in my own edit. The fixture already declared `list` further down (returning the leak-canary rows the test asserts on), so the `list: async () => ({ items: [] })` I added while satisfying the new `'derived'` construction gate was a shadowed duplicate: dead code as written, and in the opposite declaration order it would have replaced the canary rows with an empty page and quietly weakened the leak assertion. Removed — that file is now byte-identical to `main`, and its 9 tests pass because the pre-existing `list` already satisfies the gate.

`bfacfeb6` differs from `3cd58da6` by exactly that one deleted line; no product code changed.

bokelley:
Acknowledged — the six alias threads resolved and the shadowed `list` duplicate removed on `bfacfeb6`. Clean one-line delta with no product-code change; thread is in good shape.

---

bokelley:
Auto-merge is temporarily disabled pending the already-requested human maintainer review.

The immediately preceding #2911 merge demonstrated that an `aao-secretariat[bot]` `APPROVED` review can satisfy the repository's one-review rule and complete auto-merge before the requested human responds. Automated Ladon review remains valuable evidence, but it does not satisfy the coordinator's human-only merge gate for this adoption-sensitive change.

Keep the existing CI and human review request running. Re-enable merge automation only after an eligible human approves the exact final head and all required checks are terminal green. No source or review decision is changed by this hold.

bokelley:
Coordinator safety hold: this PR is now Draft with auto-merge off because the current automated review can satisfy the repository review rule without an exact-head human maintainer approval. Source and head are unchanged. Please keep it Draft until the requested human maintainer has reviewed the exact head; then re-run/confirm exact CI before any Ready or merge action.

PR #2918, opened 2026-09-14
fix: address SDK adoption reports

### LINKED ISSUE #2903: Transport diagnostics can still block completion on stalled bodies and async observers (rc.36)
Verified in published `@adcp/sdk@14.0.0-rc.36`, npm gitHead `90ad47b67401d35e0a9ffc99fd0e76252e026a9a`, Node 26.4.0. This is a remaining critical-path problem, not a claim that #2503's tee-cancellation fix is absent: rc.36 already uses non-awaited `reader.cancel()` after the 64 KiB limit and skips SSE capture.

Two bounded, entirely in-memory reproductions against both published CJS and ESM artifacts show:

1. A JSON response whose headers and first chunk are available does not return from the diagnostic fetch until EOF (or the size cap). A stalled body below 64 KiB has no capture time bound. The original response is withheld because `wrapFetchWithTransportDiagnostics` awaits `responseBodySnippet`, which awaits `reader.read()`.
2. A never-settling async `onTransportActivity` callback prevents the enclosing result from settling, even after its function and response body finish. `TransportActivityHandler` explicitly permits `Promise<void>`; `withTransportDiagnostics` waits for every observer promise in its `finally` block.

### Minimal reproduction of the body path

This imports the shipped implementation helper for a focused reproduction; applications normally enable the public `onTransportActivity` option.

```js
import { createRequire } from 'node:module';
import { dirname, join } from 'node:path';
import { pathToFileURL } from 'node:url';
import { setTimeout as delay } from 'node:timers/promises';
const require = createRequire(import.meta.url);
const root = dirname(require.resolve('@adcp/sdk/package.json'));
const { withTransportDiagnostics, wrapFetchWithTransportDiagnostics } =
  await import(pathToFileURL(join(root, 'dist/lib/protocols/transportDiagnostics.mjs')));

let controller;
const stream = new ReadableStream({ start(c) {
  controller = c;
  c.enqueue(new TextEncoder().encode('{"ok":true}'));
}});
const diagnosticFetch = wrapFetchWithTransportDiagnostics(async () =>
  new Response(stream, { headers: { 'content-type': 'application/json' } }));
let returned = false;
const result = withTransportDiagnostics({
  agentId: 'example', protocol: 'mcp', onTransportActivity: () => {},
}, async () => {
  const response = await diagnosticFetch('https://seller.example/mcp');
  returned = true;
  return response.text();
});
await delay(100);
console.log(returned); // false: even the Response is withheld
controller.close();  // deterministic cleanup of the test double
console.log(await result); // {"ok":true}
```

For the observer path, return an already-complete JSON `Response`, use an `onTransportActivity` promise with a manually held resolver, and track when the function above returns. The function completes, but `withTransportDiagnostics` remains pending until the test releases the observer. Both cases reproduced in CJS and ESM; diagnostics-disabled controls complete. No network or provider call was made.

### Requested behavior / DX

Enabling observability should not add an unbounded dependency to request completion. Bound body-preview capture by time as well as size, preserve the original stream, and make observer completion best-effort or explicitly bounded. If a caller wants flush guarantees, expose/document a separate bounded flush rather than making every request await its logging backend indefinitely.

Regression coverage should include a stalled sub-limit JSON/text body, a never-settling observer, intact original body, bounded/truncated or omitted preview, SSE behavior, and both published module formats. Preserve redaction and #2503's existing fix. A maintainer-selected finite capture budget is fine; this report does not prescribe a particular timeout.

### LINKED ISSUE #2904: rc.36 postal Zod fallback rejects valid countries outside the registered list
Verified against published `@adcp/sdk@14.0.0-rc.36` (npm gitHead `90ad47b67401d35e0a9ffc99fd0e76252e026a9a`) using its public schemas, Zod 4.1.12, AJV 8.20.0 and ajv-formats 3.0.1. This is a distinct remaining postal code-generation problem from #2628's structural `allOf` field loss.

The pinned AdCP 3.2.0-rc.2 `postal-country-system.json` accepts `postal_code` or `custom` for countries outside the registered-country list. Its fallback is a `not: { enum: [...] }` refinement of the string country. The generated Zod fallback instead requires `country: z.object({}).passthrough().optional()`. That branch cannot accept the same country's string value required by the outer intersection.

### Public reproduction

```js
const { PostalCountrySystemSchema, PostalAreaSchema } = require('@adcp/sdk/schemas');

console.log(PostalCountrySystemSchema.safeParse({
  country: 'PT', system: 'postal_code',
}).success); // false; pinned JSON Schema accepts
console.log(PostalCountrySystemSchema.safeParse({
  country: 'PT', system: 'custom',
}).success); // false; pinned JSON Schema accepts
console.log(PostalAreaSchema.safeParse({
  country: 'PT', system: 'postal_code', values: ['1000-001'],
}).success); // false; pinned JSON Schema accepts
```

Controls: `{country:'US', system:'zip'}` passes both; `{country:'US', system:'custom'}` fails both. All three PT cases above fail the SDK with `invalid_union` while AJV accepts them. The packaged postal-country/area JSON Schemas are structurally equal to the published schemas at protocol commit `929ceb16090b2f11ad20348b889780097173b00a` (formatting differs); this is SDK generation drift, not a protocol ambiguity.

Sources:
- [Pinned country/system schema](https://github.com/adcontextprotocol/adcp/blob/929ceb16090b2f11ad20348b889780097173b00a/dist/schemas/3.2.0-rc.2/core/postal-country-system.json)
- [Pinned postal-area schema](https://github.com/adcontextprotocol/adcp/blob/929ceb16090b2f11ad20348b889780097173b00a/dist/schemas/3.2.0-rc.2/core/postal-area.json)
- Published `dist/lib/types/schemas.generated.mjs`, `PostalCountrySystemSchema` fallback around line766; equivalent CJS output.

Please preserve primitive `not`/enum exclusion through generation and add public-export parity tests for a non-registered country, both allowed fallback systems, a registered-country invalid fallback, and containing `PostalArea`. Keep outer required/string/pattern and non-empty postal-values constraints. Adopters should not need a country-specific cast or patched generated artifact for valid wire input.

### LINKED ISSUE #2905: Discovery DX: give validateAdAgents a caller cancellation and whole-operation deadline contract
SDK `14.0.0-rc.36`, source `90ad47b67401d35e0a9ffc99fd0e76252e026a9a`. This is a source/declaration-verified additive API request for `validateAdAgents`, separate from the client read-method cancellation repaired after #2308 and absolute task deadlines in #2453.

`SingleAgentClient.getCapabilities(options)` now accepts/checks a caller signal. The separately exported `validateAdAgents(publisherDomain, options)` still exposes only per-request `timeoutMs`, body limit, User-Agent, log level and URL builder. `ValidateAdAgentsOptions` has no `signal`; the implementation starts direct `.well-known/adagents.json`, authoritative-location, `ads.txt`, and manager-domain fetches without caller cancellation.

An adopter doing one deadline-bound discovery operation cannot use a consistent cancellation contract across these helpers. `timeoutMs` provides a useful finite per-request bound, but it is not the caller's cancellation or an absolute budget for the whole multi-request discovery. Racing the returned promise does not cancel underlying I/O. I did not run a live discovery request for this report.

### Proposed additive seam

Accept `signal?: AbortSignal` on `ValidateAdAgentsOptions` and thread it through all existing safe-fetch paths. Reuse one absolute operation deadline when supplied, while retaining individual request/body limits. Keep the SSRF-safe transport, redirect/manager-domain rules and validation behavior intact; no raw caller fetch bypass is needed. An aborted discovery has no remote task ID, so this must not dispatch A2A `tasks/cancel`.

Please document the relationship between caller abort, per-request timeout and whole-operation deadline, including the stable error/result classification and whether a fallback may start after cancellation (it should not).

Acceptance coverage can be completely synthetic: pre-aborted signal produces zero I/O; abort during each discovery phase reaches the underlying safe fetch; no later fallback starts after abort; the complete operation cannot reset its deadline at each hop; existing successful direct/manager cases and SSRF/redirect refusal controls remain unchanged.

### LINKED ISSUE #2906: SDK 14 adoption DX: a release-bound upgrade worksheet and thin existing-platform recipe
Adopter feedback from reviewing the published SDK `14.0.0-rc.36` artifact (`90ad47b67401d35e0a9ffc99fd0e76252e026a9a`), its README, 13-to-14 migration guide, buyer/seller quickstarts, declarations, and Reporting integration guides. This is a DX proposal, not a claim that these existing guides or APIs are absent.

What is working well: focused runtime and per-tool type entrypoints, explicit schema imports, persona quickstarts, the server-default-versus-supported-ceiling explanation, durable-store guidance, and #2888's new narrow consumer-status store port. The latter plus the calendar recipe materially reduced our integration work. The five consumer-status schema gaps from that review are fixed in the released artifact.

The remaining friction is connecting those pieces into **one supported path for an existing application**, rather than learning through generated-file inspection and compatibility patches.

### 1. A compact, release-bound upgrade worksheet

The README's prerelease install range and beta-era entry link are useful for exploration; production adopters also need an exact release recipe. Put the following together for each release: npm version/integrity, default wire release, compatible peer releases, required peer/runtime versions, and the canonical migration notes. These facts already exist in different places; the request is a concise joined view, not another source of truth.

A worked rc.33/rc.35 → rc.36 example should call out `3.2.0-rc.1` → `3.2.0-rc.2`, coordinated peer upgrade, preserving a server's explicit default, and the release's malformed-surrogate idempotency-fingerprint change. Generate the worksheet from existing metadata where possible. Success: an adopter can decide what changes on the wire and what durable compatibility they must retain without searching the changelog, code and schema cache separately.

### 2. An existing-platform recipe with explicit ownership boundaries

Keep the current buyer/seller starters, and add a thin-adopter path: existing authenticated request context + existing transaction/store + one SDK handler/client, with public imports and deterministic local doubles. Start with one task and grow to asynchronous completion/recovery. Link to the detailed durability and Reporting guides rather than duplicate them.

Show which concerns the SDK owns, which application contracts must be implemented, which defaults are process-local, and which examples are production-capable only after durable adapters are installed. A sample should compile against the **packed npm artifact**, in the documented module-resolution mode, without private `dist/...` imports or type casts. Use the same public entrypoints recommended in the README. Keep schema validators opt-in so the basic example retains the documented small type footprint.

### 3. A supported way to reuse application-owned discovery evidence

We already resolve/cache seller capabilities under a bounded, tenant-scoped preflight, but SDK client operations can discover again through their own cache. We found `getCapabilities`/`refreshCapabilities` and scoped-fetch protections; the public client declarations do not offer the `primeCapabilities` seam we currently carry locally.

Please document an existing supported route if there is one, or consider a narrow validated snapshot/cache adapter. Its contract needs endpoint + authorization/transport scope + negotiated version + freshness/invalidation, and must preserve synthetic-capability provenance and tool-schema behavior. Do not accept an arbitrary globally shared allow snapshot or bypass the existing trusted-fetch scope isolation. Success: compose a current, appropriately scoped capability observation without patching generated CJS/ESM/declarations or making a redundant probe; stale or mismatched evidence still causes fresh discovery/refusal.

### 4. One error/cancellation/observability story

A short matrix and executable examples across client task, discovery, async wait and callback would help: supported signal/deadline option, thrown error versus failed result, when retries reuse identity, and how to attach diagnostics safely. The concrete remaining issues are #2903 (diagnostics on the critical path) and #2905 (`validateAdAgents` cancellation). #2904 records a separately reproduced schema bug. These should not be hidden by examples that catch everything or convert every failure into a string.

Suggested acceptance: a consumer-only smoke project installed from `npm pack` demonstrates the documented imports, exact version selection, cancellation, scoped capability reuse, current authorization before replay, a mixed-status response and a durable recovery path, without provider credentials. This can reuse existing acceptance vectors and store doubles rather than create another full conformance framework.

bokelley:
## Summary

- bound transport diagnostic body capture and observer flushing so diagnostics cannot stall CJS or ESM requests
- preserve valid postal-code fallbacks for countries outside the protocol's registered country patterns
- add caller cancellation and whole-operation deadlines to adagents discovery across direct, authoritative, ads.txt, and manager phases
- add scoped, expiring capability evidence reuse plus an existing-platform guide, runnable providerless integration, and deterministic 14.x release worksheet

## Validation

- expert DX, protocol, code, and security reviews: no remaining blockers
- `npm run ci:docs-check`
- `npm run ci:doc-links`
- `npm run ci:codegen-strict`
- `npm run format:check`
- `npm run typecheck`
- `npm run typecheck:examples`
- focused regression suites: 168 passing
- clean-room `npm run verify:package` with pnpm 10.18.3: Node CJS/ESM, Bun MCP, browser bundle, peer floors, packed examples, and providerless submitted-task recovery passing

The repository's `npm run review:codex -- --all --base main` entrypoint was invoked, but its nested Codex sandbox could not start because `bwrap` is unavailable on this worker; it returned no findings. Independent in-workspace and Claude expert reviews completed successfully.

Closes #2903
Closes #2904
Closes #2905
Closes #2906

---

[Open workspace in Conductor](https://app.conductor.build/workspace/5d8c49df-bd12-461e-a865-48a8759252ae)


bokelley:
Coordinator safety hold: the legacy approval-capable Ladon run posted a Secretariat bot APPROVED review on this exact head while the repository still lacks a trusted human-only gate. I marked the PR Draft and confirmed auto-merge is off. This is not a patch finding and does not erase review history. An authorized administrator must neutralize legacy bot approvals/runs through the documented auditable procedure, then require fresh exact-head CI and actual non-author human review before lifting this hold.

PR #2922, opened 2026-09-14
feat(reporting): rc.3 buyer-side consumer-status loop

bokelley:
Buyer side of rc.3 Reliable Reporting. #2914 (the seller ledger) is merged; this now targets `main` directly. #2923 stays stacked on it.

## rc.3 coverage matrix

| rc.3 requirement | Implemented in | Tested by |
|---|---|---|
| `content_mismatch` with `scope_media_buy_missing` | `reporting/content-mismatch.ts` `detectReportingContentMismatch` | *scope_media_buy_missing needs rows, because the metadata check is a tautology*, *an explicit zero row represents the buy, so it is not missing* |
| … `coverage_short` | same | *coverage_short fires from metadata alone*, *coverage_short is unconditional — allow_partial already froze the reduced set* |
| … `metric_missing` | same | *metric_missing compares against observed row metrics, not control totals* |
| … `schema_nonconformant` | same | *schema_nonconformant reports a row validation failure, not a pin difference* |
| … `currency_mismatch` | same | *currency_mismatch fires from control-total units*, *a control total with no declared unit does not disagree* |
| … `period_mismatch` | same | *period_mismatch is decided from row time-dimension values, not the envelope*, *the period is half-open, so a row exactly at the end instant is outside* |
| Spec precedence (`metric_missing` before `schema_nonconformant`) | same, ordered evaluation | *metric_missing wins over schema_nonconformant, per the spec precedence note* |
| `content_mismatch` is never a measurement dispute | same — no arm compares a delivered value against a buyer expectation | *never fires on a delivered value the buyer merely disagrees with* |
| Post statuses **by the deadline**, not only before scope close | `reporting/reconciliation.ts` `planReportingConsumerStatuses` + `reportingConsumerStatusDeadline` (period end + the buyer-pinned `automatedRecoveryWindowSeconds`), posted in `reconcileReporting` via the optional `client.syncReportingStatus` | *a pinned recovery window produces the deadline and flips overdue across it*, *an unpinned recovery window marks nothing overdue rather than everything* |
| An exact `content_mismatch` retry survives the revision being superseded | ordering in `ledger/status-ingest.ts` — the batch replay short-circuits at the `getConsumerStatusBatchReplay` return, before `validateStatus` runs its rc.3 currency check | *an exact content_mismatch retry replays after the revision is superseded* (`reporting-status-ingest-validation.test.js`) |
| `received` proves consumption instead of echoing the ledger | `reconciliation.ts` `consumeReportingRevision` — pages `reporting_rows` for the exact revision through the new optional `client.getMediaBuyDelivery` and recomputes SHA-256(JCS({reporting_revision_id,row_count,control_totals,reporting_rows})) | *posts revision_missing, then says nothing, then supersedes it with a consumed received* |
| A revision that does not hash to its published digest is `unreadable`/`integrity_mismatch` | `attestConsumerStatusPlan` | *a revision whose rows do not hash to its digest is unreadable, not received* |
| No exact-revision reader ⇒ nothing attested, nothing posted | `attestConsumerStatusPlan` → `suppressed: 'consumption_unavailable'` | *without an exact-revision reader the buyer plans received but never attests it* |
| `obligation_missing` / `revision_missing` dated at or after `expected_at` | `reportingExpectedAt` — obligation's `expected_at`, else `period.end` + the new `deliverySlaSeconds` pin | *obligation_missing derives expected_at from the buyer pin and the seller accepts it* (with a companion assertion that period-end dating is refused), *obligation_missing is dated at expected_at, not the period end* |
| `status_as_of` never moves backwards along the chain | `earliestStatusAsOf` floor + `latestInstant` | *posts revision_missing, then says nothing, then supersedes it with a consumed received* |
| A statement names the caller's exact current leaf | `currentConsumerLeaf` over the new `ReportingLedger.consumerStatuses` | same test — asserts the two-statement chain and its `supersedes_reporting_status_id` |
| An unchanged claim is not re-posted | `consumerStatusSuppression` → `suppressed: 'unchanged'` | same test, second and fourth passes |
| Item-local rejections are surfaced | `failedConsumerStatuses` on the result | *an item-local rejection is surfaced with its errors, not silently dropped* |
| A statement is reconstructible after a lost response | `consumerStatusId` (claim-derived, no `status_as_of`) + `consumerStatusBatchKey` (body-derived `idempotency_key`) | *a lost response does not produce a second statement, and an exact retry replays* — re-plans against a ledger that discloses nothing about the chain, so identity carries it alone |
| Row-level `mismatch_code`s reachable through the loop | second `detectReportingContentMismatch` pass in `attestConsumerStatusPlan` with `rowEvidenceFor` | *a metric promised by the pinned definition and absent from the rows is metric_missing* |
| `unreadable` failure codes | `consumeReportingRevision` | *a read that throws is unreadable/transport_failed*, *a reader that cannot serve the exact revision is unreadable/reader_incompatible*, *a revision whose rows do not hash to its digest is unreadable, not received* |
| A failed batch does not discard earlier batches | `recordConsumerStatusBatchFailure` + `break` | *a lost response does not produce a second statement, and an exact retry replays* |
| The buyer's own budget is not a seller failure | `suppressed: 'budget_exhausted'` | *running out of the read budget stays silent rather than accusing the seller* |
| Surface `consumer_status_pending` | `readReportingConsumerStatusPending` — a separate summary read, since the field is summary-only and reconciliation reads periods | returns `undefined` rather than throwing when the seller doesn't advertise it |
| Surface issue `opened_at` / `issue_state` / `external_ref` | `collectReportingEscalations` → `ReportingEscalationV1` | carried through the reconciliation result; the emitting side is tested in #2914 |
| Surface `operations_contact` | same, attached per issue, plus `requiresHumanContact` for the `contact_*` family | as above |

`test/lib/reporting-content-mismatch.test.js` (19 cases), the retry-ordering case in `test/lib/reporting-status-ingest-validation.test.js`, and `test/lib/reporting-buyer-consumer-status-loop.test.js` (11 cases) — a new end-to-end suite that drives the real reconciler against the SDK's **own** producer, `get_reporting_status` handler, exact-revision `get_media_buy_delivery` reader and `sync_reporting_status` ingest. Nothing is stubbed, so a statement the buyer plans has to survive the same `validateStatus` a real seller would run it through. `MemoryLedgerStore` moved to `test/helpers/` so both halves of the loop can share it.

## The part worth reviewing hardest

`content_mismatch` is easy to get wrong in the permissive direction, and getting it wrong is expensive: it pins the buyer's own caller-scoped view at `action_required` until the buyer backs down, and it puts a claim in front of a seller who can neither resolve nor ignore it.

The spec is explicit that it is **never a measurement dispute** — "A consumer MUST NOT use content_mismatch to dispute how many impressions the seller counted; that is a measurement disagreement, handled by measurement_terms and makegood_policy." So the detector is built to be structurally incapable of it: **nothing compares a delivered number against a buyer-side expectation.** Only names, units, membership, coverage, period, and schema identity. There's a test asserting that a revision with wildly different totals is *not* a mismatch, precisely so a future change that adds a value comparison fails.

**Four of the six codes require row evidence the caller actually gathered**, and stay silent without it. This is the correction in `86124881`: an earlier draft decided `scope_media_buy_missing` by comparing the obligation's `media_buy_ids` against the revision's, but `reporting-revision.json` defines that field as the denominator *"inherited from the obligation, including buys with zero rows"* — so against a conformant seller the comparison is a tautology that can never fire. `metric_missing`, `schema_nonconformant` and `period_mismatch` are likewise row-level predicates that revision metadata cannot decide. Each is gated on the matching field of `ReportingRowEvidenceV1`; a buyer that did not read rows cannot accuse. Only `coverage_short` and `currency_mismatch` are decidable from metadata alone.

Two deliberate behaviours worth checking against the spec rather than intuition:

- **`coverage_short` is unconditional, including under `allow_partial`.** The obligation's frozen `coverage.covered_package_ids` *is* the effective denominator once the generation is accepted, so a revision narrower than that frozen set is still short of what was promised. (An earlier draft exempted `allow_partial`; that was wrong.)
- **A media buy carried by an explicit zero row is represented, not missing.** That zero row is precisely what lets a revision distinguish zero delivery from an omitted buy, which is the spec's exculpating condition — so the predicate is "absent from the rows *and* unrepresented by a zero row", not "absent from a total".

And `unreadable` is deliberately **not** planned: it means the buyer could not consume named bytes, which only the caller's own reader knows. Inferring it from the ledger would be guessing about someone else's failure.

## Safe-by-default choices

- `client.syncReportingStatus` is **optional**. Without it the reconciler still plans every status and reports it on the result — existing adopters see new fields and no new behaviour.
- Only **overdue** statuses are posted. Posting early would churn the chain for periods the buyer may still resolve itself.
- **An unpinned recovery window means nothing is overdue.** `automated_recovery_window_seconds` is advertised on the delivery capabilities, not on the obligation, so the ledger cannot supply it. Had a missing pin defaulted to overdue, every reconcile would post a status for every period immediately — the exact opposite of posting by the deadline.
- `postedConsumerStatuses` lists only what the seller recorded. `sync_reporting_status` is a partial-success batch, so a failed item must never be counted as posted.
- `consumerStatusPending` swallows a read failure into `undefined` rather than failing reconciliation — it is visibility, not evidence.
- `ExpectedReportingPeriod.committedMetrics` / `.metricUnits` are optional, and omitting either **disables** that check. A buyer that never recorded the metric list must not claim a promised metric is absent.
- The posted status ID is content-derived and includes a deterministic `statusAsOf`, so a retry after a transport failure is the same statement rather than an idempotency conflict.

## Note on #2911

Its ingest change already closed the seller-side half of the `content_mismatch` binding rule (the digest check now applies to any status carrying `observed_revision_content_sha256`, not just `received`), which was a finding I'd earmarked for this PR. Nothing left to do there.

---


bokelley:
## Review stack found this PR unsound as first pushed. Fixed in `dc8adcbc` + `718025cc`.

Ran the protocol and code reviewers in parallel. They converged, and the honest summary is that **most of the mismatch arms were testing a different predicate than the spec defines** — two of them could never have fired correctly. Recording the findings in full because the failure mode here is one-directional: a false `content_mismatch` forces the caller's view to `action_required`, and per `consumer_mismatch_lifecycle` the seller may not clear it while the statement is the current leaf.

### The two that could never work

**`scope_media_buy_missing` was a tautology.** I compared the obligation's `media_buy_ids` against the revision's. But `reporting-revision.json` defines the revision's copy as the denominator *"inherited from the obligation, including buys with zero rows"*, and `x-adcp-validation.coverage` adds `coverage.media_buy_ids MUST equal media_buy_ids`. Against a conformant seller those are equal **by construction**. I verified this directly rather than take it on report:

```
revision.media_buy_ids: "Exact frozen media-buy denominator inherited from the
obligation, including buys with zero rows."
```

Worse, my doc comment had the spec's qualifier backwards — the explicit zero row is what *exculpates* an absent buy, not what makes it present — and a test encoded the inverted reading.

**`metric_missing` used the wrong denominator.** It resolved `metrics[].name` against `control_totals`, which are profile-defined aggregates scoped to `coverage.covered_package_ids` — a different, usually smaller set. An adopter following my JSDoc exactly would have emitted a false `content_mismatch` for every promised metric the seller did not also publish as a control total.

`period_mismatch` and `schema_nonconformant` were similarly testing the envelope and the pin rather than row time-dimension values and row validation.

### The fix: stop pretending metadata can answer row-level questions

Four arms now require `ReportingRowEvidenceV1` — what the caller's reader actually observed — and **stay silent without it**. Only `coverage_short` and `currency_mismatch` are decidable from metadata alone. Silence is deliberate: not accusing is recoverable, accusing wrongly is not.

`coverage_short` also loses its `allow_partial` skip. Both reviewers independently flagged it: the obligation's frozen `covered_package_ids` *is* the reduced denominator under `allow_partial`, so a revision narrower than it is still short. The skip locked in a false negative and a test locked it in alongside.

### Three more, same root cause — guessing where a field lives

| defect | reality |
|---|---|
| deadline read `obligation.automated_recovery_window_seconds` | The field is on `reporting-delivery-capabilities.json`; `reporting-obligation.json` is `additionalProperties: false` and has no such key. It was **always `undefined`**, and the fallback treated unknown as *overdue* — so with `syncReportingStatus` wired, every reconcile would have posted a status for every period immediately. Now a buyer-held pin, and an absent pin means **not** overdue. |
| `operations_contact` read from `ledger.scope.reporting_delivery` | `scope` is `additionalProperties: false` with no such key. Dead code that could never populate, while the changeset claimed it worked. Now a buyer-held pin. |
| `period.source_timezone` hardcoded `'UTC'` | It is part of the chain's logical key, so a wrong value **forks the chain silently** rather than failing. Read from the obligation, with a pin for the omitted-obligation case. |

### One I introduced while fixing another

`dc8adcbc` made `reporting_status_id` content-derived so exact retries reuse it. But `status_as_of` was still `new Date()` — so an "exact retry" reused the same ID with *different* content, which the spec calls an idempotency conflict outright. I turned "forks the chain" into "every retry hard-fails". `status_as_of` is now derived from immutable evidence: the revision's observation instant for `received` (which is also what the spec asks for — *"when the named revision first became consumable"*), the deadline otherwise.

Also fixed: statuses post in one batch of up to 100 rather than one request each against a shared budget, so a later timeout no longer discards the record of what the seller already accepted; and the summary read is gated on the seller advertising the task, strips the periods-only keys the summary view forbids, and runs inside the reconcile budget.

### Still open, deliberately

- **`unreadable` has no seam.** It means the buyer's reader failed on named bytes, which only the caller knows, and `consumerStatuses` is currently output-only. A buyer whose read failed will plan `revision_missing`, which is the wrong status. Fixing it means designing an input seam, and I would rather do that deliberately than invent an API under time pressure. Flagging rather than shipping a wrong status silently.
- **`observed_revision_content_sha256` is echoed from the seller's own record** rather than independently recomputed. The spec requires the consumer to recompute it, and parroting the seller's claim removes the evidentiary value. `reporting-revision.json#digest_requirement` gives the recipe, so this is buildable — it needs the row bytes, which is the same seam as above.
- **`metrics[].name` vs `control_totals`** may be a genuine spec ambiguity rather than an SDK bug. Per CLAUDE.md's `spec → mock → SDK` order that is worth filing on `adcontextprotocol/adcp` rather than guessing; happy to open it.

109 tests pass across the reporting suite.

bokelley:

PR #2923, opened 2026-09-14
docs(media-buy): rc.3 changeset audit + targeting null-clear docs

bokelley:
**Stacked on #2922.** The rc.3 changeset audit the brief asked for, the one docs gap it surfaced, and four follow-up fixes from the #2922 review stack. The audit itself found no runtime change needed beyond what already merged; the reporting fixes below are separate, and are here so they make the same prerelease.

## rc.3 coverage matrix — audit of the remaining changesets

| rc.3 changeset | Runtime impact | Evidence |
|---|---|---|
| **`add-media-buy-frequency-cap-scope` (adcp#7449)** — MediaBuy-level `frequency_cap`, one counter across packages | **No runtime change needed.** rc.3 is the first bundle to publish `core/media-buy-available-action-id.json`, so codegen emits `STRUCTURED_ONLY_MEDIA_BUY_ACTIONS = ["update_media_buy_frequency_cap"]` and the `frequency_cap → update_media_buy_frequency_cap` binding into `ACTIONS_BY_FIELD`. | `src/lib/media-buy/preflight.ts:211` already resolved that action **from the generated table** rather than hardcoding it (#2889), so it activated on the pin alone. `update-fields.generated.ts:158,172`. |
| … `media_buy.aggregate_frequency_capping` capability | **No runtime change needed** — passthrough. | Present in the generated capability type (`tools.generated.ts`, 2 refs) and in `create-media-buy-request` / `buy-products-request` / `get-adcp-capabilities-response`. Grepped `src/` for any capability allowlist that could drop it: none. |
| … `update_media_buy_frequency_cap` action | **No runtime change needed.** | Dispatches buy-level `frequency_cap`; verified distinct from the package-level path below. |
| **`add-scoped-frequency-cap-support` (d44756a)** — `overlay_support.frequency_cap_support`, constraint-aware discovery | **No runtime change needed** — passthrough. | `frequency_cap_support?: FrequencyCapConstraints` on `core.generated.ts:4423`; `FrequencyCapConstraints` / `FrequencyCapRequirements` typed (7 refs). `required_overlay_support` is forwarded as-is in `negotiation/buyer.ts` and `media-buy/compatibility.ts`, so constraint-aware discovery reaches buyers unmodified. |
| … package-qualified `update_frequency_caps` | **No runtime change needed** — already package-qualified. | `update-fields.generated.ts:108,190` binds it to `packages[].targeting_overlay.frequency_cap`, and `preflight.ts:375-379` emits `packages[${index}].targeting_overlay.frequency_cap`. The buy-level and package-level caps are separate actions with separate field bindings; neither shadows the other. |
| **`targeting-input-null-clear-semantics` (adcp#7466)** — omitted / replaced / `null`=clear | **Needed a runtime change; it merged in #2907.** | `resolveTargetingInput` / `applyTargetingInput` / `hasTargetingClears`, plus two real bugs closed in `createMediaBuyStore` where a `null` clear command could reach durable state and then be echoed into a `get_media_buys` response. Also a codegen fix: `TargetingOverlayInput.device_platform` / `.device_type` were emitted as scalar enums where the wire carries arrays. |
| … store merge semantics | **Verified correct, then hardened.** | `mergeFromUpdate` already implemented the three states; the gaps were `persistFromCreate`'s request fallback and the `new_packages` arm. `test/lib/targeting-input-null-clear.test.js`. |
| … update-fields generation | **No change needed.** | `ACTIONS_BY_FIELD` binds targeting through `update_targeting` / `update_frequency_caps` unchanged; the null-clear distinction is a value-level contract, not an action-level one. |

Net: **one changeset needed runtime work and it already merged; the other two are passthrough**, because the frequency-cap dispatch was built to read the generated table instead of hardcoding action names. That earlier decision is why a protocol-level feature addition cost nothing here.

## Follow-up fixes to the buyer loop (#2922 review)

Four correctness fixes landed here so they ship in the same prerelease as the audit. Each row names the test that backs it, all in `test/lib/reporting-buyer-consumer-status-loop.test.js`.

| Fix | Implemented in | Tested by |
|---|---|---|
| A malformed seller `expected_at` is no longer re-emitted as the buyer's `status_as_of` — it must match RFC 3339, not merely parse | `reconciliation.ts` `reportingExpectedAt` + `consumerStatusSchedule` (`missingPin` now names `obligation.expected_at`) | *a malformed seller expected_at is not re-emitted as the buyer status_as_of* |
| A chain the buyer could not walk suppresses as `chain_indeterminate` whether or not a head resolved, instead of attesting `received` on it | `reconciliation.ts` — the indeterminacy test moved out of the `!revision` branch | *a forked revision chain suppresses rather than blaming the seller* |
| A superseded sibling can no longer turn a sound current revision into `revision_missing` | `REVISION_DISQUALIFYING_REASONS` no longer carries `REVISION_CHAIN_SCOPE_MISMATCH` | *a superseded predecessor does not make a valid current revision missing* |
| Budget guard rethrows non-`RangeError`; row-size estimate recurses instead of charging a nested object 16 bytes | `consumeReportingRevision` / `approximateRowBytes` | covered indirectly by *a buyer-set record limit suppresses instead of accusing the seller*; the rethrow is a defect-surfacing path with no constructible fixture |
| `.changeset/reporting-rc3-buyer-reconciler.md` said `suppressed: 'budget_exhausted'`; the code says `local_budget_exhausted` | changeset text | n/a — doc correction |

On the third row: `assertReportingLedgerGraph` refuses an off-scope revision before planning ever runs, so the disqualifying-set entry was unreachable in practice. The narrowing is defence in depth and the test covers the in-scope predecessor case it actually protects — stated plainly rather than claiming coverage of a path I could not construct.

## The gap the audit surfaced

The null-clear helpers shipped with tests and a starter call site but **no prose** — `resolveTargetingInput` appeared in no guide and no skill, so the only way to learn that `null` means "clear" and omission means "preserve" was to read the source. Both failure modes are silent, which is the worst combination for an undocumented distinction.

- `MEDIA-BUY-3.2-COMPATIBILITY.md`: the three-state table for create *and* update, worked snippets for all three helpers, the "returns `undefined`, never `{}`" rule, and the DR-0020 point that `null` cannot remove inherent product scope.
- `skills/build-seller-agent/`: the short form under shape gotchas — a seller generated from that skill is exactly the code that gets this wrong.

**Deliberately not added to `docs/TYPE-SUMMARY.md`**, despite my saying I would on #2919. That file is generated by `scripts/generate-agent-docs.ts` from the schema index and carries *protocol* types; `createMediaBuyStore` and the AAO helpers are absent for the same reason. Adding SDK helpers would have meant hand-editing a generated file to make it inconsistent with its own scope. Flagging the reversal rather than quietly dropping it.

Prompted by #2919 reworking the starter to reject supplied overlays outright — more correct per DR-0020, and it removes the repo's only worked example, so the docs needed to exist before that lands.

---


bokelley:

Issue #2878, opened 2026-09-12
Storyboard runner sends auth-override probes as MCP tools/call even when the protocol under test is A2A (14.0.0-rc.35)


KonstantinMirin:
## Summary

In `@adcp/sdk` 14.0.0-rc.35 the storyboard runner sends every step that carries an `auth:` override through `rawMcpProbe`, whatever protocol is under test. During an A2A run the `security_baseline` probes (`probe_unauth`, `probe_invalid_api_key`) therefore POST an MCP `tools/call` JSON-RPC envelope at the A2A agent URL. A conformant A2A server answers that with JSON-RPC `-32601 Method not found` and HTTP 200, and the runner grades the 200 as "Agent returned 200 on an unauthenticated protected call". The same agent passes both probes on MCP.

## Where

`dist/lib/testing/storyboard/runner.js`, in the step executor:

```js
const useRawProbe = rawProbeHeaders !== void 0;
...
if (useRawProbe) {
  const probe = await rawMcpProbe({ agentUrl: runState.agentUrl, toolName: effectiveStep.task, ... });
  ...
  responseRecord = { transport: "mcp", ... };
```

and later

```js
transport: useRawProbe ? "mcp" : options.protocol === "a2a" ? "a2a" : "mcp",
```

`dist/lib/testing/storyboard/probes.js` exports a `rawA2aProbe` that builds a JSON-RPC envelope for an arbitrary method, and nothing in the package calls it.

## Reproduction

1. Serve an A2A agent whose JSON-RPC route rejects a missing credential with HTTP 401 and `WWW-Authenticate: Bearer` when asked with a 1.0 `SendMessage` envelope. (Verified with `curl` against the agent: 401 and the header are present.)
2. Run the storyboard with `--protocol a2a` and a test kit that sets `auth.api_key`.
3. `security_baseline` reports `probe_unauth` and `probe_invalid_api_key` failed with "returned 200". The run's `responseRecord` shows `transport: "mcp"` for those steps and the body is `{"error":{"code":-32601,"message":"Method not found"}}`.

## Expected

When `options.protocol === "a2a"`, an auth-override step sends a 1.0 `SendMessage` envelope through `rawA2aProbe` at the RPC endpoint named by the agent card's `supportedInterfaces[0].url`, forwards the trusted fetch so a test CA is honoured, and records `transport: "a2a"`.

## Workaround

[...]

PR #2958, opened 2026-09-18
fix(conformance): grade request-signing vectors by protocol and by verdict

### LINKED ISSUE #2954: Request-signing probe ignores options.protocol and has no 'a2a' transport, so every signed_requests vector errors on an A2A run (14.0.0-rc.35)
## Summary

#2878 fixed the auth-override probes so they dispatch as A2A when the protocol under test is A2A. That fix landed in 14.0.0-rc.35 — `runner.js:3116` now branches `options.protocol === "a2a" ? rawA2aAuthProbe() : rawMcpProbe(...)`, and `rawA2aProbe` is called at `runner.js:3101`.

The **request-signing probe was not included in that sweep**. It is the one remaining probe family that ignores `options.protocol`, so on an A2A run every `signed_requests` vector is dispatched as MCP against the A2A URL and fails before it is graded.

## Where

`dist/lib/testing/storyboard/request-signing/probe-dispatch.js:29`:

```js
function resolveVectorTransport(rsOpts) {
  return rsOpts.transport ?? "mcp";
}
```

The caller already has what it needs — `runner.js:3790`:

```js
httpResult = await probeRequestSigningVector(step.id, runState.agentUrl, options);
```

`options` carries `options.protocol`, but `resolveVectorTransport` reads only `options.request_signing`. `probe-dispatch.js` contains no reference to `protocol` at all.

There is also a second, deeper half. `grader.d.ts:94` types the mode as:

```ts
transport?: 'raw' | 'mcp';
```

so there is no `'a2a'` member to select even once the protocol is known. Unlike #2878 — where `rawA2aProbe` already existed and simply wasn't called — this one has no A2A dispatch path to switch to.

## Observed

One agent, `adcp_version: 3.1.1`, `sdk_version: 14.0.0-rc.35`, graded on both cards:

| card | `signed_requests` failures | shape |
|---|---|---|
| mcp | 27 | real grades (`expected 401 …, got 200`) |
| a2a | 31 | `request_signing_probe threw: MCP initialize precondition failed: Streamable HTTP error: Error POSTing to endpoint: <!doctype html> … 405 Method Not Allowed` |

All 31 A2A failures are the same thrown probe; none is a grade. The `<!doctype html>` is our framework's 405 page — the POST never reached an AdCP handler, because it was an MCP `initialize` sent to an A2A endpoint.

Excluding `signed_requests`, both cards fail an identical 21 checks. So this probe accounts for the **entire** MCP/A2A divergence in our run, which is what made it easy to mistake for a verifier bug on our side.

## Impact

Any agent that advertises `request_signing.supported: true` and exposes an A2A card fails ~31 universal checks it cannot pass by any implementation. Since `grading.pass_criteria` fails the storyboard on any vector missing its expected outcome, a conformant verifier cannot pass `signed_requests` on that card.

## Possible resolutions

1. **Add an `'a2a'` transport** that dispatches each vector as an A2A `message/send` envelope, mirroring what `'mcp'` mode does for `tools/call`, and have `resolveVectorTransport` read `options.protocol` when `request_signing.transport` is unset. This is the one that actually grades A2A verifiers.
2. **Minimum stop-gap:** have `resolveVectorTransport` read `options.protocol`, and when no mode can grade the run, return the existing `transport_ungradable` skip (`grader.js:163`) instead of throwing — so an infrastructure mismatch stops being reported as a conformance failure.

Happy to test a patch: we already run this runner in CI over mcp, a2a and rest, and carry a local `patch-package` patch against `@adcp+sdk+14.0.0-rc.35`.

Related: #2878 (same class, auth probes, fixed), #2931, #2875. Spec-side companions: adcontextprotocol/adcp#7567 (vector bodies), adcontextprotocol/adcp#7571 (filed there before I traced the runner to this repo — this issue supersedes it).

### LINKED ISSUE #2955: signed_requests storyboard: in-library vector 025 (jwks_override) always fails the synthesized http_status=401 step, even when the grader passes it
## Summary

In the `signed_requests` storyboard, negative vector `025-jwk-alg-crv-mismatch` cannot pass. The grader deliberately grades it in-library (it ships a `jwks_override`, so no HTTP probe is sent) and reports `passed: true, http_status: 0`. The synthesized step for every negative vector carries a single `http_status: 401` validation, and the dispatch layer copies `http_status: 0` into the step's probe result. The validator compares `0 === 401`, so the step fails while the grader's own verdict says pass. The agent under test is never contacted for this vector, so no implementation can change the outcome.

Reproduced on `@adcp/sdk` 13.0.3 and 13.0.4 (`dist/` files byte-identical), and the same code is on `main`.

## Where (main)

- `src/lib/testing/storyboard/request-signing/synthesize.ts:58-69` — `synthesizeNegativeStep` emits `{ check: 'http_status', value: 401 }` for every negative vector, including `jwks_override` ones.
- `src/lib/testing/storyboard/request-signing/grader.ts:561-609` — `gradeJwksOverrideNegative` runs `verifyRequestSignature` in-process and returns `http_status: 0` on both the pass and fail branches. The doc comment says this is intentional: "a black-box HTTP grade can't surface these failure modes. Grading the library verifier is the only path that exercises what the vector is testing".
- `src/lib/testing/storyboard/request-signing/probe-dispatch.ts:84-95` — maps the result to `{ status: result.http_status, error: result.passed ? undefined : ... }`. Its doc comment claims the mapping lets "the existing validation pipeline (`http_status`, `http_status_in`) work unchanged", but `status` is 0 here and `passed` is not consulted downstream.
- `src/lib/testing/storyboard/validations.ts:1823-1825` — `validateHttpStatus` passes only when `hr.status === expected`.

## Observed

AdCP 3.1 bundle, `adcp storyboard run <mcp-url>` from the 13.0.3 CLI, agent advertising `request_signing.supported: true`, `covers_content_digest: "either"`:

- 33 `signed_requests` steps pass (positives, replay 016, rate-abuse 020, multi-valued Content-Type 022, revocation, etc.).
- `negative-025-jwk-alg-crv-mismatch` fails with `reason: "Probe validations failed."`, validation error `Expected HTTP 401, got 0`. No grader diagnostic is attached, because the grader passed it.

The same failure appears at the 3.0 bundle.

## Expected

A vector the grader passes should produce a passing step. Either:

1. `synthesizeNegativeStep` emits a different validation for `jwks_override` vectors (e.g. an `error_code`-style check against the `www-authenticate` header the dispatcher already synthesizes, or a `probe_passed` check), or
2. `probeRequestSigningVector` maps in-library results to a synthetic `status: 401` when `result.passed` is true (and keeps `0` plus `error` when it is not), so the existing 401 validation holds, or
3. `jwks_override` vectors are excluded from storyboard synthesis and graded only through `gradeOneVector` / the `verify-vector` CLI, with the storyboard marking them `not_applicable`.

## Workaround status

`request_signing.skipVectors` (`operator_skip`) avoids the failure programmatically, and #668 already lists 025 there as "exercises SDK library, not the agent". But `bin/adcp.js storyboard run` sets no `request_signing` options at all (grep for `request_signing` / `skipVectors` in `bin/adcp.js` on main returns nothing), so CLI-driven CI has no way to apply it. The remaining option for CLI users is an out-of-band allowlist of the step id.

### LINKED ISSUE #2956: 3.0-line storyboard run: request-signing vectors default to the raw transport and the CLI cannot select MCP, so an MCP-only agent cannot pass signed_requests
## Summary

On the AdCP 3.0 line of the SDK (verified on `@adcp/sdk` 6.19.1 and 7.11.7, the current `adcp-3.0` dist-tag), the `signed_requests` storyboard's request-signing vectors are probed over the `raw` transport by default, and `bin/adcp.js storyboard run` gives the operator no way to change that. An MCP-only agent therefore cannot pass the 3.0 `signed_requests` storyboard: every vector is POSTed to `<agent-url>/adcp/<tool>`, a path such an agent does not serve, and every step fails with a 404 before any signature is verified.

The 13.x line already defaults to the MCP transport (`probe-dispatch.ts` `resolveVectorTransport` returns `rsOpts.transport ?? 'mcp'`), so the same agent passes 33 of the 3.1 `signed_requests` steps. This is a 3.0-line gap only.

## Where (7.11.7, identical in 6.19.1)

- `dist/lib/testing/storyboard/request-signing/grader.js:33` and `:211` — `const buildOpts = { baseUrl: agentUrl, transport: options.transport ?? 'raw' }`.
- `dist/lib/testing/storyboard/request-signing/builder.js` `retargetUrl` — takes the vector's own path (`/adcp/create_media_buy` etc.) and prefixes the agent URL's mount path in front of it, so with `http://127.0.0.1:8080/mcp` as the agent URL every probe goes to `http://127.0.0.1:8080/mcp/adcp/create_media_buy`.
- `dist/lib/testing/storyboard/request-signing/probe-dispatch.js:51` — only forwards `transport` when `options.request_signing.transport` is set.
- `bin/adcp.js` — contains no `request_signing` or `skipVectors` handling at all (`grep -c 'request_signing\|skipVectors' bin/adcp.js` prints 0). `--transport` is an alias for `--protocol` and never reaches the request-signing options.

## Observed

`node node_modules/@adcp/sdk/bin/adcp.js storyboard run http://127.0.0.1:8080/mcp --auth <token> --asserts-seeded-state --timeout 120 --summary-output storyboard-3.0.json` against an agent that serves MCP at `/mcp` and advertises `request_signing.supported: true`:

- `signed_requests` capability-discovery step passes.
- All 38 gradable vector steps fail; one is skipped. Representative failure from the summary:

```json
{
  "step_id": "positive-001-basic-post",
  "reason": "expected a 2xx status, got 404. Vector: Basic POST, Ed25519, no content-digest coverage",
  "detail": null
}
```

The same agent, same run, on the 3.1 bundle with the 13.0.3 CLI: 33 vector steps pass, 5 are `profile_excluded`, and the only failures are the two `covers_content_digest` profile mismatches (007/018) plus the in-library vector 025 (#2955).

## Expected

One of:

1. The 3.0-line `storyboard run` selects the transport from the protocol it is already running the storyboard over (MCP for an MCP agent URL), as the 13.x line does by default.
2. `bin/adcp.js storyboard run` exposes the `request_signing` options (`transport`, `skipVectors`, `skipRateAbuse`) as flags, so an operator can pass `--signing-transport mcp`.

## Workaround status

The only options today are to drive `runStoryboard` programmatically with `request_signing: { transport: 'mcp' }`, or to allowlist the whole 3.0 `signed_requests` storyboard in CI and rely on the 3.1 run for the graded proof. We are doing the latter.

bokelley:
Fixes #2954, fixes #2955, fixes #2956.

Three release-blocking bugs in the `signed_requests` conformance path. All three were reported by an adopter running the runner in CI across mcp/a2a/rest, and all three make a conformant agent fail vectors it cannot influence.

## #2954 — A2A runs dispatched signed-request probes as MCP

`resolveVectorTransport` was the last probe family that ignored `options.protocol` (the auth probes were fixed in #2878/#2902), so every `signed_requests` vector on an A2A run was framed as a JSON-RPC `tools/call` envelope, POSTed at the A2A endpoint, and failed with `405 Method Not Allowed` before grading — 31 universal checks an agent could not pass by any implementation.

**Shipped: the triaged stop-gap, reported as a coverage gap that cannot read as a pass.** `resolveVectorTransport` takes the resolved protocol and returns no framing for an A2A run that didn't pick one explicitly. Each affected vector skips as `signing_transport_unavailable` → canonical **`fixture_unavailable`**. Five behaviors carry that through, each one a false-assurance path that review caught and each with its own regression:

| Behavior | Why it was needed |
|---|---|
| No raw A2A replay of vector 028 | Replaying a fixture's JSON-RPC body at an A2A endpoint is a hand-rolled A2A dispatch (AGENTS.md forbids it), and the official `@a2a-js/sdk` handlers reject the unsigned raw shape anyway. 028 is now `signing_transport_unavailable` on A2A, still graded on MCP. |
| Storyboard cannot report `overall_passed: true` | The in-library self-check (025) graded green beside zero wire coverage, so a standalone `signed_requests` run reported a pass — and exit 0 — having verified nothing remote. |
| Track stays `partial` when every step is unavailable | `computeTrackStatus` returned `skip` for an all-skipped track, and `computeOverallStatus` ignores skipped tracks, so sibling tracks alone carried the run to `passing`. The `fixture_unavailable` check now runs **before** the all-skipped return. |
| Gate is per vector, not per storyboard | A routed `agents` run pairs a top-level `protocol: 'a2a'` with per-agent transports; a whole-storyboard gate stranded a perfectly gradable MCP seller (and hid the gap behind one synthetic skip). |
| `onlyVectors`/`skipVectors` applied before transport availability | Otherwise a vector the operator never selected came back `fixture_unavailable` — a manufactured coverage gap that inflates the unavailable count and drags the track to `partial` over vectors nobody asked to grade. |

Plus: `adcp storyboard step` renders an unavailable step as `⏭️  Not verified` with the remedy instead of a green `Passed` (a skipped step carries `passed: true`).

**Coverage that survives A2A:** only `025-jwk-alg-crv-mismatch`, decided in-library with no wire exchange — and its step title says `SDK verifier self-check, agent not contacted`, because that step grades the SDK, not the agent.

**Why no `'a2a'` transport.** The fixtures carry REST request bodies only, and no AdCP spec text defines an A2A framing for them. (An earlier revision cited adcp#7567 — that issue is schema-invalid vector bodies, not an A2A binding; adcp#7571 raised the A2A gap and was closed as superseded by adcp-client#2954. No open spec issue covers the binding today, so the rationale stands on the fixture contents.) Composing an envelope in the SDK — hand-rolled, or captured out of the official client, which also can't preserve the intentionally-invalid header/body bytes ~21 negative vectors depend on — would bake an SDK-invented binding into every adopter's conformance run.

**Scope note on the pre-existing `transport_ungradable` carve-outs.** Two vectors (026 non-ASCII host, profile-3.2/002 malformed authority) are skipped on *every* run — no HTTP client can put those bytes on the wire — and keep their existing `not_applicable` mapping. Flipping them would turn every `signed_requests` run, MCP included, `partial` over two vectors while the other 38 still verify the agent; that is a different question from these three issues and is left to a follow-up, with the reasoning recorded at the mapping.

So this PR closes #2954 with the stop-gap plus honest reporting. A real A2A transport is follow-on work gated on the spec defining the binding.

## #2955 — in-library vector 025 always failed its synthesized step

`gradeJwksOverrideNegative` decides `jwks_override` vectors against the SDK verifier (the malformed JWK is never published by the agent) and reports `http_status: 0` by contract. Synthesis emitted `http_status: 401` for every negative vector, so vector 025 failed while the grader passed it.

Adds a `probe_passed` validation check that grades the verdict the dispatch layer actually reports (`HttpProbeResult.error` unset ⇒ pass), and emits it from `synthesizeNegativeStep` for `jwks_override` vectors only. Probed negatives keep their 401 assertion. No synthetic HTTP status is invented anywhere — per the protocol-expert triage, fabricating a 401 from a library boolean would conflate the library-verification and wire planes (RFC 9421 §3).

## #2956 — CLI could not reach the request-signing options

`bin/adcp.js storyboard run` never populated `request_signing`, so CLI-driven CI had no way to select the vector transport or exclude a vector. Adds:

- `--signing-transport raw|mcp` — distinct from `--transport`/`--protocol`; help text says so explicitly. Omit it and the vector transport is inferred from the resolved protocol (MCP → `mcp`, A2A → `signing_transport_unavailable`, reported as coverage unavailable per the table above).
- `--signing-skip-vectors <ids>` and `--signing-skip-rate-abuse`.

Threaded through every storyboard-run path (single, `--file`, multi-instance, agents-routed, `--local-agent`, `storyboard step`), with `ComplyOptions.request_signing` added as a typed passthrough so the default assessment route is covered too. Malformed values exit 2 rather than falling back silently. Also corrects the stale `BuildOptions.transport` JSDoc that still advertised `'raw' (default)`, spelling out why the builder primitive and the grader entry points intentionally default differently.

**Backport determination (asked for in the issue):** yes, the published 3.0 line needs its own. `adcp-3.0` = `@adcp/sdk@7.11.7`, published from `origin/backport-2114-adcp-3.0` (`a18bbb73`). That branch still has `transport: options.transport ?? 'raw'` in `grader.ts` and no `request_signing` wiring in `bin/adcp.js`, so nothing on `main` reaches it. Not done here — it is a separate PR against a separate branch, and no `dist/` artifact was hand-edited. Until it lands, 3.0-line operators grade `signed_requests` from the 3.1 CLI or drive `runStoryboard` with `request_signing: { transport: 'mcp' }`; that workaround is now documented in `docs/guides/VALIDATE-YOUR-AGENT.md`.

## Tests

All verified to fail before the fix (by reverting the built behavior and re-running):

- `test/lib/request-signing-vector-transport.test.js` — protocol-aware resolution; the per-vector gap and its canonical `fixture_unavailable` mapping; **028 refused on A2A** and **still graded on MCP** with its verbatim JSON-RPC body witnessed on the wire; **storyboard verdict** false beside a passing SDK self-check; **all-unavailable track** stays `partial` and keeps `computeOverallStatus` off `passing`; **sibling-track rollup**; **routed agents** (`routedAgentOptions({transport:'mcp'}, {protocol:'a2a'})` still grades); **`onlyVectors`/`skipVectors` before transport availability**, with the selected vector keeping its truthful verdict; rate-abuse cap sourced from the contract.
- `test/lib/request-signing-jwks-override-grading.test.js` — `probe_passed` synthesis, the SDK-self-check step title, the grade→validation round trip, failure and no-probe-result paths.
- `test/lib/cli-storyboard-signing-flags.test.js` (fast, ~5s) — flag semantics against the shared parser seam, one exit-2 spawn, help text (including the `signing_transport_unavailable` wording), skip-line rendering, and the single-step verdict that must never print `Passed` for an ungraded step.
- `test/lib/cli-storyboard-signing-threading.test.js` (slow group, ~18s) — `--file` threading, `--signing-transport raw` observed as a per-operation probe path, the **assessment path** (`storyboard run <url> signed_requests`, no `--file`) proving `runFullAssessment` forwarding, and a false-assurance guard: a mock that accepts negatives fails them and the run reports `partial`.

CI fixes in this revision: the skip renderer and flag parser moved into the already-shipped `bin/adcp-storyboard-summary.js` instead of a new published file, so the packed-file budget is back at 6135/6135 (no budget raise); the spawn-heavy tests were split into the slow-group file so every fast-suite file stays under the 60s ceiling (4.8s / 17.8s measured).

Verification: `npm run check:package-size`, `format:check`, `lint`, `typecheck`, `ci:docs-check` (no generated drift), `ci:doc-links`, `check:skill-sync`, plus the focused suites (202 tests green).

**`rateAbuseCap` adjudication (not clamped, deliberately).** The knob drives a `for (i < cap)` probe loop, but this PR does not widen its reach: `StoryboardRunOptions.request_signing.rateAbuseCap` is public on `origin/main` (`types.ts:1723`), `GradeOptions.rateAbuseCap` likewise, and `complyImpl` on `main` does not destructure `request_signing`, so it already flowed through `...testOptions` into `runOptions` — independently confirmed by the reviewer's observation that a `comply()`-level test passes unchanged on `origin/main`. No CLI flag exposes it before or after. Per the "document evidence rather than expand scope" instruction it is documented, not clamped; a test pins the default to the shipped contract's bounded `grading_target_per_keyid_cap_requests`.

## Review

Dual-stack per `docs/development/REVIEW-STACKS.md`, run across three rounds (`review:codex` code/security/protocol personas with the diff piped via stdin — the repo-read path needs `bwrap`, unavailable in this sandbox — plus Claude code-reviewer passes). Every blocker the reviews raised changed the patch:

- **Track-level false assurance** → per-vector `fixture_unavailable` reporting, replacing a whole-storyboard skip that produced `overall_passed: true | passed 0 | failed 0 | skipped 40`.
- **All-unavailable track evaporating into `skip`** → `fixture_unavailable` check hoisted above the all-skipped return.
- **Standalone storyboard passing off an SDK self-check** → `UNVERIFIED_COVERAGE_SKIP_REASONS` blocks `overall_passed`.
- **Raw A2A replay of 028** → removed, with the AGENTS.md rule and the official-handler rejection recorded at the call site.
- **`onlyVectors` after the transport gate** → operator selection applied first.
- **Routed-agent override**, **assessment-path test that didn't discriminate**, **false adcp#7567 citation**, **JUnit-invisible self-check title**, **unredacted `skip.detail` seam**, **stale `transport_ungradable` wording in `--help`**, **missing flags in the guide/skill**, **`patch` bump on additive public surface** (now `minor`, with the rationale in the changeset) — all addressed.

Changeset included (`minor`).

---

---

[Open workspace in Conductor](https://app.conductor.build/workspace/e0a8127b-ee3c-4882-976c-5386fb98b3e0)


KonstantinMirin:
Reporter of #2954 — thanks for turning these around quickly. Two things: a measurement that bears on the "why no `'a2a'` transport" decision, and a third defect the same runs surfaced that isn't in this PR.

## 1. A measured A2A framing, offered as data rather than as a request to reverse the decision

We carry a local `patch-package` patch to finish our CI run, and it does include an `'a2a'` transport, so we have a number for what that coverage is worth. Same deployment, same agent, both cards:

| card | before | after |
|---|---|---|
| mcp | 56 passed / 25 failed | 59 / 21 |
| a2a | 31 / 52 — all 31 signing vectors thrown in the MCP `initialize` precondition | 59 / 21 |

The two cards' failure sets are now **identical — symmetric difference zero**. The 21 that remain are the same non-signing gaps both surfaces always had. That identity is the interesting part: it says the mcp/a2a divergence in a `signed_requests` run was the grader, not the agent, and that an A2A verifier is gradable to the same resolution as an MCP one once the vectors are framed.

The framing we used, with its ceilings, since they bear on your "SDK-invented binding" objection — which I think is right as stated:

- a native 1.0 `SendMessage` JSON-RPC envelope, not the 0.3 `message/send` family (an agent whose JSON-RPC routes are built without the 0.3 compat adapter doesn't serve that family at all), `parts: [{ data: { skill: <operation>, parameters: <vector body> } }]`
- `A2A-Version: 1.0` merged **before** the signature is computed — that ordering is the whole point, since Content-Digest covers the body and the signature base covers the headers, so a version header added after signing sits outside the base on a request the verifier reconstructs with it
- the RPC endpoint guessed as `/a2a` when the base URL carries no path, because the signing probe is deliberately bearer-less and doesn't fetch the card. A real ceiling; reading `supportedInterfaces[0].url` off the card the way `rawA2aAuthProbe` does since #2902 is the correct fix
- the positive canonicalization vectors skip on a2a for the same reason they skip on mcp — both route every vector to one endpoint, so the URL edge never reaches the wire. We widened `MCP_FLATTENED_VECTORS` to cover both enveloped transports
- and it is hand-rolled, which AGENTS.md forbids, and the official client can't carry the intentionally-invalid header/body bytes ~21 negatives depend on. So the binding really is a spec question, as you wrote

Posting it so the call is made against a measurement rather than an unknown. If you'd like the patch as a draft PR to look at, say the word. If the answer is "not until the spec defines the binding", that's a fine answer — the stop-gap here is the right shipping state, and we'll keep carrying the patch locally.

One small thing on #2955: your `probe_passed` check is better than what we did locally (we simply emitted no validations for `jwks_override` vectors and let the verdict land on `error`). Grading the reported verdict explicitly is the right shape.

## 2. A third defect this PR doesn't cover: `agentCapability` is never wired

Opened as #2963.

`GradeOptions.agentCapability` and its `capabilityMismatch` preflight have existed since #668, but `probe-dispatch.ts` never passes them. So `negative/007` (needs `covers_content_digest: 'required'`) and `negative/018` (needs `'forbidden'`) are both sent to every agent, and since an agent declares one value, they fail on every run — on MCP too, so it's orthogonal to the A2A question.

Wiring it exposed a second bug in the predicate itself: it keys the skip on the vector's *declared* `covers_content_digest`, and the shipped fixtures set that to `'required'` on `positive/002` (a covered digest that must be accepted), `negative/010` (a falsified digest) and `negative/023` (a malformed digest header) — none of which asserts the policy. Against an `'either'` agent all three were dropped. The function's own doc comment already states the correct rule ("it can't pass vectors 007 or 018"), so #2963 just narrows the code to its documentation.

It touches `probe-dispatch.ts`, so it conflicts with this PR in one spot — the block right after `const rsOpts = …`. `grader.ts` merges clean. Happy to rebase after this lands, or to fold it in here if you'd rather ship one change.

KonstantinMirin:
We ran this against our agent before commenting, so the numbers below are measured rather than inferred. Three of the four defects we hit are fixed here and the A2A stop-gap does what the description says — we confirmed it reports a coverage gap rather than a false pass, which we appreciate.

One defect is not addressed, and it is the one that costs a conformant agent the most.

## `agentCapability` never reaches the grader

`capabilityMismatch` and `agentCapability` appear nowhere in this diff (the only hits are test fixtures). The machinery already ships in `grader.mjs` — the storyboard path simply never hands it the agent's declared `request_signing`. So vectors keep being graded against a profile the agent does not implement:

- **`negative-007`** requires `covers_content_digest: "required"`; **`negative-018`** requires `"forbidden"`. An agent declares exactly one value, so **one of that pair fails for every agent, whatever it declares.** They are mutually exclusive by construction.
- **`negative-028`** asserts a `protocol_methods_required_for` bucket. An agent that declares nothing there — as the storyboard's own text says it may, *"the runner skips when the agent does not declare the bucket"* — still gets the vector dispatched and failed.

That is **3 permanent failures on every card, for every agent, that no implementation can move.** Merging this closes #2954/#2955/#2956 and leaves that behind.

## Measured

Against our seller, with everything in this PR applied and nothing else changed, `signed_requests` on the mcp card still carries those 3. With the capability wired through, they resolve to `capabilityMismatch` skips — which is the runner's own designed outcome, not a special case — and `signed_requests` reaches **0 failures**.

Our declared value is `covers_content_digest: "either"`, which is the one that grades the most: `"required"` would additionally fail the three positives that do not sign a digest, and `"forbidden"` would fail `positive/002`, which does. So this is not us picking a lenient profile to dodge vectors — `"either"` costs us 007 and 018 and we still cannot pass both.

## We have opened the fix

#2963 wires `agentCapability` through. One caution from our own implementation of it, which cost us a re-run: the predicate must exclude a vector **only** when its `verifier_capability` names a `covers_content_digest` value the agent did not declare. Our first attempt matched on "does this vector involve content-digest" and wrongly dropped `positive-002` and `negative-010`, both of which are gradable under `"either"` — 002 covers the digest and must be accepted, 010 has a mismatch and must be rejected. Over-exclusion inflates a pass rate as effectively as under-exclusion deflates it.

Happy to fold #2963 into this PR instead if you would rather ship it as one change.

KonstantinMirin:
**Retracting the `/a2a` caveat in my comment above.** I wrote it as a ceiling of the approach. It isn't — it's a shortcut in our local patch. The endpoint is never assumed; it comes off the agent card.

A2A is card-first by design, and the card carries more than a URL. Ours declares, at `src/a2a_server/adcp_a2a_server.py:602`:

```python
AgentInterface(url=server_url, protocol_binding="JSONRPC", protocol_version="1.0")
```

served from `/.well-known/agent-card.json` (`src/app.py:378`). A grader therefore needs no knowledge of `/a2a`, or of any path: fetch the card, read `AgentInterface.url`, `protocol_binding` and `protocol_version`, dispatch there. Any host, any port, any path, any scheme. Our patch defaults to `/a2a` only because our own card publishes exactly that URL (`get_a2a_server_url() or "http://localhost:8091/a2a"`), so on the measured run card-derived and guessed resolve to the same endpoint and the same bytes. `rawA2aAuthProbe` has done the correct thing since #2902 by reading `supportedInterfaces[0].url`.

**I'm confident about this rather than asserting it, because we got it wrong once and it was expensive.** From `config/nginx/nginx-tls-test.conf.template:69`:

> The storyboard agent. Fronted for A2A's sake: A2A is card-first, and the card renders https for any non-loopback host, so dialed plaintext on :8080 it published `https://adcp-server-storyboard:8080/a2a` and every card-derived call failed. Behind this front the scheme the card publishes is the scheme the origin speaks.

The commit that fixed it records the blast radius: 25 checks failed outright, but one of them was the capability probe, and the runner *selects storyboards* from that declaration — so the A2A axis executed 25 storyboards where MCP executed 44, and the same unread declaration left `signed_requests` ungated for 40 more failures. The card was correct and the origin was not. The fix was to make the origin match the card. Overriding the card was never on the table.

**The same lesson applies one level down, to the payload — and this is why I'd argue it isn't a hand-rolled binding.**

The SDK already maps AdCP onto A2A for every ordinary tool call, at `src/lib/protocols/a2a.ts:743`:

```ts
const legacyWire = client.protocolVersion?.startsWith('0.') ?? false;
const invocation = legacyWire ? { skill: toolName, parameters } : { skill: toolName, input: parameters };
```

Note what selects the spelling: `client.protocolVersion`, which comes from the card. Our patch hardcoded `{ skill, parameters }` — the 0.x form — while sending `A2A-Version: 1.0`. It graded green only because our dispatcher is lenient (`data.get("input") or data.get("parameters", {})`, `src/a2a_server/adcp_a2a_server.py:249`). Against a strict 1.0 agent our envelope would have been wrong by the SDK's own rule. Second proof of the same point, in the same patch: every value a grader decides for itself — scheme, host, port, path, binding, version, and the payload key that depends on the version — is a value that will be wrong for some agent.

So the shape worth offering is not "add an SDK-invented A2A framing". It is: resolve endpoint, binding and version from the card exactly as a real buyer does, then reuse the invocation the SDK already builds at `a2a.ts:743`, apply the vector's mutation, and compute the signature *after* the envelope is assembled — that ordering is the whole point, since Content-Digest covers the body and the signature base covers the headers.

That narrows the genuinely new surface to one thing: the vectors' bytes have to survive assembly, and the official client can't carry the intentionally-invalid headers ~21 negatives depend on. That is the real blocker, and it's a smaller one than "no A2A framing is defined" — the discovery path and the invocation mapping both already exist in this repo. What's missing is a way to push deliberately-broken bytes through them.

Offer stands, and it's yours to decline: happy to put this up as a draft PR, against this branch or against `main`, whichever is less disruptive. If the answer is still "not until the spec settles the binding," no argument from me — the stop-gap here is the right shipping state either way.

PR #2963, opened 2026-09-18
fix(storyboard): grade content-digest vectors against the selected verifier profile

KonstantinMirin:
## Summary

Salvages the independently valid content-digest grading correction from the original branch and rebases it onto current `main`.

An operator-selected verifier profile with `covers_content_digest: either` must still grade vectors that test covered-request acceptance, digest mismatch, and malformed digest headers. Only vectors whose expected outcome depends on a stricter `required` or `forbidden` policy are excluded.

## Safety boundary

This branch deliberately drops the original live-capability wiring. The agent's advertisement is not the operator-selected verifier grading profile. Feeding the whole live `request_signing` block into `agentCapability` lets an under-declaring agent suppress 39 of 40 vectors. Current main continues to read only the spec-defined `protocol_methods_required_for` applicability dimension from live capabilities.

## Verification

- Rebased on current main (`e1843665118c4f5b065b1b8710441c68b7ba376d`)
- `npm run build`
- 89 focused request-signing/blocker tests passed
- Conventional commit/title checks passed
- Patch changeset included

No merge requested; exact-head CI and review are required.


bokelley:
Takeover update: I rebased this PR onto current main and narrowed it to the first, independently valid commit. The live-advertisement wiring was removed because exact-head review demonstrated that a permissive/under-declared capability block could suppress 39 of 40 signing vectors. The remaining change only fixes `either` content-digest policy grading for an explicitly selected verifier profile. Fresh exact-head CI is now running; I will not merge without a clean review and explicit approval.

PR #2967, opened 2026-09-19
fix(storyboard): sign the request the official A2A client emits

KonstantinMirin:
## What an A2A run grades today

Nothing. Every request-signing vector reports `signing_transport_unavailable`.

#2958 made that **fail-closed, and it was right**: before it, the runner POSTed an MCP `initialize` at an A2A endpoint, took the `405`, and reported it as a signature failure. Reporting a coverage gap beats reporting a false verdict.

But it is a stop-gap, and it is permanent until the dispatch is wired. This wires it — in the shape #2958's own comment in `probe-dispatch.ts` described as the correct one:

> *"Wiring it means signing a request the official client issues — applying each vector's RFC 9421 headers to that client's own request"*

## Nothing here frames an A2A envelope

`@a2a-js/sdk`'s `ClientFactory` resolves the agent card and builds the call. The request is captured at the SDK's own `fetchImpl` seam, and **those exact bytes are signed**. Every protocol decision belongs to the official client:

| decision | who makes it |
|---|---|
| endpoint | the card's `supportedInterfaces[]` |
| JSON-RPC method name | the SDK, per the card's `protocolVersion` |
| `a2a-version` header | the SDK, from that same version |
| proto-JSON encoding | the SDK (`role` → `"ROLE_USER"`, never `1`) |
| payload key | the SDK (`input` on 1.x, `parameters` on 0.x) |

Observed rather than assumed, and asserted in tests. One `cancelTask` call, two cards:

```
card protocolVersion=1.0   → {"jsonrpc":"2.0","method":"CancelTask","params":{},"id":1}    a2a-version: 1.0
card protocolVersion=0.3.0 → {"id":1,"jsonrpc":"2.0","method":"tasks/cancel","params":{}}  a2a-version: 0.3
```

The second line is why this matters for vector 028. `tasks/cancel` is the method AdCP's `protocol_methods_*` namespace names — `security.mdx` @ 3.1.1 :1045 cites *"A2A 0.3.0 §7.x defines task-lifecycle methods (`tasks/cancel`, `tasks/get`, `tasks/resubscribe`)"*. The official client emits it **natively** for a 0.3 agent. No envelope is framed to produce it.

## Ordering: the version header is inside the signature base

Capture happens before signing, so `a2a-version` is part of the request the signer sees. A header added after signing sits outside the base the verifier reconstructs, and every vector would fail for a reason unrelated to the agent's verifier.

## Fail-closed is preserved, not replaced

`resolveVectorTransport` **still returns no framing for A2A**. It cannot answer the real question — whether the agent's card resolves to a JSONRPC interface — because that is a fetch and the function is synchronous.

The decision moved into the async dispatcher:

* card resolves → the official client dispatches, and the coverage gap closes;
* card does not resolve, or declares no JSONRPC interface → nothing is framed on a guess, and the vector keeps #2958's `signing_transport_unavailable` reporting **with every guardrail around it intact**.

**All 44 tests in `request-signing-vector-transport.test.js` pass unchanged.** That is deliberate: this does not delete behaviour you shipped on purpose. It is the "keep it reachable as a fallback" option from the #2964 thread, implemented rather than asked about.

## Tests

7 new. Each asserts that a protocol decision was the SDK's, not this repo's — a card is served over loopback and the SDK resolves it, so a passing test is evidence the SDK chose the endpoint, the method name, the version header and the encoding.

The availability gate is graded both ways: a resolvable card makes dispatch available and names the endpoint it resolved; an unreachable agent stays unavailable rather than inventing an endpoint from the agent URL.

`npm run test:node:fast`: 125 pass, 0 fail. (`handler-controlled-flow.test.js` fails on unmodified `main` as well — verified by stashing this branch.)

## On #2964

Closed with three objections. The first was fair and this PR is the answer to it: that branch hand-built the envelope, this one does not build one at all.

The other two are worth re-examining against this repo's own source before they are applied to this PR:

* **"vector 028 carries an MCP `tasks/cancel` envelope."** `tasks/cancel` is an A2A method — MCP defines no such method — and `builder.ts:295-299` already says so: *"The fixture body is already a JSON-RPC envelope with `method: "tasks/cancel"` — NOT a `tools/call` envelope."* 028 is the one vector that is natively A2A.
* **"an invented binding the contract does not specify."** The shipped MCP path re-frames the same REST fixtures into a `tools/call` envelope built by hand in `wrapMcpEnvelope`, with `params.name` derived from the fixture's URL path. If re-framing a REST fixture is inventing a binding, that one is invented too. This PR invents neither — it signs what the official client emits.

`@a2a-js/sdk` stays a peer dependency; the import is dynamic and inside the A2A path, so grading an MCP agent never loads it.

---

## Verified against a live agent — both cards now grade identically

Run on a real seller (prebid/salesagent, which implements the verifier transport-independently), this SDK built from this branch:

```
mcp: passed=59 failed=23 not_selected=10 graded=82
a2a: passed=59 failed=23 not_selected=10 graded=82
```

**Symmetric difference 0.** Same 82 vectors graded, same 59 passing, same 23 failing. That equality is the property worth having: two cards that grade different vector sets compare nothing.

The remaining 23 are the same non-signing gaps both surfaces always had, plus two vectors — `negative-007-missing-content-digest` and `negative-018-digest-covered-when-forbidden` — that fail **identically on both transports** because this build lacks the `agentCapability` wiring in #2963. Their symmetry is the evidence they are not transport defects.

### Three defects this measurement found, each fixed here

Getting to that equality took three corrections, and the first two were only findable against a live verifier:

1. **`32367ae` — duplicate `Content-Type` on the wire.** A plain object spread is case-sensitive; the fixture's `Content-Type` and the client's `content-type` both survived it, and the request carried the header twice. A conformant verifier refuses that at checklist step 1 as `request_signature_header_malformed`, before evaluating anything the vector grades. This was **16 of 27** A2A failures. A2A went 42p → 66p.

2. **`c49689d` — the fixture's deliberate malformation was being overwritten.** Merging the other way fixed the duplicate but broke `negative-022-multi-valued-content-type`, whose multi-valued `Content-Type` *is* the fault under test. The transport now fills only gaps (`a2a-version`, `accept`) and never replaces a header the fixture names.

3. **`26870d7` — URL-edge vectors were graded on A2A.** A2A graded 90 vectors to MCP's 82, and the extra 8 were the canonicalization positives (005–008), reported as **passes**. Their edge lives in the vector URL's port/path/query, which any single-endpoint transport neutralizes — that skip existed but was gated on `transport === 'mcp'`, and A2A routes every vector to the one endpoint its card names for exactly the same reason. Coverage claimed and not delivered; the gate now reads "not `raw`", which is what the condition always meant.

### On the Ladon medium finding (`ff091ad`)

Correct, and fixed. The 022 guard located the vector and then built `loaded.positive[0]`, asserting only `a2a-version`/`accept` — gap-fills present under either merge order, so it passed with the bug in place. It now hands the transport a `content-type` differing from the fixture's and asserts the fixture's survives, so only one order satisfies it. Verified non-vacuous: flipped, it fails on `the fixture's Content-Type must survive`; restored, it passes.

---

[Open workspace in Conductor](https://app.conductor.build/workspace/aa50b332-8f42-4385-9761-e7e09b18fa33)


KonstantinMirin:
The medium finding was right, and it's fixed in `ff091ad`.

The guard located vector 022 and then built `loaded.positive[0]`, asserting only `a2a-version` and `accept`. Both are gap-fills present under either merge order, so the test passed identically with the bug in place — it named the regression and guarded nothing. That is the worst kind of test, because the green mark is what stops anyone looking.

It now hands the transport a `content-type` that differs from the fixture's (`application/vnd.transport-must-not-win`, a value no fixture uses) and asserts the **fixture's** value survives. Only one merge order satisfies that. Checked rather than claimed: with the order flipped back the test fails on `the fixture's Content-Type must survive`, and passes again when restored.

Worth recording why the order matters at all, since it is the opposite of the intuition: the transport is a witness, not a translator. Where a vector names a header it is naming it **on purpose** — 022's multi-valued `Content-Type` is the fault under test — so a transport that overwrites it makes the agent answer about a different defect and the vector grades nothing. The client's headers fill only what the fixture does not set.

Since the last review the branch has also been measured against a live seller that implements its verifier transport-independently:

```
mcp: passed=59 failed=23 not_selected=10 graded=82
a2a: passed=59 failed=23 not_selected=10 graded=82
```

Symmetric difference 0 — same vectors, same results, both cards. Getting there surfaced two further defects this branch now carries fixes for: a case-sensitive header merge that sent `Content-Type` twice and was refused at checklist step 1 (16 of 27 A2A failures), and the URL-canonicalization skip being gated on MCP alone, which let A2A report vectors 005–008 as passes when the edge never reached the wire. Details and commit-by-commit numbers are in the updated description.

bokelley:
Issue #2973 was filed covering the `legacyCompat` knob missing from the conformance runner. Part of that issue — the hardcoded `{ enabled: true }` in `request-signing/a2a-dispatch.ts` (lines 138, 212) — overlaps with this PR's diff. The author of #2973 is you, and you acknowledged it's yours to fix here ("so that is mine to fix either way").

Suggesting this be folded before merge so #2973's scope doesn't split across two PRs. The broader work (runner guard fix, card-derived default in `createNativeA2AClientFromCardUrl`, CLI flag) is tracked separately on #2973 as ready-to-implement.

Blocked-on: this PR — #2973 resurfaces on merge for the remaining scope.

---

KonstantinMirin:
Folded the `a2a-dispatch.ts` half of #2973, in `60d2a7c`. Also merged `main` in (`8ea1259`) — this branch was five commits behind, including #2963 and #2968, which touch the same grading path; clean merge.

**#2973's open question, answered by measurement rather than reading.** That issue asked whether `legacyCompat: true` means "also accept 0.3" or "speak 0.3", and noted the right fix depends on which. One `cancelTask` against `@a2a-js/sdk`, four combinations:

| card | `enabled` | method emitted | `a2a-version` |
|---|---|---|---|
| `1.0` | `true` | `CancelTask` | `1.0` |
| `1.0` | `false` | `CancelTask` | `1.0` |
| `0.3.0` | `true` | `tasks/cancel` | `0.3` |
| `0.3.0` | `false` | `CancelTask` | `1.0` |

It is "also accept". `true` is the card-following setting — a 1.0 card is never downgraded, the first two rows are byte-identical — and it is what gives a 0.3 card the `tasks/*` family the spec names for it (security.mdx @ 3.1.1 :1045). `false` is the value that ignores the card: the last row is a 1.0 frame sent to an agent that published 0.3, silently, with no refusal.

So the hardcoded value was right and flipping it would have been a regression. What was missing was the reason. The two literals are now one frozen named constant with the table recorded at it, and a test pins both directions. It is deliberately not operator-configurable at this layer — a knob selecting framing the agent never advertised produces a verdict about an agent that does not exist. #2973 keeps its broader scope: the library-wide default in `createNativeA2AClientFromCardUrl`, the CLI flag, and the runner guard.

**Measured against a live seller before pushing**, since this PR is what grades one. Built this branch, packed it, installed it over the pinned SDK in a storyboard worktree, and ran the full stack — run `innet_200926_1358`, salesagent `c1ef49bcb`:

```
mcp  101 passed / 15 failed
a2a  101 passed / 15 failed
webhook_emission: 0 failures   signed_requests: 0 failures   (both protocols)
```

Identical to the baseline taken on the pinned SDK, and the two protocols' failure sets have symmetric difference 0. The 15 that remain are pre-existing `media_buy`/`core` gaps unrelated to signing or transport.

One thing worth reporting from that exercise: the seller's local `patch-package` patch against rc.35 no longer applies to this build, and that is the intended outcome rather than a problem — every hunk it carried is now upstream (#2958, #2963, this PR's dispatch and brand.json commits). The run above used no local patch at all; the 101/101 is this build standing on its own.

`npm run test:node:fast`: 144 files, one failure — `handler-controlled-flow.test.js`, which fails identically on unmodified `main` (all 12 subtests pass; the file-level harness is what errors).

bokelley:
Comment-resolution audit on current head `da20cdcf`:

- Vector 028 now routes through `passthrough` in A2A mode, so the captured official-client endpoint, headers, and body reach `applyTransport`; regression coverage pins that behavior.
- Agent-card discovery uses the guarded transport with an abort deadline, including injected fetches that ignore abort; a timeout regression test covers it.
- A2A availability is scoped to the current run, and card-discovery failures are redacted, truncated, and included in the coverage-gap detail instead of being reported as a permanent global capability result.
- A2A vector parsing now reuses the parsed body and has a single explicit malformed-body boundary rather than double-parsing.
- The Content-Type precedence regression test is non-vacuous: it supplies a conflicting transport value and asserts the fixture value survives exactly once.

All five inline threads are resolved. The separate #2973 concern is also closed after verifying that #2895 exposed the override, #2968 forces native compliance mode, and the official SDK follows a 1.0 card under compatibility mode. No unresolved review threads remain.

